PR #7, opened 2016-11-19
Use equals() for String comparisons instead of '==' and Remove unused imports

cosmofrit:
Corrected some String comparisons to use equals() instead of '=='


cosmofrit:
I've reverted the commit for unused imports

v-suhame:
Thanks for your contribution.

PR #8, opened 2016-11-19
Implement Driver#getParentLogger

marschall:
Implement getParentLogger

The driver and data source use JUL for logging but do not implement
#getParentLogger.

This commit contains the following changes:

 * implement Driver#getParentLogger
 * implement DataSource#getParentLogger
 * use "com.microsoft.sqlserver.jdbc" as the parent logger
 * instead of spaces and tabs on the same line for indentation use
   spaces only
 * remove the comments
   "The driver currently does not implement JDDBC 4.1 APIs".
   While technically true these specific methods are implemented now.
 * follow driver naming conventions rather than Java naming conventions
 * follow driver conventions modifier location rather than Java
   conventions
 * no tests added since I couldn't find any existing tests for the
   driver oder data source class


v-xiangs:
@marschall the PR is accepted, thanks for your contribution

PR #14, opened 2016-11-20
Avoid unnecessary calls to String copy constructor

marschall:
The driver currently contains 20 calls to the String(String) copy
constructor. To quote from the Javadoc

> Unless an explicit copy of {@code original} is needed, use of this
> constructor is unnecessary since Strings are immutable.

Calling the constructor only results in unnecessary pressure of both
the allocator and the heap.


v-suhame:
Thanks for your contribution @marschall

PR #25, opened 2016-11-24
Fix StreamError class comment

marschall:
The StreamError class comment says it's serializable because it is held
by SQLServerException. Both statements are wrong.


v-xiangs:
@marschall  good catch, thank you for the fix

PR #26, opened 2016-11-24
Use Charset throughout

marschall:
The driver spends a lot of time converting from Charset to String and
from String to Charset. Every time checked exceptions have to be
caught. In addition the Charset lookup code is quite involved. It would
be simpler to use Charset everywhere.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLCollation.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLCollation.java
@@ -20,6 +20,7 @@
 package com.microsoft.sqlserver.jdbc;
 
 import java.io.*;
+import java.nio.charset.Charset;
 import java.util.*;
 import java.text.MessageFormat;
 
@@ -49,7 +50,7 @@ final class SQLCollation implements java.io.Serializable
     private final Encoding encoding;
 
     // Utility methods for getting details of this collation's encoding
-    final String getCharset() { return encoding.charsetName(); }
+    final Charset getCharset() throws SQLServerException { return encoding.charset(); }
     final boolean supportsAsciiConversion() { return encoding.supportsAsciiConversion(); }
     final boolean hasAsciiCompatibleSBCS() { return encoding.hasAsciiCompatibleSBCS(); }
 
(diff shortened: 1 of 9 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/SQLCollation.java
```diff
@@ -587,9 +590,9 @@ final Encoding checkSupported() throws UnsupportedEncodingException
             // This works for all of the code pages above in SE 5 and later.
             try
             {
-                " ".getBytes(charsetName);
+                charset = Charset.forName(charsetName);
```
v-xiangs: instead of assigning the charset object to charset variable, is it possible to do the assignment outside the checkSupported() method? e.g. we could do the assignment after checkSupported() is called in charset()

marschall: I can do this.

src/main/java/com/microsoft/sqlserver/jdbc/SQLCollation.java
```diff
@@ -602,7 +600,23 @@ final Encoding checkSupported() throws UnsupportedEncodingException
...
 
-    final String charsetName() { return charsetName; }
+    final Charset charset() throws SQLServerException
+    {
+        try
+        {
+            checkSupported();
+            if (charset == null)
+            {
+                charset = Charset.forName(charsetName);
```
v-xiangs: @marschall thank you very much for your great idea and fix. I just have one more question and hope you could consider. Is it possible to delete the local variable 'private Charset charset'? I am not sure why we need to check if the variable is null here. instead of assigning it to charset and returning charset., could we just do "return Charset.forName(charsetName);" ?

marschall: Yes I considered this. It would be possible but:
'private Charset charset' ist not a local variable but an instance/member variable. `Encoding#charset()` is called quite frequently and by many users both directly and indirectly. If you look at the implementation of `Charset#forName` it is quite involved and the worst case is quite bad, eg. some of the charsets are in the Extended Encoding Set. I would be more comfortable with looking them up just once.

v-xiangs: @marschall thank you for your feedback.
I totally agree to what you said, we definitely need to look up once.

However, what I meant was something like this.:
```
        try
    	{
    		checkSupported();
    	} 
    	catch (UnsupportedEncodingException e) 
              ...
    	return Charset.forName(charsetName);
```
the look up is already called within `checkSupported` (by `Charset.isSupported()`)
I am not sure why we need the member variable, or why we need to check if the variable is null.

v-xiangs: @marschall thanks for the very detailed explanation, now I understand why you have introduced a new member variable here. I wanted the code to be consistent with the current implementation, but your cache implementation is better than the current. I am going to accept the PR once the testing is done. thanks again.

v-xiangs:
@marschall  thank you very much

Issue #2, opened 2016-11-14
Wildcard in import statements


iamatinkerer:
In many files, the import statements use wildcard to import all the classes in the packages

E.g. https://github.com/Microsoft/mssql-jdbc/blob/master/src/main/java/com/microsoft/sqlserver/jdbc/KerbAuthentication.java
`import java.util.*;`

It is a good practice to specify the specific classes in these imports.

v-xiangs:
@iamatinkerer Thank you very much for pointing this out. You are right, it is a good practice to specify classes in imports. We really appreciate your participation and hope you could help us to fix it.

iamatinkerer:
@v-xiangs 
>hope you could help us to fix it.

Can I take this issue and send a PR then?

ajlam:
@iamatinkerer - we are working on getting a development branch created. Once that's available, you can send a PR there. Please stay tuned and thank you for your help 😄

v-ahibr:
@iamatinkerer 
We've just set up a dev branch. You can create a pull request there.

v-nisidh:
Closing this issue as it is fixed by #52

PR #31, opened 2016-11-25
Optimize imports

gstojsic:
This is in relation to **Wildcard in import statements (https://github.com/Microsoft/mssql-jdbc/issues/2)**. I used Intellij Idea's Optimize imports function to remove the wildcards. There are 4 files in the dev branch which I believe have been mistakenly committed with CRLF's instead of LF's so I corrected those too. These files are:
`KerbAuthentication.java`
`SQLServerConnectionPoolProxy.java`
`SQLServerDataSource.java`
`SQLServerPooledConnection.java`
I verified that these were committed with CRLF's by using:
`curl -o SQLServerConnectionPoolProxy.java -L https://github.com/Microsoft/mssql-jdbc/raw/dev/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnectionPoolProxy.java`
and then:
`hexdump -C SQLServerConnectionPoolProxy.java`
It is clearly visible that the hex dump contains 0d 0a sequences. Other files contain only 0a.


v-xiangs:
@gstojsic hello thank you for fixing both import issue and CRLF issue.
I am wondering if you could split the PR into 2 different PRs, one for each issue. That would be much easier for us to test it and accept the PR. Thanks.

gstojsic:
I'll do the CRLF issue first then.

v-xiangs:
@gstojsic thank you for your understanding, can I go ahead and close this PR then?

gstojsic:
sure

PR #32, opened 2016-11-25
Corrected CRLF to LF

gstojsic:
There are 4 files in the dev branch which I believe have been mistakenly committed with CRLF's instead of LF's so I corrected those. These files are:
KerbAuthentication.java
SQLServerConnectionPoolProxy.java
SQLServerDataSource.java
SQLServerPooledConnection.java
I verified that these were committed with CRLF's by using:
curl -o SQLServerConnectionPoolProxy.java -L https://github.com/Microsoft/mssql-jdbc/raw/dev/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnectionPoolProxy.java
and then:
hexdump -C SQLServerConnectionPoolProxy.java
It is clearly visible that the hex dump contains 0d 0a sequences. Other files contain only 0a.
The content of the files should essentially be the same. Please check this and let me know.
Thanks.


v-xiangs:
@gstojsic thank you very much for the fix, this is a great finding.

PR #34, opened 2016-11-26
Optimize imports

gstojsic:
This is in relation to **Wildcard in import statements**(#2). I used Intellij Idea's Optimize imports function to remove the wildcards. Also unused imports have been removed.


v-xiangs:
@gstojsic thank you for your contribution. Our team is currently working on removing unused imports and wildcards and will be checked in very soon.

PR #35, opened 2016-11-26
Remove unused imports

marschall:
Several classes contain imports that are not unused. They can be
removed without effect.


v-xiangs:
@marschall thank you for your contribution. Our team is currently working on removing unused imports and wildcards and will be checked in very soon.

marschall:
@v-xiangs that is great to hear, I'm looking forward to it. Will you also be fixing the indentation (currently it's a mix between spaces and tabs), remove trailing spaces and use a consistent method style (currently it's a mix between braces on the same line and braces on the next line) or should I file separate issues for these?

Issue #37, opened 2016-11-27
Jdbc driver not handling schema properly when using tvp parameters in stored procedures


gstojsic:
This issue is referred to [here](https://connect.microsoft.com/SQLServer/Feedback/Details/3102655). Essentially the driver throws an error complaining that Sql server could not find the proper TVP type. The test that illustrates the problem can be found [here](https://github.com/gstojsic/mssql-jdbc-test/blob/master/src/test/java/com/skunkworks/MssqlJdbcTestApplicationTests.java#L82).

v-xiangs:
@gstojsic The issue is fixed by PR #47 based on your idea. thank you

PR #38, opened 2016-11-27
Tvp schema error fix.

gstojsic:
This is in reference to issue (https://github.com/Microsoft/mssql-jdbc/issues/37). Essentially the `getParameterTypeName(int param)` method returns only the type name. This causes the error when the TVP is in a schema different than dbo. I have taken the `SS_TYPE_SCHEMA_NAME` parameter returned by the `sp_sproc_columns_100` stored procedure for the schema name and built the tvp name from that.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerParameterMetaData.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerParameterMetaData.java
@@ -784,7 +784,15 @@ public final class SQLServerParameterMetaData implements ParameterMetaData {
 			}
 			else {
 				verifyParameterPosition(param);
-				return rsProcedureMeta.getString("TYPE_NAME");
+				String schema = rsProcedureMeta.getString("SS_TYPE_SCHEMA_NAME");
+				String typeName = rsProcedureMeta.getString("TYPE_NAME");
+				String tvpName = null;
+				if (schema != null) {
+					tvpName = "[" + schema + "].[" + typeName + "]";
+				} else if (typeName != null) {
+					tvpName = "[" + typeName + "]";
+				}
+				return tvpName;
 			}
 		}
 		catch (SQLException e) {


v-xiangs:
@gstojsic thank you very much for your contribution. As a design decision, we decided to support TVP with `setStructured()` method only. So, TVP does not work with `setObject()`

The fix that you provided is a real fix for the case with stored procedure. We can get the TVP schema from stored procedure. However, TVP should work with Prepared Statement (without stored procedure) as well. With Prepared Statement, `setObject()` does not work with ResultSet or SQLServerDataTable.

gstojsic:
I did not really understand the point of your comment. The fix is really for the case with the stored procedure which now does not work properly because it takes only the type name into account without the schema. The fix does not affect the Prepared statement case. Did you want me to provide the fix for the Prepared statement also?

Also I did not understand the first part. TVP's are handled in setObject methods as I can see from the source code. For example: 
`com.microsoft.sqlserver.jdbc.SQLServerPreparedStatement#setObject(int, java.lang.Object)`
which is an implementation of `java.sql.PreparedStatement#setObject(int, java.lang.Object)` calls:
`com.microsoft.sqlserver.jdbc.SQLServerPreparedStatement#setObjectNoType`
which has a line:
`if (JavaType.TVP == javaType)`
that is obviously meant to handle TVP's. Did you mean you will be removing TVP handling from `setObject()` methods?

v-xiangs:
@gstojsic I have talked with our team here, we all think the fix that you provided is great and we really appreciate it. After testing is done on our side, we will accept your PR and make TVP support `setObject()` with stored procedure. Also, thanks for pointing out we have implemented `setObject()` to handle TVP in the source code, we will make some changes there, to throw exception explicitly when setObject() is used with ResultSet.

v-xiangs:
hello @gstojsic we very appreciate your idea on how to fix this issue, it is awesome. however, the fix is causing regression failures on our test lab. We will look into those regression failures and provide another fix based on your idea. For now, we are going to close this PR. Thank you again.

PR #39, opened 2016-11-27
Added automatic detection of REALM in SPN needed for Cross Domain authentication

pierresouchay:
The original driver only computes the SPN without its REALM. That is why the driver
fails in a cross-domain authentication (ex: user@REALM1 try to log on server@REALM2
while REALM2 and REALM1 are in trust).

This commit solves this by trying to compute the REALM when REALM has not been
provided in the SPN. Which includes both generated SPN and User-Provided SPN.

It also enable Kerberos authentication when only IP is provided as long as reverse
DNS are present since when SPN is provided, and REALM lookup did fail, it will
also try with canonical name and if it works, override the hostname in SPN (feature
not activated when user did provide an SPN)

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/KerbAuthentication.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/KerbAuthentication.java
@@ -19,11 +19,19 @@
  
 package com.microsoft.sqlserver.jdbc;
 import java.util.logging.*;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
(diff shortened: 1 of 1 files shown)


pierresouchay:
Note: Compatibility with IBM JVM could be trivially implemented using: com.ibm.security.jgss.mech.krb5.Krb5RealmUtil.mapHostToRealm( String )

(But I have not the interest nor the IBM JVM to test)

pierresouchay:
Replaced by https://github.com/Microsoft/mssql-jdbc/pull/40

PR #47, opened 2016-11-30
make setObject() handle TVP with schema rather than [odb], when TVP is used with stored procedure

v-xiangs:
special thanks to @gstojsic , thank you for providing the idea for us.


v-nisidh:
Adding References: This is in reference to issue ([#37](https://github.com/Microsoft/mssql-jdbc/issues/37)). @gstojsic proposed fixed in PR ([#38](https://github.com/Microsoft/mssql-jdbc/pull/38))

PR #50, opened 2016-12-01
Upgrade azure-keyvault to 0.9.7

JamieMagee:
Just a minor version bump fro `0.9.3` to `0.9.7`. The changelog can be found [here](https://github.com/Azure/azure-sdk-for-java/blob/0.9/ChangeLog.txt).

As far as I can tell the only changes are introducing caching, and a bugfix to do with JSON deserialization.


v-xiangs:
@thank you very much for the upgrade

PR #55, opened 2016-12-02
Update Maven plugins

JamieMagee:
Version bumps for `maven-compiler-plugin` and `maven-source-plugin`. The changelogs are mostly bugfixes for both.


v-xiangs:
@JamieMagee thank you for your contribution

PR #56, opened 2016-12-02
Remove final from private methods

JamieMagee:
Private methods can't be overridden, so declaring them final doesn't make sense


tonytamwk:
Good private final methods are meaningless in Java

v-xiangs:
@JamieMagee thank you for your contribution.

Issue #59, opened 2016-12-03
Set the jre8 version as default


gstojsic:
Currently issuing the command:
`mvn clean install`
for which most java developers expect that it should build something, issues an ugly build error. This is because a profile must be set in order for the build to succeed. Currently there are two profiles, one for jre8 and one for jre7. Consider making the jre8 default so that the above command builds the jre8 version by default.

v-xiangs:
fixed by PR #60 thank you

PR #60, opened 2016-12-03
Set the jre8 version as default

gstojsic:
this is in relation to issue #59

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
 	
 	<groupId>com.microsoft.sqlserver</groupId>
 	<artifactId>mssql-jdbc</artifactId>
-	<version>6.1.1</version>
+	<version>6.1.1-SNAPSHOT</version>
 	
 	<packaging>jar</packaging>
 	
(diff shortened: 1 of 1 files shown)


v-xiangs:
@gstojsic thank you very much for the great idea to have a default build. our team likes your idea a lot. just one suggestion, instead of moving profile `build42` to the <build> section, can we just add <activeByDefault> to the profile `build42` to make it default? thanks.

gstojsic:
Of course you could make the `build42` the active profile. In fact that was my first solution but then I decided for the submitted one because it requires less code (the xml stuff is already very noisy). This reduced the pom by 9 lines and the other solution would increase the noise and add several lines. Secondly, it is clearer for beginners to understand (you need not care about maven profiles if you are not interested in jre7 stuff to make sense of the pom). Also I believe that the solution I submitted is closer to the way maven builds were intended to work - try to use defaults as much as possible and specify overrides only when needed. Anyway, feel free to amend the pom as you see fit. Having the `build42` set as the active profile would also be a step forward IMHO.

v-xiangs:
@gstojsic thank you for your valuable thought. after talking to the team, we decided to keep 2 profiles because we think using profiles is better for maintenance when new JDK comes out.
However, it is a great idea to make jre8 the default profile. Do you mind to help us adding '<activeByDefault>' to 'build42' profile and keep anything else the same as it is? thank you.

gstojsic:
There you go. Thanks for considering the idea.

v-xiangs:
@gstojsic thank you for your contribution. We really appreciate all your ideas and fixes.

PR #63, opened 2016-12-05
Remove redundant type casts

JamieMagee:
Getting rid of redundant type casts makes the code a lot easier to read, and there is no downside to removing them.


v-xiangs:
@JamieMagee thank you for your contribution

PR #71, opened 2016-12-07
Test framework

v-nisidh:
Added Test Framework, so we can add JUnit test cases. 

@v-suhame Can you review changes?


v-nisidh:
@v-suhame As per our discussion I deleted DBCreationUtil.java

PR #73, opened 2016-12-07
Read SQL Server error message if status flag has DONE_ERROR set.

v-suhame:
When using client side cursors, if end of TDS packet is reached, check the status flag for error and handle it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,7 +16,6 @@ local.properties
 .classpath
 .settings/
 .loadpath
-*.class
 
 # External tool builders
 .externalToolBuilders/
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ local.properties
 .classpath
 .settings/
 .loadpath
+*.class
 
 # External tool builders
 .externalToolBuilders/
--- a/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
@@ -6844,6 +6844,17 @@ final class TDSReader
 		return currentPacket.payload[payloadOffset] & 0xFF;
 	}
 
+
+	final short peekStatusFlag() throws SQLServerException {
+		// skip the current packet(i.e, TDS packet type) and peek into the status flag (USHORT)
+		if (payloadOffset + 3 <= currentPacket.payloadLength) {
+			short value = Util.readShort(currentPacket.payload, payloadOffset + 1);
+			return value;
+		}
+
+		return Util.readShort(readWrappedBytes(3), 1);
+	}
+	
 	final int readUnsignedByte() throws SQLServerException
 	{
 		// Ensure that we have a packet to read from.
--- a/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
@@ -6852,7 +6852,10 @@ final class TDSReader
 			return value;
 		}
 
-		return Util.readShort(readWrappedBytes(3), 1);
+		// as per TDS protocol, TDS_DONE packet should always be followed by status flag
+		// throw exception if status packet is not available 
+		throwInvalidTDS();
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -16,6 +16,7 @@ local.properties
 .classpath
 .settings/
 .loadpath
+*.class
```
v-xiangs: I guess we can create a separate PR for this

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -6844,6 +6844,17 @@ final int peekTokenType() throws SQLServerException
 		return currentPacket.payload[payloadOffset] & 0xFF;
 	}
 
+
+	final short peekStatusFlag() throws SQLServerException {
+		// skip the current packet(i.e, TDS packet type) and peek into the status flag (USHORT)
+		if (payloadOffset + 3 <= currentPacket.payloadLength) {
```
v-xiangs: should we check it is not the end of payload using ensurePayload() ?

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -6844,6 +6844,17 @@ final int peekTokenType() throws SQLServerException
...
+			short value = Util.readShort(currentPacket.payload, payloadOffset + 1);
+			return value;
+		}
+
+		return Util.readShort(readWrappedBytes(3), 1);
+	}
+
```
v-xiangs: the method is called "peek", should we not use Util.readXXX() to get the payload in the case of "peek"? I guess with Util.readXXX() , it actually "pop" the data?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResultSet.java
```diff
@@ -327,13 +327,25 @@ boolean onError(TDSReader tdsReader) throws SQLServerException
...
+			boolean onDone(TDSReader tdsReader) throws SQLServerException {
+				// When initializing client-cursored ResultSets, a DONE token
+				// following the column metadata indicates an empty result set.
+				rowCount = 0;
+
+				// Continue to read the error message if DONE packet has error flag
+				int packetType = tdsReader.peekTokenType();
+				if (TDS.TDS_DONE == packetType) {
+					short status = tdsReader.peekStatusFlag();
+					if ((status & 0x0002) != 0) {
```
v-xiangs: can we add a comment to explain why we need this & 0x0002 ?

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
v-suhame: ensurePayload() makes sense only for packet type. Since we already have determined the packet type to be TDS_DONE, there must be a status flag that follows as per TDS standards.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResultSet.java
v-suhame: sure, will do that.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResultSet.java
```diff
@@ -327,13 +327,26 @@ boolean onError(TDSReader tdsReader) throws SQLServerException
...
-            }
+			boolean onDone(TDSReader tdsReader) throws SQLServerException {
+				// When initializing client-cursored ResultSets, a DONE token
+				// following the column metadata indicates an empty result set.
+				rowCount = 0;
+
+				// Continue to read the error message if DONE packet has error flag
+				int packetType = tdsReader.peekTokenType();
+				if (TDS.TDS_DONE == packetType) {
+					short status = tdsReader.peekStatusFlag();
```
v-suhame: Function of methods in TDSReader are restricted to read operation, it doesn't do any data validation, it's good to keep it that way.

PR #78, opened 2016-12-09
fix a bug when the value of queryTimeout is bigger than the max value of integer

v-xiangs:


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -1457,6 +1459,34 @@ else if (0 == requestedPacketSize)
...
+				try
+				{
+					int n = (new Integer(activeConnectionProperties.getProperty(sPropKey))).intValue();
+					if (n>=defaultTimeOut){
+						queryTimeoutSeconds = n;
+					}
+					else
+					{
+						MessageFormat form = new MessageFormat(SQLServerException.getErrString("R_invalidQueryTimeout"));
+						Object[] msgArgs = {activeConnectionProperties.getProperty(sPropKey)};
```
v-nisidh: I think in case of wrong querytimeout we should fallback to some valid querytimeout by logging error message.

v-xiangs: I think we should be consistent with our current implementation of LoginTimeout and LockTimeout. In both case of LoginTimeoutand and LockTimeout, we throw exceptions.

PR #84, opened 2016-12-10
Remove explicit boxing and unboxing

### LINKED ISSUE #13: Use static factory methods for wrapper types
The driver currently instantiates wrapper types (Integer, Long, …) by directly calling the constructor using `new` rather than by calling the static `valueOf` factory methods. This is discouraged and deprecated in Java 9. To quote from the [Integer constructor](http://download.java.net/java/jdk9/docs/api/java/lang/Integer.html#Integer-int-)

> It is rarely appropriate to use this constructor. The static factory valueOf(int) is generally a better choice, as it is likely to yield significantly better space and time performance.

JamieMagee:
Explicit boxing and unboxing in unnecessary since Java 5. See https://docs.oracle.com/javase/tutorial/java/data/autoboxing.html

This also fixes #13


ajlam:
@JamieMagee - thanks for the suggestion. We need to add a few more test cases for casting. We'll revisit this PR once they've been added.

v-xiangs:
Hello @JamieMagee , could you please restore your branch if possible? so we can reopen this PR. It was deleted unintendedly.

JamieMagee:
@v-xiangs Sure, I've restored it now.

v-xiangs:
@JamieMagee Thank you very much for the quick response.

JamieMagee:
I've updated the branch to resolve any merge conflicts.

v-xiangs:
@JamieMagee Thank you for your hard works !

peterbae:
Hi @JamieMagee, thanks for the contribution! A couple of suggestions:

1. In FailOverInfo.java, I think we could use Integer.parseInt instead of instantiating another Integer object (and then unboxing it into int again since portNumber is int).
2. For SqlTypeValue.java, could we try replacing all the tabs (since they're inconsistent across different machines) from lines 19 through 38 with spaces instead, and pad the spaces in between parameters (with spaces) so that the layout of the variables stay the same as before (as in, the 3 parameters for each data type would match the spacing of the minValue, maxValue and nullValue comments on line 19, respectively)? The removal of explicit autoboxing part is good. If the tabs look different on your machine and you're not sure about the original spacing, I can edit the file for you as well.

Other than those 2, the changes look good.

JamieMagee:
@peterbae Thanks for the feedback. I've been a little busy, most recently with oneweek, but I'll make the changes as soon as I have some time.

peterbae:
@JamieMagee Thanks for the reply. We wanted to include this in our next release so I've made the small formatting changes to one file. I will merge this PR after the tests have passed.

peterbae:
Test results are good. Merging in.

JamieMagee:
@peterbae Thanks for all your help with getting this PR merged 👍

Issue #67, opened 2016-12-06
Replace synchronized increment integer patterns with AtomicInteger


tonytamwk:
Simpler code with less synchronized and better performance

v-xiangs:
@tonytamwk thank you for your suggestion. We will consider it for future enhancement.

tonytamwk:
@v-xiangs I've done the changes. Can I have right to push?

ajlam:
@tonytamwk, can you please open a PR in the dev branch?

v-nisidh:
Closing this issue as #93 is merged back to dev 30 days back. 
Thanks again @tonytamwk for your contribution.

PR #93, opened 2016-12-27
Replace synchronized increment integer patterns with AtomicInteger

tonytamwk:
This PR fixes the issue #67


sehrope:
All of these static `AtomicInteger` counter should be `final`. The original `int` values were not final because they were being modified by the `nextXyzID()` methods. Since the `AtomicInteger` objects maintain their state internally, there's no reason for them to be non-`final`.

v-xiangs:
@tonytamwk thank you for your awesome contribution

PR #12, opened 2016-11-20
Implement missing MetaData #unwrap methods

marschall:
The driver currently implements some #unwrap methods but for MetaData

 * implement DatabaseMetaData#unwrap
 * implement ParameterMetaData#unwrap
 * implement ResultSetMetaData#unwrap
 * use for spaces for indentation where spaces are used
 * use tabs for indentation where tabs are used

PR #94, opened 2017-01-02
Implement missing MetaData #isWrapperFor methods

marschall:
The driver currently implements all #unwrap methods but some of the
matching #isWrapperFor methods are missing. This is caused by the fix
for #12 which should also have included the #isWrapperFor methods.

 - implement DatabaseMetaData#isWrapperFor
 - implement ParameterMetaData#isWrapperFor
 - implement ResultSetMetaData#isWrapperFor

### DIFF
--- a/src/test/java/com/microsoft/sqlserver/jdbc/resultset/ResultSetTest.java
+++ b/src/test/java/com/microsoft/sqlserver/jdbc/resultset/ResultSetTest.java
@@ -8,11 +8,14 @@
 package com.microsoft.sqlserver.jdbc.resultset;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertSame;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.junit.jupiter.api.Assertions.fail;
 
 import java.sql.Connection;
 import java.sql.DriverManager;
 import java.sql.ResultSet;
+import java.sql.SQLException;
 import java.sql.SQLFeatureNotSupportedException;
 import java.sql.Statement;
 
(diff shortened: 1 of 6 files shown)


v-afrafi:
Thank you @marschall  for submitting a pr. The changes in the three source code files are good. But we updated the tests to use the new framework, that's why a conflict exists in your pr now.
You could either remove bvt test updates from the pr or you could get the latest code and commit changes in the test with the new test framework. Thank you again.

marschall:
I committed the tests with the new test framework, the test that's failing is `testCreatepreparedStatement` which has no connection to my changes so I'm a bit confused.

src/test/java/com/microsoft/sqlserver/jdbc/resultset/ResultSetTest.java
```diff
@@ -73,4 +77,27 @@ public void testJdbc41ResultSetMethods() throws Exception {
...
+            
+            stmt.executeUpdate("create table " + tableName + " (col1 int, col2 text, col3 int identity(1,1) primary key)");
+            
+            try (ResultSet rs = stmt.executeQuery("select * from " + tableName)) {
+                assertTrue(rs.isWrapperFor(ResultSet.class));
+                assertTrue(rs.isWrapperFor(ISQLServerResultSet.class));
+
+                assertSame(rs, rs.unwrap(ResultSet.class));
+                assertSame(rs, rs.unwrap(ISQLServerResultSet.class));
+            }
```
v-afrafi: Please drop the tables before the test finishes. It will throw exception that object already exists.

v-afrafi:
@marschall  thank you. You are correct, that build fail was not because of your changes.  The next commits you make should not cause any failures.

src/test/java/com/microsoft/sqlserver/jdbc/resultset/ResultSetTest.java
marschall: That's right, `ParameterMetaDataTest` had the same issue.

v-afrafi:
Thank you for your contribution @marschall

PR #96, opened 2017-01-04
Add failover support without db mirroring configured

kentmc:
I would like to use the JDBC failoverPartner in an application that uses Azure SQL Databases.
Our scenario is this:

Set up Database X with geo-replication on primary server A, secondary B.
In JDBC, set server=A,failoverPartner=B
If server A goes down, stop the geo-replication, and application will automatically connect to B.

Azure SQL Databases do not support database mirroring, and thus they don't send metadata at connection time about their partner. JDBC expects this metadata to be received and throws an exception if not.

I modified the code to just use the database mirror if no failoverPartner is provided by the server it connects to. This might not be right per Database Mirroring specifications, but I would like to propose it anyways to be useful in other setups than database mirroring. Maybe a flag could be introduced that will allow it, or we could go with this and issue a warning if the server does not provide a partner.


pierresouchay:
We have the same issue here, I wanted to do a similar patch ;)

anosov1960:
You are correct, the failoverPartner is not supported in Azure SQL Database. To achieve this you will need transparent failover. This is not yet public but if you'd like immediate access to this functionality you can enroll into the private preview. 
Pls email to sashan at microsoft.com if interested.

ajlam:
Closing this PR since failoverPartner is set to release for Azure SQL Database.

PR #97, opened 2017-01-05
Fips

v-nisidh:
Add FIPS Support.


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResource.java
```diff
@@ -194,6 +194,7 @@ static String getResource(String key)
 				{"R_packetSizePropertyDescription", "The network packet size used to communicate with SQL Server."},
 				{"R_encryptPropertyDescription", "Determines if Secure Sockets Layer (SSL) encryption should be used between the client and the server."},
 				{"R_trustServerCertificatePropertyDescription", "Determines if the driver should validate the SQL Server Secure Sockets Layer (SSL) certificate."},
+				{"R_trustStoreTypePropertyDescription", "Type of trust store type like JKS / PKCS12 or nay FIPS implementation KeyStore Type. etc."},
```
v-mabarw: Are these the allowable types or are there others? http://docs.oracle.com/javase/8/docs/technotes/guides/security/StandardNames.html#KeyStore

Perhaps we should simply say 'Type of trust store, such as JKS or PKCS12.'

src/main/java/com/microsoft/sqlserver/jdbc/StringUtils.java
```diff
@@ -0,0 +1,41 @@
...
+		// Hiding constructor.
+	}
+
+	/**
+	 * Checks if String is null 
+	 * @param charSequence {@link CharSequence} Can provide null
+	 * @return {@link Boolean} if provided char sequence is null or empty / blank
+	 * @since 6.1.2
+	 */
+	public static boolean isEmpty(final CharSequence charSequence) {
```
v-mabarw: Is it worth it to add an entire class for the isEmpty implementation - how much use do we expect to get out of this in the future?

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -1823,6 +1843,50 @@ void enableSSL(String host, int port) throws SQLServerException
...
+			//Here different FIPS provider supports different KeyStore type along with different JVM Implementation. 
+			isValidTrustStoreType = !StringUtils.isEmpty(trustStoreType);
+			isTrustServerCertificate = con.trustServerCertificate();
+
+			if (isEncryptOn && isValidTrustStoreType && !isTrustServerCertificate) {
+				isFipsEnabled = true;
+			} else {
+				//If FIPS is not enabled issue warning and fall-back to non-FIPS mode.
+				StringBuilder sb = new StringBuilder();
+				sb.append("Could not switch to FIPS mode due to above settings: ");
```
v-mabarw: We cannot build strings directly in the implementation - it needs to be localizable to languages other than English.  Can we add a parameterized string resource and use that (you can see what is done for the error messages, for example)?

src/test/java/com/microsoft/sqlserver/jdbc/fips/FipsEnvTest.java
```diff
@@ -0,0 +1,182 @@
...
+	 */
+	@Test
+	public void testFIPSOnOracle() throws Exception {
+		assumeTrue(ORACLE_JVM.equals(currentJVM), "Aborting test: As this is not Oracle Env. ");
+
+		assumeTrue("FIPS".equals(Utils.getConfiguredProperty("FIPS_ENV")), "Aborting test case as FIPS is not enabled. "); 
+		
+		assertTrue(isFIPS("SunJSSE"),"FIPS Should be enabled");
+		
+//		assertTrue(isFIPS("BCFIPS"),"FIPS Should be enabled");
```
v-mabarw: What's this one for (is it needed still)?

src/test/java/com/microsoft/sqlserver/jdbc/fips/FipsEnvTest.java
```diff
@@ -0,0 +1,182 @@
...
+
+		assumingThat("BCFIPS".equals(Utils.getConfiguredProperty("FIPS_ENV")),
+				() -> assertAll("All FIPS", () -> assertTrue(isFIPS("SunJSSE"), "FIPS should be Enabled."), () -> assertTrue(isFIPS("BCFIPS"), "Testing")));
+		
+		assumingThat("FIPS".equals(Utils.getConfiguredProperty("FIPS_ENV")), ()-> assertTrue(isFIPS("SunJSSE"), "FIPS Should be enabled"));
+		*/
+	}
+
+	/**
+	 * It will test
```
v-mabarw: Finish comment :)

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResource.java
v-nisidh: BCFIPS supports both PKCS12 & BCFKS. 

Ref: https://www.bouncycastle.org/fips/BCFipsDescription-20150501.pdf [2.4.8 FIPS Compliant Key Store]

src/main/java/com/microsoft/sqlserver/jdbc/StringUtils.java
v-nisidh: In future we can add more functions if we are going to use this. As per need basis. This is kind of common utilities.  In future, we can add functions like add padding, masking etc. I am seeing this class as place holder for all common / global functions related to String operations.

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
v-nisidh: I will add localized warning message in coming sprint.

PR #99, opened 2017-01-06
fix for connection close on using variant type (issue#91)

v-suhame:


sualeh:
Please attach a jar file, since I cannot build. I will test it for you.

v-suhame:
Here is the jar [mssql-jdbc-6.1.1.jre8.zip](https://github.com/Microsoft/mssql-jdbc/files/688818/mssql-jdbc-6.1.1.jre8.zip).

sualeh:
Confirmed that this fixes the reported issue.

sualeh:
When is the next release expected?

sehrope:
Would it be possible to do something similar (i.e. skipping the part of the packet dealing with the sql_variant column) but rather than throw an exception, have the driver return a `null` value for the column? At the moment if there's any sql_variant columns it's impossible to fetch any part of the result set as the metadata phase of processing the result set throws an exception.

While it'd be great to have full support for real handling of sql_variant types, I'd settle for being able to at least skip them and have the driver return `null` values instead. Having it behind a connection property (ex: `sqlVariantAsNull`) would make the behavior explicit and insulate it from future changes when proper handling of sql_variant is eventually supported.

Alternatively, having the driver not break down unless the sql_variant column is actually fetched. Could still throw an exception but it'd be in the handling of the `ResultSet.getXYZ(...)` for sql_variant types (rather than the metadata parsing stage). That way a customer user program can look at the result set metadata and know to simply not fetch those columns.

v-suhame:
Thanks @sualeh for testing it. We are expecting to release a preview of the driver next week.

v-suhame:
@sehrope Can you please raise this as an issue? We will place it in backlog and consider it for enhancement. 

Both the alternatives you suggested will take almost equivalent effort, as driver has to understand the structure of variant received in order to skip or return null.

As for the second alternative, since SQL Server sends rows of data, irrespective of if the user want to fetches or skips the variant column, driver has to read the variant in the first row to fetch any other column in the second row. If driver is going to read/skip variant anyways, we can implement it to return null rather than throwing error as per your first suggestion.

sehrope:
>  If driver is going to read/skip variant anyways, we can implement it to return null rather than throwing error as per your first suggestion.

That'd be awesome. I created issue #101 for this.

PR #103, opened 2017-01-06
Resultset test

v-afrafi:
added unit ResultSet test


src/test/java/com/microsoft/sqlserver/jdbc/resultset/ResultSetTest.java
```diff
@@ -0,0 +1,84 @@
+package com.microsoft.sqlserver.jdbc.resultset;
```
v-suhame: Add the Licensing header to all the new files

PR #105, opened 2017-01-07
Test framework update

v-suhame:


src/test/java/com/microsoft/sqlserver/testframework/sqlType/SqlSmallDateTime.java
```diff
@@ -34,7 +34,7 @@
 import java.util.concurrent.ThreadLocalRandom;
 
 public class SqlSmallDateTime extends SqlDateTime {
-    static SimpleDateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss");
+    static SimpleDateFormat dateFormat = new SimpleDateFormat("yyyyMMdd'T'HH:mm:ss");
```
v-xiangs: just a question, why we changed to yyyyMMdd instead of yyyy-MM-dd?

v-suhame: yyyyMMdd is one of non DATEFORMAT dependent, i.e., SQL Server will always convert a valid value in this format without trowing conversion error. You can find other types here,
https://msdn.microsoft.com/en-us/library/ms180878.aspx#Anchor_1
PS: I have added it as a comment for the same in SqlTypeValue.java

PR #114, opened 2017-01-12
Updates to build scipt to exclude long running test cases

v-suhame:
Build script will now exclude long running testcases that are marked with 'slow' tag. User needs to comment out the exclude property to run entire test suite.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -58,18 +58,45 @@ repositories {
     mavenCentral()
 }
 
-dependencies {
-	compile 'com.microsoft.azure:azure-keyvault:0.9.7'
-	
-	testCompile 'junit:junit:4.12',
-	'org.junit.platform:junit-platform-console:1.0.0-M3',
-	'org.junit.platform:junit-platform-commons:1.0.0-M3',
-	'org.junit.platform:junit-platform-engine:1.0.0-M3',
-	'org.junit.platform:junit-platform-launcher:1.0.0-M3',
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -58,18 +58,45 @@ repositories {
...
 
+ext.junit4Version        = '4.12'
+ext.junitPlatformVersion = '1.0.0-M3'
+ext.junitJupiterVersion  = '5.0.0-M3'
+
+junitPlatform {
+	filters {
+		tags {
+			// comment out the exclude property to run all tests
+			exclude 'slow'
```
v-xiangs: by default, does it run all the tests or skip the slow tests?

pom.xml
```diff
@@ -214,6 +208,24 @@
 				</executions>
 			</plugin>
 			
+			<plugin>
+				<artifactId>maven-surefire-plugin</artifactId>
+				<version>2.19</version>
+				<configuration>
+					<properties>
+						<!-- Comment out the excludeTags to run all tests -->
+						<excludeTags>slow</excludeTags>
```
v-xiangs: by default, does it run all the tests or skip the slow tests?

build.gradle
v-suhame: by default it will skip all the testcases with tag 'slow'. As discussed, we can come up with better alternative after we finalize on build automation system.

v-xiangs:
closing this PR since new PR is created #425 . Thank you @v-suhame for the brilliant idea 👍

PR #115, opened 2017-01-13
Replace if with switch

worldtiki:
Replace if with switch for readability and performance reasons.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/DataTypes.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/DataTypes.java
@@ -639,9 +639,9 @@ enum JavaType
             return OffsetTime.class;
         case "OffsetDateTime":
             return OffsetDateTime.class;
+        default:
+            return null;
         }
-    	
-		return null;
     }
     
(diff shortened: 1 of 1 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/DataTypes.java
```diff
@@ -627,27 +627,19 @@ private JavaType(Class<?> javaClass, JDBCType jdbcTypeFromJavaType)
...
+            return LocalTime.class;
+        case "LocalDateTime":
+            return LocalDateTime.class;
+        case "OffsetTime":
+            return OffsetTime.class;
+        case "OffsetDateTime":
+            return OffsetDateTime.class;
+        }
     	
 		return null;
```
v-nisidh: SUGGESTION: We can move this `return null;`  in default case (switch block)

worldtiki: Thanks for the suggestion. Updated the pr with the change.

v-suhame:
Thanks for your contribution @worldtiki

PR #119, opened 2017-01-16
Binary and Varbinary types for framework

v-suhame:
Support for binary and varbinary type. 

The new files have same header as the rest of test framework, we can apply the changes throughout the source once header format is finalized.

### DIFF
--- a/src/test/java/com/microsoft/sqlserver/testframework/DBTable.java
+++ b/src/test/java/com/microsoft/sqlserver/testframework/DBTable.java
@@ -252,7 +252,7 @@ public class DBTable extends AbstractSQLGenerator {
             sb.add(OPEN_BRACKET);
             for (int colNum = 0; colNum < totalColumns; colNum++) {
 
-                // TODO: add betterway to enclose data
+                // TODO: consider how to enclose data in case of preparedStatemets
                 if (passDataAsString(colNum)) {
                     sb.add("'" + String.valueOf(getColumn(colNum).getRowValue(i)) + "'");
                 }
--- a/src/test/java/com/microsoft/sqlserver/testframework/DBTable.java
+++ b/src/test/java/com/microsoft/sqlserver/testframework/DBTable.java
@@ -35,6 +35,8 @@ import java.util.StringJoiner;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
+import org.apache.commons.codec.binary.Hex;
+
 import com.microsoft.sqlserver.testframework.sqlType.SqlType;
 import com.microsoft.sqlserver.testframework.sqlType.VariableLengthType;
 import com.microsoft.sqlserver.testframework.util.RandomUtil;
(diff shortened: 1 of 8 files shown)


src/test/java/com/microsoft/sqlserver/testframework/DBTable.java
```diff
@@ -251,16 +253,22 @@ String populateTableSql() {
             for (int colNum = 0; colNum < totalColumns; colNum++) {
 
                 // TODO: add betterway to enclose data
-                if (JDBCType.CHAR == getColumn(colNum).getSqlType().getJdbctype()
+                if (JDBCType.CHAR == getColumn(colNum).getSqlType().getJdbctype()
```
v-nisidh: Suggestion: Can we have one utility method to get JDBCType

````
public JDBCType getJDBCType(DBColumn column) {
 return column.getSqlType().getJdbcType();
}
```

src/test/java/com/microsoft/sqlserver/testframework/DBTable.java
```diff
@@ -251,16 +253,22 @@ String populateTableSql() {
...
                 // TODO: add betterway to enclose data
-                if (JDBCType.CHAR == getColumn(colNum).getSqlType().getJdbctype()
+                if (JDBCType.CHAR == getColumn(colNum).getSqlType().getJdbctype() 
                         || JDBCType.VARCHAR == getColumn(colNum).getSqlType().getJdbctype()
                         || JDBCType.NCHAR == getColumn(colNum).getSqlType().getJdbctype()
                         || JDBCType.NVARCHAR == getColumn(colNum).getSqlType().getJdbctype()
                         || JDBCType.TIMESTAMP == getColumn(colNum).getSqlType().getJdbctype()
                         || JDBCType.DATE == getColumn(colNum).getSqlType().getJdbctype()
-                        || JDBCType.TIME == getColumn(colNum).getSqlType().getJdbctype())
+                        || JDBCType.TIME == getColumn(colNum).getSqlType().getJdbctype()) {
```
v-nisidh: **SUGGESTION**: In order to reduce complexity it is better to create new method having all conditions and call that method in your if condition. 
If method is reusable then you can move this method to Util class. Otherwise create this helper method in same class.

src/test/java/com/microsoft/sqlserver/testframework/DBTable.java
v-suhame: ya, that sounds good.

src/test/java/com/microsoft/sqlserver/testframework/sqlType/SqlBinary.java
```diff
@@ -0,0 +1,49 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+// IN THE SOFTWARE.
+// ---------------------------------------------------------------------------------------------------------------------------------
+
+package com.microsoft.sqlserver.testframework.sqlType;
+
+import java.sql.JDBCType;
+import java.util.concurrent.ThreadLocalRandom;
+
```
v-nisidh: Can you add appropriate java docs?

PR #120, opened 2017-01-17
BVT updates

v-afrafi:


src/test/java/com/microsoft/sqlserver/testframework/sqlType/SqlType.java
```diff
@@ -180,4 +182,31 @@ void generatePrecision() {
...
     }
+
+    /**
+     * generate char or nchar values
+     * 
+     * @param columnLength
+     * @param charSet
+     * @return
+     */
+    protected static String buildCharOrNChar(int columnLength, String charSet) {
```
v-suhame: move these new methods into SqlChar class, as this is an abstract class and we are not going to use these methods in all the classes that extends SqlType.

v-afrafi: makes sense. will do

src/test/java/com/microsoft/sqlserver/jdbc/bvt/bvtTest.java
```diff
@@ -1,678 +1,587 @@
...
+@DisplayName("BVT Test")
+public class bvtTest extends bvtTestSetup {
+    private static String driverNamePattern = "Microsoft JDBC Driver \\d.\\d for SQL Server";
+    private static DBResultSet rs = null;
+    private static DBPreparedStatement pstmt = null;
+    private static DBConnection conn = null;
+    private static DBStatement stmt = null;
+
+    ///////////////////////////////////////////////////////////////////
+    //// Connect to specified server and close the connection
```
v-suhame: can we change these comments into javadocs?

src/test/java/com/microsoft/sqlserver/jdbc/bvt/bvtTest.java
```diff
@@ -1,678 +1,587 @@
...
+        finally {
+
+            terminateVariation();
+        }
+    }
 
     ///////////////////////////////////////////////////////////////////
     // Create a statement with a query timeout
     // ResultSet.Type_forward_only,
     // ResultSet.CONCUR_READ_ONLY, executeQuery
```
v-suhame: update the comments as this test doesn't use ResultSet. Also review other comments too.

src/test/java/com/microsoft/sqlserver/jdbc/bvt/bvtTest.java
```diff
@@ -1,678 +1,587 @@
...
+    // ResultSet.Type_forward_only,
+    // ResultSet.CONCUR_READ_ONLY, executeQuery
+    // verify cursor by using next and previous and verify data
+    ///////////////////////////////////////////////////////////////////
+    @Test
+    public void testStmtForwardOnlyReadOnly() throws SQLException, ClassNotFoundException {
+
+        try {
+            conn = new DBConnection(connectionString);
+            stmt = conn.createStatement(ResultSet.TYPE_FORWARD_ONLY, ResultSet.CONCUR_READ_ONLY);
```
v-suhame: Do you think it will be better to have an enum in DBStatement/DBConnection that has all combinations of resultSetType and resultSetConcurrency? So we don't have to call and set each type (will be useful to reuse in other tests), and just have validation in btvTest

src/test/java/com/microsoft/sqlserver/testframework/DBTable.java
```diff
@@ -49,15 +49,16 @@
     String escapedTableName;
     List<DBColumn> columns;
     int totalColumns;
-    int totalRows = 2; // default row count set to 2
+    static int totalRows = 3; // default row count set to 3
     DBSchema schema;
+    ArrayList<ArrayList<Object>> rows;
```
v-suhame: any reason why we need rows and currentrow? wont the columns suffice?

v-afrafi: rows is all rows we need to verify our backend data. verification is done row by row.

src/test/java/com/microsoft/sqlserver/jdbc/bvt/bvtTest.java
v-afrafi: ya. will do.

src/test/java/com/microsoft/sqlserver/testframework/DBResultSet.java
```diff
@@ -85,4 +110,327 @@ public Object getObject(int index) throws SQLException {
...
+
+    /**
+     * 
+     * @param ordinal
+     * @param coercion
+     * @param arg
+     * @throws SQLException
+     */
+    public void verifydata(int ordinal, Class coercion, Object arg) throws SQLException {
+        Object backendData = this.currentrow().get(ordinal);
```
v-suhame: The name backendData gives an impression that it is actual data from DB, when it is actually expected data. It would be better to rename this variable.

src/test/java/com/microsoft/sqlserver/jdbc/bvt/bvtTest.java
```diff
@@ -1,678 +1,587 @@
...
+            rs1 = stmt1.executeQuery(query);
+            rs1.currentTable = table1;
+
+            String query2 = "SELECT * FROM " + table2.getEscapedTableName();
+            rs2 = stmt2.executeQuery(query2);
+            rs2.currentTable = table2;
+
+            // Interleave resultset calls
+            rs1.next();
+            rs1.verifyCurrentRow(table1);
```
v-suhame: currentTable variable in DBResultSet can be made protected, it will be less confusing. in line #429 currentTable is set and the same value is passed again in  DBResultSet#verifyCurrentRow (line #437)

src/test/java/com/microsoft/sqlserver/jdbc/bvt/bvtTest.java
```diff
@@ -1,678 +1,587 @@
...
+            conn = new DBConnection(connectionString);
+            stmt = conn.createStatement();
+
+            String query = "SELECT * FROM " + table1.getEscapedTableName();
+            rs = stmt.executeQuery(query);
+            rs.currentTable = table1;
+
+            rs.verify(table1);
+            stmt.close();
+        }
```
v-suhame: we should verify if the stmt/rs is closed here else this test will be same as the previous one.

src/test/java/com/microsoft/sqlserver/testframework/DBTable.java
v-suhame: all the expected values must be available in 
`List<DBColumn> columns `

try accessing them as
`columns.get(Columnindex).getRowValue(row);`

src/test/java/com/microsoft/sqlserver/jdbc/bvt/bvtTestSetup.java
```diff
@@ -0,0 +1,57 @@
...
+            if (null != stmt) {
+                stmt.close();
+            }
+            if (null != conn && !conn.isClosed()) {
+                conn.close();
+            }
+        }
+    }
+
+}
```
v-suhame: add a method with @AfterAll tag to clean up the 2 tables.

src/test/java/com/microsoft/sqlserver/testframework/DBResultSet.java
```diff
@@ -85,4 +117,325 @@ public Object getObject(int index) throws SQLException {
...
+
+        // getXXX - default mapping
+        Object retrieved = this.getXXX(ordinal + 1, coercion);
+
+        // Verify
+        verifydata(ordinal, coercion, expectedData, retrieved);
+    }
+
+    public void verifydata(int ordinal, Class coercion, Object expectedData, Object retrieved) throws SQLException {
+        metaData = this.getMetaData();
```
v-suhame: add javadocs for this method

src/test/java/com/microsoft/sqlserver/testframework/DBResultSet.java
```diff
@@ -85,4 +117,325 @@ public Object getObject(int index) throws SQLException {
...
+
+    /**
+     * 
+     * @param ordinal
+     * @param coercion
+     * @param arg
+     * @throws SQLException
+     */
+    public void verifydata(int ordinal, Class coercion, Object arg) throws SQLException {
+        Object expectedData = currentTable.columns.get(ordinal).getRowValue(_currentrow);
```
v-suhame: can we remove the arg variable as its not used

src/test/java/com/microsoft/sqlserver/testframework/DBResultSet.java
```diff
@@ -85,4 +117,325 @@ public Object getObject(int index) throws SQLException {
...
+                            " unexpected BINARY value, expected");
+        }
+        return true;
+    }
+
+    private Object getXXX(int idx, Class coercion) throws SQLException {
+        if (coercion == Object.class) {
+            return this.getObject(idx);
+        }
+        return null;
```
v-suhame: can we have warning logged here if there is no valid coercion? It will be easier to trace if we try to test with unimplemented coercion

src/test/java/com/microsoft/sqlserver/testframework/DBResultSet.java
v-suhame: or if you have any future TODOs for this variable add it as comment

src/test/java/com/microsoft/sqlserver/testframework/DBResultSet.java
```diff
@@ -85,4 +117,325 @@ public Object getObject(int index) throws SQLException {
...
+                        " unexpected DATETIMEOFFSET value, expected: " + expectedData + " ,received: " + retrieved);
+                break;
+
+            case java.sql.Types.BINARY:
+                assertTrue(parseByte((byte[])expectedData, (byte[])retrieved),
+                        " unexpected BINARY value, expected: " + expectedData + " ,received: " + retrieved);
+                break;
+
+            case java.sql.Types.VARBINARY:
+                assertTrue(parseByte((byte[])expectedData, (byte[])retrieved),
```
v-suhame: Can we have Arrays.equals() here instead of implementing parseByte()?

src/test/java/com/microsoft/sqlserver/testframework/DBResultSetTypes.java
```diff
@@ -0,0 +1,33 @@
...
+ *
+ */
+
+public enum DBResultSetTypes {
+
+    TYPE_FORWARD_ONLY_CONCUR_READ_ONLY(DBResultSet.TYPE_FORWARD_ONLY, DBResultSet.CONCUR_READ_ONLY), TYPE_SCROLL_INSENSITIVE_CONCUR_READ_ONLY(
+            DBResultSet.TYPE_SCROLL_INSENSITIVE, DBResultSet.CONCUR_READ_ONLY), TYPE_SCROLL_SENSITIVE_CONCUR_READ_ONLY(
+                    DBResultSet.TYPE_SCROLL_SENSITIVE, DBResultSet.CONCUR_READ_ONLY), TYPE_FORWARD_ONLY_CONCUR_UPDATABLE(
+                            DBResultSet.TYPE_FORWARD_ONLY, DBResultSet.CONCUR_UPDATABLE), TYPE_SCROLL_SENSITIVE_CONCUR_UPDATABLE(
+                                    DBResultSet.TYPE_SCROLL_SENSITIVE, DBResultSet.CONCUR_UPDATABLE), TYPE_DYNAMIC_CONCUR_OPTIMISTIC(
```
v-suhame: It will increase the readability if we don't use our formatter here!

src/test/java/com/microsoft/sqlserver/testframework/DBPreparedStatement.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    /**
+     * @param i
+     * @param bigDecimal
+     * @throws SQLException
+     */
+    public void setObject(int parameterIndex, Object targetObject) throws SQLException {
+
+        ((PreparedStatement) product()).setObject(parameterIndex, targetObject);
```
v-suhame: these javadocs needs to be updated

src/test/java/com/microsoft/sqlserver/jdbc/bvt/bvtTest.java
```diff
@@ -1,678 +1,600 @@
...
- * This program is made available under the terms of the MIT License.
- * See the LICENSE file in the project root for more information.
+ * This program is made available under the terms of the MIT License. See the LICENSE file in the project root for more information.
  */
 package com.microsoft.sqlserver.jdbc.bvt;
+
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
-import static org.junit.Assert.fail;
+import static org.junit.jupiter.api.Assertions.fail;
```
v-suhame: can you also use jupiter.api.Assertions for assertEquals and assertTrue

src/test/java/com/microsoft/sqlserver/jdbc/bvt/bvtTest.java
```diff
@@ -1,678 +1,600 @@
...
+
+            stmt.close(); // this should close the resultSet
+            try {
+                rs.next();
+            }
+            catch (SQLException e) {
+                assertEquals(e.toString(), "com.microsoft.sqlserver.jdbc.SQLServerException: The result set is closed.");
+            }
+        }
+        finally {
```
v-suhame: it will be good if we can also have an assert after catch, to verify code execution dint reach that part of code.

v-afrafi: the code will get executed after catch. we are just catching it and inside catch make sure that it has thrown the proper exception.

src/test/java/com/microsoft/sqlserver/jdbc/bvt/bvtTest.java
```diff
@@ -1,678 +1,600 @@
...
+
+        try {
+            stmt = conn.createStatement();
+        }
+        finally {
+            terminateVariation();
+        }
+    }
+
+    public static void terminateVariation() throws SQLException {
```
v-suhame: both the terminate method needs javadoc

src/test/java/com/microsoft/sqlserver/testframework/DBConnection.java
```diff
@@ -28,71 +28,110 @@
...
+
+    /**
+     * 
+     * @param query
+     * @return
+     * @throws SQLException
+     */
+    public DBPreparedStatement prepareStatement(String query) throws SQLException {
+        DBPreparedStatement dbpstmt = new DBPreparedStatement(this, internal, "preparedStatement");
+        return dbpstmt.prepareStatement(query);
```
v-suhame: the variable internal passed here must be preparedStatement object. Can we have DBPreparedStatement(DBConnection dbConnection) constructor in DBPreparedStatement like the one in DBStatement? it will reduce the confusion

src/test/java/com/microsoft/sqlserver/testframework/DBResultSet.java
```diff
@@ -25,27 +25,58 @@
...
 
     // TODO: add cursors
     // TODO: add resultSet level holdability
     // TODO: add concurrency control
+
+    public static final int TYPE_DYNAMIC = ResultSet.TYPE_SCROLL_SENSITIVE + 1;
+    public static final int CONCUR_OPTIMISTIC = ResultSet.CONCUR_UPDATABLE + 2;
+    public static final int TYPE_CURSOR_FORWARDONLY = ResultSet.TYPE_FORWARD_ONLY + 1001;
+    public static final int TYPE_FORWARD_ONLY = ResultSet.TYPE_FORWARD_ONLY;
+    public static final int CONCUR_READ_ONLY = ResultSet.CONCUR_READ_ONLY;
```
v-suhame: can you add some documentation on why some values are added?

src/test/java/com/microsoft/sqlserver/testframework/DBResultSet.java
```diff
@@ -85,4 +117,325 @@ public Object getObject(int index) throws SQLException {
...
+    /**
+     * 
+     * @return
+     */
+    public DBResultSetMetaData getMetaData() {
+        ResultSetMetaData product = null;
+        DBResultSetMetaData wrapper = null;
+        try {
+            product = resultSet.getMetaData();
+            wrapper = new DBResultSetMetaData(parent, product, name);
```
v-suhame: update the constructor in `DBResultSetMetaData `to accept just `DBResultSet`, to be consistent with other classes that extend `AbstractParentWrapper`. Also, pass valid value for both `parent `and `name `variable, they are `null `at this point.

src/test/java/com/microsoft/sqlserver/testframework/DBResultSetMetaData.java
```diff
@@ -0,0 +1,198 @@
...
+
+    /**
+     * 
+     * @param index
+     * @return
+     * @throws SQLException
+     */
+    public int getColumnType(int index) throws SQLException {
+        int current = ((SQLServerResultSetMetaData) product()).getColumnType(index);
+        return current;
```
v-suhame: can we directly return the value whenever possible rather than storing them in temp variable and then passing them? this way value will not be stored in memory while waiting for GC.

src/test/java/com/microsoft/sqlserver/testframework/DBStatement.java
```diff
@@ -51,6 +51,10 @@
         super(dbConnection, null, "statement");
     }
 
+    DBStatement(DBConnection parent, Statement internal, int type, int concurrency, int holdability) {
+        super(parent, null, "statement");
+    }
```
v-suhame: this constructor can be removed as we are not using any of the new parameters passed

src/test/java/com/microsoft/sqlserver/testframework/DBTable.java
```diff
@@ -125,14 +125,14 @@ public String getEscapedTableName() {
      * 
      * @return total rows in the table
      */
-    public int getTotalRows() {
+    public static int getTotalRows() {
         return totalRows;
```
v-suhame: why is totalRows as static? Shouldn't the number of rows be tied to a table!

src/test/java/com/microsoft/sqlserver/testframework/sqlType/SqlDateTimeOffset.java
```diff
@@ -25,74 +25,137 @@
...
+    }
+
+    public Object generateDatetimeoffset(Integer precision) {
+        if (null == precision) {
+            precision = 7;
+        }
+
+        DateTimeOffset maxDTS = calculateDateTimeOffsetMinMax("max", precision, "9999-12-31 23:59:59");
+        DateTimeOffset minDTS = calculateDateTimeOffsetMinMax("min", precision, "0001-01-01 00:00:00");
+
```
v-suhame: replace the hard coded value with SqlTypeValue.DATETIMEOFFSET.maxValue/minValue

v-suhame: It will be better to move these calls into SqlDateTimeOffset() and store them in minvalue and maxvalue. So we don't have to calculate the same value over and over while generating random datetimeoffset value

src/test/java/com/microsoft/sqlserver/testframework/sqlType/SqlType.java
```diff
@@ -39,6 +40,7 @@
     protected Object maxvalue = null;
     protected Object nullvalue = null;	// Primitives have non-null defaults
     protected VariableLengthType variableLengthType;
+    static Random r = new Random();
```
v-suhame: lets use ThreadLocalRandom instead

src/test/java/com/microsoft/sqlserver/jdbc/bvt/bvtTest.java
v-suhame: in case the exception is not thrown, the test will pass, we don't want that to happen.

PR #127, opened 2017-01-26
BulkCopy ColumnMapping test

v-ahibr:
Add junit test cases for bulkCopy column mapping


src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyColumnMappingTest.java
```diff
@@ -0,0 +1,473 @@
...
+            
+    }
+    
+    /**
+     * validate if both expected and actual value are same  
+     * @param dataType
+     * @param expectedValue
+     * @param actualValue
+     */
+    private void compareSourceDest(int dataType, Object expectedValue, Object actualValue){
```
v-afrafi: can we use the same method in BulkCopyTestUtil?

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyTestUtil.java
```diff
@@ -86,6 +174,137 @@ static void performBulkCopy(BulkCopyTestWrapper wrapper, DBTable sourceTable, bo
...
+        }
+        catch (SQLException ex) {
+            if(!fail){
+                fail(ex.getMessage());
+            }
+        }
+        finally {
+            if(dropDest){
+                stmt.dropTable(destinationTable);
+                con.close();
```
v-afrafi: let's close con and stmt out of if( dropDest ) check.

v-ahibr: The if(dropDest) condition is because some destination tables require further validation after performBulkCopy method.

i've moved con.close() to outside the if statement

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyColumnMappingTest.java
v-ahibr: yes it can be used

PR #138, opened 2017-02-07
Tvp name with stored procedure

v-afrafi:
instead of the public API setTVPName

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDataTable.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDataTable.java
@@ -257,11 +257,9 @@ public final class SQLServerDataTable {
 
     /**
      * Retrieves the column meta data of this data table.
-     * @deprecated
      * @param tvpName
      *            the name of TVP
      */
-    @Deprecated
     public void setTvpName(String tvpName) {
         this.tvpName = tvpName;
     }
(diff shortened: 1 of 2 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDataTable.java
```diff
@@ -250,17 +250,18 @@ else if (val instanceof OffsetTime)
...
-
+    
     public String getTvpName() {
         return tvpName;
     }
 
     /**
      * Retrieves the column meta data of this data table.
-     * 
+     * @deprecated
```
v-nisidh: Should give details like -> `@deprecated  As of release 6.1.x, replaced by {@link #[YOUR_METHOD]}` 
Also give reasons for deprecation which will useful for all contributors.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -2136,7 +2136,8 @@ public final void setStructured(int n,
     String getTVPNameIfNull(int n,
             String tvpName) throws SQLServerException {
         if ((null == tvpName) || (0 == tvpName.length())) {
```
v-nisidh: You can use `StringUtils.isEmpty()` method.

PR #146, opened 2017-02-14
AAD password on non windows

v-xiangs:


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -3450,83 +3467,112 @@ private SqlFedAuthToken getFedAuthToken(SqlFedAuthInfo fedAuthInfo) throws SQLSe
...
-        long expirationFileTime = 0;
-
         while (true) {
             numberOfAttempts++;
 
-            try {
-                if (authenticationString.trim().equalsIgnoreCase(SqlAuthentication.ActiveDirectoryPassword.toString())) {
-                    dllInfo = AuthenticationJNI.getAccessToken(user, password, fedAuthInfo.stsurl, fedAuthInfo.spn, clientConnectionId.toString(),
-                            ActiveDirectoryAuthentication.jdbcFedauthClientId, expirationFileTime);
+            if (authenticationString.trim().equalsIgnoreCase(SqlAuthentication.ActiveDirectoryPassword.toString())) {
```
v-nisidh: Better to have constant on left side while comparing.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -1270,8 +1285,8 @@ Connection connectInternal(Properties propsIn,
             }
 
             if ((!System.getProperty("os.name").toLowerCase().startsWith("windows"))
-                    && (!authenticationString.equalsIgnoreCase(SqlAuthentication.NotSpecified.toString()))) {
-                throw new SQLServerException(SQLServerException.getErrString("R_FedAuthOnNonWindows"), null);
+                    && (authenticationString.equalsIgnoreCase(SqlAuthentication.ActiveDirectoryIntegrated.toString()))) {
```
v-nisidh: in getFedAuthToken(..) we have similar check of ActiveDirectoryPassword. There we used authenticationString.trim(). Is there any possibility that we are getting authenticationString as null?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -3450,83 +3467,112 @@ private SqlFedAuthToken getFedAuthToken(SqlFedAuthInfo fedAuthInfo) throws SQLSe
...
                 }
+                catch (ExecutionException e) {
+                    MessageFormat form = new MessageFormat(SQLServerException.getErrString("R_ADALExecution"));
+                    Object[] msgArgs = {user, authenticationString};
 
-                // AccessToken should not be null.
-                assert null != dllInfo.accessTokenBytes;
+                    // the cause error message uses \\n\\r which does not give correct format
+                    // change it to \r\n to provide correct format
+                    String correctedErrorMessage = e.getCause().getMessage().replaceAll("\\\\r\\\\n", "\r\n");
```
v-nisidh: Do we have any Negative JUnit Test cases ? Which will expect some exception / message / cause.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -3450,83 +3467,112 @@ private SqlFedAuthToken getFedAuthToken(SqlFedAuthInfo fedAuthInfo) throws SQLSe
...
             numberOfAttempts++;
 
-            try {
-                if (authenticationString.trim().equalsIgnoreCase(SqlAuthentication.ActiveDirectoryPassword.toString())) {
-                    dllInfo = AuthenticationJNI.getAccessToken(user, password, fedAuthInfo.stsurl, fedAuthInfo.spn, clientConnectionId.toString(),
-                            ActiveDirectoryAuthentication.jdbcFedauthClientId, expirationFileTime);
+            if (authenticationString.trim().equalsIgnoreCase(SqlAuthentication.ActiveDirectoryPassword.toString())) {
+                ExecutorService executorService = Executors.newFixedThreadPool(1);
+                try {
+                    AuthenticationContext context = new AuthenticationContext(fedAuthInfo.stsurl, false, executorService);
```
v-nisidh: SUGGESTION: May be we can have `ADALAuth.getAccesstoken()` in which we can create AuthenctioanContext and acquireToken.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
v-xiangs: no, the default is NotSpecified, it cannot be null;

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
v-xiangs: good suggestion. It's done now, I have used modular approach now

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
v-xiangs: yes, we have tests for AAD error messages

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
v-xiangs: I totally agree, but there are a lot of places to change in the driver, maybe we can have a separate PR for this?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerADAL4JUtils.java
```diff
@@ -0,0 +1,55 @@
...
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+
+import com.microsoft.aad.adal4j.AuthenticationContext;
+import com.microsoft.aad.adal4j.AuthenticationException;
+import com.microsoft.aad.adal4j.AuthenticationResult;
+import com.microsoft.sqlserver.jdbc.SQLServerConnection.ActiveDirectoryAuthentication;
+import com.microsoft.sqlserver.jdbc.SQLServerConnection.SqlFedAuthInfo;
+
+class SQLServerADAL4JUtils {
```
v-nisidh: Are you sure you do not want public class? By default this will use package level visibility.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -146,46 +136,24 @@ SqlFedAuthToken getAuthenticationResult() {
...
 
-    private class ActiveDirectoryAuthentication {
-        private static final String jdbcFedauthClientId = "7f98cb04-cd1e-40df-9140-3bf7e2cea4db";
-        private static final String AdalGetAccessTokenFunctionName = "ADALGetAccessToken";
-        private static final int GetAccessTokenSuccess = 0;
-        private static final int GetAccessTokenInvalidGrant = 1;
-        private static final int GetAccessTokenTansisentError = 2;
-        private static final int GetAccessTokenOtherError = 3;
+    class ActiveDirectoryAuthentication {
+        static final String jdbcFedauthClientId = "7f98cb04-cd1e-40df-9140-3bf7e2cea4db";
```
v-nisidh: final static should be in CAPS

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerADAL4JUtils.java
v-xiangs: why would we want it to be public?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
v-xiangs: done

Issue #17, opened 2016-11-21
azure-keyvault (mssql-jdbc's dependencies)


martinm1000:
Not sure if this is alright, but I needed to **exclude azure-keyvault** since I do not use it, and this dependency was adding a lot more of its own deps to my final build.

Can you clarify those dependencies ? Maybe add this to the project README ? Thanks !

```
<dependency>
    <groupId>com.microsoft.sqlserver</groupId>
    <artifactId>mssql-jdbc</artifactId>
    <version>6.1.0.jre8</version>
    <exclusions>
        <exclusion>
            <groupId>com.microsoft.azure</groupId>
            <artifactId>azure-keyvault</artifactId>
        </exclusion>
    </exclusions>
</dependency>
```

**Dependency tree report:**

```
[INFO] | +- com.microsoft.sqlserver:mssql-jdbc:jar:6.1.0.jre8:compile
[INFO] |  \- com.microsoft.azure:azure-keyvault:jar:0.9.3:compile
[INFO] |     +- com.microsoft.azure:azure-core:jar:0.9.3:compile
[INFO] |     |  +- commons-codec:commons-codec:jar:1.10:compile
[INFO] |     |  +- commons-lang:commons-lang:jar:2.6:compile
[INFO] |     |  +- javax.mail:mail:jar:1.4.5:compile
[INFO] |     |  |  \- javax.activation:activation:jar:1.1:compile
[INFO] |     |  +- com.sun.jersey:jersey-client:jar:1.13:compile
[INFO] |     |  |  \- com.sun.jersey:jersey-core:jar:1.13:compile
[INFO] |     |  \- com.sun.jersey:jersey-json:jar:1.13:compile
[INFO] |     |     +- org.codehaus.jettison:jettison:jar:1.1:compile
[INFO] |     |     |  \- stax:stax-api:jar:1.0.1:compile
[INFO] |     |     +- com.sun.xml.bind:jaxb-impl:jar:2.2.3-1:compile
[INFO] |     |     |  \- javax.xml.bind:jaxb-api:jar:2.2.2:compile
[INFO] |     |     |     \- javax.xml.stream:stax-api:jar:1.0-2:compile
[INFO] |     |     +- org.codehaus.jackson:jackson-core-asl:jar:1.9.2:compile
[INFO] |     |     +- org.codehaus.jackson:jackson-mapper-asl:jar:1.9.2:compile
[...]

PR #148, opened 2017-02-15
make Azure Key Vault and Azure Active Directory Authentication Dependencies optional

v-xiangs:


v-xiangs:
it's also for issue #17

README.md
```diff
@@ -87,7 +87,8 @@ The driver can be downloaded from the [Microsoft Download Center](https://www.mi
 This project has following dependencies: 
 
 Compile Time:
- - `azure-keyvault` : Azure Key Vault Provider for Always Encrypted feature
+ - `azure-keyvault` : Azure Key Vault Provider for Always Encrypted feature (optinal)
```
ajlam: small typo "optional"

README.md
```diff
@@ -98,21 +99,22 @@ One can see all dependencies including Transitive Dependency by executing follow
 mvn dependency:tree
 ```
 
-###Exclude Dependencies
-If you wish to limit the number of run-time dependencies, and your project does not require the features named above, you can explicitly exclude them by adding exclusion tag.  
-***For Example:*** If you are not using *Always Encrypted Azure Key Vault feature* then you can exclude *azure-keyvault* dependency. Please see following snippet. 
+### Azure Key Vault and Azure Active Directory Authentication Dependencies
+Since the dependencies of ***Azure Key Vault*** and ***Azure Active Directory Authentication*** are optional, projects that require those 2 features need to redeclare the dependency explicitly in their pom files.
```
ajlam: Let's just say "projects that require either of the two features need to explicitly declare the dependency in their pom file".

Issue #144, opened 2017-02-11
Query Cancellation Hangs When Using Multi Subnet Failover


sehrope:
Normally we use the standard `Statement.cancel()` method to cancel running queries. If it's called on a connection made with `multiSubnetFailover=true`, it has no effect and both the original Thread executing the `Statement` and the Thread issuing the `cancel()` will hang until the statement completes.

This issue been tested and confirmed on the latest stable SQL Server JDBC driver as well as a number of older drivers (prior to the 6.x branch).

See below for a self-contained example of this. Replace the JDBC url with any recent version of SQL Server and flip the boolean `multiSubnetFailover` property between true/false to see the impact. The sample code issues a dummy `WAITFOR` statement to simulate a slow query and then tries to cancel it in a separate thread. If `multiSubnetFailover` is enabled, the cancellation never occurs as the canceler thread blocks waiting to write to the nio Channel (probably trying to write the ATTENTION cancellation packet).

Stack trace for thread: `stmt-canceller`

    java.nio.channels.Channels.writeFully(Channels.java:96)
    java.nio.channels.Channels.access$000(Channels.java:61)
    java.nio.channels.Channels$1.write(Channels.java:174)
    com.microsoft.sqlserver.jdbc.TDSChannel$ProxyOutputStream.writeInternal(IOBuffer.java:1059)
    com.microsoft.sqlserver.jdbc.TDSChannel$ProxyOutputStream.write(IOBuffer.java:1051)
    sun.security.ssl.OutputRecord.writeBuffer(OutputRecord.java:431)
    sun.security.ssl.OutputRecord.write(OutputRecord.java:417)
    sun.security.ssl.SSLSocketImpl.writeRecordInternal(SSLSocketImpl.java:876)
    sun.security.ssl.SSLSocketImpl.writeRecord(SSLSocketImpl.java:847)
    sun.security.ssl.AppOutputStream.write(AppOutputStream.java:123)
    com.microsoft.sqlserver.jdbc.TDSChannel.write(IOBuffer.java:1887)
    com.microsoft.sqlserver.jdbc.TDSWriter.flush(IOBuffer.java:4285)
    com.microsoft.sqlserver.jdbc.TDSWriter.writePacket(IOBuffer.java:4186)
    com.microsoft.sqlserver.jdbc.TDSWriter.endMessage(IOBuffer.java:3192)
[...]

PR #160, opened 2017-02-27
Convert socket adaptor to socket

v-xiangs:
for issue #144

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
@@ -2552,12 +2552,14 @@ final class SocketFinder {
             }
         }
 
-        // if a channel was selected, make the necessary updates
+     // if a channel was selected, make the necessary updates
         if (selectedChannel != null) {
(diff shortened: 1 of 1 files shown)


sehrope:
I tried out our test case and it fixes the cancellation issue when multiSubnetFailover is enabled. It no longer hangs and the cancellation happens correctly. Very nice.

Skimming through the implementation, it looks like the approach is to close the socket channel and reconnect it as a regular socket. That should be be fine though it does lead to the situation that all multi subnet enabled connections will create two socket connections to the server (the first being closed). It'd be marginally slower than directly using the channel but it shouldn't be a big deal as it's just the establishment of the TCP socket. Also, I don't think it's an issue on the server to have the first connection closed as the driver is already doing that to the rest of the resolved addresses anyway.

So looks good to me.

FYI, the commit message has a typo: `Convert socket adaptor to socket #160`. **Adaptor** should be spelled **adapter**). Looks like it's already been merged though ...

v-xiangs:
@sehrope Thank you for getting back to us. We are glad to know you are satisfied with this fix.
Forgive me about the typo. English isn't my first language, Java is :)

Issue #150, opened 2017-02-16
TimeoutTimer starts a new Thread for all queries with a query-timeout


fwi:
I think there is an opportunity to optimize the query-timeout function with relative ease.

In the ``TimeoutTimer`` class in ``IOBuffer`` I see on line 6800:

	timerThread = new Thread(this);

If I read the code correctly, this would mean that a call to
``SQLServerConnection.isValid(5)``
would start a new thread to facilitate the 5 second timeout function.
A call to ``isValid`` is executed often by connection pool implementations (with a timeout) and should be relatively cheap.

Creating and starting a new thread is a relatively expensive operation. 
This can be optimized using the ``ScheduledExecutorService``. Below a little bit of code to demonstrate.
One thing that needs special attention is the shutdown of the ``ScheduledExecutorService``: 
this must be done properly and timely to ensure there are no memory leaks when, for example, re-deploying web applications (war-files). I known this can be a bit tricky as other JDBC drivers have failed to do this properly and required work-arounds.

	package com.microsoft.sqlserver.jdbc;

	import org.junit.Assert;
	import org.junit.Test;

	import java.util.concurrent.ScheduledFuture;
	import java.util.concurrent.ScheduledThreadPoolExecutor;
	import java.util.concurrent.TimeUnit;
	import java.util.logging.Level;

	public class TimeoutTimerTest {

		@Test
		public void timeoutTest() {
			
			TdsTimeOutCommand cmd1 = new TdsTimeOutCommand("ttTest1");
			TimeoutTimerScheduled timer1 = new TimeoutTimerScheduled(1, cmd1);
			TdsTimeOutCommand cmd2 = new TdsTimeOutCommand("ttTest2");
			TimeoutTimerScheduled timer2 = new TimeoutTimerScheduled(2, cmd2);
			TimeoutTimerScheduled.prestart();
			try {
				timer1.start();
				timer2.start();
				// It appears it takes some time for timer-tasks to get scheduled:
				// this test will fail if sleep is set to 1001 milliseconds which, technically, should be enough.
				Thread.sleep(1500);
				Assert.assertNotNull("Command 1 interrupted.", cmd1.getInterruptReason());
				Assert.assertNull("Command 2 not interrupted.", cmd2.getInterruptReason());
				timer2.stop();
				Thread.sleep(1000);
				Assert.assertNull("Command 2 not interrupted after timer stopped.", cmd2.getInterruptReason());
			} catch (Exception e) {
[...]

PR #162, opened 2017-02-28
Executor service instead of thread

v-xiangs:
for issue #150

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
@@ -6853,7 +6853,7 @@ final class TimeoutTimer implements Runnable {
     }
 
     final void stop() {
-        task.cancel(false);
+        task.cancel(true);
         canceled = true;
     }
 
--- a/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
@@ -55,6 +55,8 @@ import java.util.Map.Entry;
 import java.util.Set;
 import java.util.SimpleTimeZone;
 import java.util.TimeZone;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 import java.util.concurrent.SynchronousQueue;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
@@ -6824,7 +6826,7 @@ final class TDSReader {
 final class TimeoutTimer implements Runnable {
     private final int timeoutSeconds;
     private final TDSCommand command;
-    private Thread timerThread;
+    private static final ExecutorService executor = Executors.newCachedThreadPool();
     private volatile boolean canceled = false;
 
     TimeoutTimer(int timeoutSeconds,
@@ -6836,18 +6838,16 @@ final class TimeoutTimer implements Runnable {
         this.command = command;
     }
 
-    final void start() {
-        timerThread = new Thread(this);
-        timerThread.setDaemon(true);
-        timerThread.start();
+    final void start() {        
+        executor.execute(this);
     }
 
     final void stop() {
+        executor.shutdownNow();
         canceled = true;
-        timerThread.interrupt();
     }
 
-    public void run() {
+    public void run() { 
         int secondsRemaining = timeoutSeconds;
         try {
             // Poll every second while time is left on the timer.
--- a/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
@@ -57,7 +57,9 @@ import java.util.SimpleTimeZone;
 import java.util.TimeZone;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
 import java.util.concurrent.SynchronousQueue;
+import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
@@ -6826,7 +6828,15 @@ final class TDSReader {
 final class TimeoutTimer implements Runnable {
     private final int timeoutSeconds;
     private final TDSCommand command;
-    private ExecutorService executor = Executors.newCachedThreadPool();
+    private volatile Future<?> task;
+    private static final ExecutorService executor = Executors.newCachedThreadPool(new ThreadFactory() {
+        @Override
+        public Thread newThread(Runnable r) {
+            Thread t = Executors.defaultThreadFactory().newThread(r);
+            t.setDaemon(true);
+            return t;
+        }
+    });
     private volatile boolean canceled = false;
 
     TimeoutTimer(int timeoutSeconds,
@@ -6839,15 +6849,11 @@ final class TimeoutTimer implements Runnable {
     }
 
     final void start() {
-        //if ExecutorService is shutdown already, reset it
-        if(executor.isShutdown()){
-            executor = Executors.newCachedThreadPool();
-        }
-        executor.execute(this);
+        task = executor.submit(this);
     }
 
     final void stop() {
-        executor.shutdownNow();
+        task.cancel(false);
         canceled = true;
     }
 
--- a/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
@@ -6826,7 +6826,7 @@ final class TDSReader {
 final class TimeoutTimer implements Runnable {
     private final int timeoutSeconds;
     private final TDSCommand command;
-    private static final ExecutorService executor = Executors.newCachedThreadPool();
+    private ExecutorService executor = Executors.newCachedThreadPool();
     private volatile boolean canceled = false;
 
     TimeoutTimer(int timeoutSeconds,
@@ -6838,7 +6838,11 @@ final class TimeoutTimer implements Runnable {
         this.command = command;
     }
 
-    final void start() {        
+    final void start() {
+        //if ExecutorService is shutdown already, reset it
+        if(executor.isShutdown()){
+            executor = Executors.newCachedThreadPool();
+        }
         executor.execute(this);
     }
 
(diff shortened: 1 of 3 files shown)


fwi:
I'm sorry but this was not the point of my request. A shared SCHEDULED executor service creates tasks that might be executed in the future instead of starting threads immediately (i.e. threads and tasks are separated, this pull request does not change that). That has the potential to make a real difference in performance. To be really effective, the shared scheduled executor service should be created when the first connection is created and shutdown when the last connection is closed, and is shared/re-used among all connections that need to schedule tasks.

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -6837,17 +6839,19 @@ final void TryProcessFeatureExtAck(boolean featureExtAckReceived) throws SQLServ
     }
 
     final void start() {
-        timerThread = new Thread(this);
-        timerThread.setDaemon(true);
-        timerThread.start();
+        //if ExecutorService is shutdown already, reset it
+        if(executor.isShutdown()){
+            executor = Executors.newCachedThreadPool();
```
pierresouchay: The executor does not create daemon Threads by default, thus the result may be a little different.

I would probably use a thread factory as a parameter that use daemon threads.

v-xiangs:
@pierresouchay Thank you very much! Thank you for pointing this out. I don't need to worry about shutting down executor manually if I set them as daemon threads. 
@fwi thank you for your feedback. You are right, my previous implementation did not improve much. Could you please take a look now and see if that's what you would agree? now, the executor is static final, and re-uses threads for new tasks.

fwi:
Yes, this will improve the situation: threads will get re-used which takes away the relative expensive creating and starting of threads each time. I also had a re-think about my comment on shutdown of the executor service. I think using daemon-threads is sufficient: it will not prevent an application from shutting down and in the case of war-files / redeployable web applications the general advice is to always put the JDBC-drivers in the container (the "common library directory") and not package the JDBC drivers with the war-files - if that advice is followed there will be no memory leak problem either.
Thanks for the improvement!

v-xiangs:
@fwi Thank you for getting back to us with your thoughts. Again, we really appreciate your suggestions on this improvement. Right now, we are running more tests for this implementation, and will merge the PR after the those tests are done. Thank you.

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -6824,7 +6828,15 @@ final void TryProcessFeatureExtAck(boolean featureExtAckReceived) throws SQLServ
 final class TimeoutTimer implements Runnable {
     private final int timeoutSeconds;
     private final TDSCommand command;
-    private Thread timerThread;
+    private volatile Future<?> task;
+    private static final ExecutorService executor = Executors.newCachedThreadPool(new ThreadFactory() {
+        @Override
+        public Thread newThread(Runnable r) {
+            Thread t = Executors.defaultThreadFactory().newThread(r);
```
pierresouchay: Another way would be to use a dedicated ThreadGroup (for instance with name SQLServerDriver) and to create threads regarding this behaviour.

For instance:
```java
private static final ExecutorService executor = Executors.newCachedThreadPool(new ThreadFactory() {
    final ThreadGroup tg = new ThreadGroup("SQLServerDriver");
    final AtomicInteger counter = new AtomicInteger(0);
    @Override
    public Thread newThread(Runnable r) {
        Thread t = new Thread(tg, "sqlserverdriver-"+counter.incrementAndGet()
        t.setDaemon(true);
        return t;
    }
}
```

Having a dedicated ThreadGroup and well named threads would ease debug in complex software, such as application servers when there are many threads and would help identifying the ones created by the driver.

(While ThreadGroup may be overkill, at least having a good name for threads help a lot)

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -6837,17 +6849,15 @@ final void TryProcessFeatureExtAck(boolean featureExtAckReceived) throws SQLServ
...
 
     final void start() {
-        timerThread = new Thread(this);
-        timerThread.setDaemon(true);
-        timerThread.start();
+        task = executor.submit(this);
     }
 
     final void stop() {
+        task.cancel(false);
```
fwi: This must be `cancel(true)` like in the original implementation (timerThread.interrupt was called). Else new threads will be started. E.g. if 10 queries are fired with a time-out within one second (which is the sleep-interval in the run-method), 10 threads will be created.

v-xiangs: Thank you, good catch. I agree. Going to change to true now.

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
v-xiangs: Good idea. I will talk to the team and see what name we can give to the thread group. Thank you.

v-nisidh:
@v-xiangs : Can you add Connection Pool test cases on this PR?

src/test/java/com/microsoft/sqlserver/jdbc/connection/PoolingTest.java
```diff
@@ -139,4 +145,52 @@ public void testConnectionPoolClientConnectionId() throws SQLException {
...
+    private static void connect(DataSource ds) throws SQLException {
+        Connection con = null;
+        PreparedStatement pst = null;
+        ResultSet rs = null;
+
+        try {
+            con = ds.getConnection();
+            con.isValid(5);
+
+            pst = con.prepareStatement("SELECT SUSER_SNAME()");
```
fwi: For testing timeout: `pst.setQueryTimeout(5);`

src/test/java/com/microsoft/sqlserver/jdbc/connection/PoolingTest.java
```diff
@@ -139,4 +145,52 @@ public void testConnectionPoolClientConnectionId() throws SQLException {
...
         assertEquals(Id1, Id2, "ClientConnection Ids from pool are not the same.");
     }
+    
+    @Test
+    public void testHikariCP() throws SQLException {
+        HikariConfig config = new HikariConfig();
+        config.setJdbcUrl(connectionString);
+        HikariDataSource ds = new HikariDataSource(config);
+
+        connect(ds);
```
fwi: Testing performance with 'trivial queries', so add a loop:
```
		try {
			int ttStartCount = countTimeoutThreads();
			for (int i = 0; i < 100; i++) {
				connect(ds);
			}
			int ttEndCount = countTimeoutThreads();
			if (ttEndCount - ttStartCount > 2) {
				throw new AssertionError("Too many timeout threads started.");
			}
		} finally {
			// always close the datasource
			ds.close();
		}
```

src/test/java/com/microsoft/sqlserver/jdbc/connection/PoolingTest.java
```diff
@@ -139,4 +145,52 @@ public void testConnectionPoolClientConnectionId() throws SQLException {
...
+    }
+
```
fwi: Example for thread-counts (untested):
```
	public static int countTimeoutThreads() {

		ThreadInfo[] tinfos = ManagementFactory.getThreadMXBean().getThreadInfo(ManagementFactory.getThreadMXBean().getAllThreadIds(), 0);
		// IOBuffer.TimeoutTimer.threadGroupName should be publicly accessible ...
		final String toname = "mssql-jdbc-TimeoutTimer";
		int count = 0;
		for (ThreadInfo ti : tinfos) {
			if (ti.getThreadName().startsWith(toname) && !ti.getThreadState().equals(java.lang.Thread.State.TERMINATED)) {
				count++;
			}
		}
		return count;
	}
```

src/test/java/com/microsoft/sqlserver/jdbc/connection/PoolingTest.java
v-nisidh: @fwi : Thanks for suggestions. We are trying out `Java MicroBenchmark Harness (JMH)` test cases for performance testing. We are figuring out how & when we can open performance test cases. At the same time we want our Travis-CI & Appveyour builds should not be long running.

src/test/java/com/microsoft/sqlserver/jdbc/connection/PoolingTest.java
v-xiangs: hello @fwi , could you please inform me what the benefits are by testing setQueryTimeout() instead of isValid()? from my understanding, both create a new thread for the timeout timer

src/test/java/com/microsoft/sqlserver/jdbc/connection/PoolingTest.java
fwi: Bad choice of words on my part: this is actually a functional test. Remember when we changed a little bit of code to prevent creating lots of threads when many queries are executed rapidly? How do you know there is not another mistake like that still in the code? This test simply measures the amount of threads before and after executing many queries rapidly. Measuring is knowing.

src/test/java/com/microsoft/sqlserver/jdbc/connection/PoolingTest.java
fwi: Currently "isValid" ends up using the timeout timer, but this is an implementation choice (other JDBC drivers do it differently for speed and this JDBC driver might one day too) and might change in the future. Also "isValid" is already called by the connection pool before returing a connection from the pool (see for example "validationTimeout" in [HikariCP](https://github.com/brettwooldridge/HikariCP/)).

Besides that, the original issue was about "queries with a time-out", so why not test that explicitly and make sure to use a query with a time-out?

src/test/java/com/microsoft/sqlserver/jdbc/connection/PoolingTest.java
fwi: On a side note: the datasource backed by the pool implementation does need to be closed when the test is finished. If not, connections remain open and the pool implementation itself starts threads as well that are shutdown when the datasource is closed.

fwi: Final note: verify `ttEndCount` is at least 1 to ensure the test is working (if `ttEndCount` is zero, the test did not do its intended job).

src/test/java/com/microsoft/sqlserver/jdbc/connection/PoolingTest.java
v-xiangs: that makes sense, I will change it.

src/test/java/com/microsoft/sqlserver/jdbc/connection/PoolingTest.java
v-xiangs: instead of doing `if (ttEndCount - ttStartCount > 2) { throw exception}`, I guess only verify ttEndCount is at least 1 is sufficient. In the for loop, if we increase the termination (i.g. to 100000), it may create more than 2 timeout threads.

PR #163, opened 2017-02-28
Allow authenticating using Kerberos and a Principal/Password.

### LINKED ISSUE #66: Use provided username and password for JavaKerberos
It would be great if the driver could pick up username and password from the connection settings instead of having to be provided through "kinit".

pierresouchay:
This closes #66 

This patch allow to authenticate using kerberos using the previous
methods (eg: keytab) or specifying user/password either in properties
or in connect method.

This allows to use GUIs for instance to connect using Kerberos without
having a sql user.

If fixes issue #66

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/KerbAuthentication.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/KerbAuthentication.java
@@ -87,7 +87,6 @@ final class KerbAuthentication extends SSPIAuthentication {
                     else {
                         Map<String, String> confDetails = new HashMap<String, String>();
                         confDetails.put("useTicketCache", "true");
-                        confDetails.put("doNotPrompt", "true");
                         appConf = new AppConfigurationEntry("com.sun.security.auth.module.Krb5LoginModule",
                                 AppConfigurationEntry.LoginModuleControlFlag.REQUIRED, confDetails);
                         if (authLogger.isLoggable(Level.FINER))
@@ -144,14 +143,16 @@ final class KerbAuthentication extends SSPIAuthentication {
                     AccessControlContext context = AccessController.getContext();
                     currentSubject = Subject.getSubject(context);
                     if (null == currentSubject) {
-                        lc = new LoginContext(CONFIGNAME);
+                        lc = new LoginContext(CONFIGNAME, new KerbCallback(con));
                         lc.login();
                         // per documentation LoginContext will instantiate a new subject.
                         currentSubject = lc.getSubject();
                     }
                 }
                 catch (LoginException le) {
+                    authLogger.fine("Failed to login due to " + le.getClass().getName() + ":" + le.getMessage());
                     con.terminate(SQLServerException.DRIVER_ERROR_NONE, SQLServerException.getErrString("R_integratedAuthenticationFailed"), le);
+                    return;
                 }
 
                 if (authLogger.isLoggable(Level.FINER)) {
--- a/src/main/java/com/microsoft/sqlserver/jdbc/KerbAuthentication.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/KerbAuthentication.java
@@ -139,19 +139,33 @@ final class KerbAuthentication extends SSPIAuthentication {
             }
             else {
                 Subject currentSubject = null;
+                KerbCallback callback = new KerbCallback(con);
                 try {
                     AccessControlContext context = AccessController.getContext();
                     currentSubject = Subject.getSubject(context);
                     if (null == currentSubject) {
-                        lc = new LoginContext(CONFIGNAME, new KerbCallback(con));
+                        lc = new LoginContext(CONFIGNAME, callback);
                         lc.login();
                         // per documentation LoginContext will instantiate a new subject.
                         currentSubject = lc.getSubject();
                     }
                 }
                 catch (LoginException le) {
-                    authLogger.fine("Failed to login due to " + le.getClass().getName() + ":" + le.getMessage());
-                    con.terminate(SQLServerException.DRIVER_ERROR_NONE, SQLServerException.getErrString("R_integratedAuthenticationFailed"), le);
+                    if (authLogger.isLoggable(Level.FINE)) {
+                        authLogger.fine(toString() + "Failed to login using Kerberos due to " + le.getClass().getName() + ":" + le.getMessage());
+                    }
+                    try {
+                        // Not very clean since it raises an Exception, but we are sure we are cleaning well everything
+                        con.terminate(SQLServerException.DRIVER_ERROR_NONE, SQLServerException.getErrString("R_integratedAuthenticationFailed"), le);
+                    } catch (SQLServerException alwaysTriggered) {
+                        String message = String.format("%s due to %s (%s)", alwaysTriggered.getMessage(), le.getClass().getName(), le.getMessage());
+                        if (callback.getUsernameRequested() != null) {
+                            message = String.format("Login failed for Kerberos principal '%s'. %s", callback.getUsernameRequested(), message);
+                        }
+                        // By throwing Exception with LOGON_FAILED -> we avoid looping for connection
+                        // In this case, authentication will never work anyway -> fail fast
+                        throw new SQLServerException(message, alwaysTriggered.getSQLState(), SQLServerException.LOGON_FAILED, le);
+                    }
                     return;
                 }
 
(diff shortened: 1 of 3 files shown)


pierresouchay:
A few notes:
connecting with a user does not requires the full realm as long as default_realm is properly set in /etc/krb5.conf file (for instance, if default_realm = AD.EXAMPLE.COM then, if the username has no "@" within, it will be assumed that its name is user@AD.EXAMPLE.COM

Using a fully qualified username (for instance user@AD.EXAMPLE.COM) works as well.

In the case where a keytab is used, the callback won't even be called.

The global Kerberos mechanism may be enchanced as the Driver always retry even when the password is not valid (thus the return added).

For Another Review: A possible improvement would be to stop trying when login() does fails (we don't need to wait for timeout in that case) with a proper error message.

pierresouchay:
The second commit also fixes issues with authentication taking a very long time even when Kerberos authentication is completely broken (eg: wrong keytab or wrong password)

v-suhame:
@pierresouchay Can you open this PR in dev branch?

pierresouchay:
@v-suhame done

pierresouchay:
Updated the review to fix conflict due to https://github.com/Microsoft/mssql-jdbc/pull/178

src/main/java/com/microsoft/sqlserver/jdbc/KerbAuthentication.java
```diff
@@ -140,18 +139,34 @@ private void intAuthInit() throws SQLServerException {
...
+                        authLogger.fine(toString() + "Failed to login using Kerberos due to " + le.getClass().getName() + ":" + le.getMessage());
+                    }
+                    try {
+                        // Not very clean since it raises an Exception, but we are sure we are cleaning well everything
+                        con.terminate(SQLServerException.DRIVER_ERROR_NONE, SQLServerException.getErrString("R_integratedAuthenticationFailed"), le);
+                    } catch (SQLServerException alwaysTriggered) {
+                        String message = String.format("%s due to %s (%s)", alwaysTriggered.getMessage(), le.getClass().getName(), le.getMessage());
+                        if (callback.getUsernameRequested() != null) {
+                            message = String.format("Login failed for Kerberos principal '%s'. %s", callback.getUsernameRequested(), message);
+                        }
```
v-suhame: To be consistent with the existing code, can you add this new exception error string in SQLServerResource.java?

src/main/java/com/microsoft/sqlserver/jdbc/KerbCallback.java
```diff
@@ -0,0 +1,61 @@
...
+
+    private final SQLServerConnection con;
+    private String usernameRequested = null;
+
+    KerbCallback(SQLServerConnection con) {
+        this.con = con;
+    }
+
+    private static String getAnyOf(Callback callback, Properties properties, String... names)
+            throws UnsupportedCallbackException {
```
v-suhame: use the [formatter ](https://github.com/Microsoft/mssql-jdbc/blob/master/mssql-jdbc_formatter.xml) on new code.

pierresouchay:
@v-suhame DONE in new patch

v-suhame:
@pierresouchay Can you also push changes for IBM jre?

pierresouchay:
@v-suhame As far as I know, there is no option on IBM JVM to disable callback according to https://www.ibm.com/support/knowledgecenter/en/SSYKE2_7.1.0/com.ibm.java.security.api.doc/jgss/com/ibm/security/auth/module/Krb5LoginModule.html

So I suspect it works on IBM JVM without changes

v-suhame:
Tested PR with IBM JVM, Kerberos auth with username and password doesn't work. It looks like driver should switch between `useDefaultCcache` to `true` while using ticket and `false` for principal/password.

pierresouchay:
It means probably refactoring the creation of configurations according to connection parameters (and not in a static way)

For another PR?

v-suhame:
Sounds good! I'll accept/merge this PR and leave the issue #66 open until it's fixed for IBM JVM. 
Thanks @pierresouchay, and sorry it took so long to review and merge.

pierresouchay:
That's great !

IBM owners can still create their $HOME/.java.login.config file and specify the parameters of the config to make it work.

I did actually plan to use several names for instead of hardcoded default "SQLJDBCDriver" (and so remove the static initialization) -> that would be the right time to allow switching to a default configuration that enables login with username/password for IBM JVM

bayramgns:
Hello,

Is there a road map to develop driver for also IBM JVM to be able to use connection username and password properties for the kerberos authentication?

We can provide kerberos authentication by using keytab and login config files; it' s OK. But it would be much better to authenticate by using connection username and password properties instead of keytab and login config files as it' s used for Oracle JVM.

pierresouchay:
I did not test it, but it should work, did you try it?

pierresouchay:
@bayramgns you might try by creating a .java.login.config in your $HOME directory and putting useDefaultCcache=false it seems. Which version of IBM's JVM are you using?

bayramgns:
@pierresouchay it works well for Oracle JVM. When you use Oracle JVM, you don' t need to use any keytab and login conf file;  you only need to set  up kerberos conf file and set driver connection properties. But for IBM JVM, it does not seem like that.

I use IBMSDK 7.1 and mssql-jdbc-6.2.1.jre7.jar versions.

As this change, I thought I won' t need to set any login conf property file; because you don' t need it for Oracle JVM.However, although I set login conf file as it seems below, it still does not work;

SQLJDBCDriver {  
 com.ibm.security.auth.module.Krb5LoginModule required
 useDefaultCcache=false; 
}; 

But when I set the login conf file like the below, it works well;

SQLJDBCDriver {  
 com.ibm.security.auth.module.Krb5LoginModule required
 useDefaultCcache=false 
 useKeytab="D:/GUNES/krb5.keytab"
 credsType=both
 principal="XXX@YYY"; 
};

bayramgns:
@pierresouchay  could you check the case?

pierresouchay:
What do you mean exactly?

What are you testing?

1. You are logged using Kerberos (for instance, a Windows session) and would like to access to a MS SQL Db?
2. You are not logged using Kerberos and want to use keytab "D:/GUNES/krb5.keytab" ?
3. You are not logged and want to connect using Kerberos and a login/password ?

Assuming you are in the third case, and according to this page:

https://www.ibm.com/support/knowledgecenter/en/SSYKE2_7.1.0/com.ibm.java.security.component.71.doc/security-component/jgssDocs/jaas_login_user.html

> Krb5LoginModule might prompt for a principal name, password or both. Or it might not prompt for any input whatsoever. Whether or not the login is interactive depends on the options specified in the JAAS configuration file. For example, the login proceeds non-interactively when the credential type is initiator and a Kerberos TGT is to be retrieved from a default credentials cache file. On the other hand,a user is prompt for useCcache where user can specify a different credentials cache file if useDefaultCcache is set to false. If no value is entered, , a user is prompted for a password (and, perhaps, a principal name) when a TGT is to be obtained from a Kerberos KDC. When the login is interactive, the application must specify com.ibm.security.auth.callback.Krb5CallbackHandler as the callback handler when creating the login context. The callback handler is responsible for prompting for input.

I would try the following:
```
SQLJDBCDriver {
com.ibm.security.auth.module.Krb5LoginModule required
useDefaultCcache=false
credsType=both
};
```

I don't have a Windows machine nor IBM JVM to test now, can you test this configuration and tell me whether it works?

Current default configurations are the following:
https://github.com/Microsoft/mssql-jdbc/blob/dev/src/main/java/com/microsoft/sqlserver/jdbc/JaasConfiguration.java#L26

bayramgns:
@pierresouchay I have good news :)  It worked by using IBM Java Version 8. I used to work with IBM Java 7 version before and it was getting KerberosException.

My purpose is connect to an MS SQL DB by authenticating with Kerberos protocol.I don' t want to use keytab file for username and password. I just want to set username and password connection properties instead of using keytab file.

When  using IBM Java 7, you have to create a keytab file and also provide "useKeytab" and "principal" properties in "SQLJDBCDriver .conf" file; otherwise it does not work.

But when using IBM Java 8, it works by setting connection username/password properties(no need to create keytab file) and using a "SQLJDBCDriver .conf" file as you mentioned above (no need to provide useKeytab and principal properties).

Anyway, I have no problem any more :) Thanks for help :)

pierresouchay:
@bayramgns Ok, that's good news, can you try without the config file as well (I mean simply delete the file)? 

If it does not work, we might try setting those properties in the conf, so people you not even have to create the jaas config file.

bayramgns:
@pierresouchay , it does not work without the jaas config file. I' ve removed the auth login config jvm parameter and tried; but got "com.ibm.security.krb5.KrbException, status code: 31" exception.

I guess the default value of "useDefaultCcache (true)" in JaasConfiguration class causes the error.

Sounds great if you develop for setting these properties in the conf and no need to have jaas config file any more :)

kmohanarangam:
@pierresouchay  'But when using IBM Java 8, it works by setting connection username/password properties" - I tried the same. However the driver picks up the logged in user instead of the userName and password provided in the connection string. Is your "run as" user and the username on the connection string one and the same?

PR #165, opened 2017-03-01
Fix random failure in BulkCopyColumnMapping test

v-ahibr:
Before, when two random tables are created, there was a small chance that they would have identical schemas. This caused the ImplicitMismatchCM test case to pass when it was expected to fail. Modified creation of destination table to always have a different schema from the source table.

### DIFF
--- a/src/test/java/com/microsoft/sqlserver/testframework/DBSchema.java
+++ b/src/test/java/com/microsoft/sqlserver/testframework/DBSchema.java
@@ -87,6 +87,83 @@ public class DBSchema {
             // Other types
         }
     }
+    
+    /**
+     * 
+     * @param autoGenerateSchema
+     * @param alternateSchema
+     */
+    DBSchema(boolean autoGenerateSchema, boolean alternateSchema) {
+        sqlTypes = new ArrayList<SqlType>();
+        if (autoGenerateSchema && !alternateSchema) {
+            // Exact Numeric
+            sqlTypes.add(new SqlBigInt());
+            sqlTypes.add(new SqlInt());
+            sqlTypes.add(new SqlSmallInt());
+            sqlTypes.add(new SqlTinyInt());
+            sqlTypes.add(new SqlBit());
+            sqlTypes.add(new SqlDecimal());
+            sqlTypes.add(new SqlNumeric());
+            sqlTypes.add(new SqlMoney());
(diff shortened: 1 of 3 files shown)


src/test/java/com/microsoft/sqlserver/testframework/DBSchema.java
```diff
@@ -87,6 +87,83 @@
             // Other types
         }
     }
+    
+    /**
+     * 
+     * @param autoGenerateSchema
+     * @param alternateSchema
+     */
+    DBSchema(boolean autoGenerateSchema, boolean alternateSchema) {
```
v-nisidh: Instead of copy paste you can use shuffle / reverse.

> 
>if(alternateSchema) { 
Collections.shuffle(sqlTypes);
or 
Collections.reverse(sqlTypes);
}
>

src/test/java/com/microsoft/sqlserver/testframework/DBTable.java
```diff
@@ -82,6 +82,34 @@ public DBTable(boolean autoGenerateSchema, boolean unicode) {
...
+     * 
+     * @param autoGenerateSchema
+     *            <code>true</code>: generates schema with all available
+     *            dataTypes in SqlType class
+     * @param unicode
+     *             <code>true</code>: sets unicode column names if autoGenerateSchema is also set to <code>true</code>
+     * @param alternateShcema
+     *             <code>true</code>: creates table with alternate schema
+     */
+    public DBTable(boolean autoGenerateSchema, boolean unicode, boolean alternateSchema) {
```
v-nisidh: We can reduced copy paste by changing `public DBTable(boolean autoGenerateSchema)` and `public DBTable(boolean autoGenerateSchema, boolean unicode)` 

Example: 
````
public DBTable(boolean autoGenerateSchema) {
 this(autoGenerateSchema, false,false);  
}
````

src/test/java/com/microsoft/sqlserver/testframework/DBSchema.java
v-ahibr: done

src/test/java/com/microsoft/sqlserver/testframework/DBTable.java
v-ahibr: done

Issue #16, opened 2016-11-21
LOBs are fully materialised


marschall:
BLOBs, CLOBs and NCLOBs are fully materialised into a single byte[]/String. In the case of large content (imagine a file stored in a BLOB) this puts quite a lot of pressure on both the allocator and the heap. It also defeats the point of streaming. Consider the following use case:

```java
try (InputStream stream = resultSet.getBlob("column_name").getBinaryStream()) {
    // stream content
}
```

ajlam:
@marschall - Any suggestions on how to redesign this?

marschall:
As far as I can tell the existing `ServerResultSet#getBinaryStream` and `SQLServerResultSet#getCharacterStream` methods provide streaming. One could imagine building a LOB implementation around these or their underlying concepts.

v-suhame:
Thanks @marschall! That was a good catch, we will redesign it in the future releases.

v-afrafi:
@marschall opened #172 regarding this issue.

v-afrafi:
closed by pr#172. 
Thank you @marschall for pointing out this issue.

PR #172, opened 2017-03-06
Lobs materialisation issue 

v-afrafi:
for issue  #16

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBlob.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBlob.java
@@ -44,7 +44,7 @@ public final class SQLServerBlob implements java.sql.Blob, java.io.Serializable 
 
     static private final Logger logger = Logger.getLogger("com.microsoft.sqlserver.jdbc.internals.SQLServerBlob");
 
-    static private final AtomicInteger baseID = new AtomicInteger(0);	// Unique id generator for each instance (used for logging).
+    static private final AtomicInteger baseID = new AtomicInteger(0);   // Unique id generator for each instance (used for logging).
     final private String traceID;
 
     final public String toString() {
@@ -94,7 +94,7 @@ public final class SQLServerBlob implements java.sql.Blob, java.io.Serializable 
 
     SQLServerBlob(BaseInputStream stream) throws SQLServerException {
         traceID = " SQLServerBlob:" + nextInstanceID();
-        value = stream.getBytes();
+        activeStreams.add(stream);
         if (logger.isLoggable(Level.FINE))
             logger.fine(toString() + " created by (null connection)");
     }
@@ -142,7 +142,20 @@ public final class SQLServerBlob implements java.sql.Blob, java.io.Serializable 
     public InputStream getBinaryStream() throws SQLException {
         checkClosed();
 
-        return getBinaryStreamInternal(0, value.length);
+        if( null == value)
+        {           
+            InputStream stream = (InputStream) activeStreams.get(0);
+            try {
+                stream.reset();
+            } catch (IOException e) {
+                throw new SQLServerException(null, e.getMessage(), null, 0, true);
+            }
+            return (InputStream) activeStreams.get(0);
+        }
+        else       
+        {
+          return getBinaryStreamInternal(0, value.length);
+        }   
     }
 
     public InputStream getBinaryStream(long pos,
@@ -182,6 +195,7 @@ public final class SQLServerBlob implements java.sql.Blob, java.io.Serializable 
             int length) throws SQLException {
         checkClosed();
 
+        getBytesFromStream();
         if (pos < 1) {
             MessageFormat form = new MessageFormat(SQLServerException.getErrString("R_invalidPositionIndex"));
             Object[] msgArgs = {new Long(pos)};
@@ -219,8 +233,22 @@ public final class SQLServerBlob implements java.sql.Blob, java.io.Serializable 
      */
     public long length() throws SQLException {
         checkClosed();
-
+        getBytesFromStream();
         return value.length;
+    }
+    
+    private void getBytesFromStream() throws SQLServerException 
+    {
+        if ( null == value)
+        {
+            BaseInputStream stream = (BaseInputStream) activeStreams.get(0);
+            try {
+                stream.reset();
+            } catch (IOException e) {
+                throw new SQLServerException(null, e.getMessage(), null, 0, true);
+            }
+            value =  (stream).getBytes();
+        }
     }
 
     /**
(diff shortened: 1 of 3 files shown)


v-afrafi:
@marschall Thank you for opening an issue about the implementation of lobs. Could you also review this pr and give us your thoughts on this?
As of now if user needs to create blob and call getbinaryStream, they will be directly using the stream rather than the conversion of bytes. However, if a user needs to get the length() of the blob, the stream will be converted to bytes anyways. Since this conversion is an expensive call, the stream will get closed and all other calls over blob will be from byte array of stream.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBlob.java
```diff
@@ -142,7 +142,20 @@ private void checkClosed() throws SQLServerException {
     public InputStream getBinaryStream() throws SQLException {
         checkClosed();
 
-        return getBinaryStreamInternal(0, value.length);
+        if( null == value)
+        {           
+            InputStream stream = (InputStream) activeStreams.get(0);
```
marschall: I see this several times and this looks extremely brittle. I would consider making a dedicated field for this.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerClob.java
```diff
@@ -207,6 +230,7 @@ public InputStream getAsciiStream() throws SQLException {
     public Reader getCharacterStream() throws SQLException {
         checkClosed();
 
+        getBytesFromStream();
         Reader getterStream = new StringReader(value);
         activeStreams.add(getterStream);
```
marschall: I'm not sure this is needed because the underlying stream is already registered in `activeStreams`.

marschall:
@v-afrafi I have to say I'm not familiar with TDS at all, so I don't know if missing access to the payload length is a protocol limitation. However if I have a look at both `PLPInputStream` and `SimpleInputStream` they both have a `payloadLength` field so I wonder if this could be used.

I believe it would be helpful for people if the cases where the driver loads the entire LOB into memory are documented. This can help avoid `OutOfMemoryError`s when loading large LOBs into memory.

Besides the Javadoc the following places seem appropriate to me:

https://docs.microsoft.com/en-us/sql/connect/jdbc/using-advanced-data-types
https://docs.microsoft.com/en-us/sql/connect/jdbc/improving-performance-and-reliability-with-the-jdbc-driver

v-afrafi:
@marschall thanks for reviewing this pr. will consider your recommendation and update this.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBlob.java
```diff
@@ -44,8 +44,10 @@
 
     static private final Logger logger = Logger.getLogger("com.microsoft.sqlserver.jdbc.internals.SQLServerBlob");
 
-    static private final AtomicInteger baseID = new AtomicInteger(0);	// Unique id generator for each instance (used for logging).
+    static private final AtomicInteger baseID = new AtomicInteger(0);   // Unique id generator for each instance (used for logging).
     final private String traceID;
+    
+    private InputStream outputStream = null;
```
v-suhame: rename this variable, using `outputStream` as variable name for `InputStream` is confusing.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerClob.java
```diff
@@ -120,12 +124,19 @@ private String getDisplayClassName() {
      * @param logger
      */
     SQLServerClobBase(SQLServerConnection connection,
```
v-suhame: add javadocs for the new parameter.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerClob.java
```diff
@@ -191,9 +202,23 @@ public InputStream getAsciiStream() throws SQLException {
             DataTypes.throwConversionError(getDisplayClassName(), "AsciiStream");
 
         // Need to use a BufferedInputStream since the stream returned by this method is assumed to support mark/reset
-        InputStream getterStream = new BufferedInputStream(new ReaderInputStream(new StringReader(value), US_ASCII, value.length()));
+        InputStream getterStream = null;
+        if (null == value && !activeStreams.isEmpty()) {
+            outputStream = (InputStream) activeStreams.get(0);
```
v-suhame: how about adding this `!activeStreams.isEmpty()` condition in Blob too?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerClob.java
```diff
@@ -120,12 +124,19 @@ private String getDisplayClassName() {
      * @param logger
      */
     SQLServerClobBase(SQLServerConnection connection,
-            String data,
+            Object data,
```
v-suhame: All calls to this Constructor has null as typeInfo, why is it even passed if it always going to be null? Is there any other place where an actual values is passed?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBlob.java
```diff
@@ -219,9 +234,26 @@ private InputStream getBinaryStreamInternal(int pos,
...
+    private void getBytesFromStream() throws SQLServerException {
+        if (null == value) {
+            BaseInputStream stream = (BaseInputStream) activeStreams.get(0);
+            try {
+                stream.reset();
+            }
+            catch (IOException e) {
+                throw new SQLServerException(e.getMessage(), null, 0, e);
+            }
+            value = (stream).getBytes();
```
v-suhame: remove the brackets around (stream)

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerNClob.java
```diff
@@ -21,12 +21,12 @@
...
 
     SQLServerNClob(SQLServerConnection connection) {
-        super(connection, "", connection.getDatabaseCollation(), logger);
+        super(connection, "", connection.getDatabaseCollation(), logger, null);
     }
 
     SQLServerNClob(BaseInputStream stream,
             TypeInfo typeInfo) throws SQLServerException, UnsupportedEncodingException {
-        super(null, new String(stream.getBytes(), typeInfo.getCharset()), typeInfo.getSQLCollation(), logger);
+        super(null, new String(stream.getBytes(), typeInfo.getCharset()), typeInfo.getSQLCollation(), logger, null);
```
v-suhame: Why is NClob still stored as `String` rather than stream?

v-afrafi: Thanks. I have missed this somehow.

PR #175, opened 2017-03-07
enable update counts for SELECT INTO statements

gordthompson:
fix for issue discussed on Stack Overflow [here](http://stackoverflow.com/q/42573611/2144390), where `SELECT ... INTO` statements always return an update count of -1


v-afrafi:
@gordthompson Thank you for the fix you provided! Your pr is fixing the issue you mentioned.
However, it caused a failure in our test lab which I had added the repro code as a test in this pr. Please have a look and test it. I will then go ahead and accept this pull request.  Thanks again!

gordthompson:
@v-afrafi Thanks for the follow-up. Your `RegressionTest` worked fine for me once I tweaked the `DROP TABLE` to run under SQL_2014. Using your test as an example I then added `SelectIntoUpdateCountTest` to verify the results of my original modification. Hope this helps.

v-afrafi:
@gordthompson awesome! good catch. 
Can you just move the new test under the same Regression class? Thank you.

gordthompson:
@v-afrafi 

> Can you just move the new test under the same Regression class?

Done.

Issue #157, opened 2017-02-22
Exception thrown if comments are in SQL string 


bwbond87:
Running any query with a comment (using either -- or /**/) will throw an exception. We tend to read more complex SQL from a file on the classpath so having the ability to run queries with comments is very beneficial.

Example query: "/* test \*/ top 100 user_id from users" will throw the following:
"java.sql.SQLException: com.microsoft.sqlserver.jdbc.SQLServerException: Unable to identify the table /* test */"

v-suhame:
Hi @bwbond87, I tried executing simple queries with comments, it is working fine, no exception was thrown. Even `"select /* test */ top 10 user_id from users"` runs without exception. Can you share a code snippet or repo code, so we can narrow down the cause? 

The exception `com.microsoft.sqlserver.jdbc.SQLServerException` you mentioned is from SQL Server, so the problem could likely be because of an incorrect query sent or the way query is formed while reading from the external file.

bwbond87:
I was using a library to execute the SQL with a provided DataSource. I did some digging into the library and the offending exception occurs when the method getParameterMetaData() is called from the PreparedStatement implementation.

The database I am connecting to is running SQL Server 2005, so that may be a factor. The parseStatement method that the [SQLServerParameterMetaData](https://github.com/Microsoft/mssql-jdbc/blob/master/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerParameterMetaData.java#L561)  constructor defers to looks like it's failing to parse the SQL string for SQL Server versions <= 2008. 

This snippet triggers the exception for me:

```
try (Connection conn = dataSource.getConnection()) {
    preparedStatement = conn.prepareStatement(" /* test */ select top 100 user_id from users");
    preparedStatement.getParameterMetaData();
    ResultSet resultSet = preparedStatement.executeQuery();
}
```

v-suhame:
Thanks for the details!  I was using the latest SQL Server for testing, so I wasn't able to reproduce it early. 

It is a shortcoming in the parsing of SQL query, as you pointed out. The exception is thrown only on accessing `ParameterMetaData` for SQL Server <=2008, as the `sp_describe_undeclared_parameters` stored procedure used by driver for getting metadata was introduced in the later version of SQL SERVER. 

The input query needs to be parsed for comments and nested comments, we will consider it for future enhancement. You are welcome to make changes and submit a PR.

v-xiangs:
[...]

PR #181, opened 2017-03-09
filter out comments in the beginning of the query

v-xiangs:
for issue #157

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerParameterMetaData.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerParameterMetaData.java
@@ -366,7 +366,7 @@ public final class SQLServerParameterMetaData implements ParameterMetaData {
      */
     private MetaInfo parseStatement(String sql,
             String sTableMarker) {
-        StringTokenizer st = new StringTokenizer(sql, " ,", true);
+        StringTokenizer st = new StringTokenizer(sql, " ,\r\n", true);
 
         /* Find the table */
 
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerParameterMetaData.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerParameterMetaData.java
@@ -366,7 +366,7 @@ public final class SQLServerParameterMetaData implements ParameterMetaData {
      */
     private MetaInfo parseStatement(String sql,
             String sTableMarker) {
-        StringTokenizer st = new StringTokenizer(sql, " ,\r\n*/", true);
+        StringTokenizer st = new StringTokenizer(sql, " ,\r\n", true);
 
         /* Find the table */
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerParameterMetaData.java
```diff
@@ -424,6 +436,36 @@ private MetaInfo parseStatement(String sql) throws SQLServerException {
 
         return null;
     }
+    
+    private String removeCommentsInTheBeginning(String sql, int startCommentMarkCount, int endCommentMarkCount, String startMark, String endMark) {
```
v-suhame: can you format the code with the [formatter](https://github.com/Microsoft/mssql-jdbc/blob/master/mssql-jdbc_formatter.xml)?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerParameterMetaData.java
```diff
@@ -424,6 +436,36 @@ private MetaInfo parseStatement(String sql) throws SQLServerException {
...
+        int startCommentMarkIndex = sql.indexOf(startMark);
+        int endCommentMarkIndex = sql.indexOf(endMark);
+
```
v-suhame: also, consider the case where there is no end comment. It will run into StringIndexOutOfBoundsException, drive has to handle it.

v-suhame:
this could be totally different issue, but for some reason if I test with `*/ select top 10 NAME from table`, the exception thrown mentions SQLServerException twice

> com.microsoft.sqlserver.jdbc.SQLServerException: com.microsoft.sqlserver.jdbc.SQLServerException: Unable to identify the table */ select top 1 NAME from BigDe0 for the metadata.

v-xiangs:
@v-suhame that's because the query is not valid. Multiple line comment should start with /* then end with */, right? I also tried with SQL Server 2016 with your query and it failed. So it is expected to fail with this query.

v-xiangs:
@v-suhame that might be something related to "adding causes to all SQLSeverException"... I will take a look later.

v-suhame:
Using this string in PreparedStatement throws exception
`INSERT /**/INTO " + table + " (col) VALUES(?)`

v-suhame:
For tests, can you update them to use INSERT, UPDATE, DELETE too?

v-xiangs:
@v-suhame Oh... regarding to the original issue itself, I only tested query with comments in the beginning.. Thank you for finding the new issue, I will fix it too.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerParameterMetaData.java
v-xiangs: I don't think we need to consider invalid queries

v-xiangs:
@v-suhame added tests for INSERT, UPDATE, DELETE statements

Issue #166, opened 2017-03-02
Implement PreparedStatement caching for performance


davidfrankson:
To support modern versions of Hibernate, we need a JDBC 4 driver and we are looking to switch from our existing jtds driver http://jtds.sourceforge.net/ to Microsoft’s JDBC driver.  Performance problems prevent us from switching.

When we run a load test simulating our toughest workload, Solarwinds DPA is recording 7 times the cpu/memory wait time, big spikes on SQL compiles / sec and an overall doubling of cpu usage overall.  The performance problems occur because the MS driver does not support statement pooling and does not reuse PreparedStatements.

A older comment on your jdbcteam blog summarizes this need very well, I will copy/paste it here:

> Brett.Wooldridge
> December 1, 2015 at 4:47 pm
> Hello JDBC Team,
> I am the principal developer of HikariCP (https://github.com/brettwooldridge/HikariCP), the fastest growing connection pool for Java.
> We frequently are asked to implement PreparedStatement caching within the pool, but our stance has always been and continues to be that caching belongs in the driver.  On behalf of our MS SQL Server users, I would like to request the implementation of PreparedStatement caching.
> A survey of the Top 5 databases (https://dzone.com/articles/10-most-popular-db-engines-sql) leaves only SQL Server out without prepared statement caching.
> Ironically, SQL Server's JDBC driver has a property, disableStatementPooling, which defaults to *true. Only true is supported currently (setting to false will throw an exception).
> It was literally HikariCP's stand against placing responsibility of caching in the pool that prompted PostgreSQL to implement their very excellent solution (https://github.com/pgjdbc/pgjdbc/issues/308).  
> The PostgreSQL implementation took one developer 3 days from start to finish.
> In this new Satya Nadella era, isn't this the time for Microsoft to prove that it can be just as agile as the open source databases?  The 6.0 Preview is the perfect chance to introduce this long needed feature.
> Regards,
> Brett Wooldridge

More from Brett regarding Statement Caching and why it is needed at the driver is on the HikariCP site:
https://github.com/brettwooldridge/HikariCP 

> Statement Cache
> Many connection pools, including Apache DBCP, Vibur, c3p0 and others offer PreparedStatement caching. HikariCP does not. Why?
[...]

PR #183, opened 2017-03-09
Batching of unprepare for prepared statements.

TobiasSQL:
Fix for issue #166. Prepared statement un-prepares are now batched. Also, for first execution, sp_executesql is used instead of prepared statements. Can be configured to be turned off. Default behavior is the new one. @ajlam is in-the-loop on the change.


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -5116,6 +5140,246 @@ static synchronized long getColumnEncryptionKeyCacheTtl() {
...
+     * Returns the number of currently outstanding prepared statement un-prepare actions.
+     */
+    public int outstandingPreparedStatementDiscardActionCount(){
+        return this.discardedPreparedStatementHandleQueueCount.get();
+    }
+
+    /**
+     * Forces the un-prepare requests for any outstanding discarded prepared statements to be executed.
+     */
+    public void forcePreparedStatementDiscardActions()
```
saurabh500: Nit: Open brace on the same line as function.

saurabh500: This applies to all the functions in the CR where the opening brace is on new line.

saurabh500: There a bunch of formatting issues. Can you please use an IDE to format this. Ctrl A and Crtl Shift F in Eclipse should take care of this.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -132,50 +135,62 @@ String getClassNameInternal() {
...
+            // Using batched clean-up? If not, use old method of calling sp_unprepare.
+            if(1 < this.connection.getPreparedStatementDiscardActionThreshold()){
+                // Handle unprepare actions through batching @ connection level. 
+                this.connection.enqueuePreparedStatementDiscardItem(this.prepStmtHandle, this.executedSqlDirectly);
+                this.prepStmtHandle = 0;
+                this.connection.handlePreparedStatementDiscardActions(false);
+            }
+            else{
+                // Non batched behavior (same as pre batch impl.)
+                if (getStatementLogger().isLoggable(java.util.logging.Level.FINER))
```
saurabh500: Do we need this check? I am not sure about the internal workings of the logger being used here, but doing a logger.finer(..) internally not take care of checking the logging level and figuring out if FINER should be logged or not? 
Or is this the pattern being used in the rest of the code?

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/PreparedStatementTest.java
```diff
@@ -0,0 +1,202 @@
...
+                // Set instance setting to serial execution of un-prepare actions.
+                con.setPreparedStatementDiscardActionThreshold(1);                
+
+                try (SQLServerPreparedStatement pstmt = (SQLServerPreparedStatement)con.prepareStatement(query)) {
+                    pstmt.execute();
+                }
+                // Verify that the un-prepare action was handled immediately.
+                assertSame(0, con.outstandingPreparedStatementDiscardActionCount());
+            } 
+            finally {
```
saurabh500: Why empty finally?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -63,6 +63,9 @@
     /** SQL statement with expanded parameter tokens */
     private String preparedSQL;
 
+    /** True if this execute has been called for this statement at least once */
+    private boolean bExecutedAtLeastOnce = false;
```
saurabh500: Nit: you don't need the 'b' before the name of the variable. this can be changed to `executedAtLeastOnce`

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
saurabh500: I looked through the rest of the code and this seems like the pattern to follow.

saurabh500:
cc @tvrprasad

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
TobiasSQL: Thx, fixing :-)

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
TobiasSQL: Yep, just followed what was there.

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/PreparedStatementTest.java
TobiasSQL: They are really useful ;-) Gone now.

TobiasSQL: BTW, the reason for fun is just because it is easy to comment in/out a catch-block.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
TobiasSQL: I saw the other field above called nBatchStatementDelimiter and thought there was a hungarian thing going on in this file, looking around more seems like not. Renamed to isExecutedAtLeastOnce.

TobiasSQL:
Pushed most fixes including adding connection string options prepareStatementOnFirstCall and preparedStatementDiscardActionThreshold. Need to add tests which are coming, so code coverage should be temporarily down a tad.

saurabh500:
@TobiasSQL 
> prepareStatementOnFirstCall and preparedStatementDiscardActionThreshold

I think that we need to rethink these names, both for connection string parameters and for the function calls. 
My current thought is that these parameter names don't define what really is the threshold for!

If I understand this correctly I think it should be something like EnablePrepareOnFirstStatementCall and PrepareHandleDiscardThreshold or something on these lines.

TobiasSQL:
@saurabh500, I agree, as I mentioned to you when we spoke I am not particularly happy with the names, poor fantasy on my part ;-) 

WRT the names I agree that **enablePrepareOnFirstStatementCall** definitely is clearer, so it gets my vote too. **prepareHandleDiscardThreshold** I like that you removed Action, seems unnecessary and can probably get confused with some sort of function pointer. However, I do think statement is better than handle as this is the discard threshold for the statement, all be it on the server side. The fact that it is a handle seems like an implementation detail. So I suggest **prepareStatementDiscardThreshold**.

Thoughts? I will wait with these naming changes until @ajlam signs off.

saurabh500:
Thanks for iterating.
 
Let's think about this over the weekend. :)
prepareStatementDiscardThreshold seems like we are discarding the statement or it seems like we are discarding someyhing in the statement. 
It doesn't paint a clear picture of what will be discarded. At least this is my understanding of what we discussed.

TobiasSQL:
@saurabh500 sounds good! Now, it is in fact the statement that we are discarding ;-) Remember, close has already been called on the statement, and this is the delayed action to also discard the statement on the server. Now we could potentially call it **serverPrepareStatementDiscardThreshold** to point that out :-)

saurabh500:
I like this better.

TobiasSQL:
Ok, I think everything should be done except updating names. Below are the new public methods including currently suggested new names, once we close on these I will issue a final commit.
* Connection string: prepareStatementOnFirstCall -> **enablePrepareOnFirstStatementCall** 
* Connection string: preparedStatementDiscardActionThreshold -> **serverPreparedStatementDiscardThreshold**
* instance SQLServerDataSource.get/setPrepareStatementOnFirstCall -> **get/setEnablePrepareOnFirstStatementCall** 
* instance SQLServerDataSource.get/setPreparedStatementDiscardActionThreshold -> **get/setServerPreparedStatementDiscardThreshold**
* static SQLServerConnection.INITIAL_DEFAULT_PREPARE_STATEMENT_ON_FIRST_CALL -> **INITIAL_DEFAULT_ENABLE_PREPARE_STATEMENT_ON_FIRST_CALL**
* static SQLServerConnection.INITIAL_DEFAULT_PREPARED_STATEMENT_DISCARD_ACTION_THRESHOLD -> **INITIAL_DEFAULT_SERVER_PREPARED_STATEMENT_DISCARD_THRESHOLD**
* static SQLServerConnection.get/setDefaultEnablePrepareStatementOnFirstCall -> **get/setDefaultEnablePrepareStatementOnFirstCall**
* static SQLServerConnection.get/setDefaultPreparedStatementDiscardActionThreshold -> **get/setDefaultServerPreparedStatementDiscardThreshold**
* instance SQLServerConnection.get/setPrepareStatementOnFirstCall -> **get/setEnablePrepareOnFirstStatementCall**
* instance SQLServerConnection.get/setPreparedStatementDiscardActionThreshold-> **get/setServerPreparedStatementDiscardThreshold**
* instance SQLServerConnection.forcePreparedStatementDiscardActions -> **closeOutstandingDiscardedServerPreparedStatements** 
* instance SQLServerConnection.getOutstandingPreparedStatementDiscardActionCount -> **getOutstandingServerPreparedStatementDiscardCount**

ajlam:
@TobiasSQL good with new names

TobiasSQL:
Thanks @ajlam! @saurabh500, if you are also good I will go ahead and make the changes.

TobiasSQL:
One minor tweak, we should change **enablePrepareOnFirstStatementCall** to **enablePrepareOnFirstPreparedStatementCall** not to confuse with non-prepared statements (i.e. Statement).

ajlam:
@TobiasSQL ACK

TobiasSQL:
Ready to close :-)

ajlam:
@TobiasSQL, will work with the team to review and test. Hope to get this into the next preview release.

TobiasSQL:
Thanks @ajlam!

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -82,6 +83,25 @@
...
+    /**
+     * The initial default on application start-up for if prepared statements should execute sp_executesql before following the prepare, unprepare pattern. 
+     */    
+    static final public boolean INITIAL_DEFAULT_ENABLE_PREPARE_ON_FIRST_PREPARED_STATEMENT_CALL = false; // Used to set the initial default, can be changed later. false == use sp_executesql -> sp_prepexec -> sp_execute -> batched -> sp_unprepare pattern, true == skip sp_executesql part of pattern.
+    static private Boolean defaultEnablePrepareOnFirstPreparedStatementCall = null; // Current default for new connections
+    private Boolean enablePrepareOnFirstPreparedStatementCall = null; // Current limit for this particular connection.
+
+    // Handle the actual queue of discarded prepared statements.
+    private ConcurrentLinkedQueue<PreparedStatementDiscardItem> discardedPreparedStatementHandles = new ConcurrentLinkedQueue<PreparedStatementDiscardItem>();
+    private AtomicInteger discardedPreparedStatementHandleQueueCount = new AtomicInteger(0);
```
saurabh500: It seems like `discardedPreparedStatementHandleQueueCount ` is tracking the count of items in `discardedPreparedStatementHandles `. Is this right? If yes, can you use `discardedPreparedStatementHandles.size()` to get this count, instead of declaring a new variable?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -5141,6 +5184,234 @@ static synchronized long getColumnEncryptionKeyCacheTtl() {
...
+     * Specifies the behavior for a specific connection instance. This setting controls how many outstanding prepared statement discard
+     * actions (sp_unprepare) can be outstanding per connection before a call to clean-up the outstanding handles on the server is executed.
+     * If the setting is <= 1 unprepare actions will be executed immedietely on prepared statement close. If it is set to >1 these calls will
+     * be batched together to avoid overhead of calling sp_unprepare too often. 
+     * 
+     * @param value
+     *      Changes the setting per the description.
+     */
+    public void setServerPreparedStatementDiscardThreshold(int value) {
+        this.serverPreparedStatementDiscardThreshold = value;
```
saurabh500: Nit Pick: In the code you have deferenced all the member vars and methods using `this` keyword. Looks like the files convention is to use the member or method directly. 

I leave this to the committers to state what the guideline is for JDBC.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDriver.java
```diff
@@ -270,8 +270,9 @@ public String toString() {
 	LOGIN_TIMEOUT  ("loginTimeout",    15),
 	QUERY_TIMEOUT  ("queryTimeout",    -1),
 	PORT_NUMBER    ("portNumber",      1433),
-	SOCKET_TIMEOUT ("socketTimeout",   0);
-
+	SOCKET_TIMEOUT ("socketTimeout",   0),
+    SERVER_PREPARED_STATEMENT_DISCARD_THRESHOLD("serverPreparedStatementDiscardThreshold", -1/*This is not the default, default handled in SQLServerConnection and is not final/const*/);
```
saurabh500: Nit: Indentation is off wrt other items

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDriver.java
```diff
@@ -292,18 +293,19 @@ public String toString() {
...
+	INTEGRATED_SECURITY                       ("integratedSecurity",                        false),
+	LAST_UPDATE_COUNT                         ("lastUpdateCount",                           true),
+	MULTI_SUBNET_FAILOVER                     ("multiSubnetFailover",                       false),
+	SERVER_NAME_AS_ACE                        ("serverNameAsACE",                           false),
+	SEND_STRING_PARAMETERS_AS_UNICODE         ("sendStringParametersAsUnicode",             true),
+	SEND_TIME_AS_DATETIME                     ("sendTimeAsDatetime",                        true),
+	TRANSPARENT_NETWORK_IP_RESOLUTION         ("TransparentNetworkIPResolution",            true),
+	TRUST_SERVER_CERTIFICATE                  ("trustServerCertificate",                    false),
+	XOPEN_STATES                              ("xopenStates",                               false),
+    FIPS                                      ("fips",                                      false),
```
saurabh500: Nit: Indentation is off compared to other items. 
Also why did you change the formatting of the rest of the enums? 
Having only the lines which really changed, makes code reviews easier, and prevents unnecessary merge conflicts if other devs are working on the same region of code.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDriver.java
```diff
@@ -330,50 +332,52 @@ public String toString() {
...
-        new SQLServerDriverPropertyInfo(SQLServerDriverStringProperty.FIPS_PROVIDER.toString(), 						SQLServerDriverStringProperty.FIPS_PROVIDER.getDefaultValue(), 											false, 		null),
-        new SQLServerDriverPropertyInfo(SQLServerDriverIntProperty.SOCKET_TIMEOUT.toString(),                   		Integer.toString(SQLServerDriverIntProperty.SOCKET_TIMEOUT.getDefaultValue()),         					false,      null),
-        new SQLServerDriverPropertyInfo(SQLServerDriverBooleanProperty.FIPS.toString(),                                 Boolean.toString(SQLServerDriverBooleanProperty.FIPS.getDefaultValue()),                                false,      TRUE_FALSE),
-            };
+        //                                                               													  default       																						  required    available choices
```
saurabh500: Nit: A lot of line diffs for the items in this array. Can you try to not change the items which don't affect the feature?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -137,45 +140,57 @@ private void closePreparedHandle() {
             return;
 
         // If the connection is already closed, don't bother trying to close
-        // the prepared handle. We won't be able to, and it's already closed
+        // the prepared handle.  We won't be able to, and it's already closed
```
saurabh500: Nit: Extra Space before "We"

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -63,6 +63,9 @@
     /** SQL statement with expanded parameter tokens */
     private String preparedSQL;
 
+    /** True if this execute has been called for this statement at least once */
```
saurabh500: Nit: You can use `//` for single line comment instead of using a block comment construct `/* */`

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -5141,6 +5184,234 @@ static synchronized long getColumnEncryptionKeyCacheTtl() {
...
+    
+    /**
+     * Used to keep track of an individual handle ready for un-prepare.
+     */
+    private final class PreparedStatementDiscardItem  {
+
+        public int handle; 
+        public boolean directSql;
+
+        public PreparedStatementDiscardItem(int handle, boolean directSql) {
```
saurabh500: Question: Can multiple prepare statements share the same handle if they are using the same query? 
Or a new handle is generated for new Prepared Statement? I am trying to understand if a Connection can have multiple prepare statements executing with the same handle. If it can, then do we need have the handles be partitioned / associated with the prepared statement?

saurabh500:
Thanks for the answers @TobiasSQL. 
LGTM

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDriver.java
TobiasSQL: Agreed. problem was I introduced a longer name than was there before so I wanted to keep things according to the file standard. This whole section is a bit messy in terms of tabs vs spaces etc.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDriver.java
TobiasSQL: Agreed. problem was I introduced a longer name than was there before so I wanted to keep things according to the file standard.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
TobiasSQL: Well, again, following convention in the file.

TobiasSQL:
Thx @saurabh500!

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDataSource.java
```diff
@@ -663,6 +663,24 @@ public int getQueryTimeout() {
                 SQLServerDriverIntProperty.QUERY_TIMEOUT.getDefaultValue());
     }
 
+    public void setEnablePrepareOnFirstPreparedStatementCall(boolean enablePrepareOnFirstPreparedStatementCall) {
+        setBooleanProperty(connectionProps, SQLServerDriverBooleanProperty.ENABLE_PREPARE_ON_FIRST_PREPARED_STATEMENT.toString(), enablePrepareOnFirstPreparedStatementCall);
+    }
+
```
v-suhame: Add javadocs to the new public APIs (we have missed them for most of the APIs in this class, we will add them sometime soon). If you are using eclipse IDE, typing /** (enter) will generate the basic javadocs, you can add details to it (example SQLServerDataSource#getServerNameAsACE).

TobiasSQL: Can do. I added it to all places where there were javadocs already, DataSource didn't have them so just followed the pattern.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -5141,6 +5184,234 @@ static synchronized long getColumnEncryptionKeyCacheTtl() {
...
     }
 
+    
+    /**
+     * Used to keep track of an individual handle ready for un-prepare.
+     */
+    private final class PreparedStatementDiscardItem  {
+
+        public int handle; 
+        public boolean directSql;
```
v-suhame: The access modifier `public` can be discarded from the constructor and member variables of `PreparedStatementDiscardItem`. Its more like a best practice in Java.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -5141,6 +5184,234 @@ static synchronized long getColumnEncryptionKeyCacheTtl() {
...
+        // Add the new handle to the discarding queue and find out current # enqueued.
+        this.discardedPreparedStatementHandles.add(new PreparedStatementDiscardItem(handle, directSql));
+        this.discardedPreparedStatementHandleQueueCount.incrementAndGet();
+    }
+
+
+    /**
+     * Returns the number of currently outstanding prepared statement un-prepare actions.
+     */
+    public int getDiscardedServerPreparedStatementCount() {
```
v-suhame: needs javadocs for `@return`

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
TobiasSQL: Cool, will do.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
TobiasSQL: Fixed :-)

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -82,6 +83,25 @@
     long timerExpire;
     boolean attemptRefreshTokenLocked = false;
 
+    // Threasholds related to when prepared statement handles are cleaned-up. 1 == immediately.
+    /**
+     * The initial default on application start-up for the prepared statement clean-up action threshold (i.e. when sp_unprepare is called). 
+     */    
+    static final public int INITIAL_DEFAULT_SERVER_PREPARED_STATEMENT_DISCARD_THRESHOLD = 10; // Used to set the initial default, can be changed later.
```
v-afrafi: Could we change its access to private and add a getter method for it?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -82,6 +83,25 @@
...
+    static private int defaultServerPreparedStatementDiscardThreshold = -1; // Current default for new connections
+    private int serverPreparedStatementDiscardThreshold = -1; // Current limit for this particular connection.
+
+    /**
+     * The initial default on application start-up for if prepared statements should execute sp_executesql before following the prepare, unprepare pattern. 
+     */    
+    static final public boolean INITIAL_DEFAULT_ENABLE_PREPARE_ON_FIRST_PREPARED_STATEMENT_CALL = false; // Used to set the initial default, can be changed later. false == use sp_executesql -> sp_prepexec -> sp_execute -> batched -> sp_unprepare pattern, true == skip sp_executesql part of pattern.
```
v-afrafi: same for this.

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/PreparedStatementTest.java
```diff
@@ -0,0 +1,249 @@
...
+                // Verify clean-up is happening as expected.
+                if(prevDiscardActionCount > con.getServerPreparedStatementDiscardThreshold()) {
+                    prevDiscardActionCount = 0;
+                }
+
+                assertSame(prevDiscardActionCount, con.getDiscardedServerPreparedStatementCount());
+            }  
+
+            // Verify total cache use.
+            assertSame(iterations * 4, executeSQLReturnFirstInt(con, verifyTotalCacheUsesQuery));
```
v-afrafi: out of curiosity, can you try running this test on remote sql server? or check if you get same result when debugging it?
The test sometimes fails with expecting to have 100 but retrieves different values.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
TobiasSQL: Sure, is there a specific reason for this other than changing for sake of changing? 

My take was final was analog to const and we commonly have const fields public for clarity.

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/PreparedStatementTest.java
TobiasSQL: I did that, but have not run into different results. Do you have some more details, this doesn't seem easily reproducible. What value are you seeing when you run "SELECT * FROM sys.dm_exec_cached_plans AS p CROSS APPLY sys.dm_exec_sql_text(p.plan_handle) AS s WHERE s.text LIKE '%<uuid value used in the test>%'?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
v-afrafi: Its due to errors we saw in the internal test framework where its expecting the public fields to be added to Interface level.

TobiasSQL: Got it, ok will make these methods instead :-)

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/PreparedStatementTest.java
v-afrafi: It has intermittent failure. With failure:   expected: <100> but was: <97>  97 or some other values varying each time. 
This one happened when putting break point at line 118 and running the query you mentioned on sql server. 
 plan_caching

TobiasSQL: Got it, we will need to follow up on this on the server side. For now I have changed the test to allow a certain discrepancy in cache hits. This is not a functional issue so we shouldn't block on this.

v-afrafi: Agreed. Thank you.

v-afrafi: Hi @TobiasSQL . We are finished with our internal testing and want to merge your PR. Could you please disable just this test for now? We still see failures in this test that needs investigation but as you said it is not a functional issue.  Thanks again :)

TobiasSQL:
Waiting for response?

TobiasSQL:
Sorry, missed that one. Will take care of it, can you send me an e-mail with more details on the failures so I can get the root cause figured out?

PR #187, opened 2017-03-10
Remove obsolete methods from DriverJDBCVersion

### LINKED ISSUE #61: Remove obsolete methods from DriverJDBCVersion
If I understand correctly the current codebase supports jdbc 4.1 and jdbc 4.2 specs. This means that there is no need to check if the driver supports jdbc 4.1 and lower since 4.1 is the earliest jdbc version implemented by the driver. Therefore I believe that the methods in DriverJDBCVersion:
`checkSupportsJDBC4`
and
`checkSupportsJDBC41`
are obsolete and can safely be removed. There are just under 200 references in the code for the two methods and removing them would make the code much cleaner. The methods are no-ops and are probably removed by the compiler anyway.

marschall:
The current codebase supports JDBC 4.1 and JDBC 4.2 specifications.
This means that there is no need to check if the driver supports JDBC
4.1 and lower since 4.1 is the earliest JDBC version implemented by the
driver. Therefore the methods in DriverJDBCVersion checkSupportsJDBC4
and checkSupportsJDBC41 are obsolete and can safely be removed.

Closes #61

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBlob.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBlob.java
@@ -106,8 +106,6 @@ public final class SQLServerBlob implements java.sql.Blob, java.io.Serializable 
      * multiple times, the subsequent calls to free are treated as a no-op.
      */
     public void free() throws SQLException {
-        DriverJDBCVersion.checkSupportsJDBC4();
-
         if (!isClosed) {
             // Close active streams, ignoring any errors, since nothing can be done with them after that point anyway.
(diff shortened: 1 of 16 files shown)


marschall:
`com.microsoft.sqlserver.jdbc.unit.statement.StatementTest.TCStatementParam.testRowError()` fails on AppVeyor with
```
com.microsoft.sqlserver.jdbc.SQLServerException: Lock request time out period exceeded.
```
but succeeds on Travis

v-nisidh:
As Timeout timer is missing Exception is in random nature only on Appveyor,  trying to close & reopen pull request.

PR #189, opened 2017-03-10
Typos in SQLServerConnectionPoolProxy

marschall:
There are a few easy to fix typos in SQLServerConnectionPoolProxy.


marschall:
The AppVeyor builds fail because of
```java
org.opentest4j.AssertionFailedError: Timeout timer is missing.
```
while the Travis builds succeed. I don't think this is caused by Javadoc changes.

PR #193, opened 2017-03-10
Removing ant

v-suhame:


v-xiangs:
also delete Ant from sample readme.md?

Issue #139, opened 2017-02-07
Add IOException as a cause of SQLServerException


aharbunou:
Hi, 
I have one comment for error handling in case of IOException. Is it possible to add the exception (SocketTimeoutException for example) to SQLServerException as a cause? We want to separate database timeouts from the rest of errors from SQL Server i.e. we need cause of SQLServerException. I will be very appreciated if you could implement this.
Thanks,
Andrei

v-xiangs:
@aharbunou Hello. We are very glad to give help. Before we get started, we just want to clarify with you, do you want to add only SocketTimeoutException to SQLServerException  as a cause? or all kinds of timeout exceptions (socket timeout, login timeout, etc.) as cause? or all SQLServerExceptions have a cause? Thank you.

aharbunou:
Hi @v-xiangs. Ideally, I would like to have cause of SQLServerException exception at any place where it possible. SocketTimeoutException is something that must have for me :) 
Thanks,
Andrei

v-xiangs:
@aharbunou Thank you for confirming with us. It will require a lot of work for all the SQLServerException exceptions to have causes. Also, we need plenty of time to test them. For now, we will create a PR to add SocketTimeoutException as cause to meet your needs. And we will leave the issue open and work on adding causes to all SQLServerException exceptions. Thank you.

v-xiangs:
hello @aharbunou , we created a PR to add causes to all SQLServerException exceptions. If you have time, could you please check if it meets your needs?

aharbunou:
Hi @v-xiangs. Everything looks good for me. It is enough to improve my current error handling. I will create ticket If I need something else :) Thank you for all your work! I really appreciate it!

v-xiangs:
@aharbunou We are glad to know it meets your needs. I am closing this issue and you are very welcome to create new tickets.

PR #202, opened 2017-03-14
have cause of SQLServerException exception at any place where it possible

v-xiangs:
for issue #139

### DIFF
--- a/src/test/java/com/microsoft/sqlserver/jdbc/exception/ExceptionTest.java
+++ b/src/test/java/com/microsoft/sqlserver/jdbc/exception/ExceptionTest.java
@@ -64,8 +64,8 @@ public class ExceptionTest extends AbstractTest {
         SQLServerConnection conn = null;
         try {
             conn = (SQLServerConnection) DriverManager.getConnection(connectionString);
-
-            dropWaitForDelayProcedure(conn);
+            
+            Utils.dropProcedureIfExists(waitForDelaySPName, conn.createStatement());
             createWaitForDelayPreocedure(conn);
 
             conn = (SQLServerConnection) DriverManager.getConnection(connectionString + ";socketTimeout=" + (waitForDelaySeconds * 1000 / 2) + ";");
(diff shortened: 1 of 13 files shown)


src/test/java/com/microsoft/sqlserver/jdbc/exception/ExceptionTest.java
```diff
@@ -0,0 +1,123 @@
...
+        }
+        finally {
+            if (null != conn) {
+                conn.close();
+            }
+        }
+    }
+
+    private void dropWaitForDelayProcedure(SQLServerConnection conn) throws SQLException {
+        String sql = " IF EXISTS (select * from sysobjects where id = object_id(N'" + waitForDelaySPName
```
v-nisidh: Use new Utility for delete object. `Utils.dropObjectIfExists(...)`

src/test/java/com/microsoft/sqlserver/jdbc/exception/ExceptionTest.java
```diff
@@ -0,0 +1,123 @@
...
+            assertTrue(null != e.getCause(), "Cause should not be null.");
+            assertTrue(e.getCause() instanceof UnsupportedEncodingException, "Cause should be instance of UnsupportedEncodingException.");
+        }
+    }
+
+    /**
+     * 
+     * @return location of resource file
+     */
+    static String getCurrentClassPath() {
```
v-nisidh: Can you move this to Util class as it is very much common to whole framework. 

@v-suhame : Possibly we already might be having this functionality of `getCurrentClassPath` somewhere.

v-xiangs: we have `getCurrentClassPath()` in BulkCopyCSVTest as well. I moved them into Utils class.

src/test/java/com/microsoft/sqlserver/jdbc/exception/ExceptionTest.java
v-xiangs: This PR is pretty old, before we have this `Utils.dropObjectIfExists(...)` method.

v-xiangs: changed to `Utils.dropProcedureIfExists` after merging/updating with Dev branch

PR #209, opened 2017-03-20
Tnir changes

v-xiangs:


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -166,7 +168,7 @@ public String toString() {
         Closed      // indicates that the connection has been closed.
     }
 
-    private final static float TIMEOUTSTEP = 0.08F;    // fraction of timeout to use for fast failover connections
+    private final static float TIMEOUTSTEP = 0.125F;    // fraction of timeout to use for fast failover connections
```
saurabh500: Doesn't this change the time slice for MultiSubnetfailover as well?

saurabh500: We want to retain the timeslicing behavior for MSF and not change the timeslice values there. 
For TNIR you will have to define another step.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -1586,6 +1587,39 @@ else if (null == currentPrimaryPlaceHolder) {
...
+                    }
+                }
+
+                // if only one IP is resolved and user didn't set TNIR explicitly then set TNIR to false
+                if ((1 == inetAddrs.length) && !userSetTNIR) {
+                    useTnir = false;
+                }
+
+                //calculate timeout interval for the first attempt
+                if(0 == attemptNumber){
```
saurabh500: Nit: Run the changed code through code formatter. There are some spacing differences with the rest of the file.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -1595,17 +1629,18 @@ else if (null == currentPrimaryPlaceHolder) {
                     connectionlogger.fine(toString() + " This attempt No: " + attemptNumber);
                 }
                 // end logging code
-
+
```
saurabh500: Nit: Extra Space.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -1595,17 +1629,18 @@ else if (null == currentPrimaryPlaceHolder) {
                     connectionlogger.fine(toString() + " This attempt No: " + attemptNumber);
                 }
                 // end logging code
-
+                
                 // Attempt login.
                 // use Place holder to make sure that the failoverdemand is done.
-
+
```
saurabh500: Nit: Extra Space.

v-xiangs:
@saurabh500 Hello Saurabh, I have separated TNIR timeslice from MSF timeslice and formatted the code.
For the DNS resolution, do you think I need to roll back to previous implementation to do DNS resolution with each connection re-try? Thank you.

saurabh500:
@v-xiangs 
> For the DNS resolution, do you think I need to roll back to previous implementation to do DNS resolution with each connection re-try?

I have been talking to the folks from the server side. It turns out that time-slicing is needed in case of Single DNS IP address as well. 
This is because a DB availability group can have a cluster in a single subnet. When a failover occurs, the Windows Fail over clustering moves the IP address of the old primary to the new primary replica. However only a single IP address is available. I need to research more in this area. 

I think we should take out the DNS logic from the changes completely and as you suggested, continue with the DNS resolution with each connection re-try. 
What do you think? 

I am reviewing the rest of the code.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -1498,7 +1504,10 @@ private void login(String primary,
         // for all other cases, including multiSubnetFailover
         final boolean isDBMirroring = (null == mirror && null == foActual) ? false : true;
         int sleepInterval = 100;  // milliseconds to sleep (back off) between attempts.
-        long timeoutUnitInterval;
+        long timeoutUnitInterval = 0;
```
saurabh500: These additional variable allocations can be avoided. You can enhance the logic by checking for `isDBMirroring` `userParallel` `useTnir` to decide whether TIMEOUTSTEP or TIMEOUTSTEP_TNIR needs to be used and continue to use intervalExpire to populate the correct time out in this context.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -1721,12 +1765,22 @@ else if (null == currentPrimaryPlaceHolder) {
...
 
-            if (useParallel || useTnir) {
+            if (useParallel) {
                 intervalExpire = System.currentTimeMillis() + (timeoutUnitInterval * (attemptNumber + 1));
             }
             else if (isDBMirroring) {
                 intervalExpire = System.currentTimeMillis() + (timeoutUnitInterval * ((attemptNumber / 2) + 1));
             }
+            else if (useTnir) {
+                long timeSlice = timeoutUnitInterval * ((long) Math.pow(2, attemptNumber));
```
saurabh500: Instead of using `Math.pow` , I would recommend using 1 << attemptNumber 
Logical shift should be more performant.

saurabh500: Also, should this be attemptNumber - 1 ?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
v-xiangs: I use those additional variables because I couldn't find a way to decide whether TIMEOUTSTEP or TIMEOUTSTEP_TNIR needs to be used, since I only know whether TNIR is on or off after [DNS resolution](https://github.com/Microsoft/mssql-jdbc/pull/209/files#diff-0a4455f5ffd8e31e9e1badaa7822375eR1657)

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
v-xiangs: It's a good idea to use shift operator. I changed to it base on your suggestion.

For the second comment: I had the same thought too. But our existing behaviour is 2,3,4 instead of 1,2,3,4, it starts at 2. Therefore, instead of 1,2,4, I thought I should keep the existing behaviour to make it start at 2 too, as 2,4,8.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -1635,6 +1639,40 @@ else if (null == currentPrimaryPlaceHolder) {
                     currentConnectPlaceHolder = currentPrimaryPlaceHolder;
                 }
 
+                try {
+                    InetAddress[] inetAddrs = InetAddress.getAllByName(currentConnectPlaceHolder.getServerName());
+
+                    // if only one IP is resolved and user didn't set TNIR explicitly then set TNIR to false
+                    if ((1 == inetAddrs.length) && !userSetTNIR) {
```
saurabh500: @v-xiangs I made a comment about DNS resolution for TNIR to be removed here https://github.com/Microsoft/mssql-jdbc/pull/209#issuecomment-291694425 
The problem is that Single IP address could be possible in case of a Single Subnet where Timeslicing may be needed for AGs

v-xiangs: Oh, sorry, I missed this comment. If that's the case, I guess I need to roll back a lot of changes I have made for this PR, for example, logics for TIMEOUTSTEP and TIMEOUTSTEP_TNIR. Do you think we should close this PR and create other one after we have a final solution?

saurabh500: We still want to keep the TIMEOUTSTEP and TIMEOUTSTEP_TNIR where we have different timeslices for TNIR and also turn TNIR off for FedAuth. 
If you think that it is cleaner to have a new PR, then I am fine with it. 

Basically the solution right now is 
1. Turn Tnir off for Fed Auth
2. Change TNIR time slices
3. Keep TNIR on if the user has explicitly specified it in the connection string.

v-xiangs:
@saurabh500 Thank you for your quick response.

Please let me confirm the solution with you. I just want to make sure I will do the right thing.
1. Turn Tnir off for Fed Auth
2. Change the TNIR multiplier to 0.125 instead of 0.08
3. Change the time slices to a multiple of 2,4,8 instead of 2,3,4
4. In case the timeout for the first slice is less than 500 ms then bump it up to 500 ms
5. Keep TNIR on if the user has explicitly specified it in the connection string.

saurabh500:
A small correction

``
Change the time slices to a multiple of 2,4,8 instead of 2,3,4
In case the timeout for the first slice is less than 500 ms then bump it up to 500 ms
``

Change the time slices to a multiple of 1,2,4,8 instead of 1,2,3,4

Explanation: For the first attempt we should connect to the first IP addressed as is the behavior right now, however the timeout should be = 1st timeslice and if the 1st timeslice is < 500 ms, bump it to 500 ms. As a result we increase the timeout for the first slice. From the 2nd timeslice, we do parallel open, the way it is right now.

v-xiangs:
Hi @saurabh500 , I guess you might miss [this](https://github.com/Microsoft/mssql-jdbc/pull/209#discussion_r110042141) from my previous comment. That's the reason why I make it start at 2. I definitely can change the behaviour to a multiple of 1,2,4,8 as you mentioned.

saurabh500:
Yes, I did miss out the comment. 
I hadn't realized that the MSF steps were 2,3,4 for time slice calculation in JDBC.

Let's stick to your proposal of 2,4,8.
Without the TNIR redesign, the number of timeslices allocated would be really 4 for a timeout of 30 sec
image


With the change the number of slices will reduce by 1. 
image


You are right, we dont want to reduce the minimum time slice multiplier to 1, if a stable JDBC version has shipped with a Timeslice of 2 for the first try.

v-xiangs:
@saurabh500 thank you for confirming this with me. I really appreciate all your thoughtful suggestions and code reviews. I feel this PR is too messy for me to roll back some logics, so I created a new PR #240 . Therefore, I am going to close this PR. I hope you won't mind if I add you as Reviewer on the new PR #240 .

saurabh500:
Sounds Good.

PR #212, opened 2017-03-21
Make SQLServerException constructors public

mhaderman:
Make SQLServerException constructors public so that interfaces like ISQLServerBulk record can be implemented outside of the package.


v-nisidh:
Thanks @mhaderman for your contribution. Can you open another pull request for Dev branch ?

Issue #207, opened 2017-03-17
SQLServerException thrown twice


v-xiangs:
found by @v-suhame 
if run `PreparedStatement.getParameterMetaData()` with `*/ select top 10 NAME from table`, the SQLServerException is thrown twice
```
com.microsoft.sqlserver.jdbc.SQLServerException: com.microsoft.sqlserver.jdbc.SQLServerException: Unable to identify the table */ select top 1 NAME from BigDe0 for the metadata.
```

PR #213, opened 2017-03-21
fix SQLServerException wrapped by another SQLServerException

v-xiangs:
for issue #207


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerParameterMetaData.java
```diff
@@ -628,6 +628,9 @@ private void checkClosed() throws SQLServerException {
                 }
             }
         }
+        catch (SQLServerException e) {
+            throw e;
```
v-nisidh: Just add one line comment like `SQLServerException should not again wrapped by SQLServerException`

Issue #214, opened 2017-03-21
test failures when running tests against SQL Server 2008 SP4


v-xiangs:
There are 6 test failures against SQL Server 2008 SP4.
Those failures do not exist against SQL Server 2016.
```
com.microsoft.sqlserver.jdbc.unit.statement.StatementTest:
* testLastUpdateCountFalse()
* testStatementInsertExecInsert()
* testLastUpdateCountTrue()
* testPreparedStatementInsertExecInsert()

com.microsoft.sqlserver.jdbc.resultset.ResultSetTest
* testTesultSetWrapper()

com.microsoft.sqlserver.jdbc.parametermetadata.ParameterMetaDataTest
* testParameterMetaDataWrapper()
```

v-xiangs:
fixed by @gordthompson. Thank you.

PR #215, opened 2017-03-22
Fix test failures on pre-2016 servers

gordthompson:
Replace DROP TABLE IF EXISTS calls for compatibility with pre-2016 servers. ( Issue #214 )


v-xiangs:
@gordthompson Thank you for your contribution. Now, all the tests succeed against SQL Server 2008 with your fix. Thank you.

Issue #195, opened 2017-03-13
Reuse parameter metadata when using Always Encrypted


v-suhame:
If `columnEncryptionSetting` is enabled, mssql-jdbc driver calls sp_describe_parameter_encryption each time PreparedStatement is executed. 

Below SQL profiler trace was generated when creating PreparedStatement and using it twice to execute.

Trace for mssql-jdbc:
![alt tag](https://cloud.githubusercontent.com/assets/22776421/23874145/d363ffdc-07f1-11e7-9efb-a9eb70e94efb.png)

Trace for SqlClient:
![image](https://cloud.githubusercontent.com/assets/22776421/23874485/ffb333cc-07f2-11e7-904b-8697c6dda4a4.png)

Calling the stored procedure just once and reusing the metadata will improve the drivers performance with Always Encrypted significantly.

v-suhame:
@v-xiangs Thanks for the fix!

PR #220, opened 2017-03-24
retrieve paramater encryption metadata if they are not retrieved yet

v-xiangs:
for issue #195

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
@@ -89,6 +89,11 @@ public class SQLServerPreparedStatement extends SQLServerStatement implements IS
 
     /** Flag set to true when statement execution is expected to return the prepared statement handle */
     private boolean expectPrepStmtHandle = false;
+    
+    /**
+     * Flag set to true when all encryption metadata of inOutParam is retrieved
+     */
+    private boolean encryptionMetadataISRetrieved = false;
 
     // Internal function used in tracing
     String getClassNameInternal() {
@@ -202,6 +207,7 @@ public class SQLServerPreparedStatement extends SQLServerStatement implements IS
      * @param sql
      */
     /* L0 */ final void initParams(String sql) {
+        encryptionMetadataISRetrieved = false;
         int nParams = 0;
 
         // Figure out the expected number of parameters by counting the
@@ -219,6 +225,7 @@ public class SQLServerPreparedStatement extends SQLServerStatement implements IS
     /* L0 */ public final void clearParameters() throws SQLServerException {
         loggerExternal.entering(getClassNameLogging(), "clearParameters");
         checkClosed();
+        encryptionMetadataISRetrieved = false;
         int i;
         if (inOutParam == null)
             return;
@@ -417,8 +424,9 @@ public class SQLServerPreparedStatement extends SQLServerStatement implements IS
         if ((Util.shouldHonorAEForParameters(stmtColumnEncriptionSetting, connection)) && (0 < inOutParam.length) && !isInternalEncryptionQuery) {
             
             // retrieve paramater encryption metadata if they are not retrieved yet
-            if (null == inOutParam[0].getCryptoMetadata()) {
+            if (!encryptionMetadataISRetrieved) {
                 getParameterEncryptionMetadata(inOutParam);
+                encryptionMetadataISRetrieved = true;
             }
 
             // maxRows is set to 0 when retreving encryption metadata,
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
@@ -424,7 +424,7 @@ public class SQLServerPreparedStatement extends SQLServerStatement implements IS
         if (!encryptionMetadataIsRetrieved) {
             hasNewTypeDefinitions = buildPreparedStrings(inOutParam, false);
         }
-        
+
         if ((Util.shouldHonorAEForParameters(stmtColumnEncriptionSetting, connection)) && (0 < inOutParam.length) && !isInternalEncryptionQuery) {
 
             // retrieve paramater encryption metadata if they are not retrieved yet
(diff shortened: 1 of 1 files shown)


v-xiangs:
We do not have GitHub test for AE yet, so I didn't add a test here. Even if we have AE test, how can we test it explicitly? I tested it manually.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -415,7 +415,11 @@ final void doExecutePreparedStatement(PrepStmtExecCmd command) throws SQLServerE
 
         boolean hasNewTypeDefinitions = buildPreparedStrings(inOutParam, false);
         if ((Util.shouldHonorAEForParameters(stmtColumnEncriptionSetting, connection)) && (0 < inOutParam.length) && !isInternalEncryptionQuery) {
-            getParameterEncryptionMetadata(inOutParam);
+            
+            // retrieve paramater encryption metadata if they are not retrieved yet
+            if (null == inOutParam[0].getCryptoMetadata()) {
+                getParameterEncryptionMetadata(inOutParam);
+            }
```
v-suhame: this fix will work only if the first parameter is encrypted. Consider the case where first parameter is not encrypted, then it will fall back to the old behavior as `inOutParam[0].getCryptoMetadata()` is null.

v-xiangs: thank you. fixed with flag

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -89,6 +89,11 @@
 
     /** Flag set to true when statement execution is expected to return the prepared statement handle */
     private boolean expectPrepStmtHandle = false;
+    
+    /**
+     * Flag set to true when all encryption metadata of inOutParam is retrieved
+     */
+    private boolean encryptionMetadataISRetrieved = false;
```
v-suhame: update the variable name to camel case, `encryptionMetadataIsRetrieved`

v-xiangs: ah.. typo

v-suhame:
double check if [this](https://github.com/Microsoft/mssql-jdbc/pull/220/files#diff-a51517a0e4bf7e029ed04f9476a37f89R434) needs to go inside the if block.

v-xiangs:
@v-suhame we use this if block to increase performance. The line of code you mentioned does not break anything or degrade performance. It just set 2 values

v-suhame:
apply the changes for batch statement too.

v-suhame:
It does not break anything, but I don't see any reason why either this [line ](https://github.com/Microsoft/mssql-jdbc/pull/220/files#diff-a51517a0e4bf7e029ed04f9476a37f89R434)or the one [next](https://github.com/Microsoft/mssql-jdbc/pull/220/files#diff-a51517a0e4bf7e029ed04f9476a37f89R437) to it should be called. We are resting the necessary fields after calling the `sp_describe_parameter_encryption`, so there is no obvious need to reset them when it is not called.

v-xiangs:
@v-suhame I noticed this method too but we do not need to apply changes here: since it's called only when numBatchesExecuted is equal to 0

v-xiangs:
It does break things if we do not call this [line](https://github.com/Microsoft/mssql-jdbc/pull/220/files#diff-a51517a0e4bf7e029ed04f9476a37f89R437). This new logic was introduced to resolve inconsistent problem between (AE on, no encryption) and (AE on and encryption used).
Here is an example:
when AE is on and the column is `smallmoney`, when the column is not encrypted, we should send it as Money; when the column is encrypted, we have to send it as SmallMoney. If we remove that line, it will send it as smallmoney for the second preparedStatement, which is not correct.

v-suhame:
thanks @v-xiangs, that clarifies it. 

If `buildPreparedStrings` has to be called with `renewDefinition` as true every time AE is on, then it would be better to make a single call to `buildPreparedStrings` rather than calling it twice, once with renew as false and later override it with renew as true.

Also, [hasNewTypeDefinitions ](https://github.com/Microsoft/mssql-jdbc/pull/220/files#diff-a51517a0e4bf7e029ed04f9476a37f89R423) should be updated with the result of `buildPreparedStrings`, else driver will prepare the query every single time preparedStatement is executed. In current AE implementation, `sp_prepexec` is called for each pstmt execution, unlike non-AE case where `sp_prepexec`  is called once followed by `sp_execute`

v-xiangs:
@v-suhame Thank you for the very thoughtful code review. It does help me to remember the AE logic.

For the first question:
We have to call `buildPreparedStrings` twice. 
The first call updates `inOutParam` and then `inOutParam` will be used in `getParameterEncryptionMetadata`, otherwise it will throw exception because parameter information is unknown. Also, for the first call, `renewDefinition` has to be false in order to give proper parameter definition.
For the second call, as I have already explained, is necessary to resolve inconsistent parameter definition problem when AE is on and column is not encrypted. The parameter `renewDefinition` as to be true. 

For the second question:
The purpose of `buildPreparedStrings` is to update the information of `inOutParam`, there is no need to update `hasNewTypeDefinitions` since it does not change the its value.

v-suhame:
Let me detail what I meant in the early comment.

If there is a non encrypted smallmoney column, just having AE **on** in connection, and calling `pstmt.execute` twice makes these calls

```
exec sp_prepexec @p1 output,N'@P0 decimal(38,2)',N'INSERT INTO [table] ( [smallMoney_col] ) VALUES(@P0)        ',12.34
exec sp_prepexec @p1 output,N'@P0 decimal(38,2)',N'INSERT INTO [table] ( [smallMoney_col] ) VALUES(@P0)        ',12.34
```

with AE **off**, pstmt is not prepared every time. 
```
exec sp_prepexec @p1 output,N'@P0 decimal(38,2)',N'INSERT INTO [table] ( [smallMoney_col] ) VALUES(@P0)        ',12.34
exec sp_execute 1,12.34
```

As you are aware, `sp_execute` is faster than `sp_prepexec`. It would be better to stick to it. I was suggesting those changes for this to happen.

v-xiangs:
@v-suhame I understand what you meant now. It does improve performance a lot. I am going to make the change now. Thank you.

Issue #230, opened 2017-04-03
SQLServerBulkCopy doesn't check BigDecimal scale


Arthurm1:
When using `SQLServerBulkCopy` to import a `BigDecimal` into a `Numeric` column the scale isn't checked.

If the column metadata is defined as a `Decimal(10, 4)` and then `new BigDecimal("1").setScale(2, RoundingMode.HALF_UP)` is passed the table will end up with a value of 0.01 instead of 1.0

I think it would be safer if the `IOBuffer.writeBigDecimal` was passed `bulkScale` by the `SQLServerBulkCopy.writeColumnToTdsWriter` method and then threw an exception when the `bigDecimalVal.scale()` differed from `bulkScale`

v-suhame:
Yes, using `bulkScale` value in `IOBuffer#writeBigDecimal` is the best option. 
Thanks @Arthurm1! I will submit a fix.

v-suhame:
@Arthurm1 can you review PR#233?

v-suhame:
Closing this issue with fix #251. Thanks again @Arthurm1 !

PR #233, opened 2017-04-05
Big decimal scale precision

v-suhame:
Fix to handle precision and scale for TVP and BulkCopy. Fixes issue #230

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
@@ -19,6 +19,7 @@ import java.io.Reader;
 import java.io.UnsupportedEncodingException;
 import java.math.BigDecimal;
 import java.math.BigInteger;
+import java.math.RoundingMode;
 import java.net.Inet4Address;
 import java.net.Inet6Address;
 import java.net.InetAddress;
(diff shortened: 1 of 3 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -3312,67 +3313,60 @@ void writeDouble(double value) throws SQLServerException {
...
+         * precision. Say, for BigDecimal("0.1"), if scale in metdadata is 0, then ArithmeticException would be thrown if RoundingMode is not set
+         */
+        try {
+            bigDecimalVal = bigDecimalVal.setScale(scale, RoundingMode.HALF_UP);
+        }
+        catch (ArithmeticException e) {
+            MessageFormat form = new MessageFormat(SQLServerException.getErrString("R_valueOutOfRange"));
+            Object[] msgArgs = {SSType.DECIMAL};
+            throw new SQLServerException(form.format(msgArgs), SQLState.DATA_EXCEPTION_DATETIME_FIELD_OVERFLOW, DriverError.NOT_SET, e);
+        }
```
Arthurm1: Either use `RoundingMode.ROUND_UNNECESSARY` or remove the `try/catch` as it will never be thrown with `RoundingMode.HALF_UP`.   My preference is to use `ROUND_UNNECESSARY` as it forces the user to decide how to deal with loss of precision rather than having it silently lose precision.

Also the exception message has a SQL State of `DATA_EXCEPTION_DATETIME_FIELD_OVERFLOW`.  Is there a `DATA_EXCEPTION_DECIMAL_FIELD_OVERFLOW` equivalent that can be used instead?

v-suhame: There is no SQLState specific to decimal, so I have removed the current `DATA_EXCEPTION_DATETIME_FIELD_OVERFLOW`. Thanks for catching it!

We are finalizing on the `RoundingMode` to use, I'll update you once we make a decision.

PR #234, opened 2017-04-05
Tvp server cursor fix

### LINKED ISSUE #224: TVP error with server cursor
populating table with TVP with resultset fails for the following case: with Resultset that uses server cursor (for example, `TYPE_FORWARD_ONLY` and `CONCUR_UPDATABLE`),  if the same connection is used for source result set and the insert query. 
If different connections are used, then it passes.

v-xiangs:
fixes #224

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
@@ -4799,10 +4799,10 @@ final class TDSWriter {
                     TDSReader tdsReader = tdsChannel.getReader(command);
                     int tokenType = tdsReader.peekTokenType();
 
-                    StreamError databaseError = new StreamError();
-                    databaseError.setFromTDS(tdsReader);
-
                     if (TDS.TDS_ERR == tokenType) {
+                        StreamError databaseError = new StreamError();
+                        databaseError.setFromTDS(tdsReader);
+
                         SQLServerException.makeFromDatabaseError(con, null, databaseError.getMessage(), databaseError, false);
                     }
 
--- a/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
@@ -4535,11 +4535,29 @@ final class TDSWriter {
     void writeTVPRows(TVP value) throws SQLServerException {
         boolean isShortValue, isNull;
         int dataLength;
+        
+        boolean tdsWritterCached = false;
 
-        ByteBuffer tempStagingBuffer = ByteBuffer.allocate(stagingBuffer.capacity()).order(stagingBuffer.order());
-        tempStagingBuffer.put(stagingBuffer.array(), 0, stagingBuffer.position());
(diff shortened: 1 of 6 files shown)


src/test/java/com/microsoft/sqlserver/jdbc/tvp/TVPResultSetCursorTest.java
```diff
@@ -0,0 +1,177 @@
...
+        for (int i = 0; i < expectedBigDecimals.length; i++) {
+            pstmt.setBigDecimal(1, expectedBigDecimals[i]);
+            pstmt.setString(2, expectedStrings[i]);
+            pstmt.setTimestamp(3, expectedTimestamps[i]);
+            pstmt.execute();
+        }
+    }
+
+    private static void dropTables() throws SQLException {
+        stmt.executeUpdate("if object_id('" + srcTable + "','U') is not null" + " drop table " + srcTable);
```
v-nisidh: Can we use `Utils.dropTableIfExists` to drop source & dest tables ?

src/test/java/com/microsoft/sqlserver/jdbc/tvp/TVPResultSetCursorTest.java
```diff
@@ -0,0 +1,177 @@
...
+    private static String desTable = "TVPResultSetCursorTest_DestinationTable";
+
+    /**
+     * Test a previous failure when using server cursor and using the same connection to create TVP and result set.
+     * 
+     * @throws SQLException
+     */
+    @Test
+    public void testServerCursors() throws SQLException {
+        conn = DriverManager.getConnection(connectionString);
```
v-nisidh: Can you add one test case for 3 result sets (2 TVPType and one Non-TVP)  & 2 / 3 statements for these resultsets?

v-xiangs: done

v-suhame:
In ResultSet+storedProcedure+Cursor combination, no exception is thrown if there is no storedProcedure or if invalid/incompatible storedProcedure name is passed.

v-xiangs:
@v-suhame fixed exception issue with invalid SP/TVP name

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -4749,10 +4788,40 @@ else if (DataTypes.UNKNOWN_STREAM_LENGTH == dataLength)
...
+                    command.setRequestComplete(false);
+                }
             }
         }
-        // TVP_END_TOKEN
-        writeByte((byte) 0x00);
+
+        // reset command status which have been overwritten
+        if (tdsWritterCached) {
+            command.setRequestComplete(false);
```
v-suhame: Instead of resetting them to default, it will be safe to catch original value of these 3 variables and reset them. Also update these values in synchronized way just as in the existing code.

v-xiangs: Done, thank you for the findings

Issue #206, opened 2017-03-15
Make TnirFirstAttemptTimeoutMs configurable


motymichaely:
Hey,

The current hardcoded setting of TnirFirstAttemptTimeoutMs of 500ms can cause failures on high latency connections (for instance, when running hybrid  cloud environments). I am not sure how turning TransparentNetworkIPResolution off can impact connectivity to AlwaysOn servers but this doesn't sound to be a valid solution when compared to having the ability to set custom timeout value. For instance, having TransparentNetworkIPResolutionTimeoutMs property in the driver can do the trick.

Any objections?

v-xiangs:
Hello @motymichaely. Thank you for filling the issue. We have noticed the issue and started to work on it. Please stay tuned. 
In the meanwhile, you can set transparentNetworkIPResolution to false and set multiSubnetFailover to true. This will skip using of TnirFirstAttemptTimeoutMs and still be able to connection to AlwaysOn servers. Also, if it's high latency network, increasing loginTimeout at the same time will help too. Please let me know if this does not solve the problem.

motymichaely:
@v-xiangs - Thanks! Looking forward for a fix.

I've already tried `transparentNetworkIPResolution=false` together with `multiSubnetFailover=true` but it didn't solve the problem. It only worked when setting `transparentNetworkIPResolution=false` and `multiSubnetFailover=false` but this doesn't seem like a good idea for AlwaysOn servers.

v-xiangs:
@motymichaely Thank you for getting back to us. You are right, setting transparentNetworkIPResolution=false and multiSubnetFailover=false is not a good solution for AlwaysOn servers. Before we fix this issue in the driver, could you please also try increasing the loginTimeout as well? Thank you.

motymichaely:
@v-xiangs - Thanks for clarification.

Unfortunately, loginTimeout has no impact.

v-xiangs:
@motymichaely I am sorry to hear that.. please stay tuned. We will provide a fix for it as soon as possible.

v-xiangs:
Hello @motymichaely , sorry to keep you waiting for so long. If the problem still persists, would you like to try out the new PR #240 we created? With this PR, you will not be able to set `TransparentNetworkIPResolutionTimeoutMs` property, instead, we increased time slice for each connection when setting `transparentNetworkIPResolution=false`. I hope this works for you.

v-xiangs:
[...]

PR #240, opened 2017-04-06
Tnir fix

v-xiangs:
connected to Microsoft/mssql-jdbc#206

Rework on TNIR
1. Turn TNIR off for Fed Auth
2. Change the TNIR multiplier to 0.125 instead of 0.08
3. Change the time slices to a multiple of 2,4,8 instead of 2,3,4
4. In case the timeout for the first slice is less than 500 ms then bump it up to 500 ms
5. Keep TNIR on if the user has explicitly specified it in the connection string.


saurabh500:
LGTM

v-xiangs:
@saurabh500 Thank you for your review. I will merge it soon.

v-nisidh:
connected to Microsoft/mssql-jdbc#206

PR #250, opened 2017-04-11
fix TVP issue with resultset for varchar(max)

### LINKED ISSUE #248: TVP issue with ResultSet when column is varchar(max) and string value is longer than 4000
Driver throws error 
```
Exception in thread "main" com.microsoft.sqlserver.jdbc.SQLServerException: The incoming tabular data stream (TDS) remote procedure call (RPC) protocol stream is incorrect. Table-valued parameter 0 (""), row 1, column 1: Data type 0xE7 has an invalid data length or metadata length.
```
when TVP is used with ResultSet, column is `varchar(max)`, and value length is at least 4001.

It works fine when column is char(8000), varchar(8000), or nvarchar(max).

[tryTVP_RS_varcharMax_4001_Issue.txt](https://github.com/Microsoft/mssql-jdbc/files/911720/tryTVP_RS_varcharMax_4001_Issue.txt)

v-afrafi:
fixes #248

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
@@ -4664,7 +4664,8 @@ final class TDSWriter {
                         case VARCHAR:
                         case NCHAR:
                         case NVARCHAR:
-                            isShortValue = (2 * columnPair.getValue().precision) <= DataTypes.SHORT_VARTYPE_MAX_BYTES;
+                            long columnPrecision = columnPair.getValue().precision;
+                            isShortValue = (2 * columnPrecision) <= DataTypes.SHORT_VARTYPE_MAX_BYTES; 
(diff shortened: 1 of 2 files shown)


v-afrafi:
for issue #248

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -4664,7 +4664,8 @@ void writeTVPRows(TVP value) throws SQLServerException {
                         case VARCHAR:
                         case NCHAR:
                         case NVARCHAR:
-                            isShortValue = (2 * columnPair.getValue().precision) <= DataTypes.SHORT_VARTYPE_MAX_BYTES;
```
v-afrafi: columnPair.getValue().precision was 2147483647 (for varchar(max)) and the multiplication was resulting as 0.
Therefore, seeing it as a short value.

PR #251, opened 2017-04-12
Big decimal scale precision

### LINKED ISSUE #230: SQLServerBulkCopy doesn't check BigDecimal scale

v-suhame:
Fix to handle precision and scale for TVP and BulkCopy. Fixes #230

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
@@ -3323,8 +3323,7 @@ final class TDSWriter {
             int scale) throws SQLServerException {
         /*
          * Length including sign byte One 1-byte unsigned integer that represents the sign of the decimal value (0 => Negative, 1 => positive) One 4-,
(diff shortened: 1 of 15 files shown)


v-suhame:
@Arthurm1 I had to close the older PR because of merge conflicts. We have decided to stick with `RoundingMode.HALF_UP` to be consistent with SQL Server and other Microsoft client drivers of SQL Server.

v-xiangs:
except the `ArrayIndexOutOfBoundsException` we have mentioned, anything else looks good to me. Let's add some tests to this PR before merging.

PR #255, opened 2017-04-18
setNetworkTimeout checks for SQLPermission before proceeding

v-xiangs:
as be mentioned by @mrotteveel as well on this [page](https://docs.oracle.com/javase/7/docs/api/java/sql/Connection.html#setNetworkTimeout(java.util.concurrent.Executor,%20int)), setNetworkTimeout method needs to check for permission.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
@@ -210,8 +210,9 @@ public class SQLServerConnection implements ISQLServerConnection {
     }
 
     // Permission targets
-    // currently only callAbort is implemented
     private static final String callAbortPerm = "callAbort";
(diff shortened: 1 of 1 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -4663,6 +4664,20 @@ public void setNetworkTimeout(Executor executor,
         }
 
         checkClosed();
+        
+        // check for callAbort permission
```
mrotteveel: Wrong permission in comment

v-xiangs: thank you, comment is fixed.

brettwooldridge:
Looks good to me.

Issue #210, opened 2017-03-21
SQLServerException: Data type LONGNVARCHAR not supported in Table-Valued Parameter


rabbit-mom:
hi,I met a problem when I use jdbc 6.1.4.jre7-preview version.A field in a table is "xml" type,I looked up some information on the Internet, the "xml" type is  corresponding with "LONGNVARCHAR",but I use "LONGNVARCHAR" or "LONGVARCHAR" is both not feasible

v-suhame:
@rabbit-mom Thanks for raising this issue! We will add a fix to support xml, along with support for ntext, text, and image. In the meanwhile, you can use VARCHAR or NVARCHAR as the JDBC type to pass xml.

v-afrafi:
@rabbit-mom Could you test Pr #259 . This should fix the issue. Thank you.

v-afrafi:
closing this issue as solved by pr #259. Please feel free to reopen this if you face any issue.  Thanks.

PR #259, opened 2017-04-21
added support for max types for TVP

v-afrafi:
Also longnvarchar, sqlxml, image

### DIFF
--- a/src/test/java/com/microsoft/sqlserver/jdbc/tvp/TVPTypesTest.java
+++ b/src/test/java/com/microsoft/sqlserver/jdbc/tvp/TVPTypesTest.java
@@ -14,6 +14,7 @@ import java.sql.DriverManager;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 
+import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.AfterEach;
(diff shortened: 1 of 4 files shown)


v-suhame:
fix #210

src/test/java/com/microsoft/sqlserver/jdbc/tvp/TVPTypesTest.java
```diff
@@ -0,0 +1,512 @@
...
+        tvp = new SQLServerDataTable();
+        tvp.addColumnMetadata("c1", java.sql.Types.LONGNVARCHAR);
+        tvp.addRow(buffer.toString());
+
+        SQLServerPreparedStatement pstmt = (SQLServerPreparedStatement) connection
+                .prepareStatement("INSERT INTO " + charTable + " select * from ? ;");
+        pstmt.setStructured(1, tvpName, tvp);
+
+        pstmt.execute();
+
```
v-ahibr: It may be better to add verification here (and in testLongVarchar) like you have in testXML

v-afrafi: Added those two missing ones.

PR #260, opened 2017-04-21
Handle ClassCastException 

v-suhame:


v-suhame:
it is the same source code as before, just an extra try/catch, that resulted in formatting a huge chunk of code.

PR #267, opened 2017-04-27
Split out ThreePartNamesParser and cache compiled regex Pattern

sehrope:
This patch splits out ThreePartNamesParser (inner class of SQLServerCallableStatement) into a separate class. Internally, that new class caches the regex Pattern for parsing procedure identifiers (ex: database.owner.proc).

The old code (that this replaces) compiles the regex for each new statement.

The code for the new class has been mostly copied from the previous inline implementation. The only changes are:

* Reordering the properties to match their logical order (database / owner / procedure)
* Changing the property values of the parsed object to be final
* Renaming the parsing function to be a static function named "parse"

The implementation of `ThreePartName` itself is copied wholesale from the prior code.

The second commit simply renames the `translator` variable (done in two commits for readability).

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/ThreePartName.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/ThreePartName.java
@@ -1,0 +1,71 @@
+package com.microsoft.sqlserver.jdbc;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+class ThreePartName {
+    /*
+     * Three part names parsing For metdata calls we parse the procedure name into parts so we can use it in sp_sproc_columns sp_sproc_columns
+     * [[@procedure_name =] 'name'] [,[@procedure_owner =] 'owner'] [,[@procedure_qualifier =] 'qualifier']
+     *
+     */
+    private static final Pattern THREE_PART_NAME = Pattern.compile(JDBCSyntaxTranslator.getSQLIdentifierWithGroups());
+
+    private final String databasePart;
+    private final String ownerPart;
+    private final String procedurePart;
+
+    private ThreePartName(String databasePart, String ownerPart, String procedurePart) {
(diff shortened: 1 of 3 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/ThreePartName.java
```diff
@@ -0,0 +1,71 @@
...
+    private final String ownerPart;
+    private final String procedurePart;
+
+    private ThreePartName(String databasePart, String ownerPart, String procedurePart) {
+        this.databasePart = databasePart;
+        this.ownerPart = ownerPart;
+        this.procedurePart = procedurePart;
+    }
+
+    public String getDatabasePart() {
```
v-afrafi: could we remove the public keyword from the methods?

sehrope: You could and it'd still compile (as it's in the same package as the usage). They're marked public as they're part of the public contract of that class.

If anything, the modifier for the ThreePartName class should be expanded from the default of package-private to public to allow usage from other packages within the driver source. There were not any such usages thus far, as the previous version was an inner class within SQLServerCallableStatement, I didn't mark it as such and figured that could be done when other code needs similar parsing.

v-afrafi: Thanks. What I meant was let's remove the public for now as this is used inside the driver source.

sehrope:
Updated to remove the public modifiers.

I also noticed that SQLServerParameterMetaData had similar parsing code to break out proc identifiers though it was done using explicit tokenization of the dot separated string rather than using a regex. The last commit changes SQLServerParameterMetaData to also use ThreePartName. The end result should be consistent with how it was before and all tests pass.

v-afrafi:
Changes look good! Thank you @sehrope for your contribution!

PR #270, opened 2017-04-28
fix bulk copy cursor issue

### LINKED ISSUE #236: Bulk Copy issue with Result Set and Server Cursor
This issue is very similar with issue #224 .

If the same connection is used to create result set (with server cursor or SelectMethod set to "cursor") and SQLServerBulkCopy object. It fails when writing data with SQLServerBulkCopy object.

Just like issue #224 , it may also have issue when reading data with the same connection.

v-xiangs:
Fixes #236

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
@@ -1555,17 +1555,7 @@ public class SQLServerBulkCopy implements java.lang.AutoCloseable {
 
         try {
             if (!insertRowByRow) {
-                // Create and send the initial command for bulk copy ("INSERT BULK ...").
-                tdsWriter = command.startRequest(TDS.PKT_QUERY);
-                String bulkCmd = createInsertBulkCommand(tdsWriter);
-                tdsWriter.writeString(bulkCmd);
-                TDSParser.parse(command.startResponse(), command.getLogContext());
-
-                // Send the bulk data. This is the BulkLoadBCP TDS stream.
-                tdsWriter = command.startRequest(TDS.PKT_BULK);
-
-                // Write the COLUMNMETADATA token in the stream.
-                writeColumnMetaData(tdsWriter);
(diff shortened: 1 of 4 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
```diff
@@ -3163,6 +3197,24 @@ private boolean writeBatchData(TDSWriter tdsWriter) throws SQLServerException {
...
+            
+            
+            if (insertRowByRow) {
+                // read response gotten from goToNextRow()
+                ((SQLServerResultSet) sourceResultSet).getTDSReader().readPacket();
+
+                // Create and send the initial command for bulk copy ("INSERT BULK ...").
+                tdsWriter = command.startRequest(TDS.PKT_QUERY);
+                String bulkCmd = createInsertBulkCommand(tdsWriter);
+                tdsWriter.writeString(bulkCmd);
```
v-suhame: As the `bulkCmd` is going to be same with and w/o cursors, let this be in SQLServerBulkCopy#doInsertBulk(as before) and not inside a while loop.

v-xiangs: With cursor, we need to send data row by row, so it has to be inside the while loop. Also, to pass by MARS, the sending must happen after the reading, which is goToNextRow()

v-suhame: sending 'bulkCmd' in loop is fine, but driver will generate the same `insert bulk .....` command again and again in line 3208.

v-xiangs: the thing is, we have to reset the `tdsWriter` after `goToNextRow()`, and `createInsertBulkCommand` requires the new `tdsWriter`, so, it has to be here, otherwise it fails when column is encrypted

v-suhame: Got it! thanks for details Shwan!

PR #272, opened 2017-04-29
Metadata caching - Parsed SQL and prepared statement handle caching

TobiasSQL:
Still a prototype, opened PR to allow for conversation.


sehrope:
If the parse cache is going to be shared across multiple connections you'll need a way to evict invalid entries. Otherwise the following will break a long running application until it restarts:

1. App executes SQL
2. SQL metadata is cached with value M1
3. DDL change on database - now SQL would parse to M2
4. Usage of cached M1 metadata leads to an exception

Step 4 would repeatedly give an error with no way to succeed till the app is bounced and the cache is reset.

One solution to this would be to have entries evicted if methods that are using the cache throw an exception. An error would still happen if the cache is out of date but at least it'd only happen once per running application. Subsequent attempts to repeat the same operation would then succeed as the next attempt would get fresh metadata.

Alternatively, having the cache be at the Statement level mostly avoids this issue as you'd only run into it if the DDL change happens in the middle of repeated usage of the same Statement. As most JDBC usage involves creating Statements and then closing them when the unit of work is complete, the errant window would be much smaller.

Breaking it out into pros / cons:

* Shared Connection Cache
    - **Pro:** Most memory efficient
    - **Pro:** Highest cache hits as shared across entire pool
    - **Con:** Requires more complicated eviction to deal with DDL changes (evict on error?)
    - **Con:** Must be thread safe (not a huge deal but still a minor con)
* Per Connection Cache
    - **Pro:** Medium memory efficiency
    - **Pro:** Medium cache hit rate as there will be reuse for pooled connections and shared across Statements.
    - **Pro:** Eviction from connection pool provides convenient place to get rid of a stale cache (i.e. just rebuild the pool)
    - **Pro:** Don't have to worry about thread safety or synchronization as Connections are only meant to be used by one Thread at a time.
    - **Con:** Requires more complicated eviction to deal with DDL changes
    - **Con:** May require time based eviction to prevent stale SQL from remaining cached forever though could defer to connection pool (e.g. configure your pool to evict connections from the pool after a max age)
* Per Statement Cache
    - **Pro:** Simplest to implement
    - **Pro:** Smallest "error window" when dealing DDL changes that should lead to cache invalidations
     - **Pro:** Don't have to worry about thread safety or synchronization as Statements are only meant to be used by one Thread at a time.
    - **Con:** Least efficient memory wise as each Statement would have its own cache.
    - **Con:** Least efficient cache hit wise as each Statement would have its own cache.
 
Note that I'm not advocating one over the other ... just providing some clarity on options. If it can be disabled via connection properties and gracefully handle errant evictions, the shared cache would likely be the best for the common case where DDL changes are not happening.

TobiasSQL:
Wow, @brettwooldridge & @sehrope, thanks for all of the feedback!

I just added a simple implementation of actual reuse of created server-side prepared statement handles. The prepared statment handle cache is currently controlled by the _SQLServerConnection.preparedStatementHandleCacheSize_SHOULD_BE_CONNECTION_STRING_PROPERTY_ field, a value > 0 means cache is on :-). This cache should further improve performance.

We now have two caches:
1. **SQLServerPreparedStatement.parsedSQLCache** (renamed per feedback) - Multi-connection/static cache. Contains metadata from parsing the SQL text on the client-side to avoid repeated parsing.
2. **SQLServerConnection.preparedStatementHandleCache** - Per connection cache. Contains prepared statement handles to be re-used. 

Trying to reply as best I can:
1. Curious on why not to take the dependency on the Guava LRU cache. Is this mainly about download size? 
2. @brettwooldridge, I renamed **preparedStatementMetadataSQLCache** to **parsedSQLCache** per request, way clearer, thanks! 
3. I don't see why the **parsedSQLCache** should not be used across connections (and be static), it is simply tracking results of parsed strings and does not rely on connection properties. The new prepared statement handle cache I added needs to be per connection.
4. @brettwooldridge, good point of the **ensureSQLSyntax()** method, will start tracking that one as well.
5. @brettwooldridge, agree on not using the SQL text as key (doing POC first). Will move to a hash in a subsequent commit.
6. @brettwooldridge, sqlCommand is the original SQL text before processing, that is what we need to build the key on. Can't find it based on the post-processed SQL string. Agree that naming of userSQL vs. sqlCommand is not clear here (userSQL in fact NOT being user-provided-SQL...)
7. @sehrope, given that the parsedSQLCache has no relationship to the server at all, all it does is track metadata related to client-side-only parsing. So it does not need to be invalidated based on what happens on the server :-) Do you think we should clarify the name further to make it clearer that it is caching related to metadata from SQL-text only parsing?

Thanks!
-Tobias

brettwooldridge:
@TobiasSQL 

Regarding the Guava dependency, a few things.
 * Yes, one is size.  That is a 2.5MB dependency for an LRU cache that can be implemented in ~100 lines of code.
 * Another is that projects using the current v6 driver (or before) cannot simply drop it in, as they will be broken without that dependency.
 * As far as I know, no other JDBC drivers allow external dependencies.  If they do in fact require them, they shade them.  Not that I am suggesting shading Guava, because of its size -- even if you only wanted the dependent graph of classes for ``Cache``.

------------------------------------------------------------------------
Regarding the static cache, I think you are correct.  As a pure parsed SQL cache, I don't see a reason not to share VM-wide.  I was thinking ahead, and injecting my own assumptions about implementation, assuming that the metadata cache would eventually be married with the parsed SQL cache.  But that is not necessarily so, clearly.

------------------------------------------------------------------------
Regarding the ``sqlCommand``, the only place it is referenced outside of the constructor is ``buildExecuteMetaData()``:
```java
988  private ResultSet buildExecuteMetaData() throws SQLServerException {
989     String fmtSQL = sqlCommand;
990     if (fmtSQL.indexOf(LEFT_CURLY_BRACKET) >= 0) {
991        fmtSQL = (new JDBCSyntaxTranslator()).translate(fmtSQL);
992     }
993  
994     ResultSet emptyResultSet = null;
995     try {
996        fmtSQL = replaceMarkerWithNull(fmtSQL);
      ...
```
But it seems the result of line ``991`` *must* be the same as ``userSQL`` in the constructor, right?
```java
199     userSQL = translator.translate(sql);
```

------------------------------------------------------------------------
Great to see the appearance of a metadata cache ... or the beginnings of one!  🎉 

I'll need a closer reading before, so take these comments with a grain of salt...

I think the prepared statement handles can be shared across Connections, right?  I believe the server-side access plans are shared across connections, so it should be possible -- assuming an access scheme keyed by the hostname/port/database like that discussed above.

I also believe that in the end, one of the biggest performance wins will come from caching the metadata returned by ``sp_prepare`` with option ``RETURN_METADATA`` along side the prepared handle.  I'm not sure, but one short-circuit location would likely be in ``buildParamTypeDefinitions()``, and another in ``buildExecuteMetaData()``.

TobiasSQL:
Thx @brettwooldridge!

For Guava your point on not being a simple drop-in replacement anymore settles it for me. Will replace :-)

Execution plans are shared across connections, and even users but prepared handles are not. They are tied to a specific connection. You can see if you try to do sp_prepare on two connections that both will get handle number 1 back. 

Grain of salt applied :-) WRT to the metadata from sp_prepare we actually never call sp_prepare. All of our drivers now call sp_prepexec instead (which always returns metadata). I.e. we don't do the prepare until the first execution and then we prepare and execute at the same time. Following executes call sp_execute. buildExecuteMetaData seems to be only used if you want metadata for a statement without executing (in fact this should be changed from using set fmtonly to sp_describe_first_result_set).

Good point on row 991! Like you say it should be changed to:
 _fmtSQL = userSQL;_

Made the change :-)

Thanks!
-Tobias

brettwooldridge:
@TobiasSQL Most Java ORMs call the equivalent of Connection.prepareStatement() followed by stmt.getParameterMetaData().  It was the poor performance of that pattern in jTDS that led to my implementing metadata caching back in the day.  However we implement the metadata caching, it is good to keep that common pattern in many NLD.

v-afrafi:
@TobiasSQL Could you please open this pr to dev branch? Thank you.

TobiasSQL:
@v-afrafi, changed to dev, thanks for noticing!

cogman:
As an alternative to Guava's cache, how about caffeine? ( https://github.com/ben-manes/caffeine ) It is a near drop-in replacement for Guava's cache,  it was written by the guy that made the original guava cache (with lessons learned), and it has a much smaller footprint in general compared to guava.

ben-manes:
A synchronized LRU is supported by Java's `LinkedHashMap` as a tiny extension, and can be written in less than 10 LOC. However, it does not perform well under concurrent workloads.

Attempting to solve that led to [ConcurrentLinkedHashMap](https://github.com/ben-manes/concurrentlinkedhashmap). This was used heavily (Cassandra, etc), and often embedded (Groovy, Camel, etc). It is small and JDK6 based.

Guava's Cache was [retrofitted](https://github.com/ben-manes/concurrentlinkedhashmap/blob/wiki/ConcurrentCachingAtGoogle.pdf) with a CLHM's design and embedding many features worked out custom decorators. This provides many best practices, such as using memoization rather than racy get-compute-put. Unfortunately its original design around soft references hindered our efforts, since we had to work within that framework rather than from a stronger foundation.

Caffeine is a JDK8 rewrite that improves on the above. It uses a modern eviction policy, offers more features, and has had very healthy adoption. However, this richness comes at the cost of a modestly sized jar. Some do shade it, but its probably not ideal for a jdbc driver.

I'd recommend embedding CLHM as it is probably the best fit.

TobiasSQL:
@ben-manes & @cogman, thanks! Given concurrency is important for at least the parsed SQL text cache, CLHM sounds like a good option to embed. I will take a closer look at it once I get stuff into reasonable shape :-)

@brettwooldridge , thanks for pointing out the ORM pattern using getParameterMetaData(), makes sense. I reworked things to also handle caching this metadata. 

Overall I am reworking a bunch of stuff from first prototype as it has a few issues, should hopefully be able to push some changes soon.

TobiasSQL:
Cleaned up prototype and fixed a bunch of issues. Depending on feedback next POR is to replace cache impl. and start using hash for cache key.

cogman:
I would suggest against using Sha-1 for the hash key.  Sha-1 was designed to reduce collisions in a cryptographically safe manor, but it is slow.  Faster algorithms exist which have low collision rates.  Particularly, fnv3 or murmur would be preferable as high performance hashing algorithms with low collision rates.

With that said, it probably would be a better idea to use the string as the key but limit the size that will be cached, since large statements are less likely to be reused.  If you are concerned about size, then you could compress the statement.  But I wouldn't do that as a first step.

TobiasSQL:
@cogman, after running my perf suite I agree with you. Using SHA1 more or less undid the perf gain from caching the statement handle in the first place :-( Also, after pondering this a bit more, given the string is anyway used by the prepared statement instances it seems like adding a hash will actually use more memory...

brettwooldridge:
@TobiasSQL Your hash code had a bug that caused 100% cache miss.  I suggest trying it again.  A [quick and dirty bench](https://gist.github.com/brettwooldridge/3e470910c1b6b7c13d14c647cfbede70) on my circa 2013 i5 desktop yielded 1000 SHA1 hashes in 1.7ms, or roughly 1.7 microseconds per.

brettwooldridge:
@TobiasSQL Dump the Apache Commons Digest, drop the use of ``ByteBuffer`` its ``equals()`` and ``hashCode()`` are unnecessarily expensive.

Try:
```java
static private Cache< Sha1HashKey, ParsedSQLCacheItem> parsedSQLCache;
...
private static class Sha1HashKey {
   private byte[] bytes;

   Sha1HashKey(String s) {
      this.bytes = getSha1Digest().digest(s.getBytes());
   }

   public boolean equals(Object obj) {
      return Arrays.equals(bytes, (byte[]) obj);
   }

   public int hashCode() {
      return Arrays.hashCode(bytes);
   }

   private MessageDigest getSha1Digest() {
      try {
         return MessageDigest.getInstance("SHA-1");
      }
      catch (final NoSuchAlgorithmException e) {
         // this is not theoretically possible, but we're forced to catch it anyway
         throw new RuntimeException(e);
      }
   }
}
```

TobiasSQL:
Thanks @brettwooldridge, this is a bit strange since the first thing I tried was close to exactly the code that you propose (using Arrays.equals and Arrays.hashCode) which didn't work. I am sure I did something wrong, will give this a try :-)

Meanwhile I have replaced the use of the Guava cache to the CLHM tha @ben-manes suggested :-)

brettwooldridge:
@cogman

At 1.7 microseconds overhead per statement, I have a hard time believing that SHA1 calculation "undoes the perf gain from caching the statement handle"...

I would suggest that possibly other inefficiencies in the implementation, including [this code](https://github.com/Microsoft/mssql-jdbc/pull/272/commits/dab1c6bf93a572dbcd7f9582d786acc7238e758b#diff-a51517a0e4bf7e029ed04f9476a37f89L159) which seems to result in a 100% cache miss, the use of ByteBuffer as a cache key which has a fairly high overhead for ``equals()/hashCode()`` are much more likely to have affected benchmarks.

How are you benchmarking?  If I take the JIT warm-up out of my benchmark (10000 iterations are required to trigger C2 compilation) my benchmark times jump up by over 50%.  Warm-up is essential to benchmarking on the JVM; any kind of "cold" test is going to give skewed results ... potentially *very* skewed depending on the code in question.

brettwooldridge:
@TobiasSQL Curious about the statement:

>given the string is anyway used by the prepared statement instances it seems like adding a hash will actually use more memory...

Why is the string used anyway?  After transformation, there seems no point to holding onto the *original* SQL string anywhere in the prepared statement instance, right?

TobiasSQL:
Good point @brettwooldridge. First off I was assuming it was public which it is not. After my changes so far it is only used in **buildExecuteMetaData** where it seems useless to me. If I remove that reference you are right, there seems like no point to keep track of it and it can be removed.

For the statement handle cache it is different since it uses the processed SQL (userSQL) as the key. For that case using a hash still seems unnecessary from memory usage POV.

TobiasSQL:
@brettwooldridge, thanks for the catch of the 100% cache misses for the parsed SQL text. That was clearly the main issue, sorry for wasting your time! 

After re-running my simple perf suite (with warm-up) I get inclusive results on hash vs. not (hash clearly adds overhead, not clear on %), need to run some real benchmarking on this when I get back from vacation :-) 

Will push the change including CLHM replacement for Guava and hashing for parsed SQL text lookup. @brettwooldridge et al, if you have time to compare the two (simple change in SQLServerPreparedStatement in 3 places) that would awesome :-)

brettwooldridge:
@TobiasSQL Re: taking at look, will do.

Re: the handle cache using the *processed* SQL (userSQL) as the key, does it need to?  Or can it use the user provided SQL, in which case maybe the SHA1 hash could be reused and the memory overhead cut?

TobiasSQL:
@brettwooldridge, you are right, I can just hash once and keep the hash around and use it for both caches. Very good point!

TobiasSQL:
FYI, trying to push but connection is very slow from this Caribbean cruise ship I am on, **git** isn't liking it...

cogman:
@brettwooldridge  Yeah, mainly my problem with SHA-1 wasn't so much the overhead, just that it is the wrong tool for this sort of a job.  The same as it would be wrong to use a List where a Set is wanted and vice versa.

I question the usage of it, mainly because it looks like it is mostly there for memory saving purposes.  As @TobiasSQL points out, it looks like the string is being saved off anyways as part of the statement.  If that weren't the case, I would still suggest using a non-cryptographic hashing algorithm.  We don't really care if an attacker can cause a collision.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -2699,6 +2781,10 @@ public void close() throws SQLServerException {
             tdsChannel.close();
         }
 
+        // Invalidate statement cache.
+        if(null != this.preparedStatementCache)
+            this.preparedStatementCache.clear();
```
ben-manes: The `EvictionListener` is not called on an explicit removal, only automatic (hence the switch to `RemovalListener` in Guava). You probably want to iterate the keys, remove, and forward the value to your discard logic.

TobiasSQL: For this case the intention is luckily not to run the removal logic because all statement handles on the server side are discarded when the connection is closed. Thanks for reviewing, this is a pretty big change so definitely needs a thorough go over :-)

brettwooldridge:
@cogman I wouldn't say SHA1 is the wrong tool necessarily, after all git itself uses SHA1 hashes.  Certainly it is worth exploring faster hashes, but the collision rates of hashes like CityHash, Murmur3, etc. are still unacceptably high.  Any 128-bit hash is likely insufficient.  FNV in the 256-bit variant might fit.  How the performance of FNV-256 compares to SHA1 (which enjoys some specific x86 CPU instructions), I do not know.

TobiasSQL:
10 was a random pick (and hasn't really been picked yet...), we haven't discussed a default size but now is as good a time as any :-)

This cache doesn't really use a lot of memory on the client side after we hashed the key and removed the need for keeping additional copies of the SQL text. On the client-side each cache entry really only includes the SHA1 hash (20 bytes), the handle (4 bytes) as well as a few booleans and the reference to the connection plus whatever overhead is used by the cache item instances and the cache itself. Overall cheap. If we also cache parameter metadata (after a request to getParameterMetadata) then all of the information from that instance is also part of the cache entry.  

All of this is per connection so with lots of connections it adds up, depending on the H/W you run on (thinking devices/sensors) we may want to be conservative. On the server side we basically keep the hash of the SQL text, the handle, as well as the actual SQL text. So the server-side is for sure more expensive. The server side also tends to be the critical resource since it is shared. 

10 seemed like a safe first choice which should provide good handle re-use for many workloads.

cogman:
@brettwooldridge  " I wouldn't say SHA1 is the wrong tool necessarily, after all git itself uses SHA1 hashes"

Git doesn't use SHA1 hashes for cache lookups.  Nor is it particularly chuffed about the performance of it's hashing algorithm.  SHA1 in git is to provide a unique key with low collision probability, data verification, and until recently, for commit signing.

*****

After googling the SHA extensions on Intel and Ryzen, I must say I'm really impressed at what Intel and AMD have done here.  I was going to say something like "A multiplication + xor is going to be hard to beat" for FNV, however, the SHA extensions are really crazy optimized.  Some of my reading put it at around 10 cycles/byte.  Really not that bad, especially considering statements shouldn't generally be many bytes.

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/RegressionTest.java
```diff
@@ -122,6 +127,109 @@ public void testSelectIntoUpdateCount() throws SQLException {
...
+
+        /*
+         * populate table
+         */
+        sql = "insert into " + tableName + " values(" + "?,?" + ")";
+        pstmt =  (connection).prepareStatement(sql, ResultSet.TYPE_FORWARD_ONLY,
+                ResultSet.CONCUR_READ_ONLY, connection.getHoldability());
+        
+        for (int i = 1; i <= rows; i++) {
+            pstmt.setObject(1, i, JDBCType.INTEGER);
```
v-afrafi: changing this to pstmt.setObject(1, i, java.sql.Types.INTEGER); should solve the issue for java 7.

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/RegressionTest.java
```diff
@@ -122,6 +127,109 @@ public void testSelectIntoUpdateCount() throws SQLException {
...
+     * Tests update query
+     * 
+     * @throws SQLException
+     */
+    @Test
+    public void testUpdateQuery() throws SQLException {
+        SQLServerConnection con = (SQLServerConnection) DriverManager.getConnection(connectionString);
+        String sql;
+        PreparedStatement pstmt = null;
+        JDBCType[] targets = {JDBCType.INTEGER, JDBCType.SMALLINT};
```
v-afrafi: also here changing to int[] targets = {java.sql.Types.INTEGER, java.sql.Types.SMALLINT}; should solve issue for java 7 issue.

brettwooldridge:
I have a concern about the ``com.googlecode.*`` classes.  If the user has an incompatible version of those classes in their classpath before the driver's jar, we will get unexpected failures.  If we include these files, they should be relocated to a different package, such as ``mssql.googlecode.*`` or similar.

TobiasSQL:
Good catch @brettwooldridge, thx!

TobiasSQL:
As an FYI the commits coming in now are fixes based on findings from our test suites that have not been GitHub:ed yet.

brettwooldridge:
Got it.  I'll run the benchmark later today.

brettwooldridge:
@TobiasSQL Have you had a look at my ``metadataCaching2`` branch?   I replaced the SHA1 hash with CityHash128 [which performed very well, with no collisions over the corpus](https://github.com/rurban/smhasher/blob/master/doc/CityCrc128).  In a local SHA1 vs. CityHash128 benchmark, CityHash128 was ~10x faster, at roughly 160ns per hash.

The reference counting is tightened up further still.

brettwooldridge:
Here are the benchmark results, Best of Three runs:

**mssql-jdbc-6.1.7.jre8.preview.jar**
```
22:39:42,654 (DBWorkload.java:614) INFO  - Workload Histograms:
Completed Transactions:
Persist/01           [11697] *******************************************************************************
Retrieve/02          [11536] ******************************************************************************
Update/03            [11774] ********************************************************************************
Delete/04            [11657] *******************************************************************************
```

**mssql-jdbc-6.1.8-SNAPSHOT.jre8.jar** (``mssql-jdbc/metadataCaching`` branch):
```
22:49:37,774 (DBWorkload.java:614) INFO  - Workload Histograms:
Completed Transactions:
Persist/01           [12321] ********************************************************************************
Retrieve/02          [12120] ******************************************************************************
Update/03            [12112] ******************************************************************************
Delete/04            [12080] ******************************************************************************
```

**mssql-jdbc-6.1.8-SNAPSHOT.jre8.jar** (``brettwooldridge/metadataCaching2`` branch):
```
22:58:19,075 (DBWorkload.java:614) INFO  - Workload Histograms:
Completed Transactions:
Persist/01           [12564] *******************************************************************************
Retrieve/02          [12595] ********************************************************************************
Update/03            [12457] *******************************************************************************
Delete/04            [12579] *******************************************************************************
```

brettwooldridge:
Previous benchmark was the [oltpbenchmark](https://github.com/oltpbenchmark/oltpbench) ``jpab`` benchmark.  This is the *oltpbenchmark* ``epinions`` benchmark:

**mssql-jdbc-6.1.7.jre8.preview.jar**
```
23:59:17,238 (DBWorkload.java:614) INFO  - Workload Histograms:
Completed Transactions:
GetReviewItemById/01 [33985] ***************************************
GetReviewsByUser/02  [33999] ***************************************
GetAverageRatingB... [33961] ***************************************
GetItemAverageRat... [33762] ***************************************
GetItemReviewsByT... [33722] ***************************************
UpdateUserName/06    [33919] ***************************************
UpdateItemTitle/07   [33828] ***************************************
UpdateReviewRatin... [33584] ***************************************
UpdateTrustRating/09 [68105] ********************************************************************************
```

**mssql-jdbc-6.1.8-SNAPSHOT.jre8.jar** (``mssql-jdbc/metadataCaching`` branch):
```
23:50:19,993 (DBWorkload.java:614) INFO  - Workload Histograms:
Completed Transactions:
GetReviewItemById/01 [34305] ****************************************
GetReviewsByUser/02  [34392] ****************************************
GetAverageRatingB... [34216] ***************************************
GetItemAverageRat... [34175] ***************************************
GetItemReviewsByT... [34087] ***************************************
UpdateUserName/06    [34530] ****************************************
UpdateItemTitle/07   [34278] ****************************************
UpdateReviewRatin... [34270] ****************************************
UpdateTrustRating/09 [68526] ********************************************************************************
```

**mssql-jdbc-6.1.8-SNAPSHOT.jre8.jar** (``brettwooldridge/metadataCaching2`` branch):
```
23:39:41,015 (DBWorkload.java:614) INFO  - Workload Histograms:
Completed Transactions:
GetReviewItemById/01 [34670] ****************************************
GetReviewsByUser/02  [34595] ***************************************
GetAverageRatingB... [34911] ****************************************
GetItemAverageRat... [35003] ****************************************
GetItemReviewsByT... [34670] ****************************************
UpdateUserName/06    [35248] ****************************************
UpdateItemTitle/07   [34690] ****************************************
UpdateReviewRatin... [34823] ****************************************
UpdateTrustRating/09 [69330] ********************************************************************************
```

TobiasSQL:
Left comments for your PR @brettwooldridge, sorry for the delay! I am hoping we get the current caching finalized for release next week. The week after should hopefully work great, I am thinking you can then submit a PR against the dev branch (rather than metadataCaching). We mainly need to make sure all issues found by the other test cases are resolved first.

brettwooldridge:
@TobiasSQL Sure, sounds great.  I'm still testing and tweaking my ``speedify`` branch. 👍  But current results are looking outstanding.

``oltpbench/jpab`` benchmark (Hibernate) running a transaction mix of 5/75/15/5 (Persist/Retrieve/Update/Delete):

**mssql-jdbc-6.1.7.jre8.preview.jar** 🐢 
```
INFO  - Workload Histograms:
Completed Transactions:
Persist/01           [ 4848] *****
Retrieve/02          [72439] ********************************************************************************
Update/03            [14412] ***************
Delete/04            [ 4844] *****
```

**mssql-jdbc-6.1.8.jre8.preview.jar** (``speedify``) 🏇 
```
INFO  - Workload Histograms:
Completed Transactions:
Persist/01           [ 6364] *****
Retrieve/02          [96709] ********************************************************************************
Update/03            [19439] ****************
Delete/04            [ 6487] *****
```

TobiasSQL:
Beautiful! @brettwooldridge, is the comparison branch based on current state of metadataCaching, dev, or master, or something else?

brettwooldridge:
The comparison was with v6.1.7.

TobiasSQL:
@brettwooldridge, you can now open a PR against dev for your improvements :-)

TobiasSQL:
Never mind, we have some more things to go through before the merge. Expecting it next week.

TobiasSQL:
@brettwooldridge, this is a great point. Agree with the implementation, makes sense with calling setCapacity to me.

I was noodling on this as well, not sure about tying it to the discard queue they seem to have at best a distant relationship ;-) 

Two options come to mind:
1. Expose a separate static configuration knob for the parse cache.
2. Tie the parse cache to the statement cache size (for 5x statement cache size, potentially expose the ratio)

Looking around, additionally I think we need to provide a limit for the max statement text size to place in the parse cache to prevent accidental memory hogging. My suggestion is to start with a conservative and fairly large value (say 10k), we can later make it configurable in a following release.

Thoughts?

TobiasSQL:
Given that the statement cache isn't static (which I forgot) it seems to me that only option 1 above makes sense, i.e. separate configuration knob.

Issue #161, opened 2017-02-27
Temporal data types are forced as Varchar for "BulkRecord"


jackielii:
I tried to implement my own "ISQLServerBulkRecord" based on a "List" type, everything was fine until I discovered I have to convert the Date value to String because of:

[SQLServerBulkCopy.java#L2364](https://github.com/Microsoft/mssql-jdbc/blob/master/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java#L1406)
```java
case java.sql.Types.DATE:
    /*
     * If encrypted, varbinary will be returned before. The code will come here only if unencrypted. For unencrypted bulk copy if the
     * source is CSV, we send the data as varchar and SQL Server will do the conversion. if the source is ResultSet, we send the data as
     * the corresponding temporal type.
     */
    if (null != sourceBulkRecord) {
        return "varchar(" + ((0 == bulkPrecision) ? destPrecision : bulkPrecision) + ")";
    }
    else {
        return "date";
    }
```

Why are we assuming BulkRecord is "CSV" file? I have the date already, I shouldn't need to format it to string and let sql server on the other side to interpret it back to date right?

v-suhame:
@jackielii Thanks for raising this issue. It's a good question, we will fix the issue after analyzing the impact it will have on existing users.

jackielii:
@v-suhame I think the fundamental issue here is:

How can effectively implement my own class that can be accept as input to bulk copy?

v-suhame:
Can you give more detail on what your application is trying to achieve? More specifically, what you are expecting from an `effective implementation`, does it mean less overload on the SQL Server? 

With the driver's current implementation, you can still pass `Date` object to `ISQLServerBulkRecord` and it should work. SQL Server will take care of parsing the value from `Varchar` to `Date`. Even in case of `ResultSet`, the `Date` is still converted to `String` to obtain `Timestamp`.

v-suhame:
Passing smalldatetime via ISQLServerBulkRecord throws `Received an invalid column length from the bcp client for colid 1.`. By default smalldateime doesn't have scale or precision, passing zero as scale and precision throws the exception.

Repro code: 
[ImpISQLServerBulkRecord.txt](https://github.com/Microsoft/mssql-jdbc/files/945432/ImpISQLServerBulkRecord.txt)

v-xiangs:
Hello @jackielii , thank you for the report. 

[...]

PR #275, opened 2017-05-02
For invalid error message by passing smalldatetime in ISQLBulkRecord

### LINKED ISSUE #280: Passing smalldatetime via ISQLServerBulkRecord is throwing exception
Passing smalldatetime via ISQLServerBulkRecord throws `Received an invalid column length from the bcp client for colid 1..` By default smalldateime doesn't have scale or precision, passing zero as scale and precision throws the exception.
https://github.com/Microsoft/mssql-jdbc/issues/161#issuecomment-295906312

v-afrafi:
Fixes #280 
For issue #280 also mentioned here. https://github.com/Microsoft/mssql-jdbc/issues/161#issuecomment-295906312

Passing smalldatetime via ISQLBulkRecord without specifiying precision should work. 
Also sending out of range values for character type and binary type should throw proper error rather than `Received an invalid column length from the bcp client for colid 1`.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
@@ -1240,6 +1240,11 @@ public class SQLServerBulkCopy implements java.lang.AutoCloseable {
             int destColIndx,
             TDSWriter tdsWriter) throws SQLServerException {
         boolean isStreaming;
+        /**
+         * The maximum temporal precision we can send when using varchar(precision) in bulkcommand, to send a smalldatetime/datetime 
+         * value.
+         */
+        int sourceBulkRecordTemporalMaxPrecision = 50;
         SSType destSSType = (null != destColumnMetadata.get(destColIndx).cryptoMeta)
                 ? destColumnMetadata.get(destColIndx).cryptoMeta.baseTypeInfo.getSSType() : destColumnMetadata.get(destColIndx).ssType;
 
@@ -1378,14 +1383,14 @@ public class SQLServerBulkCopy implements java.lang.AutoCloseable {
                 switch (destSSType) {
                     case SMALLDATETIME:
                         if (null != sourceBulkRecord) {
-                            return "varchar(" + ((0 == bulkPrecision) ? destPrecision : bulkPrecision) + ")";
+                            return "varchar(" + ((0 == bulkPrecision) ? sourceBulkRecordTemporalMaxPrecision  : bulkPrecision) + ")";
                         }
                         else {
                             return "smalldatetime";
                         }
                     case DATETIME:
                         if (null != sourceBulkRecord) {
-                            return "varchar(" + ((0 == bulkPrecision) ? destPrecision : bulkPrecision) + ")";
+                            return "varchar(" + ((0 == bulkPrecision) ? sourceBulkRecordTemporalMaxPrecision  : bulkPrecision) + ")";
                         }
                         else {
                             return "datetime";
@@ -1648,7 +1653,11 @@ public class SQLServerBulkCopy implements java.lang.AutoCloseable {
 
         if ((Util.isCharType(srcJdbcType) && Util.isCharType(destSSType)) || (Util.isBinaryType(srcJdbcType) && Util.isBinaryType(destSSType))) {
             if (colValue instanceof String) {
-                sourcePrecision = ((String) colValue).length();
+                if (Util.isBinaryType(destSSType)) {  // if the dest value is binary and the value is of type string
+                    sourcePrecision = (((String) colValue).getBytes().length) / 2;
+                }
+                else
+                    sourcePrecision = ((String) colValue).length();
             }
             else if (colValue instanceof byte[]) {
                 sourcePrecision = ((byte[]) colValue).length;
@@ -2623,6 +2632,10 @@ public class SQLServerBulkCopy implements java.lang.AutoCloseable {
                 validateDataTypeConversions(srcColOrdinal, destColOrdinal);
             }
         }
+        //If we are using ISQLBulckRecord and the data we are passing is char type, we need to check the source and dest precision
+        else if (null != sourceBulkRecord) {
+            validateStringBinaryLengths(colValue, srcColOrdinal, destColOrdinal);
+        }
         else if ((null != sourceBulkRecord) && (null != destCryptoMeta)) {
             // From CSV to encrypted column. Convert to respective object.
             if ((java.sql.Types.DATE == srcJdbcType) || (java.sql.Types.TIME == srcJdbcType) || (java.sql.Types.TIMESTAMP == srcJdbcType)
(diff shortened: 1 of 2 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
```diff
@@ -2623,6 +2634,10 @@ private void writeColumn(TDSWriter tdsWriter,
                 validateDataTypeConversions(srcColOrdinal, destColOrdinal);
             }
         }
+        //If we are using ISQLBulckRecord and the data we are passing is char type, we need to check the source and dest precision
```
v-ahibr: small typo, should be ISQLBulkRecord

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/ImpISQLServerBulkRecord_IssuesTest.java
```diff
@@ -0,0 +1,435 @@
...
+            bcOperation.setDestinationTableName(destTable);
+            bcOperation.writeToServer(bData);
+            bcOperation.close();
+        }
+        catch (Exception e) {
+            if (e instanceof SQLServerException) {
+                assertTrue(e.getMessage().contains("The given value of type"), "Invalid Error message: " + e.toString());
+            }
+        }
+        ResultSet rs = stmt.executeQuery("select * from " + destTable);
```
v-ahibr: this returns an empty resultset since nothing was inserted. No point in the while loop

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/ImpISQLServerBulkRecord_IssuesTest.java
```diff
@@ -0,0 +1,435 @@
...
+        BulkDat bData = new BulkDat(variation);
+        String value = "aa";
+        query = "CREATE TABLE " + destTable + " (smallDATA varchar(2))";
+        stmt.executeUpdate(query);
+
+        try {
+            SQLServerBulkCopy bcOperation = new SQLServerBulkCopy(connectionString);
+            bcOperation.setDestinationTableName(destTable);
+            bcOperation.writeToServer(bData);
+            bcOperation.close();
```
v-ahibr: after bcOperation.close(), you should force the test to fail with something like:
`fail("BulkCopy executed when it it was expected to fail");`

Once this is done, you can remove the section after the catch that iterates through the resultset

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/ImpISQLServerBulkRecord_IssuesTest.java
```diff
@@ -0,0 +1,435 @@
...
+        String value = ("1954-05-22 02:44:00.0").toString();
+
+        query = "CREATE TABLE " + destTable + " (smallDATA smalldatetime)";
+        stmt.executeUpdate(query);
+
+        try {
+            SQLServerBulkCopy bcOperation = new SQLServerBulkCopy(connectionString);
+            bcOperation.setDestinationTableName(destTable);
+            bcOperation.writeToServer(bData);
+            bcOperation.close();
```
v-ahibr: after bcOperation.close(), you should force the test to fail with something like:
`fail("BulkCopy executed when it it was expected to fail");`

Once this is done, you can remove the section after the catch that iterates through the resultset

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/ImpISQLServerBulkRecord_IssuesTest.java
```diff
@@ -0,0 +1,435 @@
...
+        }
+        catch (Exception e) {
+            if (e instanceof SQLServerException) {
+                assertTrue(e.getMessage().contains("Conversion failed when converting character string to smalldatetime data type"),
+                        "Invalid Error message: " + e.toString());
+            }
+        }
+        ResultSet rs = stmt.executeQuery("select * from " + destTable);
+        while (rs.next()) {
```
v-ahibr: empty resultSet, no point in reading from the table if nothing is in it

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/ImpISQLServerBulkRecord_IssuesTest.java
```diff
@@ -0,0 +1,435 @@
...
+        variation = "testBinaryColumnAsByte";
+        BulkDat bData = new BulkDat(variation);
+        query = "CREATE TABLE " + destTable + " (col1 binary(5))";
+        stmt.executeUpdate(query);
+
+        try {
+            SQLServerBulkCopy bcOperation = new SQLServerBulkCopy(connectionString);
+            bcOperation.setDestinationTableName(destTable);
+            bcOperation.writeToServer(bData);
+            bcOperation.close();
```
v-ahibr: after bcOperation.close(), you should force the test to fail with something like:
`fail("BulkCopy executed when it it was expected to fail");`

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/ImpISQLServerBulkRecord_IssuesTest.java
```diff
@@ -0,0 +1,435 @@
...
+        variation = "testBinaryColumnAsString";
+        BulkDat bData = new BulkDat(variation);
+        query = "CREATE TABLE " + destTable + " (col1 binary(5))";
+        stmt.executeUpdate(query);
+
+        try {
+            SQLServerBulkCopy bcOperation = new SQLServerBulkCopy(connectionString);
+            bcOperation.setDestinationTableName(destTable);
+            bcOperation.writeToServer(bData);
+            bcOperation.close();
```
v-ahibr: after bcOperation.close(), you should force the test to fail with something like:
`fail("BulkCopy executed when it it was expected to fail");`

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
```diff
@@ -1240,6 +1240,11 @@ private String getDestTypeFromSrcType(int srcColIndx,
             int destColIndx,
             TDSWriter tdsWriter) throws SQLServerException {
         boolean isStreaming;
+        /**
+         * The maximum temporal precision we can send when using varchar(precision) in bulkcommand, to send a smalldatetime/datetime 
+         * value.
+         */
+        int sourceBulkRecordTemporalMaxPrecision = 50;
```
v-nisidh: is value of sourceBulkRecordTemporalMaxPrecision constant? if so then can we use final modifier for mentioned variable? better class level `static final`

v-afrafi: done

PR #283, opened 2017-05-06
Pushing code around, style guide cleanup, code path consolidation.

brettwooldridge:
@TobiasSQL Sorry for the fairly large change set.  Probably 97% of the code is your original.

Here's what I did, briefly:
 * Minor name changes and style guide formatting
 * Moved ``Sha1HashKey``, ``ParsedSQLCacheItem`` and ``parsedSQLCache`` to ``SQLServerConnection``.  It felt weird having ``SQLServerPreparedStatement.Sha1HashKey`` references in ``SQLServerConnection``, and in ``SQLServerStatement`` (of which ``SQLServerPreparedStatement`` is a subclass).
 * Some collapsing/consolidation of methods

Feels tighter to me.  All tests are passing.  Nice work by the way.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
@@ -217,13 +217,8 @@ public class SQLServerConnection implements ISQLServerConnection {
             return 0 < statementHandle.getHandle();
         }
 
-        int getHandleAndIncrementRefCount() {
-            statementHandle.incrementRefCount();
-            return statementHandle.getHandle();
-        }
-
-        void setHandle(int handle, boolean isDirectSql) {
-            statementHandle.setHandle(handle, isDirectSql);
+        PreparedStatementHandle getPreparedStatementHandle() {
+            return statementHandle;
         }
         
         boolean hasParameterMetadata() {
@@ -236,14 +231,6 @@ public class SQLServerConnection implements ISQLServerConnection {
 
         void setParameterMetadata(SQLServerParameterMetaData metadata) {
             parameterMetadata = metadata;
-        }
-
-        int decrementHandleRefCount() {
-            return statementHandle.decrementRefCount();
-        }
-
-        int getHandleRefCount() {
-            return statementHandle.getRefCount();
         }
     }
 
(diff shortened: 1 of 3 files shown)


TobiasSQL:
Thanks @brettwooldridge! Much appreciated! Will take a look and comment in a day or two when I am back in Seattle :-)

brettwooldridge:
@TobiasSQL No problem.  Given the code-shuffling, it's probably easiest to pull my changes into a local branch and just look at the ``SQLServerConnection`` and ``SQLServerPreparedStatement`` classes directly, and use the diffs as reference.

brettwooldridge:
One of the bigger changes is that reference counting is simplified into two places, ``getHandleAndIncrementRefCount()`` and ``returnCachedPreparedStatementMetadata()``.

TobiasSQL:
Ack!

brettwooldridge:
@TobiasSQL lol.  I don't think I missed any edge cases. There was at least one race condition before, but I can find any now.

TobiasSQL:
Sounds great! As you said, will likely take changes manually through local branch.

brettwooldridge:
The refactored code is almost trivial in its simplicity, if you start from the SQLServerPreparedStatement constructor and follow the various scenarios.  Key points being the constructor, closeInternal(), and onRetValue().

TobiasSQL:
Finished the merge @brettwooldridge, looks great! Very elegant & clean indeed :-) 

Have a few comments/questions that I will post in a bit. Also pushing your merged changes to the metadataCaching branch (may have missed something but tests are passing, I think I got it all).

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -5369,8 +5475,9 @@ public void closeDiscardedServerPreparedStatements() {
      * Remove references to outstanding un-prepare requests. Should be run when connection is closed.
      */
     private final void cleanupPreparedStatementDiscardActions() {
+        closeDiscardedServerPreparedStatements();
```
TobiasSQL: There is no need to call close here as all handles will be discarded by the server when the connection is closed.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -238,26 +155,38 @@ String getClassNameInternal() {
             int nRSConcur,
             SQLServerStatementColumnEncryptionSetting stmtColEncSetting) throws SQLServerException {
         super(conn, nRSType, nRSConcur, stmtColEncSetting);
+
+        if (null == sql) {
```
TobiasSQL: I am wondering if this change may break existing apps. Not sure how but if we don't throw on "null" SQL text today this may create an issue...

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -5339,23 +5445,23 @@ static synchronized long getColumnEncryptionKeyCacheTtl() {
...
-            this.getConnectionLogger().finer(this + ": Adding PreparedHandle to queue for un-prepare:" + handle);
+    final void enqueueUnprepareStatementHandle(PreparedStatementHandle statementHandle) {
+        if (loggerExternal.isLoggable(java.util.logging.Level.FINER))
+            loggerExternal.finer(this + ": Adding PreparedHandle to queue for un-prepare:" + statementHandle.getHandle());
 
         // Add the new handle to the discarding queue and find out current # enqueued.
-        this.discardedPreparedStatementHandles.add(new PreparedStatementDiscardItem(handle, directSql));
-        this.discardedPreparedStatementHandleQueueCount.incrementAndGet();
+        discardedPreparedStatementHandles.add(statementHandle);
+        if (0 < discardedPreparedStatementHandleCount.incrementAndGet())
```
TobiasSQL: I noticed when writing the eviction tests that we cannot run the un-prepare batch call while another statement is being prepared so I move this call to handlePreparedStatementDiscardActions to the closePreparedHandle.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -5339,23 +5445,23 @@ static synchronized long getColumnEncryptionKeyCacheTtl() {
...
+            handlePreparedStatementDiscardActions(false);
     }
 
-
     /**
      * Returns the number of currently outstanding prepared statement un-prepare actions.
      * 
      * @return Returns the current value per the description.
      */
-    public int getDiscardedServerPreparedStatementCount() {
```
TobiasSQL: We debated the name of these methods before the last dev release, not sure it makes sense to change this public API now.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
TobiasSQL: Never mind, this is better and will not break apps :-)

PR #287, opened 2017-05-12
Fixed possible resource leak

pierresouchay:


v-afrafi:
Thanks for the fix @pierresouchay and quick response!

PR #291, opened 2017-05-12
Cleaner and tighter.

brettwooldridge:
I did rename one or two public methods, but AFAIK they are all still pre-release, so now is the time to get them right.

I tested performance between the before/after and find no difference. I did add a multi-threaded test to try to detect any race conditions (haven't found any in a few hundred runs), and it can also serve as a decent quick test balloon on performance because it is designed to generate maximum contentention.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
@@ -158,20 +158,20 @@ public class SQLServerConnection implements ISQLServerConnection {
     }
 
      /** Get prepared statement cache entry if exists, if not parse and create a new one */
-     static ParsedSQLCacheItem getOrCreateCachedParsedSQL(CityHash128Key key, String sql) throws SQLServerException {
-         ParsedSQLCacheItem cacheItem = parsedSQLCache.get(key);
-         if (null == cacheItem) {
-             JDBCSyntaxTranslator translator = new JDBCSyntaxTranslator();
- 
-             String parsedSql = translator.translate(sql);
-             String procName = translator.getProcedureName(); // may return null        
-             boolean returnValueSyntax = translator.hasReturnValueSyntax();
-             int paramCount = countParams(parsedSql);
- 
-             cacheItem = new ParsedSQLCacheItem(parsedSql, paramCount, procName, returnValueSyntax);
-             parsedSQLCache.putIfAbsent(key, cacheItem);
-         }
- 
+     static ParsedSQLCacheItem getCachedParsedSQL(CityHash128Key key, String sql) {
+         return parsedSQLCache.get(key);
+     }
+
+     static ParsedSQLCacheItem parseAndCacheSQL(CityHash128Key key, String sql) throws SQLServerException {
+         JDBCSyntaxTranslator translator = new JDBCSyntaxTranslator();
+         
+         String parsedSql = translator.translate(sql);
+         String procName = translator.getProcedureName(); // may return null        
+         boolean returnValueSyntax = translator.hasReturnValueSyntax();
+         int paramCount = countParams(parsedSql);
+
+         ParsedSQLCacheItem cacheItem = new ParsedSQLCacheItem(parsedSql, paramCount, procName, returnValueSyntax);
+         parsedSQLCache.putIfAbsent(key, cacheItem);
          return cacheItem;
      }
  
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
@@ -36,18 +36,21 @@ import java.sql.Savepoint;
 import java.sql.Statement;
 import java.sql.Struct;
 import java.text.MessageFormat;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Enumeration;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Properties;
 import java.util.UUID;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.Executor;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
-import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.logging.Level;
 
 import javax.sql.XAConnection;
@@ -102,9 +105,9 @@ public class SQLServerConnection implements ISQLServerConnection {
     static private Boolean defaultEnablePrepareOnFirstPreparedStatementCall = null; // Current default for new connections
     private Boolean enablePrepareOnFirstPreparedStatementCall = null; // Current limit for this particular connection.
 
-    // Handle the actual queue of discarded prepared statements.
-    private ConcurrentLinkedQueue<PreparedStatementHandle> discardedPreparedStatementHandles = new ConcurrentLinkedQueue<PreparedStatementHandle>();
-    private AtomicInteger discardedPreparedStatementHandleCount = new AtomicInteger(0);
+    // A map of SQL text hash keys to session-shared prepared statement handles
+    private ConcurrentHashMap<CityHash128Key, PreparedStatementHandle> preparedStatementHandleMap = new ConcurrentHashMap<>();
+    private AtomicBoolean isUnprepareInProgress = new AtomicBoolean();
 
     private boolean fedAuthRequiredByUser = false;
     private boolean fedAuthRequiredPreLoginResponse = false;
@@ -119,32 +122,26 @@ public class SQLServerConnection implements ISQLServerConnection {
 
     private SqlFedAuthToken fedAuthToken = null;
 
-    static class Sha1HashKey {
-        private byte[] bytes;
+    static class CityHash128Key {
+        private long[] segments;
+        private int hashCode;
 
-        Sha1HashKey(String s) {
-            bytes = getSha1Digest().digest(s.getBytes());
+        CityHash128Key(String s) {
+            segments = CityHash.cityHash128(s.getBytes(), 0, s.length());
         }
 
         public boolean equals(Object obj) {
-            if (!(obj instanceof Sha1HashKey))
+            if (!(obj instanceof CityHash128Key))
                 return false;
 
-            return java.util.Arrays.equals(bytes, ((Sha1HashKey)obj).bytes);
+            return java.util.Arrays.equals(segments, ((CityHash128Key)obj).segments);
         }
 
         public int hashCode() {
-            return java.util.Arrays.hashCode(bytes);
-        }
-
-        private java.security.MessageDigest getSha1Digest() {
-            try {
-                return java.security.MessageDigest.getInstance("SHA-1");
+            if (0 == hashCode) {
+                hashCode = java.util.Arrays.hashCode(segments); 
             }
-            catch (final java.security.NoSuchAlgorithmException e) {
-                // This is not theoretically possible, but we're forced to catch it anyway
-                throw new RuntimeException(e);
-            }
+            return hashCode;
         }
     }
 
@@ -152,16 +149,16 @@ public class SQLServerConnection implements ISQLServerConnection {
     static final private int PARSED_SQL_CACHE_SIZE = 100;
 
     /** Cache of parsed SQL meta data */
-    static private ConcurrentLinkedHashMap<Sha1HashKey, ParsedSQLCacheItem> parsedSQLCache;
+    static private ConcurrentLinkedHashMap<CityHash128Key, ParsedSQLCacheItem> parsedSQLCache;
 
     static {
-        parsedSQLCache = new Builder<Sha1HashKey, ParsedSQLCacheItem>()
+        parsedSQLCache = new Builder<CityHash128Key, ParsedSQLCacheItem>()
 	        .maximumWeightedCapacity(PARSED_SQL_CACHE_SIZE)
             .build();
     }
 
      /** Get prepared statement cache entry if exists, if not parse and create a new one */
-     static ParsedSQLCacheItem getOrCreateCachedParsedSQLMetadata(Sha1HashKey key, String sql) throws SQLServerException {
+     static ParsedSQLCacheItem getOrCreateCachedParsedSQL(CityHash128Key key, String sql) throws SQLServerException {
          ParsedSQLCacheItem cacheItem = parsedSQLCache.get(key);
          if (null == cacheItem) {
              JDBCSyntaxTranslator translator = new JDBCSyntaxTranslator();
@@ -199,38 +196,24 @@ public class SQLServerConnection implements ISQLServerConnection {
     /**
      * Used to keep track of an individual handle ready for un-prepare.
      */
-    final class PreparedStatementCacheItem {
-        private final PreparedStatementHandle statementHandle = new PreparedStatementHandle();
-        private volatile SQLServerParameterMetaData parameterMetadata;
-        private volatile boolean hasExecutedSpExecuteSql;
-        volatile boolean evictedFromCache;
+    final class PreparedStatementMetadata {
+        private final SQLServerParameterMetaData parameterMetadata;
+        private final CityHash128Key hashKey;
 
-        boolean hasExecutedSpExecuteSql() {
-            return hasExecutedSpExecuteSql;
-        }
+        PreparedStatementMetadata(CityHash128Key hashKey, SQLServerParameterMetaData parameterMetadata) {
+            this.hashKey = hashKey;
+            this.parameterMetadata = parameterMetadata;
 
-        void setHasExecutedSpExecuteSql(boolean hasExecutedSpExecuteSql) {
-            this.hasExecutedSpExecuteSql = hasExecutedSpExecuteSql;
-        }
-
-        boolean hasHandle() {
-            return 0 < statementHandle.getHandle();
-        }
-
-        PreparedStatementHandle getPreparedStatementHandle() {
-            return statementHandle;
-        }
-        
-        boolean hasParameterMetadata() {
-            return null != this.parameterMetadata;
+            // keep the statement handle alive for the life of the prepared statement metadata
+            borrowStatementHandle(hashKey);
         }
 
         SQLServerParameterMetaData getParameterMetadata() {
             return parameterMetadata;
         }
 
-        void setParameterMetadata(SQLServerParameterMetaData metadata) {
-            parameterMetadata = metadata;
+        void releaseStatementHandle() {
+            returnStatementHandle(hashKey);
         }
     }
 
@@ -238,7 +221,7 @@ public class SQLServerConnection implements ISQLServerConnection {
     private int statementPoolingCacheSize = 10;
 
     /** Cache of prepared statement handles */
-    private ConcurrentLinkedHashMap<Sha1HashKey, PreparedStatementCacheItem> preparedStatementCache;
+    private ConcurrentLinkedHashMap<CityHash128Key, PreparedStatementMetadata> preparedStatementMetadataCache;
 
     SqlFedAuthToken getAuthenticationResult() {
         return fedAuthToken;
@@ -340,13 +323,8 @@ public class SQLServerConnection implements ISQLServerConnection {
     
     private static final String SET_NETWORK_TIMEOUT_PERM = "setNetworkTimeout";
 
-    private boolean sendStringParametersAsUnicode = SQLServerDriverBooleanProperty.SEND_STRING_PARAMETERS_AS_UNICODE.getDefaultValue();        // see
-                                                                                                                                               // connection
-                                                                                                                                               // properties
-                                                                                                                                               // doc
-                                                                                                                                               // (default
-                                                                                                                                               // is
-                                                                                                                                               // false).
+    // see connection properties doc (default is false).
+    private boolean sendStringParametersAsUnicode = SQLServerDriverBooleanProperty.SEND_STRING_PARAMETERS_AS_UNICODE.getDefaultValue();
 
     boolean sendStringParametersAsUnicode() {
         return sendStringParametersAsUnicode;
@@ -845,14 +823,6 @@ public class SQLServerConnection implements ISQLServerConnection {
             connectionlogger.severe(message);
             throw new UnsupportedOperationException(message);
         }
-
-        // Caching turned off?
-        if (isStatementPoolingEnabled()) {
-            preparedStatementCache = new Builder<Sha1HashKey, PreparedStatementCacheItem>()
-                                        .maximumWeightedCapacity(getStatementPoolingCacheSize())
-                                        .listener(new PreparedStatementCacheEvictionListener())
-                                        .build();
-        }
     }
 
     void setFailoverPartnerServerProvided(String partner) {
@@ -894,7 +864,7 @@ public class SQLServerConnection implements ISQLServerConnection {
         return connectionlogger;
     }
 
-    String getClassNameLogging() {
+    final String getClassNameLogging() {
         return loggingClassName;
     }
 
@@ -1653,6 +1623,18 @@ public class SQLServerConnection implements ISQLServerConnection {
             String mirror = null;
             if (null == fo)
                 mirror = failOverPartnerPropertyValue;
+
+            // Caching turned off?
+            if (0 < statementPoolingCacheSize && null == preparedStatementMetadataCache) {
+                preparedStatementMetadataCache = new Builder<CityHash128Key, PreparedStatementMetadata>()
+                                            .maximumWeightedCapacity(getStatementPoolingCacheSize())
+                                            .listener(new EvictionListener<CityHash128Key, PreparedStatementMetadata>() {
+                                                public void onEviction(CityHash128Key key, PreparedStatementMetadata statementMetadata) {
+                                                    statementMetadata.releaseStatementHandle();
+                                                }
+                                            })
+                                            .build();
+            }
 
             long startTime = System.currentTimeMillis();
             login(activeConnectionProperties.getProperty(serverNameProperty), instanceValue, nPort, mirror, fo, loginTimeoutSeconds, startTime);
@@ -2847,11 +2829,11 @@ public class SQLServerConnection implements ISQLServerConnection {
         }
 
         // Invalidate statement cache.
-        if(null != this.preparedStatementCache)
-            this.preparedStatementCache.clear();
+        if(null != this.preparedStatementMetadataCache)
+            this.preparedStatementMetadataCache.clear();
 
         // Clean-up queue etc. related to batching of prepared statement discard actions (sp_unprepare).
-        cleanupPreparedStatementDiscardActions();
+        clearPreparedStatementHandles();
 
         loggerExternal.exiting(getClassNameLogging(), "close");
     }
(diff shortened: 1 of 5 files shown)


brettwooldridge:
@TobiasSQL Again, possibly best viewed by just pulling my branch local (note it is ``metadataCaching2`` not ``metadataCaching``) and using the diffs here as a guide.

brettwooldridge:
@TobiasSQL Sorry I need poking at this, squashing and pushing tweaks.  I'll resolve the above conflicts that have surfaced.

I have separate change set I'm working on, built on top of these changes, that are showing 3-10x improvements in transactions per second across the board (both prepared statements and regular statements).  Rather than continually piling on the ``metadataCaching`` branch, I'll hold those until the branch is merged into the mainline.

brettwooldridge:
Rebased.  All tests passing.  Feel free to ask about any particular change.

brettwooldridge:
In summary...

Observation: The prepared statement handle cache and prepared metadata cache are largely orthogonal.  For example, the handle cache can be enabled, but the metadata cache disabled, and vise-versa.

 * If we ignore the metadata cache (assume disabled), the handle is borrowed in the ``SQLServerPreparedStatement`` constructor, and released in ``closeInternal()`` (as before).
 * When the metadata cache is enabled, the metadata cache acts as an additional client -- equivalent to a user prepared statement -- and borrows the statement handle when the metadata cache entry is created, and releases it when the cache entry is evicted.
 * Treating the cache a just another client simplifies the state management.  Because of the symmetry in acquisition/release it is not possible for the reference count to "over reference" or "under reference" the statement handle.

Running multi-threaded [oltpbenchmark](https://github.com/oltpbenchmark/oltpbench) harness's JPAB benchmark with ~75,000 transactions and eight clients under the profiler also showed 0 thread contentions on any of the cache collections (handle or metadata).

```
02:27:26,097 (DBWorkload.java:801) INFO  - ======================================================================
02:27:26,097 (DBWorkload.java:802) INFO  - Rate limited reqs/s: Results(nanoSeconds=40000512846, measuredRequests=76630) = 1915.7254381967982 requests/sec
02:27:26,099 (DBWorkload.java:675) INFO  - Output Raw data into file: results/jpab-newer.32.csv
02:27:26,742 (DBWorkload.java:709) INFO  - Output into file: results/jpab-newer.32.res
02:27:26,742 (DBWorkload.java:712) INFO  - Grouped into Buckets of 5 seconds
02:27:26,788 (DBWorkload.java:613) INFO  - ======================================================================
02:27:26,788 (DBWorkload.java:614) INFO  - Workload Histograms:
Completed Transactions:
Persist/01           [ 3792] *****
Retrieve/02          [53728] ********************************************************************************
Update/03            [15246] **********************
Delete/04            [ 3872] *****
```

TobiasSQL:
Thanks a ton @brettwooldridge! 

I have started going through this and merging with changes I have been working on, posting comments as I go :-)

ajlam:
@brettwooldridge, thanks a lot for your help in improving the performance of our JDBC driver! We are currently working on stabilizing the code to prepare for our next release. This includes running a bunch of automated tests in our test framework, which we haven't managed to put on GitHub yet. I've told @TobiasSQL that we need to merge these metadata caching changes ASAP to start stabilizing. Him and I agree that we should take your updated API names. @TobiasSQL is working on merging changes from your branch, and will continue to take more after stabilizing. 

Please let us know if there's anything, especially related to public API names, that you strongly feel we should get into the next release. 

Again, thank you so much for your help thus far. We really appreciate your time! 😄 

- Andrea

brettwooldridge:
@ajlam @TobiasSQL Thanks to everyone for moving the driver forward, and huge kudos to Microsoft for open sourcing the driver in the first place!

I don't envision any changes to the public API.  I *am* working on a set of caching improvements based on some profiling runs that I think will provide a large performance increase, particularly for patterns employed by ORM frameworks.  I don't have a solid completion date, but possibly within the week.

TobiasSQL:
Thanks @brettwooldridge, I just finished a push into metadataCaching fixing some issues found by tests in our old infrastructure (related to cursors). Also cleaned up public APIs a bit, would be great if you take a peek at those (especially related to "discard actions/unprepare") so we can make any changes before the next public release. As @ajlam mentioned, once we have things stabilized for the release we should continue the optimizations, hopefully we can then just take your PRs directly :-)

brettwooldridge:
Some comparables of current progress

Using the v6.1.6-preview driver:
```
04:16:06,153 (DBWorkload.java:801) INFO  - ======================================================================
04:16:06,154 (DBWorkload.java:802) INFO  - Rate limited reqs/s: Results(nanoSeconds=40001173473, measuredRequests=80252) = 2006.2411432546726 requests/sec
04:16:06,157 (DBWorkload.java:675) INFO  - Output Raw data into file: results/jpab-newer.53.csv
04:16:06,844 (DBWorkload.java:709) INFO  - Output into file: results/jpab-newer.53.res
04:16:06,844 (DBWorkload.java:712) INFO  - Grouped into Buckets of 5 seconds
04:16:06,900 (DBWorkload.java:613) INFO  - ======================================================================
04:16:06,900 (DBWorkload.java:614) INFO  - Workload Histograms:
Completed Transactions:
Persist/01           [ 4121] *****
Retrieve/02          [63980] ********************************************************************************
Update/03            [ 8090] **********
Delete/04            [ 4071] *****
```

From my ``speedify`` branch, consisting of the ``metadataCache2`` branch plus ongoing work...
```
04:10:18,202 (DBWorkload.java:801) INFO  - ======================================================================
04:10:18,202 (DBWorkload.java:802) INFO  - Rate limited reqs/s: Results(nanoSeconds=40000398344, measuredRequests=92646) = 2316.126934618309 requests/sec
04:10:18,205 (DBWorkload.java:675) INFO  - Output Raw data into file: results/jpab-newer.52.csv
04:10:19,005 (DBWorkload.java:709) INFO  - Output into file: results/jpab-newer.52.res
04:10:19,005 (DBWorkload.java:712) INFO  - Grouped into Buckets of 5 seconds
04:10:19,081 (DBWorkload.java:613) INFO  - ======================================================================
04:10:19,082 (DBWorkload.java:614) INFO  - Workload Histograms:
Completed Transactions:
Persist/01           [ 4750] *****
Retrieve/02          [74045] ********************************************************************************
Update/03            [ 9246] *********
Delete/04            [ 4615] ****
```
More work to do ... the big wins (eg. ResultSet metadata caching) aren't in there yet.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -119,49 +122,43 @@
 
     private SqlFedAuthToken fedAuthToken = null;
 
-    static class Sha1HashKey {
-        private byte[] bytes;
+    static class CityHash128Key {
```
brettwooldridge: Simply performance.  On a local benchmark I measured, for 100,000 hashes:

SHA1 Hash Total ns: 124926909 (~125ms)

CityHashCrc128 Total ns: 23457110 (~23.5ms)

CitiHash128 passes the [smhasher](https://github.com/rurban/smhasher/blob/master/doc/City128) with no collisions, which is what we're really after as much as performance.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -119,49 +122,43 @@
...
     static {
-        parsedSQLCache = new Builder<Sha1HashKey, ParsedSQLCacheItem>()
+        parsedSQLCache = new Builder<CityHash128Key, ParsedSQLCacheItem>()
 	        .maximumWeightedCapacity(PARSED_SQL_CACHE_SIZE)
             .build();
     }
 
      /** Get prepared statement cache entry if exists, if not parse and create a new one */
-     static ParsedSQLCacheItem getOrCreateCachedParsedSQLMetadata(Sha1HashKey key, String sql) throws SQLServerException {
+     static ParsedSQLCacheItem getOrCreateCachedParsedSQL(CityHash128Key key, String sql) throws SQLServerException {
```
brettwooldridge: @TobiasSQL I disagree here.  "Metadata" in the JDBC sense is typically either database metadata (``Connection.getMetaData()``), column metadata (``PreparedStatement.getParameterMetaData()``), or result metadata (``ResultSet.getMetaData()``).  In this case, the cache is simply the pre-processed SQL text of the statement, without respect to type information.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -199,46 +196,32 @@ private static int countParams(String sql) {
...
+            returnStatementHandle(hashKey);
         }
     }
 
     /** Size of the  prepared statement handle cache */
     private int statementPoolingCacheSize = 10;
 
     /** Cache of prepared statement handles */
-    private ConcurrentLinkedHashMap<Sha1HashKey, PreparedStatementCacheItem> preparedStatementCache;
+    private ConcurrentLinkedHashMap<CityHash128Key, PreparedStatementMetadata> preparedStatementMetadataCache;
```
brettwooldridge: I keep going back and forth on this.  In general, is a Good Idea<sup>tm</sup> to set them in lock-step in terms of size, but abstractly there is no implementation reason to couple them.  I'm fine with snipping the thread between them -- only two places to cut.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -5384,81 +5366,73 @@ static synchronized long getColumnEncryptionKeyCacheTtl() {
...
         public boolean isDirectSql() {
             return isDirectSql;
         }
 
-        public boolean killHandle() {
-            if(!hasHandle())
-                return false;
-            else
-                return handleRefCount.compareAndSet(0, -999);
+        public int getRefCount() {
```
brettwooldridge: WIthing giving it too much thought, that seems reasonable.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -5619,56 +5598,75 @@ public void setServerPreparedStatementDiscardThreshold(int value) {
...
+    final void unprepareUnreferencedPreparedStatementHandles(boolean force) {
+        // Skip out if session is unavailable to adhere to previous non-batched behavior
+        if (isSessionUnAvailable())
             return;
 
+        // Skip out if we are not forced and:
+        //   * another thread is already doing the work of un-preparing, OR
+        //   * number of handles in the map have not crossed the threshold (optimization)
         final int threshold = getServerPreparedStatementDiscardThreshold();
+        if (!force && (isUnprepareInProgress.get() || threshold >= preparedStatementHandleMap.size()))
```
brettwooldridge: @TobiasSQL The only issue with a queue is that a handle is either in or out.  I'm not sure how the semantics work in that case.  In theory, a handle that reaches refcount 0 is reclaimable, but that is the case immediately after a statement is executed.  Before, the tie to the metadata cache, and the eviction, artifically kept handles alive.  But if that relationship is severed, as discussed above, claiming handles as soon as they reach refcount 0 seems inefficient.

In the case of the concurrent collection, a handle that as reached refcount 0 always has the possibility of being "rescued" by a new statement execution -- but only because it remains in the live collection.  In the case of a true unprepare queue, once a handle is enqueued, its fate is sealed.

Now, maybe something like an eviction-based collection, dedicated to handles (separate from the metadata) *could* work, if the behavior is LRU in nature...

Thoughts?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -165,31 +167,24 @@ String getClassNameInternal() {
...
-            }
-        }
+        procedureName = parsedSql.procedureName;
+        bReturnValueSyntax = parsedSql.bReturnValueSyntax;
+        userSQL = parsedSql.processedSQL;
+        initParams(parsedSql.parameterCount);
+
+        // Attempt to borrow the statement handle, if statement handle caching is enabled, this will be 0 if not.
+        prepStmtHandle = connection.borrowStatementHandle(cacheKey);
+        if (0 < prepStmtHandle)
```
brettwooldridge: I think you're missing something -- unless I'm missing something. 😁 

If a new PreparedStatement is created, for which no handle exists, the handle remains at 0 (uninitialized) and therefore sp_executesql will be used.  Once sp_prepexec has run (i.e. the same statement is executed a second time) then the handle will be cached.  Subsequent PreparedStatements with the same SQL text will acquire the same handle, and skip the sp_executesql, going directly to sp_execute.  Right?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
TobiasSQL: Agreed. We can't move handles to the queue simply at refcount 0, they need to actually fall out of the cache or be explicitly discarded before being moved to the queue (which is the case right now). 

The downside with the queue approach is that handles live for some time in the queue where you could argue that they could be re-used but they won't be because they are no longer in the cache. My thought is that the queue simplifies the discard process and that these serve two distinct purposes, one is for batching un-prepares together to make sure we don't do unnecessary round trips and the cache is there to make us re-use handles to both reduce round trips and to minimize network traffic (lesser payload).

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
TobiasSQL: Correctamundo :-) The issue I have is the following scenario:
1. New stmt with SQL text & param definitions A is created & executed & closed: sp_executesql
2. New stmt with SQL text & param definitions A is created & executed & closed: ?

In this implementation #2 will do sp_executesql again where I would argue it should move to sp_prepexec so the third execution starts using a handle.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
TobiasSQL: Makes sense :-) Just wanted to dbl check. Thx!

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
TobiasSQL: I see. Then I agree with you, makes sense :-)

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
TobiasSQL: Cool :-)

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
TobiasSQL: Cool

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
brettwooldridge: @TobiasSQL I pushed a small change to account for this.  Basically, I "cheat" a little by leveraging the parsed SQL cache.  Presence of a query in the parsed SQL cache is evidence that we have executed that SQL before, even without the knowledge of handles and whotnot.

An edge-case of my change is that execution of SQL by a ``SQLServerStatement`` also creates a cache entry.  Later execution of *identical* SQL through a ``SQLServerPreparedStatement`` would be viewed as ``isExecutedAtLeastOnce``.  Not a bad thing by any means, but worth pointing out.

brettwooldridge: I was going to rebase my change onto ``mssql/metadataCaching``, but basically every change was a merge conflict, so I opted to make the spot change for this in my ``metadataCaching2`` branch.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
brettwooldridge: @TobiasSQL Do you want me to sever the coupling between the two caches and push it to my ``metadataCache2`` branch?  Or is that something you want to look at post-6.1.8?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
brettwooldridge: @TobiasSQL I think it is workable either way you want to go.  In practice, I would recommend that users set both the metadata cache size and batch handle discard thresholds to the same value, which basically ensures a large amount of handle reuse and minimizes network chatter (and server CPU due to re-prepares).

TobiasSQL: Makes sense to me also @brettwooldridge

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
TobiasSQL: I already did that in metadataCaching :-)

brettwooldridge: Sweet.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -170,7 +171,13 @@ String getClassNameInternal() {
...
 
         // Parse or fetch SQL metadata from cache.
-        ParsedSQLCacheItem parsedSql = getOrCreateCachedParsedSQL(cacheKey, sql);
+        ParsedSQLCacheItem parsedSql = getCachedParsedSQL(cacheKey, sql);
+        if (null != parsedSql) {
+            isExecutedAtLeastOnce = true;
+        }
+        else {
+            parsedSql = parseAndCacheSQL(cacheKey, sql);
+        }
```
brettwooldridge: I actually like this for another reason.  The parsed sql cache is across connections, whereas the handle cache is per-connection.  In connection pooling environments, users often have pools of say 50 (or even 100) connections, so executing PreparedStatement A only has a 1:50 chance of executing on the same connection later -- and therefore could avoid getting prepared for quite a while, even though it may be executed semi-frequently.

Using presence in a cache that cuts across connections as the indicator of ``isExecutedAtLeastOnce`` is going "recognize" multiple identical executions on seperate connections immediately.  For ORMs this is going to be a big boost, because the generated SQL is always identical -- Hibernate uses PreparedStatements across the board -- but there is additionally almost always a pool present.

TobiasSQL: I was thinking the same thing at first but then arrived at the opposite conclusion. My comment is that since prepared handles can only be reused on the same connection the fact that a statement gets re-used across connections is void. In addition we don't include parameter definitions in the parsed SQL key (correctly), which we need to in the handle cache. So we conclusion is that we really should not "skip" sp_executesql for the same text across connections.

brettwooldridge: @TobiasSQL Hang on a sec. 😁   This is a JDBC driver, right?  I think if you survey the landscape you'll find that ``PreparedStatements`` with identical SQL are *overwhelmingly* reused across connections, by a large margin.

JEE containers, Spring containers, web containers (Tomcat/Jetty), JPA persistence, almost without fail acquire a connection from a pool, execute some business logic involving static (but parameterized) SQL, via ``PreparedStatement``, and return the connection to the pool.

Skipping *sp_executesql* after observing multiple executions of identical SQL is not the same as skipping *sp_prepsql* and going directly to *sp_execute*.  Which by the way is what my ``speedify`` branch does, because it caches *input* parameter metadata and goes directly to *sp_execute* for identical SQL with identical parameter types, additionally using the *fReuseMetaData* flag (and *fNoMetaData*, because the *column info* metadata coming out is cached as well).<sup>[reference](https://msdn.microsoft.com/en-us/library/dd357576.aspx)</sup>

Functionally, using *sp_executesql* within the ``PreparedStatement`` is purely optional, but *I do* recognize the case for *dynamic* SQL executed via ``PreparedStatement``.  If it is truly dynamic, the behavior above is exactly as you would expect/want.  If it is *quasi-dynamic* wherein identical SQL *is* being executed with some frequency, I would argue that it would benefit from the optimizations that (will) come with *fReuseMetaData* and *fNoMetaData* by skipping the *sp_executesql* (and even the *sp_prepsql* when those changes land).

Keep in mind that "*is being executed with some frequency*" means multiple executions of identical SQL within the timeframe in which parsed SQL has not rolled out of the cache.  With a *lot* of (truly) dynamic SQL the cache is going to roll very often.  We talking about *identical* SQL executed *within a narrow time horizon* being "promoted" to *sp_prepsql* (regardless of connection affiliation).

brettwooldridge: Providing a real world example from just yesterday...

I was trying to run the ``oltpbench/epinions`` benchmark against my ``speedify`` branch, but I wasn't seeing any performance gain at all.  Zero.  So, I added some logging to see what was going on...

The code benchmark code does this:
 * Grab a connection from the pool (in a *try-with-resources* block)
 * Prepare a statement (``conn.prepareStatement(sql)``
 * Set parameters and execute
 * Return the connection to the pool (exit the ``try``)
 * Rinse and repeat

No *single* ``PreparedStatement`` is *ever* executed more than once.  So, no *sp_prepsql* was occurring, only *sp_executesql*.

Then I realized ... that is what my company's application, and most Java web applications, so in 90% of the cases.
 * The user types a search, enters an order, etc. and then clicks "go" (or "save", or "search")
 * The web request comes in, grab a connection from the pool
 * Prepare a statement
 * Set parameters and execute
 * Close the connection (back to the pool), and return results back to the user

Again, individual PreparedStatement instances are rarely executed multiple times, so no *sp_prepsql*.  However, there are thousands of users doing this!  That SQL needs to be prepared!

And a user of SQL Server JDBC shouldn't have to force prepare-on-first-execute across the board.  [But you can see](https://github.com/brettwooldridge/oltpbench/blob/hibernate4/config/sample_epinions_config.xml#L7), in order to actually realize *any* of the optimizations we've been making on this branch, that is exactly what I had to do.

And not only for the ``epinions`` benchmark, but also the ``jpab`` [benchmark](https://github.com/brettwooldridge/oltpbench/blob/hibernate4/src/META-INF/persistence.xml#L28).  Surveying the suite of the ~30 benchmarks in that suite, developed by different individuals, nearly all of them require that.

TobiasSQL: I see why your test would run only sp_executesql with your previous implementation. However, if you kept already executed once" as part of the handle cache you would not see that behavior, this is why I added that in the first place :-) 

I guess the difference of treating the parse cache hit as a "first execution" is that we assume that we will have >2 executions of the same handle per connection rather than > 1. This doesn't seem like a huge difference to me. We should also remember the cost of the unprepare calls. At the end of the day the re-use of the handle will only occur within the same connection.

TobiasSQL: Just to clarify, as far as I see it this would be the only difference between the two implementations, i.e. at the end of the day you will get handle re-use in both. For the "single call on connection" case the connection level is better, for the other cases I would say difference will be negligible:
```
1 call
sp_executesql | sp_prepexec
              | sp_unprepare

2 calls
sp_executesql | sp_prepexec
sp_prepexec   | sp_execute 
sp_unprepare  | sp_unprepare

3+ calls
sp_executesql | sp_prepexec
sp_prepexec   | sp_execute 
sp_execute    | sp_execute
...           | ...
sp_unprepare  | sp_unprepare
```

brettwooldridge: > I see why your test would run only sp_executesql with your previous implementation. However, if you kept "already executed once" as part of the handle cache you would not see that behavior, this is why I added that in the first place :-)

I don't understand how that would have made any difference in the *oltpbench* ``epinions`` or ``jpab`` benchmarks (or Hibernate in general), where a single PreparedStatement instance is never executed more than once.  If execution for a single statement never reaches *sp_prepsql* then no handle would be created, right?  The benchmarks below seem to bear this out.

> For the "single call on connection" case the connection level is better, for the other cases I would say difference will be negligible (happy to be proven wrong though :-)

If the difference is negligible, I don't know what we're doing. :grin:  Fortunately, the difference is not negligible.  This is the *oltpbench/epinions* benchmark with the current ``metadataCaching`` branch, with ``enablePrepareOnFirstPreparedStatementCall=false``:
```
Benchmark:     EPINIONS {com.oltpbenchmark.benchmarks.epinions.EpinionsBenchmark}
Configuration: config/sample_epinions_config.xml
Type:          SQLSERVER
Driver:        com.microsoft.sqlserver.jdbc.SQLServerDriver
URL:           jdbc:sqlserver://10.0.40.11:1433;databaseName=epinions;statementPoolingCacheSize=100;serverPreparedStatementDiscardThreshold=100;enablePrepareOnFirstPreparedStatementCall=false;
Isolation:     TRANSACTION_READ_COMMITTED
Scale Factor:  100.0

17:26:53,982 (DBWorkload.java:260) INFO  - ======================================================================
17:26:53,994 (DBWorkload.java:790) INFO  - Creating 10 virtual terminals...
[main] INFO com.zaxxer.hikari.HikariDataSource - HikariPool-1 - Starting...
[main] INFO com.zaxxer.hikari.HikariDataSource - HikariPool-1 - Start completed.
17:26:54,627 (DBWorkload.java:795) INFO  - Launching the EPINIONS Benchmark with 1 Phase...
17:26:54,639 (ThreadBench.java:325) INFO  - PHASE START :: [Workload=EPINIONS] [Serial=false] [Time=120] [Rate=10000] [Arrival=REGULAR] [Ratios=[10.0, 10.0, 10.0, 10.0, 10.0, 10.0, 10.0, 10.0, 20.0]] [ActiveWorkers=10]
17:26:54,640 (ThreadBench.java:472) INFO  - MEASURE :: Warmup complete, starting measurements.
17:28:54,636 (ThreadBench.java:428) INFO  - TERMINATE :: Waiting for all terminals to finish ..
17:28:54,639 (ThreadBench.java:233) INFO  - Starting WatchDogThread
17:28:54,787 (DBWorkload.java:801) INFO  - ======================================================================
17:28:54,787 (DBWorkload.java:802) INFO  - Rate limited reqs/s: Results(nanoSeconds=120000780351, measuredRequests=468917) = 3907.6162557312264 requests/sec
17:28:54,789 (DBWorkload.java:675) INFO  - Output Raw data into file: results/oltpbench.22.csv
17:28:58,977 (DBWorkload.java:613) INFO  - ======================================================================
17:28:58,978 (DBWorkload.java:614) INFO  - Workload Histograms:
GetReviewItemById/01 [37559] ****************************************
GetReviewsByUser/02  [37486] ****************************************
GetAverageRatingB... [37548] ****************************************
GetItemAverageRat... [37485] ****************************************
GetItemReviewsByT... [37568] ****************************************
UpdateUserName/06    [37808] ****************************************
UpdateItemTitle/07   [37365] ***************************************
UpdateReviewRatin... [37629] ****************************************
UpdateTrustRating/09 [74947] ********************************************************************************
```
Again, because no PreparedStatement is ever executed more than once, there are no handles to be shared.  Here is the same test with ``enablePrepareOnFirstPreparedStatementCall=true``:
```
GetReviewItemById/01 [46663] ***************************************
GetReviewsByUser/02  [47003] ****************************************
GetAverageRatingB... [46806] ****************************************
GetItemAverageRat... [47010] ****************************************
GetItemReviewsByT... [47139] ****************************************
UpdateUserName/06    [46886] ****************************************
UpdateItemTitle/07   [46643] ***************************************
UpdateReviewRatin... [47369] ****************************************
UpdateTrustRating/09 [93408] ********************************************************************************
```
And *oltpbench/jpab* (Hibernate) benchmark with ``enablePrepareOnFirstPreparedStatementCall=false``:
```
Persist/01           [ 5641] *****
Retrieve/02          [85171] ********************************************************************************
Update/03            [17168] ****************
Delete/04            [ 5780] *****
```
And with ``enablePrepareOnFirstPreparedStatementCall=true``:
```
Persist/01           [ 6352] *****
Retrieve/02          [92136] ********************************************************************************
Update/03            [18669] ****************
Delete/04            [ 6128] *****
```

TobiasSQL: Thanks for the analysis! 

I will take a deeper look into this. This statement doesn't make sense to me.
> Again, because no PreparedStatement instance is ever executed more than once, there are no handles to be shared. 

That is why we have the cache and move directly to sp_prepexec if we see that another instance has already done sp_executsql before. This should make the new instance re-use the state from the previous instance and create the handle. Subsequent instance should re-use the handle. 
```
// Because sp_executesql was already called on this SQL-text use
// regular prep/exec pattern.
if (cachedPreparedStatementHandle.hasExecutedAtLeastOnce())
	isExecutedAtLeastOnce = true;
```

This may simply be an indication that the default cache size is of 10 is too small.

TobiasSQL: Ok, after wasting my time with my own analysis I concluded you are right @brettwooldridge :-) 

What I completely missed is that the cache hit ratio with ``enablePrepareOnFirstPreparedStatementCall=true`` is 1/queriesPerCacheSize given the unique query usage is uniformly random (i.e. 100% where cache size is at least as large as total number of queries and 50% where the cache size is half of the number of unique queries). On the other hand the default of ``enablePrepareOnFirstPreparedStatementCall=false`` is 1/queriesPerCacheSize^2 (i.e. where cache size is half the number of queries the hit rate would be 25%).

When using your updated change it is close to the same as ``enablePrepareOnFirstPreparedStatementCall=false`` while still using sp_executesql when unique queries are executed. Consider me completely on-board with the chance of having isExecutedAtLeastOnce be dependent on the parsed SQL text cache :-) 

I have pushed the change to metadata caching and hope to get it into the release :-)

Thanks for being persistent @brettwooldridge!

PS I believe I fixed the final bug for the not-yet-GitHub:ed tests :-)

v-nisidh:
@TobiasSQL : I guess your metadatacaching includes this fix too. If so we can close this PR.

TobiasSQL:
Yes, I think this particular PR I have picked up most everything from into my PR. Brett has been doing a bunch of more work that he was planning to issue a PR for as soon as my changes make it back to master (when do you expect that to happen?). @brettwooldridge is that accurate?

brettwooldridge:
@TobiasSQL That is accurate.

v-nisidh:
Closing this PR as @TobiasSQL already include this in his patch.

@TobiasSQL : Your patch is already in Master & DEV.  

> Brett has been doing a bunch of more work that he was planning to issue a PR for as soon as my changes make it back to master (when do you expect that to happen?).

TobiasSQL:
@v-nisidh, oh, I did not realize (maybe I should have looked...). Hey @brettwooldridge that means we are ready for you to start PRing your speedify stuff :-)

PR #292, opened 2017-05-15
SonarQube analysis

v-xiangs:


src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -2435,7 +2435,8 @@ else if (!useTnir) {
 
         }
 
-        assert result.equals(Result.SUCCESS) == true;
+        boolean equalSuccess = result.equals(Result.SUCCESS);
```
v-nisidh: SonarQube might happy but I don't see any usage of `equalSuccess` other than assert. 

I guess following code will be more useful.

```
assert (result.equals(Result.SUCCESS))
```
**Think about this in different way:** 
What will happen if result is false **if assertions are not enabled**. 

Are we handle this in somewhere (in our exception flow)? Should we want to handle it here?

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -2640,7 +2640,9 @@ private void findSocketUsingThreading(LinkedList<Inet6Address> inetAddrs,
             int portNumber,
             int timeoutInMilliSeconds) throws IOException, InterruptedException {
         assert timeoutInMilliSeconds != 0 : "The timeout cannot be zero";
-        assert inetAddrs.isEmpty() == false : "Number of inetAddresses should not be zero in this function";
+
```
v-nisidh: I don't see here any value addition due to using boolean empty. 

As per my understanding SonarQube flags this because if assertion is **not enabled** then whole condition will be omit. 
Now question : If assertions are not enabled is it ok to skip this line from execution? If yes then our previous implementation was right.

Think of imaginary scenario. -> we are removing objects from list. 

**Wrong Assert**
```
assert myList.remove(myObj);
```

**Right Assert**
```
boolean objRemoved = myList.remove(myObj);
assert objRemoved;
```

In first case `myList.remove(myObj)` will be omitted if assertions are not enabled.

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -3684,7 +3684,10 @@ void writeWrappedBytes(byte value[],
         // what remains in the current staging buffer. However, the value must
         // be short enough to fit in an empty buffer.
         assert valueLength <= value.length;
-        assert stagingBuffer.remaining() < valueLength;
+        
+        int remaining = stagingBuffer.remaining();
```
v-nisidh: In future please try to club commits whenever necessary. Like this one and next one where you are reusing `remaining` variable.

v-nisidh:
@v-xiangs :  like to see club commits at-least file wise.

src/main/java/com/microsoft/sqlserver/jdbc/AE.java
```diff
@@ -45,19 +45,19 @@
...
 /**
  * Represents a unique CEK as an entry in the CekTable. A unique (plaintext is unique) CEK can have multiple encrypted CEKs when using multiple CMKs.
- * These encrypted CEKs are represented by a member vector.
+ * These encrypted CEKs are represented by a member ArrayList.
  */
 class CekTableEntry {
     static final private java.util.logging.Logger aeLogger = java.util.logging.Logger.getLogger("com.microsoft.sqlserver.jdbc.AE");
 
-    Vector<EncryptionKeyInfo> columnEncryptionKeyValues;
+    ArrayList<EncryptionKeyInfo> columnEncryptionKeyValues;
```
v-nisidh: Can you provide unit test cases for this change?

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -2341,7 +2341,8 @@ else if (!useTnir) {
                         inet4Addrs.add((Inet4Address) inetAddr);
                     }
                     else {
-                        assert inetAddr instanceof Inet6Address : "Unexpected IP address " + inetAddr.toString();
+                        boolean instanceOfIPv6 = inetAddr instanceof Inet6Address;
```
v-nisidh: Please see my comments related to assert

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
v-xiangs: there is no usage of the variable. however, I think what sonar says is a good practice. In this case, the `equals()` method only check for equivalence and return a boolean value. if we think about another case: a method returns a boolean value and also modifies a class-level variable. In this case, if we use this method with assert, it is likely to cause issues if `-ea` is not enabled explicitly. Instead, we could always separate the method and the `assert` to make sure the method behaves correctly.  

I don't see any significant advantages in your suggestion over the original code.

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
v-xiangs: reason explained in the previous comment about `assert`

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
v-xiangs: reason explained in the previous comment about `assert`

src/main/java/com/microsoft/sqlserver/jdbc/AE.java
v-xiangs: this method is used in every AE test case

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -2308,12 +2308,12 @@ else if (!useTnir) {
             // Code reaches here only if MSF = true or (TNIR = true and not TNIR first attempt)
 
             if (logger.isLoggable(Level.FINER)) {
-                String loggingString = this.toString() + " Total no of InetAddresses: " + inetAddrs.length + ". They are: ";
+                StringBuilder loggingString = new StringBuilder(this.toString() + " Total no of InetAddresses: " + inetAddrs.length + ". They are: ");
```
v-xiangs: done

v-xiangs:
@sehrope thank you. I have changed the method base on your suggestion.

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
v-nisidh: I agree about rule & purpose.  I feel that this might not be good scenario for this rule because you are not performing any operation. Performing any operation in assert is **bad coding**

**Think about this in different way:**
What will happen if result is not Result.SUCCESS if assertions are not enabled.
Are we handling this scenario / condition in somewhere (may be in our exception flow)? Should we want to handle it here? 

if not then what value addition we are doing here?

v-xiangs: if we do not handle the scenario when result is not Result.SUCCESS and assertion is not enabled, I guess we should remove the assertion in the original code. But the assertion is there already..

v-xiangs:
created another PR for this fix

PR #294, opened 2017-05-17
Fix password logging

v-ahibr:
FIPS property trustStorePassword was being logged

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/Util.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/Util.java
@@ -374,8 +374,14 @@ final class Util {
                         if (null != name) {
                             if (logger.isLoggable(Level.FINE)) {
                                 if ((false == name.equals(SQLServerDriverStringProperty.USER.toString()))
-                                        && (false == name.equals(SQLServerDriverStringProperty.PASSWORD.toString())))
-                                    logger.fine("Property:" + name + " Value:" + value);
+                                        && (false == name.equals(SQLServerDriverStringProperty.PASSWORD.toString()))) {
+                                    if (!name.contains("password") && !name.contains("Password")) {
+                                        logger.fine("Property:" + name + " Value:" + value);
+                                    }
+                                    else {
+                                        logger.fine("Property:" + name);
+                                    }
+                                }
                             }
                             p.put(name, value);
                         }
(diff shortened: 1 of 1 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/Util.java
```diff
@@ -373,9 +373,14 @@ else if (ch == ':')
                         name = SQLServerDriver.getNormalizedPropertyName(name, logger);
                         if (null != name) {
                             if (logger.isLoggable(Level.FINE)) {
-                                if ((false == name.equals(SQLServerDriverStringProperty.USER.toString()))
-                                        && (false == name.equals(SQLServerDriverStringProperty.PASSWORD.toString())))
```
v-xiangs: just for my own curiosity, why removed `&& (false == name.equals(SQLServerDriverStringProperty.PASSWORD.toString()))` ?

src/main/java/com/microsoft/sqlserver/jdbc/Util.java
```diff
@@ -373,9 +373,14 @@ else if (ch == ':')
...
-                                    logger.fine("Property:" + name + " Value:" + value);
+                                if (false == name.equals(SQLServerDriverStringProperty.USER.toString())) {
+                                    if (!name.contains("password") && !name.contains("Password")) {
```
v-xiangs: maybe we can also use `name.toLowerCase().contains("password")`

v-ahibr: either works.
only reason i wrote it this way is that a similar if condition is used for  SQLServerDataSource

```
private void setStringProperty(Properties props,
            String propKey,
            String propValue) {
        if (loggerExternal.isLoggable(java.util.logging.Level.FINER) && !propKey.contains("password") && !propKey.contains("Password")) {
            loggerExternal.entering(getClassNameLogging(), "set" + propKey, propValue);
        }
        else
            loggerExternal.entering(getClassNameLogging(), "set" + propKey);
        if (null != propValue)
            props.setProperty(propKey, propValue);
        loggerExternal.exiting(getClassNameLogging(), "set" + propKey);
    }
```

src/main/java/com/microsoft/sqlserver/jdbc/Util.java
v-ahibr: Originally, the condition checked if we are working with the database password and then decides not to log it.

Now, we check if the property name includes the string "password" so the first if condition is not needed anymore

PR #299, opened 2017-05-18
Fix javadoc

v-xiangs:


v-afrafi:
Should we add javadocs for SQLServerDataSource#setQueryTimeout and getQueryTimeout here as well?

v-xiangs:
@v-afrafi agree, also for `SQLServerDataSource#setSocketTimeoutand` and `getSocketTimeout` as well

PR #301, opened 2017-05-18
SonarQube fixes

v-afrafi:


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerXAResource.java
```diff
@@ -791,7 +792,7 @@ else if (-1 != version.indexOf('.')) {
     /* L0 */ public Xid[] recover(int flags) throws XAException {
         XAReturnValue r = DTC_XA_Interface(XA_RECOVER, null, flags | tightlyCoupled);
         int offset = 0;
-        Vector<XidImpl> v = new Vector<XidImpl>();
+        ArrayList<XidImpl> v = new ArrayList<XidImpl>();
```
v-ahibr: the variable name doesnt mean anything anymore now that its an array list instead of a vector. It might be better to rename it to something like al

PR #305, opened 2017-05-19
Use Javadoc for API documentation

marschall:
Java offers two kinds of multi line comments /** and /*.
The first is called Javadoc and ends up in generated API documentation
and is displayed by IDEs. The second is a multi line comment that does
not show up.

There are several places there /* instead of /** is used for API
documentation.

 [1] https://stackoverflow.com/questions/29815636/and-in-java-comments
 [2] http://javadude.com/articles/comments.html


marschall:
As an example have a look at:
https://static.javadoc.io/com.microsoft.sqlserver/mssql-jdbc/6.1.7.jre8-preview/microsoft/sql/Types.html

PR #306, opened 2017-05-20
UTC should be a class

marschall:
UTC is an enum with one constant and a static variable. It has the
following comment

> The enum type delays initialization until first use.

This is demonstrably wrong.

The static variable is initialized when the class is initialized. If we
look at the decompiled static initializer of UTC we see that.

```
  // access flags 0x8
  static <clinit>()V
   L0
    LINENUMBER 522 L0
    NEW com/microsoft/sqlserver/jdbc/UTC
    DUP
    LDC "INSTANCE"
    ICONST_0
    INVOKESPECIAL com/microsoft/sqlserver/jdbc/UTC.<init> (Ljava/lang/String;I)V
    PUTSTATIC com/microsoft/sqlserver/jdbc/UTC.INSTANCE : Lcom/microsoft/sqlserver/jdbc/UTC;
    ICONST_1
    ANEWARRAY com/microsoft/sqlserver/jdbc/UTC
    DUP
    ICONST_0
    GETSTATIC com/microsoft/sqlserver/jdbc/UTC.INSTANCE : Lcom/microsoft/sqlserver/jdbc/UTC;
    AASTORE
    PUTSTATIC com/microsoft/sqlserver/jdbc/UTC.ENUM$VALUES : [Lcom/microsoft/sqlserver/jdbc/UTC;
   L1
    LINENUMBER 524 L1
    NEW java/util/SimpleTimeZone
    DUP
    ICONST_0
    LDC "UTC"
    INVOKESPECIAL java/util/SimpleTimeZone.<init> (ILjava/lang/String;)V
    PUTSTATIC com/microsoft/sqlserver/jdbc/UTC.timeZone : Ljava/util/TimeZone;
    RETURN
    MAXSTACK = 4
    MAXLOCALS = 0
```

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
@@ -517,11 +517,13 @@ class GregorianChange {
     }
 }
 
-// UTC/GMT time zone singleton. The enum type delays initialization until first use.
-enum UTC {
-    INSTANCE;
+final class UTC {
 
+    // UTC/GMT time zone singleton.
     static final TimeZone timeZone = new SimpleTimeZone(0, "UTC");
+
+    private UTC() {
+    }
 }
 
 final class TDSChannel {


v-xiangs:
Hello @marschall , thank you for your findings and knowledge sharing. However, I created a sample program and it seems like Enum does delay the initialization of the variable.. 

The following is my sample program.. could you please tell me what I do wrong? Thank you!

```
public class try_ENUM_initialization {

    public static void main(String[] args) throws InterruptedException {
        Hello hello = new Hello();

        Thread.sleep(3000);

        hello.say();
    }
}

class Hello {

    enum hhh {
        haha;

        static final Long ll = System.currentTimeMillis();
    }

    public Hello() {
        System.out.println("init");
        System.out.println(System.currentTimeMillis());
    }

    public void say() {
        System.out.println("say hello");
        System.out.println(hhh.ll);
    }
}
```

marschall:
Indeed class loading is delayed until the variable is accessed.

marschall:
Sorry for closing too early

Consider the following example

```java
public class DemonstraceEnumInitialization {

  public static void main(String[] args) throws InterruptedException {
    System.out.println("not accessing constants");

    TimeZone tz;
    tz = 1 < 2 ? null : UtcEnum.timeZone;
    System.out.println(tz);
    tz = 1 < 2 ? null : UtcClass.timeZone;
    System.out.println(tz);

    System.out.println("accessing constants");
    System.out.println(UtcEnum.timeZone);
    System.out.println(UtcClass.timeZone);

  }
}

enum UtcEnum {
  INSTANCE;

  static final TimeZone timeZone = utcTimeZone();

  private static TimeZone utcTimeZone() {
    System.out.println("UtcEnum.utcTimeZone()");
    return new SimpleTimeZone(0, "UTC");
  }

}

class UtcClass {
  static final TimeZone timeZone = utcTimeZone();

  private static TimeZone utcTimeZone() {
    System.out.println("UtcClass.utcTimeZone()");
    return new SimpleTimeZone(0, "UTC");
  }

}
```

The output is:
```
not accessing constants
null
null
accessing constants
UtcEnum.utcTimeZone()
java.util.SimpleTimeZone[id=UTC,offset=0,dstSavings=3600000,useDaylight=false,startYear=0,startMode=0,startMonth=0,startDay=0,startDayOfWeek=0,startTime=0,startTimeMode=0,endMode=0,endMonth=0,endDay=0,endDayOfWeek=0,endTime=0,endTimeMode=0]
UtcClass.utcTimeZone()
java.util.SimpleTimeZone[id=UTC,offset=0,dstSavings=3600000,useDaylight=false,startYear=0,startMode=0,startMonth=0,startDay=0,startDayOfWeek=0,startTime=0,startTimeMode=0,endMode=0,endMonth=0,endDay=0,endDayOfWeek=0,endTime=0,endTimeMode=0]
```

What we see is
- class loading does not happen until the constant is actually used
- using an enum makes no difference, it behaves the same as a class

v-xiangs:
Hello @marschall thank you for your sharing. I also noticed there was no difference for static variable between enum or class in term of initialization delay. 

My opinion was leaving it as is since there is no difference in behaviour or error caused by this. However, if there is any reason that we should use class over enum, please share with us. Then we can change it after testing in our test lab. Thank you!

v-xiangs:
Also, there is one more reason I prefer enum over class. Demonstrated by the following code:
```
public class DemonstraceEnumInitialization  {

    public static void main(String[] args) throws InterruptedException {
        System.out.println("not accessing constants");

        Long tz;
        tz = 1 < 2 ? null : UtcEnum.longV;
        System.out.println(tz);
        tz = 1 < 2 ? null : UtcClass.longV;
        System.out.println(tz);

        new UtcClass();

        System.out.println("accessing constants");

        Thread.sleep(3000);

        System.out.println(UtcEnum.longV);
        System.out.println(UtcClass.longV);

    }
}

enum UtcEnum {
    INSTANCE;
    static Long longV = System.currentTimeMillis();
}

class UtcClass {
    static Long longV = System.currentTimeMillis();
}
```
For class, it is possible to call the constructor. Once constructor is called, the static variable is initialized. With enum, we can make sure the variable is initialized only when it is called explicitly. What do you think?

marschall:
> For class, it is possible to call the constructor. Once constructor is called, the static variable is initialized.

While it is true that the enum prevents calling the constructor it does not prevent using the enum constant. Using the enum constant initializes the class just like calling the constructor. If you change the code to use the enum constant it behaves the same way as with a constructor

```java
public class DemonstrateEnumInitialization  {

    public static void main(String[] args) throws InterruptedException {
        System.out.println("not accessing constants");

        Long tz;
        tz = 1 < 2 ? null : UtcEnum.longV;
        System.out.println(tz);
        tz = 1 < 2 ? null : UtcClass.longV;
        System.out.println(tz);

        UtcEnum utc = UtcEnum.INSTANCE;

        System.out.println("accessing constants");

        Thread.sleep(3000);

        System.out.println(UtcEnum.longV);
        System.out.println(UtcClass.longV);

    }
}

enum UtcEnum {
    INSTANCE;
    static Long longV = System.currentTimeMillis();
}

class UtcClass {
    static Long longV = System.currentTimeMillis();
}
```

I would add four things
1. The UTC class is package scoped, it can only be seen by the driver and not by clients. The potential for misuse is minimal.
1. Using the constructor is almost guaranteed to generate a warning in the IDE. Either the "Non-static access to static member" check for the "Unused object allocation" check which is unfortunately disabled.
1. Using an enum to prevent instantiation is not idiomatic Java. I have not seen this pattern anywhere else. The pattern everybody else uses is the one suggested in [Effective Java](https://www.amazon.com/Effective-Java-2nd-Joshua-Bloch/dp/0321356683) which is to use a private constructor. This is also the pattern used in the `GregorianChange`, `Nanos` and `TDS` classes in the same file. I will update the PR with this.
1. Using an enum to prevent instantiation is not consistent with existing driver code as demonstrated by the `GregorianChange`, `Nanos` and `TDS` classes in the same file.

v-xiangs:
Hello @marschall Thank you for your reply. I am convinced by point 3
> The pattern everybody else uses is the one suggested in Effective Java which is to use a private constructor.

This definitely addresses my concern. I am going to test the changes and merge it once test is done. Thank you very much for your sharing and explanation.

PR #307, opened 2017-05-20
Improve enum usage

marschall:
Usage of the Java Enum feature can be improved by the driver. This
commit includes the following changes:

 - use ordinal where possible
 - make enum fields final where possible


v-afrafi:
Thank you for you contribution @marschall!

Issue #315, opened 2017-05-25
TVP issue: milliseconds are missed in destination table


v-xiangs:
If column is `time(x)` and TVP is used (with either ResultSet, Stored Procedure or SQLServerDataTable). The milliseconds or nanoseconds are not copied into the destination table.

For example:
if source column time(4) is: 15:39:27.6167
destination column time(4) will be: 15:39:27.0000

it only happens for time column. datetime2  and datetimeoffset are working correct.

v-xiangs:
fixed

PR #316, opened 2017-05-26
fix TVP time issue

v-xiangs:
fix issue #315


v-xiangs:
will create another PR for this fix

PR #317, opened 2017-05-26
Tvp time 4 fix rtw

v-xiangs:
for issue #315

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/TVP.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/TVP.java
@@ -112,7 +112,12 @@ class TVP {
             Object[] rowData = new Object[colCount];
             for (int i = 0; i < colCount; i++) {
                 try {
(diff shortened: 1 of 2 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/TVP.java
```diff
@@ -112,7 +112,12 @@ boolean isNull() {
             Object[] rowData = new Object[colCount];
             for (int i = 0; i < colCount; i++) {
                 try {
-                    rowData[i] = sourceResultSet.getObject(i + 1);
+                    if (java.sql.Types.TIME == sourceResultSet.getMetaData().getColumnType(i + 1)) {
+                        rowData[i] = sourceResultSet.getTimestamp(i + 1);
```
v-suhame: Looks good. can you just add a comment explaining why getTimestamp is used?

v-xiangs: Thank you! I added comment for it. Will merge the PR soon.

PR #318, opened 2017-05-26
SQLServerBulkCSVFileRecord accepting InputStream 

### LINKED ISSUE #311: SQLServerBulkCSVFileRecord should accept an InputStream as one of the constructor overloads
SQLServerBulkCSVFileRecord takes a path to a file on the local filesystem. This works fine when you are doing a bulk upload of a local file. However, this is problematic if the file exists in a remote filesystem, for eg, S3 or HDFS. In that case, we have to copy the file locally, then upload it. Yes, there is the interface, but implementing it requires implementing 95% of what the SQLServerBulkCSVFileRecord class provides out of the box (handling metadata, and parsing input from stream according to that metadata).

All this will be solved if there was an overload available that took an InputStream.

Also, since some of this code will be needed by implementers of the interface, it should be put in a common utility class so that it can be leveraged. I am specifically talking about  the follwoing methods:

SQLServerBulkCSVFileRecord:addColumnMetadataInternal()
SQLServerBulkCSVFileRecord:getRowData()

v-afrafi:
Fixes #311


src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyCSVTest.java
```diff
@@ -72,16 +80,46 @@ static void setUpConnection() {
...
+    /**
+     * test simple csv file for bulkcopy by passing a file from url
+     * 
+     * @throws SQLException
+     */
+    @Test
+    @DisplayName("Test SQLServerBulkCSVFileRecord with passing file from url")
+    void testCSVFromURL() throws SQLException {
+        try {
+            InputStream csvFileURL = new URL(
```
v-nisidh: Suggestion. Instead of `csvFileURL` can you use `csvFileIn` or `csvFileInputStream` or `csvFileStream`

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyCSVTest.java
```diff
@@ -131,6 +169,7 @@ void testCSV() {
                 stmt.dropTable(destTable);
             }
         }
+
```
v-nisidh: Can you add a test case for `firstLineIsColumnNames` having false value?

PR #320, opened 2017-05-30
Caching mvn dependencies for Appveyor

v-ahibr:
This is done to speed up builds

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -9,8 +9,7 @@ services:
   - mssql2016
   
 cache:
-  directories:
-    - $HOME\.m2 -> pom.xml
+  - $HOME\.m2 -> pom.xml
 
 build_script:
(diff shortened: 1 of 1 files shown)


appveyor.yml
```diff
@@ -7,10 +7,13 @@ environment:
 
 services:
   - mssql2016
+  
+cache:
+  - C:\Users\appveyor\.m2 -> pom.xml
```
v-nisidh: I think it should be directory caching. By following snippet it will cache whole directory.

```
cache:
  directories:
    - $HOME/.m2
```
or
```
cache:
 - $HOME/.m2
```

v-ahibr: changing to $HOME didnt work. 

With that update, appveryor can't find the cache, downloads all maven dependencies, and then creates a new build cache on every build.

Reverting back to original implementation

PR #321, opened 2017-05-31
Fixes Classloader leakage issue

### LINKED ISSUE #314: ActivityCorrelator is causing a classloader leak
[ActivityCorrelator](https://github.com/Microsoft/mssql-jdbc/blob/master/src/main/java/com/microsoft/sqlserver/jdbc/ActivityCorrelator.java) uses ThreadLocal, but it never calls ThreadLocal.remove() to clean it up. As a result, if a thread pool is used, the classloader that loaded ActivityCorrelator can never be garbage collected thus creating a memory leak.

Example of the path from GC root:
```
this     - value: org.apache.catalina.loader.WebappClassLoader #3
 <- <classLoader>     - class: com.microsoft.sqlserver.jdbc.ActivityId, value: org.apache.catalina.loader.WebappClassLoader #3
  <- <class>     - class: com.microsoft.sqlserver.jdbc.ActivityId, value: com.microsoft.sqlserver.jdbc.ActivityId class ActivityId
   <- value     - class: java.lang.ThreadLocal$ThreadLocalMap$Entry, value: com.microsoft.sqlserver.jdbc.ActivityId #2
    <- [54]     - class: java.lang.ThreadLocal$ThreadLocalMap$Entry[], value: java.lang.ThreadLocal$ThreadLocalMap$Entry #137
     <- table     - class: java.lang.ThreadLocal$ThreadLocalMap, value: java.lang.ThreadLocal$ThreadLocalMap$Entry[] #28
      <- threadLocals (thread object)     - class: org.apache.tomcat.util.threads.TaskThread, value: java.lang.ThreadLocal$ThreadLocalMap #28
```

v-nisidh:
connect to #314

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
@@ -44,10 +44,10 @@ import java.util.Locale;
 import java.util.Map;
 import java.util.Properties;
 import java.util.UUID;
+import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.Executor;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
-import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.logging.Level;
 
 import javax.sql.XAConnection;
@@ -2702,6 +2702,7 @@ public class SQLServerConnection implements ISQLServerConnection {
         // Clean-up queue etc. related to batching of prepared statement discard actions (sp_unprepare).
         cleanupPreparedStatementDiscardActions();
 
+        ActivityCorrelator.reset();
         loggerExternal.exiting(getClassNameLogging(), "close");
     }
 
@@ -5516,5 +5517,15 @@ final class SQLServerConnectionSecurityManager {
         if (null != security) {
             security.checkLink(dllName);
         }
+    }
+    
+    /**
+     * Although {@link ActivityCorrelator} used in different part of programs releasing connection might be good point to release {@link ThreadLocal}
+     * resource. We are releasing {@link ThreadLocal} resource in {@link SQLServerConnection#close()} too.
+     * @since 6.1.8
+     */
+    @Override
+    protected void finalize() throws Throwable {
+        ActivityCorrelator.reset();
     }
 }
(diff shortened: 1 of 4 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -5517,4 +5518,14 @@ public void checkLink() throws SecurityException {
...
         }
     }
+    
+    /**
+     * Although {@link ActivityCorrelator} used in different part of programs releasing connection might be good point to release {@link ThreadLocal}
+     * resource. We are releasing {@link ThreadLocal} resource in {@link SQLServerConnection#close()} too.
+     * @since 6.1.8
+     */
+    @Override
+    protected void finalize() throws Throwable {
```
marschall: [Oracle recommends against using finalize](https://stuartmarks.wordpress.com/2017/04/17/deprecation-of-object-finalize/) to the point where it's deprecated in Java 9. They recommend try-with-resources or `PhantomReference` instead.

marschall:
I wonder if this actually works in an application server with a connection pool. As far as I can tell this only works if the thread that creates the value is the one that removes the value.

The issue that I see is that [ThreadLocal.remove()](https://docs.oracle.com/javase/8/docs/api/java/lang/ThreadLocal.html#remove--) removes only the value of the current thread.

Removing the value is done by calling `#close()` on the connection. In an application server with a connection pool this is usually a thread belonging to the application server. The values seem to be created by statement calls. So this would likely be done by worker threads. These are likely different threads than the ones managing the connection pool. So calling `#close()` on the connection will not remove the values created in the worker threads.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
v-nisidh: Thanks @marschall : New learning !!! I will remove finalize.

v-nisidh:
@marschall Thanks. Next week I will test on tomcat & try to find out solution for connection pool. 

Actually previously I was thinking of shutdown hook like functionality but as it deals only with Current Thread need to find out some other alternative.

marschall:
@v-nisidh you'll probably have to consider at least two scenarios:
- the driver is deployed in the application server and the application server connection pool is used, this is the recommended deployment scenario, in this case the leak should not happen as the driver is not redeployed or undeployed
- the driver and connection pool is deployed with the application, this is not recommended as it's prone to class loader leaks, likely this is what is used in #314

v-nisidh:
Thanks @marschall : I able to detect leakage in connection pool. I am closing PR right now. After my fixes I will open same or new PR.

v-nisidh:
Although this is partial fix, it is good to have this fix. Still in case of connection pool, issue is still there. So #314 will remain open.

ajlam:
@nsidhaye, closing this PR as #465 contains this code fix. Please review #465 if you have time 😄

PR #342, opened 2017-06-15
Get rid of datatype converter

v-xiangs:
We can easily get rid of DatatypeConverter to resolve Java 9 module issue.

The driver is doing some unnecessary conversions: 
`SQLServerConnection#generateInterfaceLibVersion` converts driver version numbers (4 integers) to hex string, then `DatatypeConverter#parseHexBinary` converts them back to 4 bytes. We can remove this unnecessary conversion.

the original method works when every driver version number is smaller than 226. Tested with the new method, it is working fine with all valid version numbers.


src/test/java/com/microsoft/sqlserver/jdbc/connection/DriverVersionTest.java
```diff
@@ -0,0 +1,107 @@
...
+
+        assertEquals(originalBytes, newBytes, "Original: " + originalBytes + "; New: " + newBytes);
+    }
+
+    /**
+     * the original method that converts version number to hex string
+     * 
+     * @return
+     */
+    private String generateInterfaceLibVersion() {
```
v-nisidh: Like the idea of moving old implementation to Unit Testcases for testing implementation.

src/test/java/com/microsoft/sqlserver/jdbc/connection/DriverVersionTest.java
```diff
@@ -0,0 +1,107 @@
...
+    @Test
+    public void testConnectionDriver() {
+        // the original way to create version byte array
+        String interfaceLibVersion = generateInterfaceLibVersion();
+        byte originalVersionBytes[] = DatatypeConverter.parseHexBinary(interfaceLibVersion);
+
+        String originalBytes = Arrays.toString(originalVersionBytes);
+
+        // the new way to create version byte array
+        byte newVersionBytes[] = {(byte) build, (byte) patch, (byte) minor, (byte) major};
```
v-nisidh: Good to use with new method for getting newVersionBytes[]

v-xiangs:
closing this since new PR #349 is created

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -4058,7 +4056,8 @@ final boolean complete(LogonCommand logonCommand,
         byte appNameBytes[] = toUCS16(appName);
         byte serverNameBytes[] = toUCS16(serverName);
         byte interfaceLibNameBytes[] = toUCS16(interfaceLibName);
-        byte interfaceLibVersionBytes[] = DatatypeConverter.parseHexBinary(interfaceLibVersion);
+        byte interfaceLibVersionBytes[] = {(byte) SQLJdbcVersion.build, (byte) SQLJdbcVersion.patch, (byte) SQLJdbcVersion.minor,
```
v-xiangs: but Util class is not public.. test cannot access it.

Issue #343, opened 2017-06-15
Incorrect parameter count for INSERT with subquery


chadberchek:
When an INSERT statement contains a subquery with parameters, they are not included in the count returned by `PreparedStatement.getParameterMetaData().getParameterCount()`.

Example:
Create a table: `CREATE TABLE EXPERIMENT (ID NUMERIC(9,0));`
Run the following test, given an established `Connection conn`:
```java
@Test
public void parameterCountForInsertWithSubquery() throws SQLException {
    try (PreparedStatement ps = conn.prepareStatement(
        "insert into experiment (id) select ? where not exists (select * from experiment where id = ?)");
    ) {
        Assert.assertEquals(2, ps.getParameterMetaData().getParameterCount());
    }
}
```
Output:
```
java.lang.AssertionError: 
Expected :2
Actual   :1
```

Please note that the query actually executes fine if you set the two parameters on `ps` and call `executeUpdate()`.

Tested with driver 6.1.0.jre8 and also a local build of 8103dc386f1210f93216c8a37fa4e018f98bdf38 against Microsoft SQL Server 2008 R2 (SP2) - 10.50.4000.0 (X64)

ajlam:
Thanks for filing this issue @chadberchek. We need to do some investigation into the support we have for SQL Server 2008 and 2008R2.

v-xiangs:
Hello @chadberchek , we created a PR #350 for this issue. Hope this PR fixes this issue for you.

v-xiangs:
@chadberchek Hello, we are closing this issue since a fix is merged. Please feel free to re-open it or create new issues if the problem persists.

PR #350, opened 2017-06-19
Set fmtonly on to support complex query

v-xiangs:
fix issue #343 

the original way of using `Set fmtonly on` does not work with complex query or multiple tables. This PR uses a workaround by using `LEFT JOIN` query.

What the PR does is, converting the original query to a `LEFT JOIN` query and retrieve parameter metadata.


v-xiangs:
new PR #373 is created to get rid of Revert-MetadataCaching commit

PR #351, opened 2017-06-21
Fix parameter meta data with whitespace characters

### LINKED ISSUE #344: Errors in getParameterMetaData() when non-space characters in query
I discovered a variety of scenarios where `PreparedStatement.getParameterMetaData()` either throws an exception or returns an incorrect parameter count in a subsequent call to `getParameterCount()`, all of which appear to be related to the presence or absence of whitespace characters (spaces, newlines, tabs).

Below is a test suite Gist which demonstrates some of the issues. Five of the seven tests fail for me (the other two are just to demonstrate that the errors are only due to whitespace differences). Note that each test first executes the statement to prove that the statement really works and is not the problem; only fetching the parameter metadata is the problem. Please note that most of the issues with with line breaks also occur with tabs.

https://gist.github.com/chadberchek/1d0c686890a69e171e6e01d0ec6fffdb#file-experiment-java

Here is a summary of my results:

Test | Result
-----|--------
insertWithLineBreakBeforeTableName (build of 8103dc386f1210f93216c8a37fa4e018f98bdf38) | SQLServerException: Incorrect syntax near the keyword 'WHERE'.
insertWithLineBreakBeforeTableName (6.1.0.jre8) | SQLServerException: Unable to identify the table insert into experiment (id) values (?) for the metadata.
insertWithNoSpaceAfterTableName | SQLServerException: Invalid column name 'id'.
updateWithLineBreakAfterWhere | SQLServerException: Incorrect syntax near the keyword 'where'.
updateWithLineBreakBeforeFirstParameter | Parameter count mismatch Expected :2 Actual   :0
updateWithLineBreakAfterUpdate | SQLServerException: Unable to identify the table update experiment set id = ? where id = ? for the metadata.

Tested with driver 6.1.0.jre8 and also a local build of 8103dc386f1210f93216c8a37fa4e018f98bdf38 (for JDK8) against Microsoft SQL Server 2008 R2 (SP2) - 10.50.4000.0 (X64).

v-xiangs:
Fixes #344

### DIFF
--- a/src/test/java/com/microsoft/sqlserver/jdbc/parametermetadata/ParameterMetaDataWhiteSpaceTest.java
+++ b/src/test/java/com/microsoft/sqlserver/jdbc/parametermetadata/ParameterMetaDataWhiteSpaceTest.java
@@ -1,0 +1,144 @@
+/*
+ * Microsoft JDBC Driver for SQL Server
+ * 
+ * Copyright(c) Microsoft Corporation All rights reserved.
+ * 
+ * This program is made available under the terms of the MIT License. See the LICENSE file in the project root for more information.
+ */
+package com.microsoft.sqlserver.jdbc.parametermetadata;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import java.sql.DriverManager;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Statement;
+
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+import org.junit.platform.runner.JUnitPlatform;
+import org.junit.runner.RunWith;
+
+import com.microsoft.sqlserver.jdbc.SQLServerConnection;
+import com.microsoft.sqlserver.testframework.AbstractTest;
+import com.microsoft.sqlserver.testframework.util.RandomUtil;
+
+@RunWith(JUnitPlatform.class)
+public class ParameterMetaDataWhiteSpaceTest extends AbstractTest {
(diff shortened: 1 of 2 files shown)


chadberchek:
I was just looking at the PR and it occurred to me that it might still not handle tabs correctly since I did not see those in the regexes that were modified. I didn't provide separate test cases for those but I mentioned it in the comments. The tests would just be the same as the ones I showed but with tab characters of course. I'm not sure what characters the server accepts as delimiters, but it might be worth seeing if it corresponds to the `\s` built-in character class or if there are any other characters that might be missing.

v-xiangs:
@chadberchek thank you very much for the excellent finding. As you suggested, I added support for tab and form feed. I think that should be all the whitespaces we need to support. 

I am going to add your test code into our test suite. Thank you.

v-xiangs:
Test is added now. @chadberchek again, I very appreciate the test you provided.

src/test/java/com/microsoft/sqlserver/jdbc/parametermetadata/ParameterMetaDataWhiteSpaceTest.java
```diff
@@ -0,0 +1,144 @@
...
+    @BeforeAll
+    public static void BeforeTests() throws SQLException {
+        connection = (SQLServerConnection) DriverManager.getConnection(connectionString);
+        stmt = connection.createStatement();
+        createCharTable();
+    }
+
+    @AfterAll
+    public static void dropTables() throws SQLException {
+        stmt.execute("if object_id('" + tableName + "','U') is not null" + " drop table " + tableName);
```
v-nisidh: Please use `Utils.dropObjectIfExists(tableName,"IsTable",stmt)` function to drop table.

v-xiangs: done

v-xiangs:
new PR #371 is created to get rid of Revert-MetadataCaching

PR #362, opened 2017-06-27
Early prototype of MARS

v-ahibr:


v-ahibr:
There are still too many logs being printed for travic ci builds to pass, but I don't want to remove any more of them. The remaining logs will be very helpful for further implementing MARS

PR #371, opened 2017-07-06
Fix parameter meta data with whitespace characters

### LINKED ISSUE #344: Errors in getParameterMetaData() when non-space characters in query

v-xiangs:
new PR equivalent to #351 which fixes #344

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerParameterMetaData.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerParameterMetaData.java
@@ -73,9 +73,10 @@ public final class SQLServerParameterMetaData implements ParameterMetaData {
      *            the list of columns
      * @param columnStartToken
      *            the token that prfixes the column set
+     * @throws SQLServerException 
      */
     /* L2 */ private String parseColumns(String columnSet,
-            String columnStartToken) {
+            String columnStartToken) throws SQLServerException {
         StringTokenizer st = new StringTokenizer(columnSet, " =?<>!\r\n\t\f", true);
         final int START = 0;
         final int PARAMNAME = 1;
@@ -130,9 +131,10 @@ public final class SQLServerParameterMetaData implements ParameterMetaData {
      *            the sql syntax
      * @param columnMarker
      *            the token that denotes the start of the column set
+     * @throws SQLServerException 
      */
     /* L2 */ private String parseInsertColumns(String sql,
-            String columnMarker) {
+            String columnMarker) throws SQLServerException {
         StringTokenizer st = new StringTokenizer(sql, " (),", true);
         int nState = 0;
         String sLastField = null;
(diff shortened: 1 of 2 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerParameterMetaData.java
```diff
@@ -330,7 +330,7 @@ private String escapeParse(StringTokenizer st,
         String fullName;
         nameFragment = firstToken;
         // skip spaces
-        while (" ".equals(nameFragment) && st.hasMoreTokens()) {
+        while ((0 == nameFragment.trim().length()) && st.hasMoreTokens()) {
```
ulvii: Do we need a null check here?

v-xiangs: In this case, if we look at the caller of the method, the variable will never be null since it's a token returned from StringTokenizer (instead of returning null, NoSuchElementException will be thrown if there are no more tokens in this tokenizer).

You are right, it's a good practice to always check for null.

ulvii: Yes, currently `firstToken` cannot be `null`, but there might be more caller methods in the future :)

v-xiangs: Yes, you are right :) Done

PR #372, opened 2017-07-06
fix for queries without parameters using preparedStatement

### LINKED ISSUE #370: Regression when running DDL (CREATE SCHEMA or CREATE VIEW) in PreparedStatement
The following code:

```java
try (Connection c = new com.microsoft.sqlserver.jdbc.SQLServerDriver().connect(url, properties)) {
    try (PreparedStatement s1 = c.prepareStatement("create schema x");
         PreparedStatement s2 = c.prepareStatement("drop schema x")) {
        System.out.println(s1.execute());
        System.out.println(s2.execute());
    }
}
```

... works with:

```xml
<dependency>
    <groupId>com.microsoft.sqlserver</groupId>
    <artifactId>mssql-jdbc</artifactId>
    <version>6.1.0.jre8</version>
</dependency>
```

... but it no longer works with:

```xml
    <version>6.2.0.jre8</version>
```

The error I'm getting is this:

```
com.microsoft.sqlserver.jdbc.SQLServerException: Incorrect syntax near the keyword 'schema'.
	at com.microsoft.sqlserver.jdbc.SQLServerException.makeFromDatabaseError(SQLServerException.java:258)
	at com.microsoft.sqlserver.jdbc.SQLServerStatement.getNextResult(SQLServerStatement.java:1547)
	at com.microsoft.sqlserver.jdbc.SQLServerPreparedStatement.doExecutePreparedStatement(SQLServerPreparedStatement.java:528)
	at com.microsoft.sqlserver.jdbc.SQLServerPreparedStatement$PrepStmtExecCmd.doExecute(SQLServerPreparedStatement.java:461)
	at com.microsoft.sqlserver.jdbc.TDSCommand.execute(IOBuffer.java:7151)
	at com.microsoft.sqlserver.jdbc.SQLServerConnection.executeCommand(SQLServerConnection.java:2689)
	at com.microsoft.sqlserver.jdbc.SQLServerStatement.executeCommand(SQLServerStatement.java:224)
	at com.microsoft.sqlserver.jdbc.SQLServerStatement.executeStatement(SQLServerStatement.java:204)
	at com.microsoft.sqlserver.jdbc.SQLServerPreparedStatement.execute(SQLServerPreparedStatement.java:445)
	at SQLServer.main(SQLServer.java:69)
```

A workaround is to resort to static statements. The same is true for `CREATE VIEW` statements:

```java
// Works on both versions
try (Statement s1 = c.createStatement();
     Statement s2 = c.createStatement()) {
    System.out.println(s1.execute("create view x as select 1 a"));
    System.out.println(s2.execute("drop view x"));
}

// Works only on version 6.1.0
try (PreparedStatement s1 = c.prepareStatement("create view x as select 1 a");
     PreparedStatement s2 = c.prepareStatement("drop view x")) {
    System.out.println(s1.execute());
    System.out.println(s2.execute());
}
```

I think that's quite a significant regression. Other DDL statements might be affected as well.

v-afrafi:
Fixes #370

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
@@ -704,7 +704,8 @@ public class SQLServerPreparedStatement extends SQLServerStatement implements IS
         tdsWriter.writeRPCStringUnicode(preparedSQL);
 
         // <formal parameter defn> IN
-        tdsWriter.writeRPCStringUnicode((preparedTypeDefinitions.length() > 0) ? preparedTypeDefinitions : null);
+        if (preparedTypeDefinitions.length() > 0) 
+            tdsWriter.writeRPCStringUnicode(preparedTypeDefinitions);
     }
 
     private void buildServerCursorExecParams(TDSWriter tdsWriter) throws SQLServerException {
@@ -966,7 +967,6 @@ public class SQLServerPreparedStatement extends SQLServerStatement implements IS
             if (needsPrepare 
                 && !connection.getEnablePrepareOnFirstPreparedStatementCall() 
                 && !isExecutedAtLeastOnce
-                && preparedTypeDefinitions.length() > 0
             ) {
                 buildExecSQLParams(tdsWriter);
                 isExecutedAtLeastOnce = true;
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
@@ -966,6 +966,7 @@ public class SQLServerPreparedStatement extends SQLServerStatement implements IS
             if (needsPrepare 
                 && !connection.getEnablePrepareOnFirstPreparedStatementCall() 
                 && !isExecutedAtLeastOnce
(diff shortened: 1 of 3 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -966,6 +966,7 @@ private boolean doPrepExec(TDSWriter tdsWriter,
             if (needsPrepare 
                 && !connection.getEnablePrepareOnFirstPreparedStatementCall() 
                 && !isExecutedAtLeastOnce
+                && preparedTypeDefinitions.length() > 0
```
TobiasSQL: We should not make this check here. sp_executesql works fine without arguments. The problem is in the buildExecSQLParams method. Basically this  works fine sp_prepexec/buildPrepExecParams but not for sp_executesql:
`tdsWriter.writeRPCStringUnicode((preparedTypeDefinitions.length() > 0) ? preparedTypeDefinitions : null);`

So we should just change that to:
`if (preparedTypeDefinitions.length() > 0)
tdsWriter.writeRPCStringUnicode(preparedTypeDefinitions);`

v-afrafi: Cool, thanks @TobiasSQL . That was my initial thought too. Will commit the changes.

PR #374, opened 2017-07-07
Batch support is broken, when setNull is used

rPraml:
*Use case*
In our application we got sporadically a PRIMARY KEY violation, which we could not explain. After investigating we found out, that this happens since we updated to 6.2.0 (6.1.7.jre8-preview works)

*Steps to reproduce*
do several pstmt.setNull & pstmt.setString commands ln a certain order

*How to fix*
- Either avoid that the "Could not find prepared statement with handle X" error occurs. It seems that there is a workaround when this error occurs https://github.com/Microsoft/mssql-jdbc/blob/0b64232330748ecdffd2302886eddb25b6776c76/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java#L531
- OR fix the retry. It seems that the last statement is (partially) executed und executing the same batch again leads into the duplicate-key-exception

### DIFF
--- a/src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/BatchExecutionTestWithNull.java
+++ b/src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/BatchExecutionTestWithNull.java
@@ -1,0 +1,222 @@
+/*
+ * Microsoft JDBC Driver for SQL Server
+ * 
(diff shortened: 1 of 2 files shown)


v-afrafi:
@rPraml Thank you for providing the test. We are investigating this.

v-afrafi:
Closing this pr as got merged with pr #393. Thank you @rPraml for providing the valuable repro!

Issue #381, opened 2017-07-12
Missing Locale for checking properties passed by user. 


v-afrafi:
The driver is missing Locale in some String functions which produces an incorrect parameter being passed in other languages. 
For instance:
 https://github.com/Microsoft/mssql-jdbc/blob/master/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java#L1360

Reference of issue:
https://docs.microsoft.com/en-us/sql/connect/jdbc/microsoft-jdbc-driver-for-sql-server

peterbae:
fixed by #384

PR #384, opened 2017-07-14
Adding locale fix for turkey locale issue when lowercasing an "i"

peterbae:
using toLowerCase() without any locale makes it implicitly toLowerCase(Locale.getDefault()). This is fine in most counties but in Turkey, the default locale converts uppercase "I" to a dotless "i", which is not the same as the lowercase "i" in English. Passing Locale.ENGLISH (or Locale.US) in toLowerCase() solves this problem. I have made these changes to all the applicable places in the codebase.

Fixes issue #381


nsidhaye:
@peterbae: Please ignore my 1st suggestion. From mobile I could not able to edit comments. Although 2nd suggestion is still valid.

peterbae:
Hi Nikhil, thanks for the suggestion. I think providing a wrapper function for toLowerCase() and adding more locale flexibility is a good idea, but currently we only want to explicitly use English as our locale to avoid case conversion problems, and I think this change solves that problem. I'll come back to that feature later when it becomes necessary.

PR #388, opened 2017-07-17
fixes metadata caching regression with DDL queries using pstmt

v-afrafi:
Adding it to dev branch also.


v-xiangs:
this PR is equivalent to PR #372

PR #422, opened 2017-07-31
New connection property: sslProtocol

ulvii:
Adding a new connection property to let the users specify TLS protocol keyword. Possible values: "TLS", "TLSv1", "TLSv1.1", "TLSv1.2". The default is "TLS". These keywords might behave differently depending on JRE( Oracle, IBM, SAP ), so the users should read the documentation before using them. 

Example use-case: 
When **Suite B** or **SP800-131a** standards are used, only TLSv1.2 must be enabled.


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDriver.java
```diff
@@ -117,6 +117,46 @@ else if (value.toLowerCase(Locale.US).equalsIgnoreCase(ColumnEncryptionSetting.D
     }
 }
 
+enum SSLProtocol {
```
nsidhaye: I like enum... Good One.

src/test/java/com/microsoft/sqlserver/jdbc/connection/SSLProtocolTest.java
```diff
@@ -0,0 +1,104 @@
...
+        }
+    }
+
+    /**
+     * Test with unsupported protocols.
+     * 
+     * @throws Exception
+     */
+    @Test
+    public void testConnectWithWrongProtocols() throws Exception {
```
nsidhaye: **SUGGESTION:** You can use JUnit's ability of parametrized test. 

In future JUnit might give (or may be available) to pass parameters from some configuration file so no need to touch test code to add extra tests just like [TestNG](http://testng.org/doc/)

````
   @DisplayName("TestForWrongValues") 
    @ParameterizedTest
    @ValueSource(strings={"SSLv1111","SSLv2222","SSLv3111", "SSLv2Hello1111","TLSv1.11","TLSv2.4","HTTPS"}) 
    public void testConnectWithWrongProtocol(String sslProtocol) throws Exception {
        testWithUnSupportedProtocols(sslProtocol); 
    }
````

You need to upgrade JUnit 1.0.0-M4 & add following dependency

````
               <dependency>
			<groupId>org.junit.jupiter</groupId>
			<artifactId>junit-jupiter-params</artifactId>
			<version>5.0.0-M4</version>
			<scope>test</scope>
		</dependency>	
````

Comment (ulvii) [inline src/test/java/com/microsoft/sqlserver/jdbc/connection/SSLProtocolTest.java]: Hi @nsidhaye,
Thank you for suggestion. I modified the test a bit to not introduce a new dependency and make it compatible with previous JUnit versions.

**Comment by ulvii** [comment]
Link to Wiki page: https://github.com/Microsoft/mssql-jdbc/wiki/SSLProtocol

File: src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDriver.java [hunk 3]
```diff
@@ -117,6 +117,46 @@ else if (value.toLowerCase(Locale.US).equalsIgnoreCase(ColumnEncryptionSetting.D
...
+    }
+
+    public String toString() {
+        return name;
+    }
+
+    static SSLProtocol valueOfString(String value) throws SQLServerException {
+        SSLProtocol protocol = null;
+
+        if (value.toLowerCase(Locale.ENGLISH).equalsIgnoreCase(SSLProtocol.TLS.toString())) {
```
jorsol: The toLowerCase(Locale.ENGLISH) is unnecesary since equalsIgnoreCase is used.

PR #425, opened 2017-08-01
Give option to skip certian tests base on Tag

v-xiangs:
give option to user to skip certain tests base on tags. (revised base on PR #114 )

For example, running the following command to build the driver will skip tests that are tagged as `slow`:
```
mvn clean install -Pbuild42 -DskipTestTag=slow
```

If `-DskipTestTag` is not provided in the command, all tests will be executed.

***Will need to update ReadMe after next release.***

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -314,10 +314,10 @@
 				<groupId>org.apache.maven.plugins</groupId>
 				<artifactId>maven-surefire-plugin</artifactId>
 				<version>2.19</version>
-				
+
 				<configuration>
 					<properties>
-						<excludeTags>slow</excludeTags>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -42,6 +42,8 @@
 
 	<properties>
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+		<junit.platform.version>1.0.0-M3</junit.platform.version>
```
nsidhaye: Super !!! 

I like this. In future you will able to use maven versioning plugin to upgrade latest 3rd party dependencies

Issue #426, opened 2017-08-01
Metadata caching failure on SQL Server 2008


v-xiangs:
## Driver version or jar name
6.2.0 or 6.3.0 (with metadata caching)

## SQL Server version
SQL Server 2008

## Client operating system
Windows 10

## Java/JVM version
Oracle JDK1.8

## Table schema
The second line fails
```
this.executeSQL(con, "CREATE TABLE #update1 (col INT);INSERT #update1 VALUES (1);");
this.executeSQL(con, "CREATE PROC #updateProc1 AS UPDATE #update1 SET col += 1; IF EXISTS (SELECT * FROM #update1 WHERE col % 5 = 0) THROW 99586, 'Prepared handle GAH!', 1;");
```

## Problem description
When building the driver on SQL Server 2008, one test fails with this error:
```
com.microsoft.sqlserver.jdbc.SQLServerException: Incorrect syntax near 'THROW'.
```

I believe it is related to metadata caching. It only fails on SQL Server 2008.

## Expected behavior and actual behavior
Please tell us what should happen and what happened instead

## Repro code
The test is in the package:
com.microsoft.sqlserver.jdbc.unit.statement.PreparedStatementTest#testStatementPooling

v-xiangs:
this is not a bug in driver. a test uses a query that is not supported on SQL Server 2008. PR #427 fixes it. Closing this issue for now.

PR #427, opened 2017-08-01
Fix test failures on sql server 2008

v-xiangs:
for issue #426 

it's actually not a bug in the driver. The query in the test is not supported on SQL Server 2008. 

Exclude the test for SQL Server 2008 now.

This PR also fixes other failures on SQL Server 2008


src/test/java/com/microsoft/sqlserver/jdbc/preparedStatement/BatchExecutionWithNullTest.java
```diff
@@ -117,13 +117,10 @@ public void testSetup() throws TestAbortedException, Exception {
 
     @AfterAll
     public static void terminateVariation() throws SQLException {
-
+        connection = DriverManager.getConnection(connectionString);
```
ulvii: Why do we need to do this?

v-xiangs: otherwise it fails on 2008 because the connect is null

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/PreparedStatementTest.java
```diff
@@ -170,6 +170,8 @@ public void testStatementPooling() throws SQLException {
             }
         }
 
+        Utils.skipTestForSQLServer2008((SQLServerConnection) DriverManager.getConnection(connectionString));
```
ulvii: Instead of skipping the test, we can make it compatible with SQL Server 2008 :)

You can try replacing  `throw`  with `RAISERROR`

v-xiangs: man, you are so good 👍

v-xiangs: Hello @TobiasSQL, just want to confirm with you. If we change throw to RAISERROR , the test should still be valid, right ?

ulvii: Apparently `RAISERROR` fails too. This might be an actual driver issue. We decided to not merge it until we investigate the test more deeply.

TobiasSQL:
@v-xiangs, in `retryBasedOnFailedReuseOfCachedHandle()` we check for error number 99586 specifically to enable this test case. This is presumably why the change to `RAISERROR` for the test case did not work (it will return error number 50000). 

You can fix this by making sure to add the error message to the instance somewhere before running `RAISERROR`.
```
IF NOT EXISTS (SELECT * FROM sys.messages WHERE message_id = 99586) 
	EXEC sp_addmessage 99586, 16, 'JDBC test case error';
```
After this you can use `RAISERROR(99586, 16, 1);` instead of `THROW` :-)

v-xiangs:
Hello @TobiasSQL Thank you very much for the explanation :smile: we very appreciate your help.

After I modified the test, now it fails with another error message, I guess it's a bug in the driver which is related to metadata caching. Right now I am working on a fix for it, but it would be great if you could have time to take a look too :thumbsup: Thank you.
[try_issue_426.txt](https://github.com/Microsoft/mssql-jdbc/files/1209232/try_issue_426.txt)

v-xiangs:
closing this PR since new PR #436 is created.

PR #325, opened 2017-06-01
Sslprotocol

v-nisidh:
Please DO NOT MERGE. This Pull Request is only for review for approach taken for IBM FIPS & SSL Protocol

PR #433, opened 2017-08-05
Removing connection properties: fipsProvider and fips

ulvii:
**This is a breaking change and should not be merged. Creating this PR early to allow discussions.** 
1. Removing connection property `fipsProvider`
Currently, this property is only used to specify the name of a provider which guarantees that the implementation of the type requested is from the named provider. Note that, the specified provider must still be registered in the security provider list.
```
KeyStore.getInstance(trustStoreType, fipsProvider);
````
If `fipsProvider` is not specified,`KeyStore.getInstance` traverses the list of registered security Providers, starting with the most preferred Provider.

2. Removing connection property `fips`.
Once FIPS mode is enabled for JAVA (either from JAVA options or java.security), all the driver needs to know is `trustStoreType`. There is no need to set `fips` to true, because the driver does not really enable FIPS. 

To sum up, after proposed changes, user will have to enable FIPS mode for JAVA and pass `trustStoreType` to driver.

### CODE REVIEW

**Comment by ulvii** [comment]
Hi @v-nisidh/ @nsidhaye, please feel free to leave feedback :)

File: src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java [hunk 1]
```diff
@@ -1710,14 +1697,12 @@ void enableSSL(String host,
...
 
-                // if the host name in cert provided use it or use the host name Only if it is not FIPS
-                if (!isFips) {
-                    if (null != hostNameInCertificate) {
-                        tm = new TrustManager[] {new HostNameOverrideX509TrustManager(this, (X509TrustManager) tm[0], hostNameInCertificate)};
-                    }
-                    else {
-                        tm = new TrustManager[] {new HostNameOverrideX509TrustManager(this, (X509TrustManager) tm[0], host)};
-                    }
+                // if the host name in cert provided use it or use the host name
```
nsidhaye: In case of FIPS our driver should delegate responsibility of validating certificate to JVM. 

In this case we are using our own TrustManager i.e. `HostNameOverrideX509TrustManager` which will throw following exception : 

````
FIPS mode: only SunJSSE TrustManagers may be used
````
Ref: [OpenJDK: SSLContextImpl](https://github.com/openjdk-mirror/jdk7u-jdk/blob/master/src/share/classes/sun/security/ssl/SSLContextImpl.java#L126)

If I am correctly remember in order to remove error `FIPS mode: only SunJSSE TrustManagers may be used` we review our FIPS implementation and added fips attribute. 

If you want to get rid of fips attribute then you have to come up with function which will detect if JVM is in FIPS mode or not. You can find logic in our test cases as well as on internet. 

I will suggest instead of merging back in master / dev branch you can come up with `FIPS` or `FIPSRnD` branch and test with different FIPS implementation. If users are comfortable with then you can merge back in main stream

Remember to test with at-least following FIPS implementations: 

1. BouncyCastle 
1. IBM
1. OpenJDK 9  (Not Tested yet)

File: src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java [hunk 2]
```diff
@@ -1577,9 +1577,7 @@ void enableSSL(String host,
         String tmfDefaultAlgorithm = null;  // Default algorithm (typically X.509) used by the TrustManagerFactory
         SSLHandhsakeState handshakeState = SSLHandhsakeState.SSL_HANDHSAKE_NOT_STARTED;
 
-        boolean isFips = false;
         String trustStoreType = null;
-        String fipsProvider = null;
```
nsidhaye: Removing fipsProvider will not directly impact. But think of following cases: 

You can configure BouncyCastle FIPS provider in two ways : 
1.  By configuration files (java.security)
1.  Run time by passing FIPS provider .

I added fipsProvider for 2nd case.  

## Also on Oracle site mentioned following: 

**Configuring SunJSSE for FIPS Mode**

SunJSSE is configured in FIPS mode by associating it with an appropriate FIPS 140 certified cryptographic provider that supplies the implementations for all cryptographic algorithms required by SunJSSE. This can be done in one of the following ways:

1. edit the file ${java.home}/lib/security/java.security and modify the line that lists com.sun.net.ssl.internal.ssl.Provider to list the provider name of the FIPS 140 certified cryptographic provider. For example if the name of the cryptographic provider is SunPKCS11-NSS, change the line from

      security.provider.4=com.sun.net.ssl.internal.ssl.Provider`

    to

      security.provider.4=com.sun.net.ssl.internal.ssl.Provider SunPKCS11-NSS

    The class for the provider of the given name must also be listed as a security provider in the java.security file.

1. at runtime, call the constructor of the SunJSSE provider that takes a java.security.Provider object as a parameter. For example, if the variable cryptoProvider is a reference to the cryptographic provider, call new com.sun.net.ssl.internal.ssl.Provider(cryptoProvider).

  1. at runtime, call the constructor of the SunJSSE provider that takes a String object as a parameter. For example if the cryptographic provider is called SunPKCS11-NSS call new com.sun.net.ssl.internal.ssl.Provider("SunPKCS11-NSS"). A provider with the specified name must be one of the configured security providers.

Within a given Java process, SunJSSE can be used either in FIPS mode or in default mode, but not both at the same time. Once SunJSSE has been initialized, it is not possible to change the mode. This means that if one of the runtime configuration options is used (option 2 or 3), the configuration must take place before any SSL/TLS operation.

Ref: [Oracle FIPS](https://docs.oracle.com/javase/8/docs/technotes/guides/security/jsse/FIPS.html)

ulvii: Hi @nsidhaye ,
Thank you so much for detailed comments. We appreciate the time and effort. 

As I mentioned above, if `fipsProvider` is not specified,`KeyStore.getInstance()` traverses the list of registered security Providers - `Security.getProviders()` and returns the matching `KeyStore` object. 

Example, in case of option 3, user would make sure `BouncyCastleFipsProvider` is installed before creating a connection.

```
Security.addProvider(new org.bouncycastle.jcajce.provider.BouncyCastleFipsProvider());
Security.addProvider(new com.sun.net.ssl.internal.ssl.Provider ("BCFIPS"));
ds.setTrustStoreType("BCFKS");
connection = (SQLServerConnection) ds.getConnection();
```
When the drivers calls `KeyStore.getInstance(trustStoreType)` **before TLS Handshake**, `BCFIPS` will already be in the list of installed providers.

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
ulvii: Thank you for pointing this out. I am able to reproduce the exception, will investigate and get back to you.

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
ulvii: After further investigation, I figured out the root cause of the issue, 
Provider `IBMJCEFIPS` does not contain a keystore. The users should use non-IBMJCEFIPS keystores to store the keys. See [this page](https://www.ibm.com/support/knowledgecenter/en/SSYKE2_7.0.0/com.ibm.java.security.component.70.doc/security-component/progpractices.html). This means, the provider that contains keystore is not necessarily the same as the provider that contains FIPS. Below is a sample app that would work for IBM users with the current implementation.
```
ds.setTrustStoreType("JCEKS");
ds.setFIPS(true);
ds.setFIPSProvider("IBMJCE");  //Not IBMJCEFIPS
connection = (SQLServerConnection) ds.getConnection();
```
The suggestion above does not require any breaking changes, but the connection property `fipsProvider` has a confusing name.

nsidhaye: @ulvi :

In this case as FIPSProvider is non-FIPS it's not fip connection. One can test it by [FipsEnvTest#isFips](https://github.com/Microsoft/mssql-jdbc/blob/master/src/test/java/com/microsoft/sqlserver/jdbc/fips/FipsEnvTest.java#L142)
> ds.setTrustStoreType("JCEKS");
ds.setFIPS(true);
ds.setFIPSProvider("IBMJCE");  //Not IBMJCEFIPS
connection = (SQLServerConnection) ds.getConnection();

For one of user I forked out SSLProtocol which having 2 fixes: 
1. Configurable SSL Protocol (Which you already worked on)
2. Made FIPSProvider optional as IBMFIPS not like FIPSProvider while calling `Keystore.getInstance` . (Need to test what happens in JDK 9 & Bouncy Castle for configurable configuraions)

Unfortunately, we did not finalize on solution with user, but this worked on user environment with appropriate SSL certificate.

If you use my  [SSLProtocol fork](https://github.com/Microsoft/mssql-jdbc/pull/325) and use following code for IBM FIPS.

```
//ds.setTrustStoreType("JCEKS"); (As per my understanding IBM works well with JKS too in FIPS)
ds.setFIPS(true);
//ds.setFIPSProvider("IBMJCE");  For IBM FIPS one should not set FIPSProvider.
connection = (SQLServerConnection) ds.getConnection();
```
Possibly @v-xiangs might have some information. Feel free to call me.

ulvii: Hi @nsidhaye ,
Thank you for the comments.

> In this case as FIPSProvider is non-FIPS it's not fip connection.

`fipsProvider` connection property is only used to load the keys. Users will still have to create the keys using a FIPS-approved provider. Non FIPS provider keystores can be used to **store** those keys and this will not affect FIPS compliant mode. Please, take a look at [this article.](https://www.ibm.com/support/knowledgecenter/en/SSYKE2_7.0.0/com.ibm.java.security.component.70.doc/security-component/progpractices.html)

>  One can test it by FipsEnvTest#isFips

Of course this test will fail if you pass in non-FIPS provider. However, as I mentioned above, non FIPS provider is used to **only store the keys.**

> If you use my SSLProtocol fork and use following code for IBM FIPS.

Let me explain what happens when you don't set `fipsProvider` property. In this case, since you did not specify the provider, the driver will traverse the registered providers and find the matching keystore. So if you set `keystoreType` to `JKS`, the matching provider that contains this Keystore is a non-FIPS provider. 

> Made FIPSProvider optional as IBMFIPS does not like FIPSProvider while calling Keystore.getInstance(...)

`Keystore.getInstance(storeType, "IBMJCEFIPS")` does not work, because `IBMJCEFIPS` does not contain any keystores. **That's why we need to use non-FIPS provider to store the keys.** If the given provider contains the specified `storeType`, `Keystore.getInstance(storeType, provider)` is perfectly valid. See [this page](https://www.ibm.com/support/knowledgecenter/SSYKE2_7.0.0/com.ibm.java.security.component.70.doc/security-component/JceDocs/keystore.html). 

I am definitely open to a discussion over a call. Please email me your contact details at v-ulibra@microsoft.com.

Thanks again for valuable comments.

ulvii:
Closing this PR and creating a new one which removes `fipsProvider` connection property only.  #460

PR #436, opened 2017-08-09
Process all responses before proceed and fix retry logic for metadata caching

v-xiangs:
fix 2 issues in the driver related to metadata caching. 

The issues are found by replacing `throw` with `raiserror` in the test, in order to run the test against SQL Server 2008

Repo code for first issue:
[try_issue.txt](https://github.com/Microsoft/mssql-jdbc/files/1210020/try_issue.txt)

Another issue happens when using batching. The data is inserted/updated more times than expected.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
@@ -6476,23 +6476,13 @@ final class TDSReader {
         payloadOffset = 0;
         return true;
     }
-    
+
     boolean hasNextPacket() throws SQLServerException {
-        if (null == currentPacket.next) {
-            return false;
-        }
-        else {
-            return true;
-        }
+        return (null != currentPacket.next);
     }
 
     boolean endOfCurrentPacket() {
-        if (payloadOffset == currentPacket.payloadLength) {
-            return true;
-        }
-        else {
-            return false;
-        }
+        return (payloadOffset == currentPacket.payloadLength);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


v-xiangs:
Hello @TobiasSQL , we know you are very busy, but if you could have a few minutes to take a look at this PR, it would be very helpful to us. Thank you 😃

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -6476,6 +6476,24 @@ private boolean nextPacket() throws SQLServerException {
...
+    
+    boolean hasNextPacket() throws SQLServerException {
+        if (null == currentPacket.next) {
+            return false;
+        }
+        else {
+            return true;
+        }
+    }
+
```
cheenamalhotra: Method `hasNextPacket()` content can be compressed to:
`return (null!=currentPacket.next);`

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -6476,6 +6476,24 @@ private boolean nextPacket() throws SQLServerException {
...
+    boolean endOfCurrentPacket() {
+        if (payloadOffset == currentPacket.payloadLength) {
+            return true;
+        }
+        else {
+            return false;
+        }
```
cheenamalhotra: Method `endOfCurrentPacket()` content can be compressed to:
`return (payload == currentPacket.payloadLength);`

v-xiangs:
Things to verify for this fix:
1. Make sure we don't lose error messages
2. Only retry if the call was "exec sp_execute 14;" and not "INSERT t VALUES(1); exec sp_execute 14;"
3. Manually get the handle and explicitly kill the handle for a specific error type.

v-xiangs:
Things to verify for this fix:

- [ ]  Make sure we don't lose error messages
- [ ] Only retry if the call was "exec sp_execute 14;" and not "INSERT t VALUES(1); exec sp_execute 14;"
- [ ] Manually get the handle and explicitly kill the handle for a specific error type.

v-xiangs:
close it due to new PR #543

Issue #308, opened 2017-05-20
Statement leak in SQLServerPreparedStatement.getMetaData()


marschall:
I believe there is a statement leak in `SQLServerPreparedStatement.getMetaData()`. If no result set is present `#getMetaData()` calls `#buildExecuteMetaData()` which in tern creates and executes a statement and returns a result set. `#close()` on this result set and statement is never called because the caller of `#getMetaData()` only sees a `ResultSetMetaData` object which has no `#close()` method.

The following code can be used to reproduce the issue

```java
try (Connection con = DriverManager.getConnection(connectionString);
     PreparedStatement stmt = con.prepareStatement("SELECT 1")) {
    ResultSetMetaData metaData = stmt.getMetaData();
    assertEquals(1, metaData.getColumnCount());
}
```

peterbae:
Fixed in #455

PR #455, opened 2017-08-18
Fixes all statement leaks in the driver.

peterbae:
Close all statements in the driver that are initialized locally but are never / sometimes not closed.

Fixes issue #308.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
@@ -244,12 +244,12 @@ public final class SQLServerDatabaseMetaData implements java.sql.DatabaseMetaDat
         SQLServerResultSet rs = null;
         try {
             rs = ((SQLServerStatement) connection.createStatement()).executeQueryInternal(query);
-
         }
         finally {
-            if (null != orgCat) {
+            if (null != rs)
+                rs.close();
+            if (null != orgCat)
                 connection.setCatalog(orgCat);
-            }
         }
         return rs;
     }
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
@@ -246,8 +246,9 @@ public final class SQLServerDatabaseMetaData implements java.sql.DatabaseMetaDat
             rs = ((SQLServerStatement) connection.createStatement()).executeQueryInternal(query);
         }
         finally {
-            if (null != orgCat)
+            if (null != orgCat) {
                 connection.setCatalog(orgCat);
+            }
         }
         return rs;
     }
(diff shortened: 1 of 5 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
```diff
@@ -244,12 +244,12 @@ private SQLServerResultSet getResultSetFromInternalQueries(String catalog,
         SQLServerResultSet rs = null;
         try {
             rs = ((SQLServerStatement) connection.createStatement()).executeQueryInternal(query);
-
         }
         finally {
-            if (null != orgCat) {
+            if (null != rs)
+                rs.close();
```
v-xiangs: this method returns the resultset, so I guess we should not close this resultset here

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -1035,6 +1036,10 @@ else if (resultSet != null) {
                 SQLServerException.makeFromDriverError(connection, this, form.format(msgArgs), null, true);
             }
         }
+        finally {
+            if (null != stmt)
+                stmt.close();
```
v-xiangs: this statement creates resultset which will be returned, closing this statement also close the returned resultset, which is not what we want I guess

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
```diff
@@ -244,12 +244,10 @@ private SQLServerResultSet getResultSetFromInternalQueries(String catalog,
         SQLServerResultSet rs = null;
         try {
             rs = ((SQLServerStatement) connection.createStatement()).executeQueryInternal(query);
-
         }
         finally {
-            if (null != orgCat) {
+            if (null != orgCat)
```
peterbae: I don't think we ever put brackets when we could omit them. so i omit them here as well

peterbae: I think we always omit the bracket when we can, so i did the same here.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerParameterMetaData.java
```diff
@@ -563,6 +563,8 @@ private void checkClosed() throws SQLServerException {
         assert null != st;
         stmtParent = st;
         con = st.connection;
+        SQLServerStatement s = null;
```
peterbae: i guess s doesnt need to be instantiated to null but i did it for consistency

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -98,6 +98,9 @@
 
     /** The prepared statement handle returned by the server */
     private int prepStmtHandle = 0;
+    
+    /** Statement used for getMetadata(). Declared as a field to facilitate closing the statement. */
+    private SQLServerStatement internalStmt = null;
```
peterbae: having internalStmt as a field is the only way to close it after the user is done working with the ResultsetMetadata.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
v-afrafi: actually we started adding back the brackets even with one line of code, so lets keep the brackets here  😄

peterbae: done.

Issue #453, opened 2017-08-18
Error message errors


v-xiangs:
## Driver version or jar name
all versions

## SQL Server version
N/A

## Client operating system
N/A

## Java/JVM version
N/A

## Table schema
N/A

## Problem description
Some error messages inside SQLServerResource.java file have grammatical/spelling issues. Please see @ulvii 's comments in PR #452 . We decide to merge #452 first and create an issue to remind us the rest.

Please note, changing error messages may break client programs if those error messages are public to the client.

## Expected behavior and actual behavior
N/A

## Repro code
N/A

ulvii:
Fixed by #459

PR #459, opened 2017-08-24
Fixing a few mistakes in SQLServerResource

ulvii:
#453 

Please let me know if there are any more issues in the file.


src/test/java/com/microsoft/sqlserver/jdbc/datatypes/TVPWithSqlVariantTest.java
```diff
@@ -340,7 +340,7 @@ public void testNull() throws SQLServerException {
             tvp.addRow((Date) null);
         }
         catch (Exception e) {
-            assertTrue(e.getMessage().startsWith("Inserting null value with column"));
+            assertTrue(e.getMessage().startsWith("Use of TVPs containing null sql_variant columns is not supported."));
         }
```
cheenamalhotra: I would again recommend using same SQLServerResource > getResource() to compare error messages, so we don't have to change Strings at every place it is referenced. Would make sense to make a separate PR for this.

Issue #438, opened 2017-08-12
FIPS Compliant Mode


ulvii:
We recently (version 6.1.2) introduced 3 new connection properties for [FIPS](https://github.com/Microsoft/mssql-jdbc/wiki/FIPS) mode: `fips`, `trustStoreType`, `fipsProvider`. The implementation assumes `fipsProvider` contains `trustStoreType` [when calling ](https://github.com/Microsoft/mssql-jdbc/blob/97ab207fffaf23c967ecbe775423f6262dab9eaf/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java#L1651)`Keystore.getInstance(trustStoreType,fipsProvider)`.  However, not all providers that contain FIPS-complaint algorithms also provide keystores, such as `IBMJCEFIPS`. [See this page](https://www.ibm.com/support/knowledgecenter/en/SSYKE2_7.0.0/com.ibm.java.security.component.70.doc/security-component/progpractices.html). In this case, users have to use non-FIPS IBMJCE provider (`IBMJCE`) to store the keys, which means they also have to set `fipsProvider=IBMJCE`. This can be confusing, because `IBMJCE` is not the actual FIPS provider, but it is the provider that contains keystore. 

Below are a few possible solutions to solve the issue and  we would really appreciate any feedback from the community. 

1. Keep the current implementation and update the documentation describing the differences. This way we would not be introducing breaking changes, but the connection property `fipsProvider`  could still be confusing to users because of the naming. 
2. Rename the connection property `fipsProvider` to `trustStoreProvider`. This would break the existing FIPS applications, but the name of the property would be less confusing. 
3. Remove the connection property `fipsProvider`. If provider is not specified, `Keystore.getInstance(trustStoreType)` traverses the list of registered providers and creates a new instance of matching Keystore. So, instead of creating a new Keystore from the specified provider, the driver would  create it from the most preferred (the one that comes first in security providers list) provider. This would obviously be a breaking change too, but also would let us avoid any possible confusions. 

I am also posting links to the related PRs below. 

https://github.com/Microsoft/mssql-jdbc/pull/97
https://github.com/Microsoft/mssql-jdbc/pull/135
https://github.com/Microsoft/mssql-jdbc/pull/325
https://github.com/Microsoft/mssql-jdbc/pull/433

We would appreciate opinions on the proposed solutions. Please feel free to let us know if you have other suggestions 👍

nsidhaye:
[...]

PR #460, opened 2017-08-25
Removing connection property - fipsProvider

ulvii:
Based on the discussions on #433 and #438.


src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -1828,57 +1822,40 @@ void enableSSL(String host,
...
-                if (!isValidFipsProvider || !isValidTrustStoreType) {
-                    isValid = false;
-                    strError = SQLServerException.getErrString("R_invalidFipsProviderConfig");
-                    
+                // In case of valid trust store we need to check TrustStoreType.
+                if (!isValidTrustStoreType) {
+                    isValid = false;               
                     if (logger.isLoggable(Level.FINER))
-                        logger.finer(toString() + " FIPS provider & TrustStoreType should pass with TrustStore.");
+                        logger.finer(toString() + "TrustStoreType is requiered alongside with TrustStore.");
```
sehrope: s/requiered/required/

src/test/java/com/microsoft/sqlserver/jdbc/fips/FipsTest.java
```diff
@@ -72,32 +71,12 @@ public void fipsEncryptTest() throws Exception {
         }
         catch (SQLServerException e) {
             Assertions.assertTrue(
-                    e.getMessage().contains("Could not enable FIPS due to either encrypt is not true or using trusted certificate settings."),
+                    e.getMessage().contains("Unable to verify FIPS mode settings."),
```
cheenamalhotra: I would suggest getting these same String error messages from SQLServerResource.java in test cases, such that your code becomes: `e.getMessage().contains(SQLServerResource.getResource("R_invalidFipsConfig")). `
It would remove all possibilities of errors and add maintainability.

src/test/java/com/microsoft/sqlserver/jdbc/fips/FipsTest.java
```diff
@@ -52,8 +52,7 @@ public void fipsTrustServerCertificateTest() throws Exception {
         }
         catch (SQLServerException e) {
             Assertions.assertTrue(
-                    e.getMessage().contains("Could not enable FIPS due to either encrypt is not true or using trusted certificate settings."),
-                    "Should create exception for invalid TrustServerCertificate value");
+                    e.getMessage().contains("Unable to verify FIPS mode settings."));
```
cheenamalhotra: Any specific reason for removing `"Should create exception for invalid TrustServerCertificate value"` from line 56?

src/test/java/com/microsoft/sqlserver/jdbc/fips/FipsTest.java
ulvii: `getResource()` is not a public API

cheenamalhotra: Yeah.. I missed that, in that case we need a custom resource bundle for our test framework. I have created issue #466 to track this change.

PR #465, opened 2017-08-31
Classloader leak 314

peterbae:
Fixes issue #314 

I put a comment as to why I took this approach for solving this problem in that issue as well.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/ActivityCorrelator.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/ActivityCorrelator.java
@@ -8,6 +8,7 @@
 
 package com.microsoft.sqlserver.jdbc;
 
+import java.util.HashMap;
 import java.util.UUID;
 
 /**
@@ -15,21 +16,32 @@ import java.util.UUID;
  */
 final class ActivityCorrelator {
 
-    private static ThreadLocal<ActivityId> ActivityIdTls = new ThreadLocal<ActivityId>() {
-        protected ActivityId initialValue() {
-            return new ActivityId();
+    private static HashMap<Long, ActivityId> ActivityIdTlsMap = new HashMap<Long, ActivityId>();
+    
+    static void checkAndInitActivityId() {
+        long uniqueThreadId = Thread.currentThread().getId();
+        
+        //Since the Id for each thread is unique, this assures that the below code is run only once per *thread*,
+        //the first time this constructor is called.
+        if (!ActivityIdTlsMap.containsKey(uniqueThreadId)) {
+            ActivityIdTlsMap.put(uniqueThreadId, new ActivityId());
         }
-    };
+    }
 
     // Get the current ActivityId in TLS
     static ActivityId getCurrent() {
+        checkAndInitActivityId();
+        
         // get the value in TLS, not reference
-        return ActivityIdTls.get();
+        long uniqueThreadId = Thread.currentThread().getId();
+        
+        return ActivityIdTlsMap.get(uniqueThreadId);
     }
 
     // Increment the Sequence number of the ActivityId in TLS
     // and return the ActivityId with new Sequence number
     static ActivityId getNext() {
+        checkAndInitActivityId();
         // We need to call get() method on ThreadLocal to get
         // the current value of ActivityId stored in TLS,
         // then increment the sequence number.
@@ -44,10 +56,11 @@ final class ActivityCorrelator {
     }
 
     static void setCurrentActivityIdSentFlag() {
+        checkAndInitActivityId();
+        
         ActivityId activityId = getCurrent();
         activityId.setSentFlag();
     }
-
 }
 
 class ActivityId {
--- a/src/main/java/com/microsoft/sqlserver/jdbc/ActivityCorrelator.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/ActivityCorrelator.java
@@ -21,8 +21,7 @@ final class ActivityCorrelator {
     static void checkAndInitActivityId() {
         long uniqueThreadId = Thread.currentThread().getId();
         
-        //Since the Id for each thread is unique, this assures that the below code is run only once per *thread*,
-        //the first time this constructor is called.
+        //Since the Id for each thread is unique, this assures that the below code is run only once per *thread*.
         if (!ActivityIdTlsMap.containsKey(uniqueThreadId)) {
             ActivityIdTlsMap.put(uniqueThreadId, new ActivityId());
         }


marschall:
Don't you need proper synchronization when accessing a shared object (the map)?
I can't find where the values are removed from the map again. If that doesn't happen, then you end up with a memory leak.

peterbae:
Hi @marschall,

I thought about making the map synchronized - it's true that the map is shared by multiple threads, but since each entry in the map is only going to be modified by one thread (the thread can only access the entry in the map that shares the same key as its thread ID), I do not believe any two threads will ever try to access/modify the same entry in the map, and therefore this operation should be thread safe.

As for the second point, I agree. I was thinking the map would get removed when the app server gets undeployed, but it may never get undeployed, in which case the map would keep holding entries that will never be used again. I've made changes so that a thread will remove its own entry in the map when its connection gets closed (which should be thread safe for the same reason above).

marschall:
> I thought about making the map synchronized - it's true that the map is shared by multiple threads, but since each entry in the map is only going to be modified by one thread (the thread can only access the entry in the map that shares the same key as its thread ID), I do not believe any two threads will ever try to access/modify the same entry in the map, and therefore this operation should be thread safe.

That's not how thread safety in Java works at all. And it's not true since the same entry is modified in case of a hash collision. In addition the size, the modification count, the backing array and the pointer to the backing array are modified by all thread.

For a quick introduction into the JMM I recommend the following articles

https://shipilev.net/blog/2014/safe-public-construction/
https://shipilev.net/blog/2016/close-encounters-of-jmm-kind/

> As for the second point, I agree. I was thinking the map would get removed when the app server gets undeployed, but it may never get undeployed, in which case the map would keep holding entries that will never be used again. I've made changes so that a thread will remove its own entry in the map when its connection gets closed (which should be thread safe for the same reason above).

Not in an application server with a connection pool.

peterbae:
Thanks @marschall  for your insight. It was an oversight on my end to disregard the hash collision and misusing HashMap here. I've replaced the HashMap with a ConcurrentHashMap.

For the issue where the static map that holds ActivityId may not be removing the unused entries due to the scenario where the threads that are using the driver object and closing the connection being different, I believe we need a way to track the threads that are alive and periodically remove entries from the map. However, this will likely introduce performance degradation and will need further input from the team, and consider if the potential memory being held by the map is worth the performance gain. Meanwhile, since I believe the classloader leak issue itself is fixed, I think it's still good to merge this in unless more problems are found with this solution.

src/main/java/com/microsoft/sqlserver/jdbc/ActivityCorrelator.java
```diff
@@ -44,10 +56,11 @@ static ActivityId getNext() {
     }
 
     static void setCurrentActivityIdSentFlag() {
+        checkAndInitActivityId();
```
rPraml: This call is obsolete, as "getCurrent" calls "checkAndInitActivityId"

src/main/java/com/microsoft/sqlserver/jdbc/ActivityCorrelator.java
```diff
@@ -8,28 +8,40 @@
...
-        return ActivityIdTls.get();
+        long uniqueThreadId = Thread.currentThread().getId();
+        
+        return ActivityIdTlsMap.get(uniqueThreadId);
     }
 
     // Increment the Sequence number of the ActivityId in TLS
     // and return the ActivityId with new Sequence number
     static ActivityId getNext() {
+        checkAndInitActivityId();
```
rPraml: This call is obsolete

src/main/java/com/microsoft/sqlserver/jdbc/ActivityCorrelator.java
```diff
@@ -8,28 +8,40 @@
...
+        //the first time this constructor is called.
+        if (!ActivityIdTlsMap.containsKey(uniqueThreadId)) {
+            ActivityIdTlsMap.put(uniqueThreadId, new ActivityId());
         }
-    };
+    }
 
     // Get the current ActivityId in TLS
     static ActivityId getCurrent() {
+        checkAndInitActivityId();
```
rPraml: So there's only one call for checkAndInitActivityId - you can refactor this method:
```java
static ActivityId getCurrent() {
  long uniqueThreadId = Thread.currentThread().getId();
  ActivityId id = ActivityIdTlsMap.get(uniqueThreadId);
  if (id == null) {
    id = new ActivityId();
    ActivityIdTlsMap.put(uniqueThreadId, id);
  } 
  return id;
}
```
This would save some CPU-cycles

peterbae:
Thanks @rPraml, I've made the refactoring changes and also made some commenting changes.

rPraml:
@peterbae I do not see any major problems in your fix, It's true that you have a memory leak when you start & stop a lot of threads AND Util.IsActivityTraceOn() = true. But I think this behaviour is better than the classloading issue.

I also tried to find out what ActivityId is good for.
It seems to me, that ActivitiId & ActivitiCorrelator is used for logging purpose and ONLY used if "com.microsoft.sqlserver.jdbc.traceactivity" is set to "on".

I also found often this code block:
```java
if (loggerExternal.isLoggable(Level.FINER) && Util.IsActivityTraceOn()) {
  loggerExternal.finer(toString() + " ActivityId: " + ActivityCorrelator.getNext().toString());
}
```
If I interpret this right, the logger statement calls ActivityCorrelator.getNext() - which will increment the ID. This means, the ActivityId counts different if log level is FINER or not. This looks like a bug/mistake...
(Maybe ,getCurrent() instead of .getNext() is correct here)

Maybe some other people from DEV can take a look at the ActivityCorrelator. If this is really only for debugging (https://msdn.microsoft.com/en-us/library/hh269638.aspx) maybe you can use a static AtomicInteger and remove some complexity?

peterbae:
@rPraml We call ActivityCorrelator.getNext so that we increment the sequence and then log it. If we were to call getCurrent instead here, we would always be logging the same sequence since we're never incrementing it (it also seems like we're only concerned with logging when the log level is FINER), so getNext is fine here.

As for the ActivityCorrelator, it's also used in other parts of the driver to store if a message was sent to the server (in TDSWriter, for example), so it's not only used for debugging.

rPraml:
@peterbae thanks for clarifying.
It really seems that incrementing the ID is only important when logging is enabled as there is only one place in `SQLServerConnection.Prelogin` where .getNext() is called.

peterbae:
@rPraml yes. As discussed above, the issue with this solution is that the ActivityId objects (which are relatively small) are potentially being held in the static map in ActivityCorrelator in an application server with a connection pool, but in order to clean up the ActivityId objects in such scenario, the driver would see some performance degradation. I would still like to merge this fix as an improvement.

peterbae:
@marschall @rPraml please let me know if there's other issues with this fix, otherwise I will merge it in. Thanks.

PR #474, opened 2017-09-01
Query with timer can throw IllegalThreadStateException

maiken2051:
Running a query that uses a timer can cause an IllegalThreadStateException if the underlying ThreadGroup has been destroyed.

This error happens when the caller executes a query with a timeout and then doesn't execute another query with a timeout until some time has elapsed. The JVM can destroy the ThreadGroup behind the TimeoutTimer (because it contains no active threads) which causes any new query with a timeout to throw an IllegalThreadStateException.

Once the JDBC driver has entered this state any new query with a timeout will throw the exception. It's especially easy to run into this problem in demanding / multi-threaded applications.

My change makes sure the ThreadGroup is not destroyed before creating a new timer thread. If the ThreadGroup is destroyed, the code will create a new ThreadGroup.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
@@ -64,6 +64,7 @@ import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
--- a/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
@@ -7122,13 +7122,21 @@ final class TimeoutTimer implements Runnable {
     private volatile Future<?> task;
 
     private static final ExecutorService executor = Executors.newCachedThreadPool(new ThreadFactory() {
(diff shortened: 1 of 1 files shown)


v-afrafi:
@maiken2051 Thank you for your effort in creating this pr. Could you please open this to dev branch?

ajlam:
Hi @maiken2051, would you be able to reopen this PR against the dev branch?

maiken2051:
Thanks for rebasing the PR...

ajlam:
@maiken2051, just wanted to keep you posted. We are going to be doing a code review soon. Thanks again for your contributions.

cheenamalhotra:
Hi @maiken2051 Please sign CLA so that we can proceed with merging your PR.

maiken2051:
I think the CLA should be OK now?

peterbae:
Hi @maiken2051, thanks for the contribution. The changes look good, and I approved it - but I'm curious as to where I can find the information regarding JVM destroying a ThreadGroup that contains no active threads. Could you point me to where I can read more about this?

maiken2051:
Check out the source for [ThreadGroup.threadTerminated](http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/8u40-b25/java/lang/ThreadGroup.java#ThreadGroup.threadTerminated%28java.lang.Thread%29). This is from OpenJDK but I assume all JVM implementations are similar. The docs say:

> Notifies the group that the thread t has terminated. Destroy the group if all of the following conditions are true: this is a daemon thread group; there are no more alive or unstarted threads in the group; there are no subgroups in this thread group.

When a thread exits it will call this method to remove itself from the active list on its owner ThreadGroup. If it's the last active thread the group will be destroyed.

PR #485, opened 2017-09-11
removing javax.xml.bind package dependency 

v-afrafi:


src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -4934,7 +4933,7 @@ else if (DataTypes.UNKNOWN_STREAM_LENGTH == dataLength)
                 isShortValue = columnPair.getValue().precision <= DataTypes.SHORT_VARTYPE_MAX_BYTES;
                 isNull = (null == currentObject);
                 if (currentObject instanceof String)
-                    dataLength = isNull ? 0 : (toByteArray(currentObject.toString())).length;
+                    dataLength = isNull ? 0 : (ParameterUtils.HexToBin(currentObject.toString())).length;
```
v-xiangs: for this `ParameterUtils` class, are we adding new dependency to the driver?

v-afrafi: No, this is an existing class.

v-xiangs: oh never mind, it's existing code in the driver

PR #489, opened 2017-09-12
Fixes #486 : Keep unknown tokens when normalising.

### LINKED ISSUE #486: [Feature] Keep unknown property when getting url from database metadata
I'm writing features for jdbc distributed tracing. In particular, the ability to define the remote service name inside the connection string that will be parsed.
Unfortunately, unknown tokens are removed  from the final properties map.
```java
final Connection connection = DriverManager.getConnection("sqlserver://localhost;databaseName=master;foo=bar", "user", "password");
connection.getMetaData().getURL().contains("foo=bar"); //returns false
```
Indeed, in [SQLServerDriver.java#L576](https://github.com/Microsoft/mssql-jdbc/blob/19af32bb561685c681c364fbd4d2d02efdfd352b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDriver.java#L576), null is returned for unknown tokens.

A quick fix would be to return the name instead of null but it might break some other logic in the code (even if the test passes locally on my machine)

What do you think of such feature. Does it make sense ?

Thanks
(PS: it works with other databases such as MySQL)

hypnoce:


hypnoce:
This is what I could think of while reading the code. Not sure if that's what you had in mind. Thanks for your help !

hypnoce:
CLA signed !

PR #495, opened 2017-09-20
recognize CallableStatement parameter names with leading '@'

gordthompson:
When passing parameters to a CallableStatement by name, recognize names that start with '@' as reported by `DatabaseMetaData#getProcedureColumns`.

(ref: Stack Overflow question [here](https://stackoverflow.com/q/46294187/2144390))

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCallableStatement.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCallableStatement.java
@@ -1453,8 +1453,11 @@ public class SQLServerCallableStatement extends SQLServerPreparedStatement imple
 
         // handle `@name` as well as `name`, since `@name` is what's returned 
         // by DatabaseMetaData#getProcedureColumns
+        String columnNameWithoutAtSign = null;
         if (columnName.startsWith("@")) {
-            columnName = columnName.substring(1, columnName.length());
+            columnNameWithoutAtSign = columnName.substring(1, columnName.length());
+        } else {
+            columnNameWithoutAtSign = columnName;
         }
         
         // In order to be as accurate as possible when locating parameter name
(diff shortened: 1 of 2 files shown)


gordthompson:
Addresses issue #496

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCallableStatement.java
```diff
@@ -1451,6 +1451,12 @@ public NClob getNClob(String parameterName) throws SQLException {
         if (paramNames != null)
             l = paramNames.size();
 
+        // handle `@name` as well as `name`, since `@name` is what's returned 
+        // by DatabaseMetaData#getProcedureColumns
+        if (columnName.startsWith("@")) {
```
v-afrafi: we can add the check at this [line ](https://github.com/gordthompson/mssql-jdbc/blob/6789eb6fc86fe5b31a1ffd769f87065a616de2c2/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCallableStatement.java#L1472)
```
if (!columnName.startsWith("@"))
                sParam = sParam.substring(1, sParam.length());
```

The reason I am proposing this is that to have the original column name when we throw the exception at this [line ](https://github.com/gordthompson/mssql-jdbc/blob/6789eb6fc86fe5b31a1ffd769f87065a616de2c2/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCallableStatement.java#L1495) Otherwise looks great! thanks

gordthompson: @v-afrafi - Good point. Your change would actually have to be made in two places, so I tweaked it a bit.

v-afrafi:
Thanks @gordthompson for fixing this issue 🎉

PR #500, opened 2017-09-22
Replace manual array copy

JamieMagee:
Replace manual array copying with System.arraycopy().


JamieMagee:
@peterbae These changes were detected by FIndBugs as array copying that could be directly replaced by `System.arraycopy()`. Any other array copy methods contain more advanced logic, and can't be directly replaced by `System.arraycopy()`.

I'm not suggesting changing any of the others, but just these 3 simple instances.

PR #501, opened 2017-09-22
Remove redundant String.toString()

JamieMagee:
Calling toString() on a String object is redundant


v-afrafi:
Thank you @JamieMagee for your effort in creating this pr!

PR #503, opened 2017-09-22
Remove final modifier from static methods

JamieMagee:
When a static method is overridden in a subclass it can still be accessed via the super class, making a final declaration not very necessary. Declaring a static method final does prevent subclasses from defining a static method with the same signature.


v-xiangs:
@JamieMagee Hello. The purpose of those final keywords is to preventing method overriding, not preventing from accessing them via super classes. Although some classes are already declared as final (some classes are not final, i.g. `PLPInputStream`, so keeping the final keyword is very necessary in this case), we have decided to keep the final keywords on those methods declaration too. Thank you.

PR #509, opened 2017-09-25
Cleaning up tables

v-afrafi:


src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/CallableMixedTest.java
```diff
@@ -31,79 +32,71 @@
...
-        callableStatement.close();
-        rs.close();
-        stmt.close();
+                assertEquals(rs.getInt(1), 0, "Received data not equal to setdata");
+                assertEquals(callableStatement.getInt((int) 1), -2147483648, "Received data not equal to setdata");
+                assertEquals(callableStatement.getInt((int) 5), -5372, "Received data not equal to setdata");
+                rs = callableStatement.executeQuery();
+                rs.close();
+            }
+        }
```
cheenamalhotra: Before closing the try, terminate variation can be passed existing Connection & Statement objects instead of starting a new connection for dropping objects.

Issue #497, opened 2017-09-21
Creating a SQLServerDataTable is O(n^2)


alexnixon:
This is true as of the latest commit on the dev branch (e6f66e9).

The API of `SQLServerDataTable` prevents it from being reused between calls - the only way to clear it is by calling `clear()`, which throws away column metadata in addition to row data.

So you have to construct it afresh each time, and the only way to do so is by repeatedly calling `addColumnMetaData`. But each invocation calls `Util.checkDuplicateColumnName`, which iterates over all added columns so far, resulting in the `SQLServerDataTable` construction being O(n^2).

Possible fixes are:
1) Store a `Set` of column names in `SQLServerDataTable` and check against that when adding columns, so the n^2 behavior goes away
2) Provide a method to clear the rows in a `SQLServerDataTable`, so it's possible to write code in a way that incurs this cost less often.

Personally I wouldn't be averse to both (1) and (2). Thoughts?

Thanks,
Alex

(I haven't benchmarked this to prove it's an issue for me - it's probably not - but I thought this behavior surprising so worth noting).

ajlam:
Thanks for filing this issue @alexnixon.

peterbae:
Hi @alexnixon, I created https://github.com/Microsoft/mssql-jdbc/pull/514 for this issue (implements your first suggestion). Could you take a look?

peterbae:
Fixed in #514. Closing the issue.

PR #514, opened 2017-09-27
Fix creating SQLServerDataTable being O(n^2) issue

peterbae:
Addresses issue https://github.com/Microsoft/mssql-jdbc/issues/497.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDataTable.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDataTable.java
@@ -79,7 +79,7 @@ public final class SQLServerDataTable {
     public synchronized void addColumnMetadata(String columnName,
             int sqlType) throws SQLServerException {
         // column names must be unique
-        checkDuplicateColumnName(columnName);
+        Util.checkDuplicateColumnName(columnName, columnNames);
         columnMetadata.put(columnCount++, new SQLServerDataColumn(columnName, sqlType));
     }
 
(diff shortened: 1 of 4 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDataTable.java
```diff
@@ -13,17 +13,20 @@
...
 import java.util.Map.Entry;
+import java.util.Set;
 import java.util.UUID;
 
 public final class SQLServerDataTable {
 
     int rowCount = 0;
     int columnCount = 0;
     Map<Integer, SQLServerDataColumn> columnMetadata = null;
+    Set<String> columnList = null;
```
alexnixon: I think naming this `columNames` would be clearer

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDataTable.java
```diff
@@ -89,11 +93,30 @@ public synchronized void addColumnMetadata(String columnName,
...
     /**
+     * Checks if duplicate columns exists, in O(n) time.
+     * 
+     * @param columnName
+     *            the name of the column
+     * @throws SQLServerException
+     *             when a duplicate column exists
+     */
+    private void checkDuplicateColumnName(String columnName) throws SQLServerException {
+        if (null != columnList) {
```
alexnixon: I'd rather an NPE than silently fail to check dupes if someone messes up null handling.

Better IMO would be defining the local variable like `Set<String> columnNames = new HashSet<>();` and omitting this check, but I don't know if that'd violate the project's coding style.

peterbae:
@alexnixon Thanks for the review. I added the same logic for TVP since it was also applicable there, and made minor changes according to your suggestions.

v-afrafi:
let's also add a test for checking duplicate column

Issue #483, opened 2017-09-08
Throw exception on first connection instead of retry for incorrect password


ajlam:
Filing on behalf of a comment left on [JDBC docs](https://docs.microsoft.com/en-us/sql/connect/jdbc/microsoft-jdbc-driver-for-sql-server) 

## Driver version or jar name
JDBC Driver 3.0

## SQL Server version
N/A

## Client operating system
N/A

## Java/JVM version
N/A

## Table schema
N/A

## Problem description
The issue is that when the SQLServerConnection class fails to make a connection on an incorrect password, we would require the SQL Server driver to throw an exception and abort the connection attempt instead of a looping retry on the same credentials. This will allow the Hikari pool to create a new connection with a new password.

Is there a way that we can SQL Server driver to throw an exception on the first connection exception instead of retrying for the duration of the set timeout? This would allow the connection pool to manage retrying connections. We were unable to find any settings for the jdbc connection string to achieve this.

We checked the latest version but there isn't a solution to this problem yet. https://github.com/Microsoft/mssql-jdbc/blob/master/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java

MightyPen:
The customer brought this concern to Andrea and I, after the customer read the following documentation article:
https://docs.microsoft.com/sql/connect/jdbc/microsoft-jdbc-driver-for-sql-server

cheenamalhotra:
Hi Sandeep Chinnari,

I tried to replicate your issue while trying to attempt connecting to SQL Server with incorrect password. I tested with driver versions 3.0 and latest 6.3.2 preview version, I was not able to replicate the issue. Driver does not seem to re-attempt establishing connection, if wrong credentials are provided. Neither does it make retry attempts, nor does it wait till timeout occurs to throw the Exception. 

I would like to help you resolve this issue, if you can provide us repro code which can be helpful to reproduce the issue on our end. Also provide details on SQL Server version, connection properties, application configurations (e.g. connection pooling) which may be helpful to trace the issue.

sandeepchinnari:
SQL Server versions tried - SQL 2012 Enterprise edition ,SQL 2008 R2 Enterprise edition
Client operating system - Linux 2.6.32 
Java/JVM version - Java 8

sandeepchinnari:
Our app is a spring boot app running on Tomcat and using Hikari CP.

ajlam:
[...]

PR #522, opened 2017-10-06
Throw exception for "Account Locked" SQL Exception 18486

cheenamalhotra:
This PR fixes customer issue #483 by throwing exception if Account locked Exception 18486 found.


sandeepchinnari:
@cheenamalhotra  Thanks for the fix. But this will come in handy when its already too late.
What happens when the password change is in progress  and concurrent threads attempt to make connections?

cheenamalhotra:
@sandeepchinnari It would be interesting to see how one can reproduce this usecase. 

As far as I understand, while request is made to Server to change password, old password is still valid. And password change is an atomic operation to update `sys.sql_logins ` table of database with new password. After which, any new attempt will fail with an error code of **18456 (Logon Failed).**

I don't see a reason, why there would be a recurrence of attempts here coz we are talking nanoseconds here. Even if there is a race condition, as soon as new password is updated, the attempt to connect will fail.

cheenamalhotra:
Hi @sandeepchinnari ! Would appreciate your confirmation, if this PR fixes your failed use-case. We would like to merge the PR soon.

PR #534, opened 2017-10-24
Updating survey link to point to new survey

David-Engel:
New sentiment survery

https://www.surveybuilder.com/s/kmdwV


David-Engel:
Modified the forwarding link instead.

Issue #186, opened 2017-03-10
Mutable static in SQLServerClobBase


marschall:
The static field `com.microsoft.sqlserver.jdbc.SQLServerClobBase#logger` is mutable and set by every constructor invocation. It is only used in `com.microsoft.sqlserver.jdbc.SQLServerClobBase#free()`. As a result with logger is used there is pretty much random. My recommendation would be to make it `static final`.

cheenamalhotra:
Hi @marschall The logger in SQLServerClobBase (abstract class) is present as an object holder for passed logger object while creating a new object of the class that extends it, SQLServerClob or SQLServerNClob. 

I agree with your point of finalizing logger objects to stop mutation, but in this case, the intention is to log messages as the extended class, and not as per the abstract class and that's the reason, it's passed by constructor.

I think the current implementation is intentional and need not be changed since this is an exceptional case, and not a regular logger scenario.

marschall:
My point is not so much about finalizing but about semantics. There is only one logger in for all instances of `SQLServerClobBase` and subclasses. Whoever created the last instance overwrites the logger of everybody else. It does not matter what you pass as an argument to the super constructor, the next call to the super constructor will overwrite it.

I do not agree with your statement that this is an exceptional case since `SQLServerClob` and `SQLServerNClob` always pass a logger. They both pass in a different logger. So it is very well possible that you get a `SQLServerNClob` logger in a `SQLServerClob` instance and vice versa.

cheenamalhotra:
I understand your point now.

The reason behind this behavior is basically because logger is static member in `SQLServerClobBase`. Removing `static` keyword will resolve this issue and result in maintaining logger for every object as instantiated. On the other hand, making it `static final` would result in logging messages with `SQLServerClobBase` abstract class and not subclasses `SQLServerClob` or `SQLServerNClob`.

I am creating a PR  which makes logger `final` in abstract class `SQLServerClobBase` and not `static` such that it is instantiated based on subclass name (using getClass().getName()). Therefore, logger is removed from subclasses since it was used for only passing through constructor. 

Let me know if the PR resolves your query.

cheenamalhotra:
Hi @marschall the fix is merged and released in v6.3.5 - please review it and let us know if it works for you. Closing the issue for now.

PR #537, opened 2017-11-01
Fix for static logger member in abstract class 'SQLServerClobBase'

cheenamalhotra:
Fixes issue #186.

This change makes logger in abstract class 'SQLServerClobBase' non-static but final such that it is instantiated for every new object, based on subclass been used to create object and is immutable.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerClob.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerClob.java
@@ -35,9 +35,6 @@ import java.util.logging.Logger;
 public class SQLServerClob extends SQLServerClobBase implements Clob {
     private static final long serialVersionUID = 2872035282200133865L;
 
-    // Loggers should be class static to avoid lock contention with multiple threads
-    private static final Logger logger = Logger.getLogger("com.microsoft.sqlserver.jdbc.internals.SQLServerClob");
-
     /**
      * Create a new CLOB
      *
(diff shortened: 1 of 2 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerClob.java
```diff
@@ -91,7 +88,9 @@ final JDBCType getJdbcType() {
     private ArrayList<Closeable> activeStreams = new ArrayList<>(1);
 
     transient SQLServerConnection con;
-    private static Logger logger;
+    
+    private final Logger logger = Logger.getLogger(getClass().getName());
```
ulvii: How about this comment that got deleted?
`// Loggers should be class static to avoid lock contention with multiple threads`

cheenamalhotra: Well - its because this logger object won't be static in this case.

marschall: There is now a logger lookup whenever an instance of `SQLServerClob` or `SQLServerNClob` is created. This could be avoided if:
1. there is a `static final` logger in `SQLServerClob` or `SQLServerNClob`, like before
1. the logger in `SQLServerClobBase` is made `private final`, like in this PR
1. `SQLServerClob` and `SQLServerNClob` pass the logger in the `super` constructor call

cheenamalhotra: Hi @marschall 

Thanks for replying back! We could go for either solution, but I removed the loggers from sub classes as: 
1. Class Lookup isn't expensive here.
2. The loggers in sub classes are not used anywhere else except being passed in constructor (but remain attached in memory with class objects)

marschall: It's not the class lookup that I was concerned about, it's get logger lookup.

cheenamalhotra: Can you please explain what you mean by Logger lookup and how will it be different in the other case - maybe I am not able to understand your point.

marschall: With a non-static logger in `SQLServerClob` or `SQLServerNClob` every time a new instance of these classes is generated the following is executed:
```java
Logger.getLogger(getClass().getName())
```
If we ignore `getClass().getName()` for now that still leaves us with `Logger.getLogger()` which does among other things:
1. `Reflection.getCallerClass()` which creates an exception and walks the entire stack
1. check for a security manager several times
1. goes through a global lock in `java.util.logging.LogManager.LoggerContext#namedLoggers`

If you make the loggers in `SQLServerClob` and `SQLServerNClob` `static` all that work happens only one during class loading.

cheenamalhotra: Hi @marschall I understand the concern for non-static logger here. Made changes in the new PR #563 as discussed above. Request you to review the PR so we can get it working.

PR #538, opened 2017-11-01
Code improvements in SQLServerDataTable > 'internalAddRow()'

cheenamalhotra:
Replaced with #990


cogman:
Pretty much exactly what I was thinking for the linked issue.

cheenamalhotra:
Fix added for #539 - Support for Bit values (1,0) added for Boolean datatype in TVP.

alexnixon:
I think this introduces regressions, see my comment on https://github.com/Microsoft/mssql-jdbc/issues/539

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDataTable.java
```diff
@@ -158,20 +158,20 @@ private void internalAddrow(JDBCType jdbcType,
 	        int nValueLen;
 	        switch (jdbcType) {
 	            case BIGINT:
-	                rowValues[pair.getKey()] = (val instanceof Number ? ((Number)val).longValue() : Long.parseLong(val.toString()));
+	                rowValues[pair.getKey()] = (val instanceof Number ? (Long)val : Long.parseLong(val.toString()));
```
alexnixon: If `val` is an `Integer` this will throw a `ClassCastException`. Same for most of the `instanceof` checks below

alexnixon:
I think this would benefit from some more tests, perhaps similar to the examples I gave on #539.

As for the implementation, something like this is what I had in mind:

```
    private static long coerceToLong(Object val, String fieldTypeName) {
        if (val instanceof Byte) {
            return ((Byte) val).longValue();
        } else if (val instanceof Short) {
            return ((Short)val).longValue();
        } else if (val instanceof Integer) {
            return ((Integer)val).longValue();
        } else if (val instanceof Long) {
            return (Long) val;
        } else if (val instanceof BigDecimal) {
            return ((BigDecimal)val).longValueExact();
        } else if (val instanceof BigInteger) {
            return ((BigInteger)val).longValueExact();
        } else if (val instanceof Float || val instanceof Double) {
            throw new IllegalArgumentException(String.format("%s fields cannot be set with floating point numbers",
                    fieldTypeName));
        } else {
            throw new IllegalArgumentException(String.format("%s fields cannot be set with object %s of type %s",
                    fieldTypeName, val, val.getClass().getName()));
        }
    }

    private static boolean coerceToBoolean(Object val, String fieldTypeName) {
        long l = coerceToLong(val, fieldTypeName);
        if (l == 0) {
            return false;
        } else if (l == 1) {
            return true;
        } else {
            throw new IllegalArgumentException(String.format("Cannot coerce value %s of type %s to %s", val,
                    val.getClass().getName(), fieldTypeName));
        }
    }

    private static short coerceToShort(Object val, String fieldTypeName) {
        Long l = coerceToLong(val, fieldTypeName);
        if (l.shortValue() == l) {
            return l.shortValue();
        } else {
            throw new IllegalArgumentException(String.format("Cannot coerce value %s of type %s to %s", val,
                    val.getClass().getName(), fieldTypeName));
        }
    }

    private static int coerceToInt(Object val, String fieldTypeName) {
        Long l = coerceToLong(val, fieldTypeName);
        if (l.intValue() == l) {
            return l.intValue();
        } else {
            throw new IllegalArgumentException(String.format("Cannot coerce value %s of type %s to %s", val,
                    val.getClass().getName(), fieldTypeName));
        }
    }
```
And likewise for other types. Then call these from the jdbcType switch, eg:

```
switch (jdbcType) {
  [ ... ]
  case BIT:
    rowValues[pair.getKey()] = coerceToBoolean(val, "BIT");
    break;
  [ ... ]
  case SMALLINT:
    rowValues[pair.getKey()] = coerceToShort(val, "SMALLINT");
    break;
  case TINYINT:
    rowValues[pair.getKey()] = coerceToShort(val, "TINYINT");
    break;
}
```
As a disclaimer I haven't tested this.

cheenamalhotra:
Hi @alexnixon Thanks for your suggestions above. I agree to check for object instance for specific types for casting, but with this code, there are 2 downsides:

- Support for String datatype will be gone for setting all numeric types. Is that intended? 
_(Will be a breaking change for support to String values in TVP)_ 
If not, I would like to add another layer of conversion from String values to support String inputs.

- There will be an extra data conversion to long and then to desired type again (for else cases) - can anything be done to prevent that? 
_(Not a major impact, just an observation)_

alexnixon:
Thanks for the reply @cheenamalhotra 

Indeed it would break support for strings. If I was designing from scratch I would omit it, but I understand you might want to keep that for backwards compatibility. So I guess that's a choice for the project maintainer (which might be you?).

But if it is to be included I would still argue for being strict, i.e. ensure it throws an exception for out-of-range values and decimals.

I think the time/memory of an extra long conversion would be negligible, especially given we're about to write to a database, I just wrote the example code that way as it's shorter.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDataTable.java
```diff
@@ -152,34 +152,31 @@ private void internalAddrow(JDBCType jdbcType,
...
-                BigDecimal bd = null;
-                if (null != val) {
-                    bd = new BigDecimal(val.toString());
+    	if(null != val) {
+	        SQLServerDataColumn currentColumnMetadata = pair.getValue();
+	        boolean isColumnMetadataUpdated = false;
+	        int nValueLen;
+	        switch (jdbcType) {
+	            case BIGINT:
+	                rowValues[pair.getKey()] = (val instanceof Number ? (Long)val : Long.parseLong(val.toString()));
```
TheConstructor: Should this really check for `Number` and then cast to `Long`? I would either change to `instanceof Long` or use `((Number)val).longValue()` (which may introduce an unnecessary reboxing if `val` is `Long`)

PR #540, opened 2017-11-03
Updating connectionResiliency branch with the changes from dev branch

ulvii:


src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/NamedParamMultiPartTest.java
```diff
@@ -132,6 +135,7 @@ public void update6() throws Exception {
         Object data = cs.getObject("p_out");
         assertEquals(data, dataPut, "Received data not equal to setdata");
     }
+    */
```
cheenamalhotra: Please disable tests by adding @ Disabled tag - so it will be easier to search for all disabled tests and fix them after complete implementation.

PR #541, opened 2017-11-04
Version Update Configuration Rules.

nsidhaye:
Generally Maven Project use version plugin to update dependencies. Maven Version gives you way to automate update dependencies. 

Most of times we need only stable dependencies. In this PR I am adding `maven-version-rules.xml` so we can configure which version we need to omit from automatic update.

Ref: 
Any Maven Project can use following commands: 
`mvn versions:display-dependency-updates` : Shows newer version of dependencies 
`mvn versions:use-latest-releases` :  Update dependencies to latest version 
`mvn versions:update-properties` :  Update dependencies by updating properties


cheenamalhotra:
Thanks @nsidhaye for the updates. 
We are good to merge this PR with our next release.

Thanks for your contribution.

PR #542, opened 2017-11-06
add "connectionDateformat" connection parameter

gordthompson:
(Prompted by [this Stack Overflow question](https://stackoverflow.com/q/47111591/2144390).)

It is not uncommon for applications to try and be database-agnostic by dealing with date/time values as strings. Sqoop seems to be one of those applications.

Normally, using a string date literal like `'2017-11-05'` is considered safe because the leading year value suggests that the format is `yyyy-mm-dd`. `yyyy-dd-mm` is almost unheard of ...

... except with SQL Server. If the SQL Server instance is installed with "British English" (or perhaps "English (UK)"?) as the default locale then the default DATEFORMAT is "dmy". When it is passed a `yyyy-xx-xx` string literal for a DATETIME value it is interpreted as `yyyy-dd-mm`, not `yyyy-mm-dd`:

```java
connUrl = "jdbc:sqlserver://localhost:49242;databaseName=myDb";
try (Connection conn = DriverManager.getConnection(connUrl, uid, pwd)) {
    String sql = "SELECT MONTH(CONVERT(DATETIME, '2017-11-05')) AS month_num";
    System.out.println(sql);
    try (Statement st = conn.createStatement()) {
        ResultSet rs = st.executeQuery(sql);
        rs.next();
        System.out.printf("month_num is %d%n", rs.getInt("month_num"));
    }
```

result:

```
SELECT MONTH(CONVERT(DATETIME, '2017-11-05')) AS month_num
month_num is 5
```

A [workaround](https://support.microsoft.com/en-us/help/173907/inf-how-to-set-the-day-month-year-date-format-in-sql-server) is to send a `SET DATEFORMAT 'ymd'` statement immediately after opening the connection. Unfortunately, Sqoop has that option for Oracle connections, but apparently not for other JDBC drivers.

This patch adds a "connectionDateformat" parameter for the connection URL and for a SQLServerDataSource. After adding `;connectionDateformat=ymd` ...

```java
connUrl = "jdbc:sqlserver://localhost:49242;databaseName=myDb;connectionDateformat=ymd";
try (Connection conn = DriverManager.getConnection(connUrl, uid, pwd)) {
    String sql = "SELECT MONTH(CONVERT(DATETIME, '2017-11-05')) AS month_num";
    System.out.println(sql);
    try (Statement st = conn.createStatement()) {
        ResultSet rs = st.executeQuery(sql);
        rs.next();
        System.out.printf("month_num is %d%n", rs.getInt("month_num"));
    }
```

... the result is:

```
SELECT MONTH(CONVERT(DATETIME, '2017-11-05')) AS month_num
month_num is 11
```

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
@@ -356,7 +356,7 @@ public class SQLServerBulkCopy implements java.lang.AutoCloseable {
      * @throws SQLServerException
      *             If a connection cannot be established.
      */
-    public SQLServerBulkCopy(String connectionUrl) throws SQLServerException {
+    public SQLServerBulkCopy(String connectionUrl) throws SQLServerException, SQLException {
         loggerExternal.entering(loggerClassName, "SQLServerBulkCopy", "connectionUrl not traced.");
         if ((connectionUrl == null) || "".equals(connectionUrl.trim())) {
             throw new SQLServerException(null, SQLServerException.getErrString("R_nullConnection"), null, 0, false);
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
@@ -999,7 +999,7 @@ public class SQLServerConnection implements ISQLServerConnection {
     /**
      * Check if the connection is closed Create a new connection if it's a fedauth connection and the access token is going to expire.
      * 
-     * @throws SQLServerException
+     * @throws SQLServerException 
      */
     /* L0 */ void checkClosed() throws SQLServerException {
         if (isSessionUnAvailable()) {
@@ -1073,7 +1073,33 @@ public class SQLServerConnection implements ISQLServerConnection {
 
         for (int retryAttempt = 0;;) {
             try {
-                return connectInternal(propsIn, pooledConnection);
+                Connection result = connectInternal(propsIn, pooledConnection);
+                
+                String connectionDateformatValue = propsIn.getProperty(SQLServerDriverStringProperty.CONNECTION_DATEFORMAT.toString(), "").trim();
+                if (connectionDateformatValue.length() > 0) {
+                    switch (connectionDateformatValue) {
+                        // valid values per https://support.microsoft.com/en-us/help/173907/inf-how-to-set-the-day-month-year-date-format-in-sql-server
+                        case "mdy":
+                        case "dmy":
+                        case "ymd":
+                        case "ydm":
+                        case "myd":
+                        case "dym":
+                            try {
+                                Statement st = result.createStatement();
+                                st.execute(String.format("SET DATEFORMAT '%s'", connectionDateformatValue));
+                                break;
+                            } catch (SQLException se) {
+                                throw new SQLServerException("connectionDateformat SET DATEFORMAT call failed", se);
+                            }
+                        default:
+                            throw new IllegalArgumentException(
+                                    String.format("Invalid value for %s: %s", 
+                                    SQLServerDriverStringProperty.CONNECTION_DATEFORMAT.toString(),
+                                    connectionDateformatValue));
+                    }
+                }
+                return result;
             }
             catch (SQLServerException e) {
                 // Catch only the TLS 1.2 specific intermittent error.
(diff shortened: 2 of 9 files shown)


ulvii:
I would let the applications handle this case. @v-xiangs , @cheenamalhotra what do you think?

cheenamalhotra:
Hi @gordthompson Thanks for bringing this change request. I did go through the Stack Overflow discussion, and I understand from your description, changing Default Language to "English" instead of "British English" does make the desired change, and looks a fair solution to the problem.

I would say we want to give complete control over to the developer to decide how and when do they want to make this change. Providing a connection property is an option, but not an ideal solution in this case. If anyone wants to execute this statement on a connection, that can be done programmatically (as it will be in rare cases) and that would not add complexity to the driver.

Also, we would like to seek community's feedback over this connection property and whether it does make a big difference to the driver.

@ajlam @sehrope @TobiasSQL @marschall Appreciate your thoughts on this new connection property introduction to the driver.

marschall:
> It is not uncommon for applications to try and be database-agnostic by dealing with date/time values as strings.

This is horrible and wrong. The same goes for dealing with date/time values as numbers.

> Sqoop seems to be one of those applications.

Then Sqoop needs to be fixed.

gordthompson:
@cheenamalhotra re: "changing Default Language ... looks a fair solution to the problem"

Except that 
1. server configuration changes are often difficult (if not impossible) for humble end users, and
2. such a change *could* have side-effects.
&nbsp;

@ulvii re: "let the applications handle this case"
@cheenamalhotra re: "that can be done programmatically"

The whole point of the Pull Request is that doing it "programmatically" does not currently seem to be possible in Sqoop, at least not for SQL Server.
&nbsp;

@marschall

If string literals for date/time values are so wrong, then **(1)** why does SQL Server support this in the first place?

```sql
CREATE TABLE #tmp (id INT PRIMARY KEY, dt DATETIME);
INSERT INTO #tmp (id, dt) VALUES (1, NULL);
UPDATE #tmp SET dt = '2017-11-05 13:14:15' WHERE id = 1;  -- "horrible and wrong"
SELECT MONTH(dt) AS month_num FROM #tmp WHERE id = 1; 
DROP TABLE #tmp;
```

And **(2)** why does it interpret `yyyy-xx-xx` as `yyyy-dd-mm` (only for DATETIME, not for DATE or DATETIME2) when the default DATEFORMAT for "British English" is `dmy`, not `ydm`.

sehrope:
I see the utility of this though if you go down the road of adding connection properties for setting the date format, someone could argue for the same thing for numeric formats or timestamp formats. That has nothing to do with them having more consistent formats either. Someone may just want a way to set them to match the format of their input data. Plus as new values get added server side for any of those, the driver would need to be updated to accommodate them (or remove any validation).

Rather than having specific hooks for each property perhaps a generic way to execute a "post-connect" SQL command or set of commands. Not sure I really like that idea either as it seems like something that should be handled at the app level. That way the app can deal with potential errors as well. But at least it's just one connection property and hook to maintain rather than potentially N of them.

cheenamalhotra:
I think main concern is that **Date Format** being one such factor that gets affected by locale and we modify it with specifying custom value in connection properties.

I would rather recommend something more generic, eg. "**Regional**" as we have in [Microsoft ODBC Driver connection params](https://docs.microsoft.com/en-us/sql/relational-databases/native-client/applications/using-connection-string-keywords-with-sql-server-native-client#odbc-driver-connection-string-keywords), which when **yes**, uses client settings when converting currency, date, and time data to character data, while when **no**, would make driver use JDBC standard strings to represent currency, date, and time data that is converted to character data.

gordthompson:
@cheenamalhotra 

I understand your reasoning, but this patch is specific to the unusual handling of date strings for certain locales, and is specifically addressed by a T-SQL [SET Statement](https://docs.microsoft.com/en-us/sql/t-sql/statements/set-statements-transact-sql) ([SET DATEFORMAT](https://docs.microsoft.com/en-us/sql/t-sql/statements/set-dateformat-transact-sql)). The [SET Statement](https://docs.microsoft.com/en-us/sql/t-sql/statements/set-statements-transact-sql) document does not mention similar SET statements for currency or other data types that may have a variety of local string representations, so presumably they have not been sufficiently problematic for the core SQL Server team to address.

peterbae:
Hi @gordthompson,

Sorry to have left this PR hanging here for a long time. We've reviewed this PR again, and based on the discussions it seems like we'd like a more generic approach to solving this problem, rather than introducing a connection property for each specific format. We've backlogged this item for now, and will return to this in the future.

PR #543, opened 2017-11-07
Fix prepared statement metadata cache issue

v-xiangs:
also added tests to verify the number of insertion/update executed.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
@@ -2618,7 +2618,6 @@ public class SQLServerPreparedStatement extends SQLServerStatement implements IS
             if (doPrepExec(tdsWriter, batchParam, hasNewTypeDefinitions, hasExistingTypeDefinitions) || numBatchesPrepared == numBatches) {
                 ensureExecuteResultsReader(batchCommand.startResponse(getIsResponseBufferingAdaptive()));
 
-                boolean retry = false;
                 while (numBatchesExecuted < numBatchesPrepared) {
                     // NOTE:
                     // When making changes to anything below, consider whether similar changes need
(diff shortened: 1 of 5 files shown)


TobiasSQL:
This SQLServerConnection.contextChanged thing, do we really believe that we can keep this accurate? There are lots of reasons for the context on the server to change, trying to figure this out from the client side seems like a losing proposition. This was why we went with re-try logic instead, obviously with its own issues. What are the thoughts here?

v-xiangs:
@TobiasSQL Thank you so much for checking the code. I did talk to our team about the limitation of this ` SQLServerConnection.contextChanged` thing. I didn't want to send another query to SQL Server to retrieve the status of `ANSI_NULLS`, `QUOTED_IDENTIFIER` etc., which would affect the performance of the driver. Instead, I decided to detect keywords passed into the driver. The limitation is: if user calls 'use DB1' twice, the driver still assumes the context is changed although context didn't change; also, if user has keywords in text the driver would think the context is changed. I think the chance of having those scenarios is low. At least, having it works with slower performance in some scenarios is acceptable in my opinion, unless we can have better solutions. Also, context is connection based. I tested it. If I establish a connection with SSMS and change the context there, it wont affect the driver, since they use different connections.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -916,7 +905,11 @@ private void getParameterEncryptionMetadata(Parameter[] params) throws SQLServer
 
 	/** Manage re-using cached handles */
 	private boolean reuseCachedHandle(boolean hasNewTypeDefinitions, boolean discardCurrentCacheItem) {
-		
+        if (definitionChanged || connection.contextChanged) {
+            prepStmtHandle = -1; // so that hasPreparedStatementHandle() also returns false
+            return false;
```
v-afrafi: we should also add these two in here, do you agree? 
```
connection.contextChanged = false;  // to be able to use the handle after this call
connection.clearCachedPreparedStatementHandle(); //making sure the cache is cleared
```

v-xiangs: Thank you so much for the suggestion, it's much better.

v-xiangs:
@TobiasSQL Hello, the team has decided to merge this PR. Please feel free to leave comments after it's merged.

PR #547, opened 2017-11-08
Calling stored procedure directly 

AfsanehR-zz:
calling stored procedure directly instead of wrapping it around `exec sp_executesql`.

Not to be merged yet. Opening this pull request for discussion and improvement. 

**Notice that this still does not have the push for avoiding the metadata lookup with named params. Will be adding that soon to the same pr in another commit.**


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -772,6 +784,22 @@ private void buildExecParams(TDSWriter tdsWriter) throws SQLServerException {
         assert hasPreparedStatementHandle();
         tdsWriter.writeRPCInt(null, getPreparedStatementHandle(), false);
     }
+    
+    private void buildRPCExecParams(TDSWriter tdsWriter) throws SQLServerException {
+        if (getStatementLogger().isLoggable(java.util.logging.Level.FINE))
+            getStatementLogger().fine(toString() + ": calling sp_execute: PROC" + ", SQL:" + preparedSQL);
```
TobiasSQL: This should not say "calling sp_execute" since we are doing a direct SQLRPC call to the specific procedure.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -967,38 +995,57 @@ private boolean doPrepExec(TDSWriter tdsWriter,
...
-            
-            if (needsPrepare) 
+            if (needsPrepare)
                 buildServerCursorPrepExecParams(tdsWriter);
             else
                 buildServerCursorExecParams(tdsWriter);
         }
         else {
+            int paramCount = SQLServerConnection.countParams(userSQL);
+            if (needsPrepare && null != procedureName && paramCount != 0 && !isTVPType(params)) {
```
TobiasSQL: A few questions here:
1. How do we know that this is an RPC call with separate list of parameters vs. something like "EXEC spTobias @dobidoo = 2" as a string plus separately passing other parameters.
2. We should support this also for 1 == paramCount, no?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
v-afrafi: Yes! good catch.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
v-afrafi: Good question:
for queries like `EXEC spTobias @dobidoo = 2` the param count is still counted as 0. The countParams(userSQL) method will return 0. This method which was already used in the driver, counts params by the number of ? marks.  (since we are not using queries such as `EXEC spTobias (?)` and then later bind the parameters. For such queries for now, we use the current behavior of driver which is 'sp_execute'. 

I think that would require the countParams to parse the queries for parameters passed in userSQL too. 
I will try to add this with the next push.

TobiasSQL: To clarify, how do you handle this "EXEC spTobias 2, ?", where paramCount is 1?

v-afrafi: You are correct @TobiasSQL .. Good catch.. in that case we need to check the query for such scenarios. One solution is to look for '? ' marks and replace them with parameters that is passed. Instead of RPC call we can directly use a query call.
For now, I added a check for those queries and not use rpc calls, instead use the old behavior of calling sp_executesql .
What do you think about this?

v-afrafi:
I also added the fix for removing metadata lookup for named parameters. This will break the existing applications that use a mix of named and index parameters. 
For example for a proc like: 
```
CREATE PROCEDURE proc (
    @par1 INT = NULL
) AS
BEGIN
	insert into tempTable values (@par1)
END
```

If we need to use named parameters to get the output parameters, we need to register the parameter with name values:
```
callableStatement.registerOutParameter("@par1", java.sql.Types.INTEGER);
callableStatement.execute();
```
and then call:
` callableStatement.getObject("@par1");`

Registering output parameter with name and using indexes to retrieve them, however, still works.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCallableStatement.java
```diff
@@ -464,7 +498,7 @@ public int getInt(int index) throws SQLServerException {
     public int getInt(String sCol) throws SQLServerException {
         loggerExternal.entering(getClassNameLogging(), "getInt", sCol);
         checkClosed();
-        Integer value = (Integer) getValue(findColumn(sCol), JDBCType.INTEGER);
+        Integer value = (Integer) getValue(findColumn(sCol, false), JDBCType.INTEGER);
```
brettwooldridge: It seems nit picky, but could you either: 1) define two boolean constants, IS_SETTER and IS_GETTER, or 2) use an Enum, for the parameter passed to findColumn()?

Reading this diff it was like a murder mystery until the end of the file; what is that parameter going to mean?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -267,12 +267,25 @@ static ParsedSQLCacheItem  parseAndCacheSQL(Sha1HashKey key, String sql) throws
...
+     * @param sql
+     * @return
+     */
+    static boolean isRPCValid(String sql) {
+        if (sql.contains("?") && sql.contains("\'")) // means that query params are not all passed. Example: exec procName 1,?
+        {
+            return false;
+        }
+        return true;
+    }
```
brettwooldridge: Maybe?

return !(sql.contains("?") && sql.contains("\'"));

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCallableStatement.java
v-afrafi: @brettwooldridge Thank you for reviewing this! I will push in the next commit with the change to something more defining like you suggested.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCallableStatement.java
```diff
@@ -1383,11 +1417,27 @@ public NClob getNClob(String parameterName) throws SQLException {
...
+        return getArray(findColumn(sCol, false));
+    }
+    
+    /**
+     * handle `@name` as well as `name`, since `@name` is what's returned by DatabaseMetaData#getProcedureColumns
+     * 
+     * @param columnName
+     * @return
+     */
+    private String concatAtSign(String columnName) {
```
gordthompson: This doesn't concatenate an "@" to anything, it removes one (if present), so perhaps it should be called `stripLeadingAtSign`.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
TobiasSQL: How do you detect this (i.e. that we are either strictly calling SP without params or only ?s) ? I couldn't find it.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCallableStatement.java
v-afrafi: @gordthompson That's right. Changed it to the suggested name. Thanks!

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
v-afrafi: @TobiasSQL Thank you so much for following up. please check this method on callRPCDirectly in [here](https://github.com/Microsoft/mssql-jdbc/pull/547/files#diff-a51517a0e4bf7e029ed04f9476a37f89R1042) and its implementation in [here](https://github.com/Microsoft/mssql-jdbc/pull/547/files#diff-0a4455f5ffd8e31e9e1badaa7822375eR284).
Basically by counting the number of parameters (?) and commas. 
If there is no ? (meaning parameters in the query) or there sql is something like `exec proc ?, 1, 2` we directly use sp_executesql.
Do you think it makes sense?

praskutti:
Hello mssql-jdbc team, can you please let me know when this will be merged and released?

ulvii:
Hi @praskutti ,
The PR is incomplete and requires quite some changes to merge it. It is in our backlog to address them, but there is no timeline currently.

praskutti:
> Hi @praskutti ,
> The PR is incomplete and requires quite some changes to merge it. It is in our backlog to address them, but there is no timeline currently.

Thanks @ulvii . I coded a usable workaround for our requirement (to create a generic method for SQL Server proc calls that would avoid extra metadata lookups). I thought this PR would solve it, but because it will take time to release, we will roll with our method. I will post the code-sample soon.

Jeffery-Wasty:
This is still being worked on #2154.

Issue #477, opened 2017-09-05
Retrieving result set after deleting a row using deleteRow method throws SQLServerException


rickysang:
## Driver version or jar name
mssql-jdbc-6.2.1.jre8.jar

## SQL Server version
SQL Server 2012

## Client operating system
Windows 10

## Java/JVM version
java version "1.8.0_121"
Java(TM) SE Runtime Environment (build 1.8.0_121-b13)
Java HotSpot(TM) 64-Bit Server VM (build 25.121-b13, mixed mode)

## Table schema
``` sql
CREATE TABLE DeleteRowTest 
(
    Id INT PRIMARY KEY NOT NULL IDENTITY,
    Name VARCHAR(32) NOT NULL
)
GO

SET IDENTITY_INSERT DeleteRowTest ON
GO

INSERT INTO DeleteRowTest (Id, Name) VALUES
  (1, 'Jon Snow'),
  (2, 'Tony Stack'),
  (3, 'Bran Stack'),
  (4, 'Samwell Tarly')
GO

SET IDENTITY_INSERT DeleteRowTest OFF
GO
```

## Problem description
Set type of `ResultSet` to `ResultSet.TYPE_SCROLL_SENSITIVE` and set concurrency of `ResultSet` to `ResultSet.CONCUR_UPDATEABLE`, then running sql to get an updateable `ResultSet`.  

Next, perform delete operation on table using `deleteRow()` method. 

After deleting a row, retrieving result set throws SQLServerException(Cannot get a value from a deleted row).

## Expected behavior and actual behavior
Run reproduce code below, it throws a SQLServerException. I also tried run it with MySQL, it works fine. I suppose that this is not correct.
![20170905110115](https://user-images.githubusercontent.com/19810448/30043377-91fa669a-9229-11e7-80f8-67196c3baed3.png)

It can be fixed by checking whether current row has been deleted using `rowDeleted()` method before retrieving the column value. I tried this fix solution with MySQL again, MySQL driver doesn't support `rowDeleted()` feature.

Seems it's a vendor specific issue, a deleted row leave a visible hole in result set in mssql-jdbc while mysql-jdbc does not.
[http://docs.oracle.com/javase/8/docs/api/java/sql/ResultSet.html#rowDeleted--](http://docs.oracle.com/javase/8/docs/api/java/sql/ResultSet.html#rowDeleted--)

Maybe this difference should be included in document. Or is it possible to eject visible hole in result set?

## Repro code
Java code:
[...]

PR #552, opened 2017-11-16
Add check for deleted rows in resultsets for scrollable & updatable resultsets

peterbae:
fixes issue #477


rickysang:
@peterbae Thanks for the fix. Test passed.

Issue #525, opened 2017-10-11
Throw exception when Statement#setQueryTimeout is reached when no network is available.


dave-r12:
## Driver version or jar name
6.2.1.jre8

## SQL Server version
SQL Server 2012

## Client operating system
CentOS 7

## Java/JVM version
java version "1.8.0_91"

## Table schema
N/A

## Problem description
In our setup, SQL Server will issue TCP keepalives on idle connections to eagerly detect potentially bad connections. Unfortunately this behavior sometimes results in a good connection be flagged as bad if the keepalives do not reach the other host. We've seen instances where SQL Server will terminate the connection, but our application hosts think the connections are still good. (Presumably the keepalives are getting dropped.) When we then attempt to use the connection it hangs. SQL Server will silently drop the packets without responding with a TCP reset. Eventually, we reach the TCP packet retransmitted limit (aka `tcp_retries2`).

Attempting to set the `Statement#setQueryTimeout` only works if the server responds with data. If the server is not responding, then the timeout is essentially ignored.

## Expected behavior and actual behavior
I believe the timeout should trigger the Statement to terminate immediately, preferably with an exception of some kind.

Currently the code will hang until the socket times out via the retransmitted limit being reached.

## Repro code
Here is a simple Java class to execute:

```java
  public static void main(String[] args) throws Exception {
    DriverManager.class.forName("com.microsoft.sqlserver.jdbc.SQLServerDriver");

    Connection connection = DriverManager.getConnection("jdbc:sqlserver://localhost;databaseName=mydatabase", "sa", "Password");
    System.out.println("Cause the TCP connection to be terminated from SQL Server side then hit enter.");
    System.in.read();
    Statement statement = connection.createStatement();
    statement.setQueryTimeout(1);

    try {
      statement.execute("WAITFOR DELAY '00:00:3'; SELECT 1");
    } catch (SQLException e) {
      e.printStackTrace();
    }
    System.out.println("Complete");
  }
```

Here's how I terminate the TCP connection from the app host. Basically don't respond to the TCP keepalives:

```bash
# Get the port.
ss -tn | grep 1433

# Drop the packets.
sudo iptables -I INPUT 1 -p tcp --dport <paste-port-number-here> -j DROP

# Wait for a bit...
# Remove the iptables rule once the connection is terminated by SQL Server.
sudo iptables -D INPUT 1
```

To determine when SQL Server terminates the connection, I use this query:
```sql
select * from sys.dm_exec_sessions
[...]

PR #560, opened 2017-11-22
Add cancelTimeout to cancel the query timeout

AfsanehR:
Added new connection property of `cancelQueryTimeout`. This will address issue #525 where querytimeout hangs and does not throw exception due to silent dropping of connection. 
The driver will wait the total amount of cancelQueryTimeout + queryTimeout to drop the connection and close the channel. 

**Note:** The new connection property of `cancelQueryTimeout` will be effective only if the `queryTimeout ` is set in the driver.


cheenamalhotra:
Closing this PR and new PR #674 extended from this one.

Issue #556, opened 2017-11-20
Passing datetime2/date via ISQLServerBulkRecord is throwing exception


SubashRoul:
## Driver version or jar name
mssql-jdbc-6.2.1.jre8.jar

## Java/JVM version
java version "1.8.0_152"
Java(TM) SE Runtime Environment (build 1.8.0_152-b16)

## Problem description
For Azure SQL Data Warehouse, passing datetime2 via ISQLServerBulkRecord throws `Received an invalid column length from the bcp client for colid modify_time`. Same error for date type columns.

I have tried the example codes given in issue #161  with datetime2 and date columns.

For datetime2, I am passing the following parameters for datetime2 columns.
columnType : java.sql.Types.TIMESTAMP
precision : 0
scale : 0

For date,
columnType : java.sql.Types.DATE
precision : 0
scale : 0

peterbae:
Hi @SubashRoul, to implement a solution for this issue, I need to implement #161 as well, and it seems like it's going to take more time than i originally thought. I've left a comment in my PR (closed now) regarding my progress, and I will return to this issue at a later time.

rkinabhi:
Hello @peterbae @cheenamalhotra ,
is this still a work in progress?
cc @saksham1711 @tgorthi

peterbae:
It's in our backlog for now, but we don't have a timeline for this issue yet.

FelipeBR2:
Hi, as [#161](https://github.com/microsoft/mssql-jdbc/issues/161) is now closed, do you have a timeline for this topic?

Jeffery-Wasty:
While this has been in the backlog for a while, there is no current timeline. We'll take a look at it during out next planning and provide updates as necessary.

FelipeBR2:
Hello @Jeffery-Wasty , could you please prioritize this issue as it has a very high business impact for the Robert Bosch GmbH? We use the azure data warehouse as part of an analytical procedure to optimize our blocked inventory  on a global level at Bosch. We require this bulk upload because thousands of records needs to be update simultaneously which also includes the Datetime data type which is currently not working with the JDBC driver.

If you have further questions, please contact me at felipe.deguchi@br.bosch.com

Jeffery-Wasty:
Hi @FelipeBR2,

Thank you for your patience, we have a team member looking into the issue at the moment.

lilgreenbird:
hi @FelipeBR2 

[...]

PR #568, opened 2017-11-30
No longer sending varchar to server for temporal data types.

peterbae:
Fixes issue #556.

Also addresses previously closed (for inactivity) issue https://github.com/Microsoft/mssql-jdbc/issues/161.


peterbae:
Closing this PR as this doesn't fully solve the problem. After diving deeper into the problem, I have a solution for date/datetimeoffset data types, but the driver needs to implement a logic for parsing datetime/datetime2 data (in varchar format) to their respective temporal data types. Looks like the more robust approach would be to use DateTimeFormatter to parse all the potential data from CSV, and I will return to this issue after the higher priority issues have been addressed.

PR #571, opened 2017-12-04
Jdbc4.3

rene-ye:
with working module-info.java


rene-ye:
Maven-javadoc-plugin3.0.0 was released today, fixing the [compatibility problem](https://issues.apache.org/jira/browse/MJAVADOC-498) with module-info.java. It is no longer necessary to comment out the javadoc plugin, a new PR will be created to reflect these changes.

PR #573, opened 2017-12-06
Fix issue with rowDeleted() causing issues with testing

peterbae:
Further addresses issue #477.

the rowDeleted() method calls currentRowDeleted, which checks for the row being deleted outside the cursor. This also involves skipping some columns within the row which results in the resultset state being different than what the test expects and throws an error. Using getDeletedCurrentRow instead (which checks to see if the row has been deleted via deleteRow()) prevents this problem from happening.


peterbae:
Will make a new PR that reverts my original PR.

PR #576, opened 2017-12-07
module-info updates

rene-ye:


src/main/java/module-info.java
```diff
@@ -0,0 +1,15 @@
+module mssql.jdbc
+{
+	exports com.microsoft.sqlserver.jdbc;
+
+	requires adal4j;
```
marschall: I think all of these should be optional dependencies.

src/main/java/module-info.java
```diff
@@ -0,0 +1,15 @@
...
+	requires azure.keyvault;
+	requires azure.keyvault.webkey;
+
+	requires java.logging;
+	requires java.management;
+	requires java.naming;
+	requires java.security.jgss;
```
marschall: Are all of these required? I would imagine there is a good chance at least that GSS and JMX, maybe even JNDI are optional.

src/main/java/module-info.java
```diff
@@ -0,0 +1,15 @@
+module mssql.jdbc
+{
+	exports com.microsoft.sqlserver.jdbc;
+
+	requires adal4j;
+	requires azure.keyvault;
+	requires azure.keyvault.webkey;
+
```
marschall: I think you're missing a service declaration there, something like

```java
provides java.sql.Driver with com.microsoft.sqlserver.jdbc.SQLServerDriver;
```

src/main/java/module-info.java
```diff
@@ -0,0 +1,15 @@
+module mssql.jdbc
```
marschall: I know naming is hard, I just wanted to let you know there are other possible naming strategies and simply converting the artefact name to the module name in not necessarily the best option

https://github.com/jodastephen/jpms-module-names

cheenamalhotra:
Closing to be addressed later

src/main/java/module-info.java
nsidhaye: If we want real modularity then may jgss / Kerberos might be different module altogether. Well that might be challenge as our code is so much integrated. But this way we can leverage real benefit of modularity. 

Suggesting some modules: 
1. Kerberos-auth
1. encryption / azure-keyvault or just azure
1. fips
1. tvp

Some of features which may not be used in day to day operations can be turned into modules which will definitely will be useful to end users. Any way full jar is always available which can treated everything is mandatory.

PR #577, opened 2017-12-07
Removal of deprecated APIs and update travic script with jdk 9

v-afrafi:


v-afrafi:
@v-xiangs good point. The difference between those two is that Integer.parseInt() return primitive int and valueOf returns java.lang.Integer.  I mostly used Integer.parseInt(String) except for the line that we need to convert Int to string. for example [here](https://github.com/Microsoft/mssql-jdbc/pull/577/files#diff-18d0d0627e5a0b7a9abe73358d2358a8R869): using Integer.valueOf() allows us to call toString() method on the Object.

Issue #567, opened 2017-11-29
"The TDS protocol stream is not valid" exception when streaming larger content as blob from varbinary(max) column


ericpias:
## Driver version or jar name
6.2.2.jre8.  Also happens in 6.3.5 and 6.2.0.jre8.  It does NOT happen with 6.1.0.jre8 so it seems like a regression.

## SQL Server version
2014

## Client operating system
Mac and Windows.  Does not matter.

## Java/JVM version
1.8.0_45-b14

## Table schema
create table [contentdata] (
        [id] varchar(255) not null,
        [dtype] varchar(255) not null,
        [name] varchar(255),
        [type] varchar(255),
        [size] numeric(19,0),
        [data] varbinary(max),
        primary key ([id])
    );

The data column is the important one.

## Problem description
This seems to be an issue with 6.2 and later driver.

Attempting to read from the binary stream of a blob for a varbinary column with more than 40-60 kb of content throws SQL exception: "The TDS protocol stream is not valid".  Trying blob.length() first also throws the same exception.

## Expected behavior and actual behavior
Should be able to read from the binary stream.  Worked ok in v 6.1.

## Repro code
This is the simplest plain vanilla jdbc code that illustrates the issue.  The main point is that the blob's stream is no longer valid after closing the resultset when using v6.2 and later of the mssql jdbc driver whereas in v6.1 the stream is still valid and can be read from after resultset/statement close.  From my understanding, the "A Blob object is valid for the duration of the transaction in which is was created".

Code:

package com.gauss.forms.web;

import java.io.InputStream;
import java.sql.Blob;
import java.sql.Connection;
import java.sql.Driver;
import java.sql.DriverManager;
import java.sql.ResultSet;
import java.sql.Statement;

public class JdbcBlobTester {

	public static void main(String[] args) {
		// 
		Connection connection = null;  
        Statement statement = null;   
        ResultSet rs = null;  
		try {
			Class cls = Class.forName("com.microsoft.sqlserver.jdbc.SQLServerDriver");
			
			Driver driver = (Driver)cls.newInstance();
			System.out.println("Driver version: " + driver.getMajorVersion() + "." + driver.getMinorVersion());
			
			String connectionString = "some connection string;";
			
			connection = DriverManager.getConnection(connectionString, "user", "password");  
			connection.setAutoCommit(false);
			
			String selectSql = "SELECT ...";  
            statement = connection.createStatement();  
            rs = statement.executeQuery(selectSql);  

[...]

PR #595, opened 2018-01-03
Blob fix

rene-ye:
Fills the contents of a blob and makes it availible for use after the RS or statement has been closed. Addresses the TDS issue from #567.


eacarvalho:
Hi all,
Please, I'd like to know if there is an expected date to release this pull request. It will be in the 6.5.0 release? Actually the current version is 6.4.0.
Thanks a lot.

rene-ye:
This PR will be in 6.5.0-preview, scheduled to be released by end of Friday March 9th.

Lonzak:
I have a question: I don't see a final release of 6.5. Did all the changes of 6.5 went into version 7.0?

PR #605, opened 2018-01-16
Add support for a serverTimeZone property

kkofler:
The serverTimeZone property can be passed in the connection string and
allows globally setting a time zone, which is used for all date and time
conversions where no Calendar is explicitly specified, instead of using
the client JVM's time zone. Having this option set explicitly in all
clients ensures that they all agree on the time zone to use,
independently of what their default time zone is.

Forcing serverTimeZone to UTC even also helps if the database contains
times in more than one time zone, because UTC has no invalid time
ranges, so no information is lost converting between a time with
unspecified time zone and a time stamp in UTC.

Some other JDBC drivers (e.g., MySQL Connector/J and MariaDB
Connector/J) also have such an option.


kkofler:
For the CLA, I am waiting for the official OK from my employer.

kkofler:
CLA now on file, my employer OKed it.

gordthompson:
@cheenamalhotra - Related to #749 ...?

ulvii:
Hi @kkofler,

The methods that let you specify `Calendar` explicitly already provide the functionality you are asking for. That being said, are there any date/time APIs that do not have an alternative with `Calendar` support? If that's the case, please let us know, we might consider adding it.

kkofler:
Unfortunately, always explicitly specifying the `Calendar` is not really a practical option when working with an abstraction layer such as Hibernate.

ulvii:
@kkofler,

Looks like Hibernate 5.3 has a property called [`hibernate.jdbc.time_zone`](http://docs.jboss.org/hibernate/orm/5.3/userguide/html_single/Hibernate_User_Guide.html#basic-datetime-time-zone) that invokes the methods with `Calendar` support. Please give it a try.

ulvii:
Closing the PR as the driver already provides the requested functionality with `Calendar`. @kkofler, please let us know if you have any concerns. Please also feel free to create PRs, we appreciate contributions from community.

Savers80:
@ulvii i think the solution of use property hibernate.jdbc.time_zone is valid only if  use Hibernate
However if the application use hibernate and you can config hibernate property in case  you have more jdbc database connection you cannot specify time _zone for one specific database; all database must use same time_zone; i.e db1 store datetime as GMT and JVM use user.timeZone=GMT, db2 used from third-party application stores datetime as UTC+2 

In case you don't use hibernate or you can't modify hibernate configuration per single database (i.e third-party application "Appian" ) and you can't manage this scenario

I think  pull #749  is not related to this feature serverTimeZone
i.e JavaApplication1 that use JVM time zone GMT that acces to another database db2  where datetime are stored in UTC+2 , convert java date time at 00:00 GMT into database side 02:00 UTC+2
Using the requested property  'serverTimeZone=UTC+2' in this scenario the above conversion musto no take place and query filtering will consider java side 00:00 GMT as database side 00:00 UTC+2

Hope reconsider to add support for a serverTimeZone property
Or suggest a solution for the scenario showed , without modify java code of thirty Part Application, or modify/implement database view

Issue #610, opened 2018-01-25
Detect changed context is just ...


TobiasSQL:
https://github.com/Microsoft/mssql-jdbc/blob/e0e70a429aa6f84d151e5a18aef8b92f3337e269/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerStatement.java#L806

+@brettwooldridge to keep me honest.

The `checkIfContextChanged` method in `SQLServerStatement` has a bunch of issues. My understanding is that this was added to handle the case where prepared handles cannot be re-used because the server context has changed. However, first off the implementation is very flaky, for ex. if I have a stored procedure called Add**Use**r it will be detected as a context change. This also adds a bunch of overhead in terms of string manipulation & comparison with very little gain (most of the time the context will not have changed).

The previous implementation (that I did) attempted to re-create the handle if the context changed. There are pros and cons here, but I think that or simply throwing an error are the only reasonable alternatives. IMHO, the cons against this current approach are:
1. Not guaranteed to work, driver will not have the context and logic of the server to be accurate
2. Adds a bunch of overhead for a fairly edge case scenario

Thoughts?
/Tobias

brettwooldridge:
@TobiasSQL *My eyes are burning!* 😱 

Let see:
 * How many times can we call ``toUpperCase()``?  Each call horribly inefficient if the SQL is several hundred lines (not uncommon in some apps).
 * How many times can we call ``contains()`` on each one of those?  Again each call horribly inefficient, because ``contains()`` as no choice but to scan the entire string to the end.

The answer is apparently 4x ``toUpperCase()`` and 4x ``contains()``, which will generally *all* execute because it is a chain of ``||`` operators.

I see no reason not to kill it.  The proposal outlined above, of simply throwing, seems reasonable.

TobiasSQL:
Thanks @brettwooldridge , I share your sentiment exactly.

TobiasSQL:
I will add that the reason I went with re-try in the first implementation is that otherwise it would be a breaking change. I.e. someone not using cached handles would not have gotten and exception and now they do. I verified internally that the way we do it, it should be safe to re-try.

shawnxiangyusun:
[...]

PR #618, opened 2018-01-31
Re-introduce Retry logic for prepared statement caching 

AfsanehR:
- Added APIs for disableStatementPooling
- Set the disableStatementPooling to true by default
- Disable prepared statement caching by default 
- Introduce retry logic for missing handle exception. 

Fixes issue #610

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
@@ -548,7 +548,7 @@ public class SQLServerPreparedStatement extends SQLServerStatement implements IS
                 getNextResult();
             }
             catch (SQLException e) {
-                if (retryBasedOnFailedReuseOfCachedHandle(e, attempt, needsPrepare) && connection.isStatementPoolingEnabled())
+                if (retryBasedOnFailedReuseOfCachedHandle(e, attempt, needsPrepare))
                     continue;
                 else
                     throw e;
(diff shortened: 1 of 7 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -547,7 +547,7 @@ final void doExecutePreparedStatement(PrepStmtExecCmd command) throws SQLServerE
                 getNextResult();
             }
             catch (SQLException e) {
-                if (retryBasedOnFailedReuseOfCachedHandle(e, attempt))
+                if (retryBasedOnFailedReuseOfCachedHandle(e, attempt) && connection.isStatementPoolingEnabled())
```
TobiasSQL: Suggest changing `retryBasedOnFailedReuseOfCachedHandle` to take care of the `connection.IsStatementPoolingEnabled()` checking, that is part of whether or not to reuse.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -565,13 +566,18 @@ else if (EXECUTE_UPDATE == executeMethod && null != resultSet) {
     
     /** Should the execution be retried because the re-used cached handle could not be re-used due to server side state changes? */
     private boolean retryBasedOnFailedReuseOfCachedHandle(SQLException e,
```
TobiasSQL: Posting here as well as the method changed. Suggest changing retryBasedOnFailedReuseOfCachedHandle to take care of the connection.IsStatementPoolingEnabled() checking, that is part of whether or not to reuse.

v-afrafi: done. Thanks!

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
v-afrafi: done.

PR #621, opened 2018-02-05
Added additional error information

rene-ye:
Added additional error information to debug a random mismatch error which can't be reproduced locally.


rene-ye:
Closing the PR as the changes are not coming from the right branch and are mixed.

PR #627, opened 2018-02-15
Update changelog and version number for RTW 6.4 release (JRE 7 branch)

cheenamalhotra:


README.md
```diff
@@ -120,14 +120,14 @@ Projects that require either of the two features need to explicitly declare the
 <dependency>
 	<groupId>com.microsoft.sqlserver</groupId>
 	<artifactId>mssql-jdbc</artifactId>
-	<version>6.3.6.jre8-preview</version>
```
ulvii: DO you think this should be updated to jre7 since this is JRE7 branch?

README.md
```diff
@@ -80,7 +80,7 @@ We're now on the Maven Central Repository. Add the following to your POM file to
 <dependency>
 	<groupId>com.microsoft.sqlserver</groupId>
 	<artifactId>mssql-jdbc</artifactId>
-	<version>6.2.2.jre8</version>
+	<version>6.4.0.jre8</version>
```
ulvii: jre7 maybe?

README.md
cheenamalhotra: We shall keep it jre8 since this is just the same state as old dev branch used to be. We never used jre7 as we generally use latest supported versions in current branch in readme.

README.md
cheenamalhotra: Same as above.

PR #629, opened 2018-02-15
Added needsPrepare Check and removed error code 99586 check

cheenamalhotra:


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -69,6 +69,7 @@
     /** True if this execute has been called for this statement at least once */
     private boolean isExecutedAtLeastOnce = false;
 
+    private boolean needsPrepare = false;
```
TobiasSQL: Why do we need `needsPrepare` as a instance field? Should always be local variable.

cheenamalhotra: Fixed.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -529,6 +528,7 @@ final void doExecutePreparedStatement(PrepStmtExecCmd command) throws SQLServerE
         }
 
         String dbName = connection.getSCatalog();
+        boolean needsPrepare = false;
```
TobiasSQL: Let's make the default `true` to be safe (better prepare than not). Should not matter but just to be a bit defensive...

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
TobiasSQL: Great!

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
TobiasSQL: Otherwise things look good :-)

cheenamalhotra: Modified to true.

Issue #624, opened 2018-02-13
Attempting to insert an empty string into an encrypted column throws a SQLServerException


jaymajors:
## Driver version or jar name
Tried both mssql-jdbc-6.2.2.jre8.jar & mssql-jdbc-6.3.6.jre8-preview.jar

## SQL Server version
SQL Server 2016

## Client operating system
Test on Windows 10 & Windows Server 2016

## Java/JVM version
Oracle jdk 1.8.0

## Table schema

CREATE TABLE [dbo].[Person](
	[first_name] [varchar](50) COLLATE Latin1_General_BIN2 ENCRYPTED WITH (COLUMN_ENCRYPTION_KEY = [CEK_Auto1], ENCRYPTION_TYPE = Deterministic, ALGORITHM = 'AEAD_AES_256_CBC_HMAC_SHA_256') NULL,
	[last_name] [varchar](50) NULL,
	[middle_name] [varchar](50) NULL
) ON [PRIMARY]
GO

## Problem description

Trying to insert an empty string '' into an encrypted column causes the following exception to be thrown.

```
com.microsoft.sqlserver.jdbc.SQLServerException: The incoming tabular data stream (TDS) remote procedure call (RPC) protocol stream is incorrect. Parameter 3 (""): Data type 0xA7 has an invalid data length or metadata length.
	at com.microsoft.sqlserver.jdbc.SQLServerException.makeFromDatabaseError(SQLServerException.java:259)
	at com.microsoft.sqlserver.jdbc.SQLServerStatement.getNextResult(SQLServerStatement.java:1566)
	at com.microsoft.sqlserver.jdbc.SQLServerPreparedStatement.doExecutePreparedStatement(SQLServerPreparedStatement.java:554)
	at com.microsoft.sqlserver.jdbc.SQLServerPreparedStatement$PrepStmtExecCmd.doExecute(SQLServerPreparedStatement.java:488)
	at com.microsoft.sqlserver.jdbc.TDSCommand.execute(IOBuffer.java:7352)
	at com.microsoft.sqlserver.jdbc.SQLServerConnection.executeCommand(SQLServerConnection.java:2724)
	at com.microsoft.sqlserver.jdbc.SQLServerStatement.executeCommand(SQLServerStatement.java:224)
	at com.microsoft.sqlserver.jdbc.SQLServerStatement.executeStatement(SQLServerStatement.java:204)
[...]

PR #632, opened 2018-02-21
Fix for issue when attempting to insert an empty or null value on encrypted column

ulvii:
Workaround for #624.
Currently, when inserting empty string or `null` into encrypted variable-length columns, the driver sets the data length to `0` and the type to `type(0)` (`varchar(0)` for example). The workaround is to set the length to 1 and the type to type(1).


peterbae:
The workaround looks fine. After looking around the code, it seems like the valueLength is 0 if the user is using a callable statement and using the parameter as an output parameter. Have you considered this case / looked into what giving n/varchar(1) to the server implies?

peterbae:
I think it should be fine since I just tried the same scenario with output parameter, and it seems to work fine, but I wanted you to consider this scenario as well.

ulvii:
I tested the PR with both PreparedStatement and CallableStatement.

Issue #607, opened 2018-01-22
Kerberos constrained delegation and connections pool. 


krlm:
## Driver version or jar name
6.2.2-jre8

## SQL Server version
2016

## Client operating system
Linux, Ubuntu 18.04 (same problem on 16.10)

## Java/JVM version
Java(TM) SE Runtime Environment (build 1.8.0_162-ea-b03), Oracle

## Table schema
Not relevant, using `SELECT SYSTEM_USER` query

## Problem description
I'm trying to use mssql-jdbc in a Spring boot application with Kerberos authentication. It works fine when I'm using unconstrained mode, (I guess it's because tickets are renewable then) but when I switch to constrained delegation mode then only first iteration of [this](https://github.com/apache/tomcat/blob/JDBC_POOL_1_0_8_4/java/org/apache/tomcat/jdbc/pool/ConnectionPool.java#L427) loop is passing, next iteration is throwing following exception: 

```java
java.lang.IllegalStateException: This credential is no longer valid
    at sun.security.jgss.GSSCredentialImpl.getElement(GSSCredentialImpl.java:549) ~[na:1.8.0_162-ea]
    at sun.security.jgss.GSSContextImpl.initSecContext(GSSContextImpl.java:200) ~[na:1.8.0_162-ea]
    at sun.security.jgss.GSSContextImpl.initSecContext(GSSContextImpl.java:179) ~[na:1.8.0_162-ea]
    at com.microsoft.sqlserver.jdbc.KerbAuthentication.intAuthHandShake(KerbAuthentication.java:163) ~[mssql-jdbc-6.2.2.jre8.jar:na]
    at com.microsoft.sqlserver.jdbc.KerbAuthentication.GenerateClientContext(KerbAuthentication.java:401) ~[mssql-jdbc-6.2.2.jre8.jar:na]
    at com.microsoft.sqlserver.jdbc.SQLServerConnection.sendLogon(SQLServerConnection.java:4049) ~[mssql-jdbc-6.2.2.jre8.jar:na]
    at   
    ...
    at com.microsoft.sqlserver.jdbc.SQLServerDriver.connect(SQLServerDriver.java:569) ~[mssql-jdbc-6.2.2.jre8.jar:na]
    at org.apache.tomcat.jdbc.pool.PooledConnection.connectUsingDriver(PooledConnection.java:310) ~[tomcat-jdbc-8.5.23.jar:na]
[...]

PR #636, opened 2018-02-23
Kerberos Constrained Delegation Impersonated Credential Expiry fix

rene-ye:
no longer discards the impersonated credential after every request. Now checks that it times out and allows the impersonator to request multiple times. Addresses issue #607.

Also updated Maven Felix plugin from 3.4.0 to 3.5.0. The host for 3.4.0 seems unstable when running with Appveyor.


src/main/java/com/microsoft/sqlserver/jdbc/KerbAuthentication.java
```diff
@@ -390,6 +391,7 @@ private String findRealmFromHostname(RealmValidator realmValidator,
             int port,
             GSSCredential ImpersonatedUserCred) throws SQLServerException {
         this(con, address, port);
+        isUserCreatedCredential = true;
```
ulvii: Can we pass this flag to the constructor from `logon()`? 
` if (null != ImpersonatedUserCred)`

rene-ye: changes pushed

PR #648, opened 2018-03-08
Fix for Prepared Statement handle not found intermittent exception

cheenamalhotra:
With current implementation of the driver, if a prepared statement is executed multiple times which changes in type definitions, e.g, calling setBigDecimal() with different precision values, results in creating new handles for every changed precision type def. This results in

- Increased creation of handles, as handles created for old type definitions are discarded, reuse of already created handles (with old type def) does not happen.
- Intermittent failures with exception `Could not find prepared statement with handle x`. This specifically happens when the handle `x` created (with new type defs) is the same number as the one just discarded (with old type defs). The database instead of creating a new handle looks for previous one.

This fix does not discard existing handle in above scenario (solves both issues), but instead calls resetPrepStmtHandle() when new type definitions are found to reset handle for current query and expect new handle from database.


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -948,18 +948,14 @@ private boolean reuseCachedHandle(boolean hasNewTypeDefinitions, boolean discard
 		
 		// New type definitions and existing cached handle reference then deregister cached handle.
 		if(hasNewTypeDefinitions) {
-			if (null != cachedPreparedStatementHandle && hasPreparedStatementHandle() && prepStmtHandle == cachedPreparedStatementHandle.getHandle()) {
-				cachedPreparedStatementHandle.removeReference();
-	            cachedPreparedStatementHandle.setIsExplicitlyDiscarded();
-			}
+			resetPrepStmtHandle();
```
TobiasSQL: We do need to keep the previous code that does `cachedPreparedStatementHandle.removeReference`, in fact maybe code should be refactored so that call does `resetPrepStmtHandle`. If we don't remove the reference it will leak. It is clear that the main problem with the previous code was the call to `cachedPreparedStatementHandle.setIsExplicitlyDiscarded`, that should not be called here.

cheenamalhotra: I added call to `removeReference` back. This needs to be here. Thanks for catching that!

Also I saw another random issue with executeBatch() calls, that failed with `Could not find prepared statement with handle x`. We were able to reproduce this issue by load testing Statement tests and saw same pattern in all exceptions. After long investigations, I came up with a fix in my recent push. 

I made a change to `retryBasedOnFailedReuseOfCachedHandle` by passing `isBatch` parameter, which does not throw exception if a batch is being processed. There are intermittent failures if we do not add this, since if a handle is fetched from cache, and `needsPrepare=true` as a result of type def change, it throws exception while executing batch.

Let me know your thoughts on this. We have not seen any failure with this fix after running tests in multi-threaded environments with load testing.

TobiasSQL:
@cheenamalhotra , I took a look at this. Have two questions and a suggestion.

**Question 1:** I am missing something on the `isBatch` addition you made. If `needsPrepare` is `true` we never want to re-try given that the error cannot be due to missing handle, no? What am I missing here? 
 
**Question 2:** Why are we adding `dbName` to the source for the statement handle hash? This seems closely akin to trying to try to interpret the SET-options. Are we sure that `dbName` will always have an impact here, and that it can be trusted to be accurate?

On the suggestion side I think we can simplify the remove handle ref. logic. This isn't something new in your pull request but given we are starting to have more and more checks on `cachedPreparedStatementHandle` with regards to clean-up I thought it may be a reasonable thing to do. Basically my suggestion is to make `PreparedStatementHandle.removeReference` private given it is all handled by `PreparedStatementHandle.returnCachedPreparedStatementHandle` and then on the `SQLServerPreparedStatement` side of the house move all clean-up logic into `resetPrepStmtHandle`. 

Something along these lines:
```java
    private boolean resetPrepStmtHandle(boolean discardCurrentCacheItem) {

        PreparedStatementHandle handle = cachedPreparedStatementHandle;

        boolean statementPoolingUsed = null != handle; 

        // Return to pool and decrement reference count
        if (statementPoolingUsed) {
            // Make sure the cached handle does not get re-used more.
            if(discardCurrentCacheItem)
                handle.setIsExplicitlyDiscarded();
    
            connection.returnCachedPreparedStatementHandle(handle);
        }

        // Remove references
        cachedPreparedStatementHandle = null; 			
        prepStmtHandle = 0;

        return statementPoolingUsed;
    }
```

Now in `closePreparedHandle` we will simply do this for statement pooling:
```java
           // Handle unprepare actions through statement pooling.
            if (resetPrepStmtHandle(false)) {
                // Done
            }
            // If no reference to a statement pool cache item is found handle unprepare actions through batching @ connection level. 
            else ...
```

And in `reuseCachedHandle` we can do:
```java
// New type definitions and existing cached handle reference then deregister cached handle.
// If current cache item should be discarded make sure it is not used again.
if(hasNewTypeDefinitions || discardCurrentCacheItem) {
        resetPrepStmtHandle(discardCurrentCacheItem);
        if(discardCurrentCacheItem)
                return false;
}
		
// Check for new cache reference.
```

Thoughts?

TobiasSQL:
Ping?

cheenamalhotra:
Hi Tobias, I was away last week and looking into the changes now. Will update you soon!

cheenamalhotra:
Hi @TobiasSQL 

Sorry it took a while to get this working. Answers to your questions:
- **Question 1**: Allowing retry if batch is processed by `isBatch` check.
So the reason why I added this check was because we were getting some exceptions when batches were being processed with caching enabled. As you know when we discussed #610 the reason why we wanted to throw exception if needsPrepare is true is in [this](https://gist.github.com/AfsanehR/53f3d6bcab785694162a2016e157682d) test case where `exec sp_execute` or `RAISEERROR` is called inside SQL query that should go for retry of handle creation. Except that if a batch call is sent internally, with a prepared handle to statement, during executeBatch call, that still qualifies for retrying handle creation. The use case that fails is very complex but I can say allowing this retry does fix all scenarios where exceptions are thrown in batch processing. I also tested different possibilities where this would fail, but till now I haven't discovered anything like that. Please let me know if you find a scenario that would fail with this change.
- **Question 2**: This was added to address exception scenarios, by limiting handle sharing between different databases but since we have a better fix of not discarding handles, I tested with removed 'dbName' and everything looks good.

I did push changes as suggested by you recently, I could not incorporate the exact method as you mentioned above due to different conditions where `resetPrepStmtHandle` was called, but most of it I tried to taken in.

Request you to review the PR changes once more, and share your thoughts.

cheenamalhotra:
@TobiasSQL Please review changes if you find some time!

Thanks, Cheena.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
```diff
@@ -267,11 +267,11 @@ private CallableStatement getCallableStatementHandle(CallableHandles request,
             hassoc = new HandleAssociation(catalog, CS);
             HandleAssociation previous = handleMap.put(request, hassoc);
             if (null != previous) {
-                ((SQLServerCallableStatement) previous.stmt).handleDBName = previous.databaseName;
+                //((SQLServerCallableStatement) previous.stmt).handleDBName = previous.databaseName;
```
AfsanehR: let's remove these commented lines.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -934,38 +935,28 @@ private void getParameterEncryptionMetadata(Parameter[] params) throws SQLServer
...
-			if (null != cachedPreparedStatementHandle && hasPreparedStatementHandle() && prepStmtHandle == cachedPreparedStatementHandle.getHandle()) {
+		// If current cache items needs to be discarded or New type definitions found with existing cached handle reference then deregister cached handle.
+		if (discardCurrentCacheItem || hasNewTypeDefinitions) {
+			if (null != cachedPreparedStatementHandle && (discardCurrentCacheItem || (hasPreparedStatementHandle() && prepStmtHandle == cachedPreparedStatementHandle.getHandle()))) {
 				cachedPreparedStatementHandle.removeReference();
-	            cachedPreparedStatementHandle.setIsExplicitlyDiscarded();
 			}
-			cachedPreparedStatementHandle = null; 			
+            
+            // Make sure the cached handle does not get re-used more if it should be discarded
```
ulvii: Formatting issues. Please apply the formatter to the modified/added lines.

Issue #655, opened 2018-03-12
geometry parameters reported as having java.sql.Types = -157 (and -158 for geography) on SQL Server 2012 and 2014


jpoernomoatsafe:
## Driver version or jar name
mssql-jdbc-6.5.0.jre8.jar (preview)

## SQL Server version
SQL Server 2012 and SQL Server 2014

## Client operating system
Win 10 x64

## Java/JVM version
java version "1.8.0_144"

## Table schema
CREATE TABLE [jpoetest] ([_creation_instance] varchar(200), [GEOM] geometry)

## Problem description
Working with SQL Server 2012 and 2014, when trying to insert to a geometry column, parameter type for geometry columns are returned as `-157`. (geography columns are `-158`)

```
Connection connection = getConnection();
PreparedStatement statement = connection.prepareStatement("INSERT INTO [jpoetest] ([_creation_instance], [GEOM]) VALUES (?, ?)");
ParameterMetaData paramMetaData = statement.getParameterMetaData();
int sqlType = paramMetaData.getParameterType(2);
```

When talking to SQL Server 2008, `paramMetaData.getParameterType()` would return `java.sql.Types.VARBINARY`

I haven't tried SQL Server 2016 and 2017. 

I've searched for which `sql.Types` would `-157` map to, but it doesn't seem to be a valid `sql.Types` value.

Interestingly, SQL Server 2008, 2012, and 2014 all return `paramMetaData.getParameterTypeName()` correctly. ie. "geometry" or "geography"

## Expected behavior and actual behavior
expected: Geometry and Geography parameters should be reported as `sql.Types.VARBINARY`, consistent with `ResultSetMetadata.getColumnType()` for these columns.
actual: Geometry parameters are being returned as type `-157`, and Geography as type `-158`

## Repro code
```
Connection connection = getConnection();
PreparedStatement statement = connection.prepareStatement("INSERT INTO [jpoetest] ([_creation_instance], [GEOM]) VALUES (?, ?)");
ParameterMetaData paramMetaData = statement.getParameterMetaData();
int sqlType = paramMetaData.getParameterType(2);
```

peterbae:
[...]

PR #657, opened 2018-03-13
655 Fix to return correct column type for spatial datatypes

peterbae:
Fixes issue #655. Also added test cases for it.

Also removed an unused import from Geometry/Geography classes.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResultSetMetaData.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResultSetMetaData.java
@@ -121,11 +121,12 @@ public final class SQLServerResultSetMetaData implements java.sql.ResultSetMetaD
         }
         
         JDBCType jdbcType = typeInfo.getSSType().getJDBCType();
+        SSType sqlType = typeInfo.getSSType();
         // in bulkcopy for instance, we need to return the real jdbc type which is sql variant and not the default Char one. 
-        if ( SSType.SQL_VARIANT == typeInfo.getSSType()){
+        if ( SSType.SQL_VARIANT == sqlType){
             jdbcType = JDBCType.SQL_VARIANT;
         }
-        if (SSType.UDT == typeInfo.getSSType()) {
+        if (SSType.UDT == sqlType) {
             if (typeInfo.getSSTypeName().equalsIgnoreCase("geometry")) {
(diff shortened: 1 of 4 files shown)


peterbae:
this appveyor failure is unrelated to this PR, it's a bug I found during internal testing that occurs only with SQL Server 2008. I will merge this PR after that issue has been solved.

peterbae:
Actually I'm going to use a workaround for this problem, it seems like it could take a while to solve.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResultSetMetaData.java
```diff
@@ -125,6 +125,14 @@ public int getColumnType(int column) throws SQLServerException {
         if ( SSType.SQL_VARIANT == typeInfo.getSSType()){
             jdbcType = JDBCType.SQL_VARIANT;
         }
+        if (SSType.UDT == typeInfo.getSSType()) {
+            if (typeInfo.getSSTypeName().equalsIgnoreCase("geometry")) {
```
rene-ye: I don't really have a problem with this, but why are we doing type check in 3 different ways. VARIANT is being checked separately outside the switch, and now GEOMETRY/GEOGRAPHY will be checked by name. Can we maybe stick to one form? This Microsoft [overview](https://docs.microsoft.com/en-us/sql/relational-databases/spatial/spatial-data-types-overview) of spatial datatypes seems to imply that it's only available in SQL Server 2012 and later, so shouldn't this be in the isKatamaiOrLater() block?

peterbae: We can't stick to one form. If you debug this part of the code, you'll see that the getSSType will give SQL_VARIANT or UDT. If it's UDT, we need to look at the SSTypeName to find out if it's Geometry or Geography, so this is the only way to check for these datatypes.

The overview page mentions that new spatial features have been introduced in SQL Server 2012, but this does not mean that spatial datatypes are supported starting from 2012 - it's supported starting from 2008. so we need this check for 2008 and up.

cheenamalhotra: Please create Constants for these strings and reference them here.

Issue #645, opened 2018-03-07
Server-side CPU Affinity problem caused by driver


greg-pendlebury:
## Driver version or jar name
6.1.5 or greater. We have only tested up to 6.2.2.1 though

## SQL Server version
SQL Server 2012

## Client operating system
Windows Server 2012 R2

## Java/JVM version
1.8.0_u152 Hotspot

## Problem description
It looks like https://github.com/Microsoft/mssql-jdbc/pull/160 (introduced in 6.1.5 onwards) introduces some highly undesirable behaviour regarding CPU affinity on servers with two NUMA nodes. That change introduces two connections being opened in sequence and one of them is then closed. The server side of the connection is pinned to a NUMA node in a round robin fashion, but because every second connection is closed it results in a horrible resource allocation imbalance on the remaining connections.

Our Java application (JBoss EAP 7) opens 300 connections at maximum and maintains them under heavy use in a connection pool. We have observed all connections allocated to a single node, or nearly all... sometimes a random connection when the pool is scaling up (like a DBA trying to work out what the hell is going on) will alter the pattern slightly and all new connections then flip to the other node.

The server is then essentially restrained to operate at half capacity because half of the CPUs are idle.

We have reproduced in JMeter as well, using different driver versions and 'multiSubnetFailover=true'. This problem does not occur on versions 6.1.4 and below. Our assumption is https://github.com/Microsoft/mssql-jdbc/pull/160 is the cause, but perhaps that is incorrect.

## Expected behavior and actual behavior
The connections should have been distributed evenly across NUMA nodes... they were instead mostly on a single node. Perfmon shows us that when people are not actively investigating by establishing their own connections it will perfectly pin all 300 connections to a single NUMA node.

## Repro code
 1) Use 'multiSubnetFailover=true' in your JDBC connection string.
 2) Connect to a server with two NUMA nodes.
 3) Maintain all open connections for the duration of the test.
 4) Run some needlessly expensive queries to exercise the CPUs.
 5) Observe half of the CPUs on the machine saturate and half remain idle.
 6) DBAs monitor affinity on the server like so:
```
--Displays connections per NUMA node for 'Microsoft JDBC Driver 6.1' client interfaces
select c.node_affinity, s.client_interface_name,  count(*) as connections
from sys.dm_exec_connections c
join sys.dm_exec_sessions s on s.session_id = c.session_id
where s.client_interface_name = 'Microsoft JDBC Driver 6.1'
[...]

PR #662, opened 2018-03-20
Use Socket instead of SocketChannel when multiSubnetFailover=true

ulvii:
Fix for both #144 and #645 .This PR reverts #160.

Switching from SocketChannels to Sockets to avoid creating 2 connections when multiSubnetFailover=true.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
@@ -2337,8 +2337,8 @@ final class SocketFinder {
                 findSocketUsingJavaNIO(inetAddrs, portNumber, timeoutInMilliSeconds);
             }
             else {
-                LinkedList<Inet4Address> inet4Addrs = new LinkedList<>();
-                LinkedList<Inet6Address> inet6Addrs = new LinkedList<>();
+                LinkedList<InetAddress> inet4Addrs = new LinkedList<>();
+                LinkedList<InetAddress> inet6Addrs = new LinkedList<>();
 
                 for (InetAddress inetAddr : inetAddrs) {
                     if (inetAddr instanceof Inet4Address) {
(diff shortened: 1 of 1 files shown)


ulvii:
Attaching the jars: [PR662_jars.zip](https://github.com/Microsoft/mssql-jdbc/files/1831182/6.5.1-SNAPSHOT_jars.zip)

greg-pendlebury:
Thanks @ulvii, one of our DBAs tested the jar with JMeter using the same test we used in demonstrating #645. He has confirmed that the NUMA affinity issues are resolved. We will try updating our full application once a release of the driver is available, but the regression testing of all the other changes since 6.1.7 might take us some time.

sehrope:
Sure we'll get this tested. Are the jars you've attached built off the tip of this PR? i.e. can I just build the PR?

I took a peek at the diff itself and it's quite minimal. The main change seems to be replacing `findSocketUsingJavaNIO(...)` with `findSocketUsingThreading(...)` which has existed at least since the open sourcing of the driver but doesn't seem to have been referenced in any active code.

ulvii:
Thanks @greg-pendlebury , @sehrope .

@sehrope , yes the jars are built on top of the PR.

sehrope:
We tested this a bit and it doesn't break query cancellation when multiSubnetFailover is enabled.

I think it'd be worthwhile to eventually rewrite how `SocketFinder` is implemented. The concept itself is fine, i.e use a thread pool to race opening sockets to all resolved addresses. The complexity is that it's done using manual locking, `.wait(...)`, `.notify()` and the code is intermixed with the parent class. It likely predates modern concurrency primitives and redoing it using something like Future or CountDownLatch would simplify things.

The current code also handles IPv4 and IPv6 sequentially. I don't see that behavior explicitly documented anywhere nor does it seem like it'd be useful. If a user has multiple address types resolving from a host name they should all be attempted in parallel right? I took a stab at a high level clean up for this. See here for a PR: #663

PR #663, opened 2018-03-21
Clean up SocketFinder.findSocket(...) handling

sehrope:
Cleans up `SocketFinder.findSocket(...)` a bit and simplifies the logic for the simple case of a single address being resolved.

This changes some of the prior functionality though I wouldn't consider it a breaking change as it's not documented functionality and the new approach seems (to me) to be more correct. Previously if a host resolved to both IPv4 and IPv6 addresses, each set would get half the time out and the IPv4 addresses would be attempted first. This PR changes the logic to handle all IP addresses concurrently using the full timeout.

It also centralizes the logic for handling the case where there's a single address. Regardless of JVM type it now attempts to directly connect to it. This changes the behavior for the IBM JDK as previously it would have attempted the NIO approach followed by a normal socket creation upon success. Again this seems like a better choice as the overall logic is simpler and end result is the same.

This PR was coded atop #662. See there for more details on the parent change.


ulvii:
Hi @sehrope ,
Thanks for the contribution. We decided to merge #662 for now, will review and test your changes for one of the upcoming releases.

ulvii:
Hi @sehrope ,

Would you mind resolving the conflicts?

sehrope:
@ulvii Done.

sehrope:
One more rebase as dev has been updated since the last one.

Issue #661, opened 2018-03-19
Exception using setMaxRows() followed by query execution when showplan_text is on


tseylerd:
## Driver version or jar name
6.0.8112.100

## SQL Server version
14.0 (Linux)

## Client operating system
macOS

## Java/JVM version
1.8 OpenJDK

## Problem description
When the SHOWPLAN_TEXT is on, driver may fail on statement execution. Especially, using setMaxRows() on statement before execution. The problem is that before querying driver sends SET ROWCOUNT to the database and expects nothing as a results. But showplan_text is on and database will return result set causing driver failure with:
```
com.microsoft.sqlserver.jdbc.SQLServerException: The TDS protocol stream is not valid. Unexpected token TDS_COLMETADATA (0x81).
	at com.microsoft.sqlserver.jdbc.SQLServerConnection.terminate(SQLServerConnection.java:2400)
	at com.microsoft.sqlserver.jdbc.SQLServerConnection.terminate(SQLServerConnection.java:2384)
	at com.microsoft.sqlserver.jdbc.SQLServerConnection.throwInvalidTDSToken(SQLServerConnection.java:2373)
	at com.microsoft.sqlserver.jdbc.TDSReader.throwInvalidTDSToken(IOBuffer.java:6558)
	at com.microsoft.sqlserver.jdbc.TDSParser.throwUnexpectedTokenException(tdsparser.java:119)
	at com.microsoft.sqlserver.jdbc.TDSTokenHandler.onColMetaData(tdsparser.java:247)
	at com.microsoft.sqlserver.jdbc.TDSParser.parse(tdsparser.java:87)
	at com.microsoft.sqlserver.jdbc.TDSParser.parse(tdsparser.java:36)
	at com.microsoft.sqlserver.jdbc.SQLServerConnection$1ConnectionCommand.doExecute(SQLServerConnection.java:2495)
	at com.microsoft.sqlserver.jdbc.TDSCommand.execute(IOBuffer.java:7505)
	at com.microsoft.sqlserver.jdbc.SQLServerConnection.executeCommand(SQLServerConnection.java:2445)
	at com.microsoft.sqlserver.jdbc.SQLServerConnection.connectionCommand(SQLServerConnection.java:2500)
	at com.microsoft.sqlserver.jdbc.SQLServerConnection.setMaxRows(SQLServerConnection.java:519)
[...]

PR #666, opened 2018-03-26
Fix for the issue when using setMaxRows() with SHOWPLAN ON

ulvii:
#661.

The fix is to Ignore the column metadata in  `connectionCommand()`, because execution plan information is returned when SHOWPLAN is ON.


ulvii:
Attaching the jars that are built on top of the PR. 
[pr666_jars.zip](https://github.com/Microsoft/mssql-jdbc/files/1849885/pr666_jars.zip)

PR #675, opened 2018-04-12
Adding Old Constructor back to AKV Provider

cheenamalhotra:
This PR adds back the old Constructor for SQLServerColumnEncryptionAzureKeyVaultProvider as below:

`SQLServerColumnEncryptionAzureKeyVaultProvider akvProvider = new SQLServerColumnEncryptionAzureKeyVaultProvider(authenticationCallback, service);`

This constructor was removed in PR #397 due to changed HttpClient implementation in AKV library version 1.0.0. With this PR we add backwards compatibility to applications using old AKV libraries with 6.0 and 6.2.1 driver versions and will now be able to upgrade without making any changes to code.


src/main/java/com/microsoft/sqlserver/jdbc/KeyVaultCredential.java
```diff
@@ -24,20 +24,30 @@
  */
 class KeyVaultCredential extends KeyVaultCredentials {
 
+	SQLServerKeyVaultAuthenticationCallback authenticationCallback = null;
```
ulvii: Indentation.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerColumnEncryptionAzureKeyVaultProvider.java
```diff
@@ -62,7 +71,35 @@ public void setName(String name) {
...
+     * 
+     * @param authenticationCallback
+     *            - Callback function used for authenticating to AAD.
+     * @param executorService
+     *            - The ExecutorService used to create the keyVaultClient
+     * @throws SQLServerException
+     *             when an error occurs
+     */
+    public SQLServerColumnEncryptionAzureKeyVaultProvider(SQLServerKeyVaultAuthenticationCallback authenticationCallback,
+            ExecutorService executorService) throws SQLServerException {
```
ulvii: Where is `executorService` used?

src/main/java/com/microsoft/sqlserver/jdbc/KeyVaultCredential.java
```diff
@@ -24,20 +24,30 @@
...
+    KeyVaultCredential(SQLServerKeyVaultAuthenticationCallback authenticationCallback) {
+        this.authenticationCallback = authenticationCallback;
+    }
+    
     public String doAuthenticate(String authorization,
             String resource,
             String scope) {
-        AuthenticationResult token = getAccessTokenFromClientCredentials(authorization, resource, clientId, clientKey);
-        return token.getAccessToken();
+    	if(authenticationCallback==null) {
```
ulvii: Maybe we could try to avoid multiple returns?

https://github.com/Microsoft/mssql-jdbc/blob/dev/Coding_Guidelines.md

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerColumnEncryptionAzureKeyVaultProvider.java
cheenamalhotra: executorService is not consumed here because the KeyVaultClientImpl now manages its own KeyVaultClientService as [here](https://github.com/Azure/azure-keyvault-java/blob/dev/azure-keyvault/src/main/java/com/microsoft/azure/keyvault/KeyVaultClientImpl.java#L246). These services are used by Retrofit to make Rest Calls.

The only reason of adding this here was to retain the old constructor params used by current applications in industry to allow no change upgrade of JDBC Driver version, but can be removed as well.

ulvii: I think this could lead to confusions. Let's make sure we discuss this with Jakub.

src/main/java/com/microsoft/sqlserver/jdbc/KeyVaultCredential.java
```diff
@@ -24,20 +24,33 @@
  */
 class KeyVaultCredential extends KeyVaultCredentials {
 
-    String clientId = null;
-    String clientKey = null;
+    SQLServerKeyVaultAuthenticationCallback authenticationCallback = null;
+    String                                  clientId               = null;
+    String                                  clientKey              = null;
+    String                                  accessToken            = null;
```
peterbae: formatting here?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerColumnEncryptionAzureKeyVaultProvider.java
```diff
@@ -64,9 +73,58 @@ public String getName() {
...
+        if (null == authenticationCallback) {
+            MessageFormat form = new MessageFormat(SQLServerException.getErrString("R_NullValue"));
+            Object[] msgArgs1 = {"SQLServerKeyVaultAuthenticationCallback"};
+            throw new SQLServerException(form.format(msgArgs1), null);
+        }
+        credentials = new KeyVaultCredential(authenticationCallback);
+        RestClient restClient = new RestClient.Builder(new OkHttpClient.Builder(), new Retrofit.Builder()).withBaseUrl(baseUrl)
+                .withCredentials(credentials).withSerializerAdapter(new AzureJacksonAdapter())
+                .withResponseBuilderFactory(new AzureResponseBuilder.Factory()).build();
+        keyVaultClient = new KeyVaultClient(restClient);
```
peterbae: If both SQLServerColumnEncryptionAzureKeyVaultProvider constructors do exactly the same thing, maybe refactor the inner part so we don't duplicate code?

...actually, we would have to revert that in our next release when we remove the deprecated API. in that case, just have the SQLServerColumnEncryptionAzureKeyVaultProvider with ExecutorService in its parameter call the other SQLServerColumnEncryptionAzureKeyVaultProvider constructor.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerColumnEncryptionAzureKeyVaultProvider.java
```diff
@@ -64,9 +73,58 @@ public String getName() {
...
+     * @param authenticationCallback
+     *            - Callback function used for authenticating to AAD.
+     * @param executorService
+     *            - The ExecutorService, previously used to create the keyVaultClient, but not in use anymore. - This param can be passed as 'null'
+     * @throws SQLServerException
+     *             when an error occurs
+     */
+    @Deprecated
+    public SQLServerColumnEncryptionAzureKeyVaultProvider(SQLServerKeyVaultAuthenticationCallback authenticationCallback,
+            ExecutorService executorService) throws SQLServerException {
```
rene-ye: Since we aren't doing anything with the executor, and the constructor is just here for backwards compatibility, shouldn't we make a call to `SQLServerColumnEncryptionAzureKeyVaultProvider(SQLServerKeyVaultAuthenticationCallback authenticationCallback)`? This might confuse people who are stepping through the code, thinking there is a very small change somewhere and that's why there are separate functions.

src/main/java/com/microsoft/sqlserver/jdbc/KeyVaultCredential.java
cheenamalhotra: That's how mssql-jdbc formatter works here.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerColumnEncryptionAzureKeyVaultProvider.java
cheenamalhotra: Fixed

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerColumnEncryptionAzureKeyVaultProvider.java
cheenamalhotra: Fixed

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerColumnEncryptionAzureKeyVaultProvider.java
```diff
@@ -64,9 +73,49 @@ public String getName() {
...
-     * Constructor that authenticates to AAD. This is used by KeyVaultClient at runtime to authenticate to Azure Key
+     * Constructor that takes a callback function to authenticate to AAD. This is used by KeyVaultClient at runtime to authenticate to Azure Key
      * Vault.
      * 
+     * This constructor is present to maintain backwards compatibility with 6.0 version of the driver. Deprecated for removal in next Stable release.
+     * 
+     * @param authenticationCallback
+     *            - Callback function used for authenticating to AAD.
+     * @param executorService
+     *            - The ExecutorService, previously used to create the keyVaultClient, but not in use anymore. - This param can be passed as 'null'
```
ulvii: "This parameter"

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerColumnEncryptionAzureKeyVaultProvider.java
```diff
@@ -64,9 +73,49 @@ public String getName() {
     }
 
     /**
-     * Constructor that authenticates to AAD. This is used by KeyVaultClient at runtime to authenticate to Azure Key
+     * Constructor that takes a callback function to authenticate to AAD. This is used by KeyVaultClient at runtime to authenticate to Azure Key
      * Vault.
      * 
+     * This constructor is present to maintain backwards compatibility with 6.0 version of the driver. Deprecated for removal in next Stable release.
```
ulvii: "stable release"

Issue #683, opened 2018-04-21
BigDecimal Bug


amirmf:
## Driver version or jar name
 mssql-jdbc-6.4.0.jre8.jar

## SQL Server version
mssql 2016 

## Client operating system
i used ubuntu and windows ....both same response

## Java/JVM version
java version "1.8.0",

## Table schema
default(dbo | public)

## Problem description
i've used BigDecimal() in allover my application and i cant change it to use manual scale method:
and when i want to insert BigDecimal() i got this erroe:
Caused by: com.microsoft.sqlserver.jdbc.SQLServerException: Error converting data type nvarchar to decimal.
	at com.microsoft.sqlserver.jdbc.SQLServerException.makeFromDatabaseError(SQLServerException.java:259)
	at com.microsoft.sqlserver.jdbc.SQLServerStatement.getNextResult(SQLServerStatement.java:1547)
	at com.microsoft.sqlserver.jdbc.SQLServerPreparedStatement.doExecutePreparedStatement(SQLServerPreparedStatement.java:548)
	at com.microsoft.sqlserver.jdbc.SQLServerPreparedStatement$PrepStmtExecCmd.doExecute(SQLServerPreparedStatement.java:479)
	at com.microsoft.sqlserver.jdbc.TDSCommand.execute(IOBuffer.java:7344)
	at com.microsoft.sqlserver.jdbc.SQLServerConnection.executeCommand(SQLServerConnection.java:2713)
	at com.microsoft.sqlserver.jdbc.SQLServerStatement.executeCommand(SQLServerStatement.java:224)
	at com.microsoft.sqlserver.jdbc.SQLServerStatement.executeStatement(SQLServerStatement.java:204)
	at com.microsoft.sqlserver.jdbc.SQLServerPreparedStatement.executeUpdate(SQLServerPreparedStatement.java:426)
	at bitronix.tm.resource.jdbc.proxy.PreparedStatementJavassistProxy.executeUpdate(PreparedStatementJavassistProxy.java)

## Expected behavior and actual behavior
As jdbc standard api this your driver should support BigDecimal()

cheenamalhotra:
Hi @amirmf 

I've created a fix for the issue reported above, request you to test and confirm if your application gets fixed with PR #684 . Attached Jars for testing as well.

[mssql-jdbc-PR684.zip](https://github.com/Microsoft/mssql-jdbc/files/1939965/mssql-jdbc-PR684.zip)

[...]

PR #684, opened 2018-04-23
Fixes issue with PreparedStatement.setBigDecimal() in the driver when no scale is passed

cheenamalhotra:
Fix for issue #683.

Issue occurs when BigDecimal value is created from double [`e.g. new BigDecimal(5.55)`] or as it is exceeds in Precision than maxAllowedPrecision (38) in SQL Server, and no scale is passed from Client. The driver in this case fails to round off digits and passes BigDecimal value as String, which when received by SQL Server fails with Exception:
`com.microsoft.sqlserver.jdbc.SQLServerException: Error converting data type nvarchar to decimal.`

This fix makes sure BigDecimal values passed to SQL Server are within precision limits and scaled accordingly.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/dtv.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/dtv.java
@@ -2212,8 +2212,8 @@ final class AppDTVImpl extends DTVImpl {
                 BigDecimal bigDecimalValue) throws SQLServerException {
             // Rescale the value if necessary
             if (null != bigDecimalValue) {
-            	int dtvScale, biScale = bigDecimalValue.scale();
-                if (null == dtv.getScale()) {
+            	Integer dtvScale, biScale = bigDecimalValue.scale();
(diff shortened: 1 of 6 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/dtv.java
```diff
@@ -2212,11 +2212,20 @@ void execute(DTV dtv,
                 BigDecimal bigDecimalValue) throws SQLServerException {
             // Rescale the value if necessary
             if (null != bigDecimalValue) {
-                Integer inScale = dtv.getScale();
-                if (null != inScale && inScale != bigDecimalValue.scale())
-                    bigDecimalValue = bigDecimalValue.setScale(inScale, RoundingMode.DOWN);
+            	Integer dtvScale, biScale = bigDecimalValue.scale();
```
ulvii: Please apply the formatter, otherwise looks good!

Issue #331, opened 2017-06-10
Support For Batch or Bulk Inserts to Azure SQL Data Warehouse (SQLDW)


rblevi01:
Microsoft support recommended this driver for connecting to Azure SQL Data Warehouse (not Azure SQL Database).  It works however inserts are very slow at 13 rows per second.  I happen to be using DataStage.  Same issue is reported by Talend users.  Is this driver designed to work with Azure SQL Data Warehouse?  If so, what configuration will cause it to use Batch or Bulk inserts?  Note that my same system and configuration when pointing to Azure SQL Database inserts 10,000 rows per second.  Looking at the data sent to Data Warehouse, it appears single row inserts are sent to Data Warehouse and a Batch is sent to Database.

v-nisidh:
Thanks @rblevi01, 

We are working on this. 

Meanwhile can you verify if [BulkCopy option](https://docs.microsoft.com/en-us/sql/connect/jdbc/using-bulk-copy-with-the-jdbc-driver) is suitable for your case?

ajlam:
@rblevi01 - out of curiosity, have you tried [PolyBase](https://docs.microsoft.com/en-us/sql/relational-databases/polybase/polybase-guide)? This is the recommended way of loading data into Azure SQL Data Warehouse.

ajlam:
@rblevi01, have you been able to try out PolyBase for data loading?

cheenamalhotra:
Closing the issue due to inactivity.

ikoblik:
@cheenamalhotra , @ajlam I noticed that this bug was closed due to inactivity, but was the originally reported issue fixed? Does MSSQL JDBC driver now support Batch updates when interacting with Data Warehouse. My team experiences similar issues, no mattar how many clients we cannot insert more than 10 records to DW.

cheenamalhotra:
Hi @ikoblik - We closed the issue as we were of the view using Polybase might have resolved customer's issue as there was no response from customer. 

I am reopening this thread as I do see many customers facing similar issue - we will investigate on providing support for better performance/Bulk Copy options.

ajilius:
The bulk copy API performs much better than 5-10 rows per second. I regularly load to ASDW using this JDBC driver API, and 40megabytes per second - row counts depend on data content, but will be in the hundreds of thousands - are achievable. Of course, the recommendation to use Polybase will improve this by 1-2 orders of magnitude.

peterbae:
[...]

Issue #531, opened 2017-10-20
SQLServerDataTable should not call toString on all objects


cogman:
## Problem description

Right now, SqlServerDataTable will take in a type, if that type is numeric, it will call toString on that type and then it will use whatever numeric type it is turning it into to parse the output string.

String/numeric parsing in java is horribly slow and uses quite a bit of memory.  Instead of doing that, DataTable should check "Is this an instance of Number" (which often, it will be), and then should use the various "xValue" methods off of number to turn it into the desired type.  toString and parsing should only be called if this method fails.

cogman:
Also, a bunch of these null checks and sets are superfluous.  Object arrays start off with null as the default value, per the language spec.  We should just do one null check at the beginning rather than sprinkling the null checks throughout all of the row code.

cheenamalhotra:
Hi @cogman Thanks for raising this issue. 

I created PR #538 to make desired changes in `SQLServerDataTable` > `internalAddRow()` function. Appreciate your review and feedback on the same.

cogman:
Pretty much exactly what I imagined it to be.  Thanks!

cheenamalhotra:
Closing issue as PR #990 is now merged.

PR #686, opened 2018-04-26
Use Bulk Copy API for batch insert operation

peterbae:
Improves the batch insert operation performace against Azure DW.

Fixes issue https://github.com/Microsoft/mssql-jdbc/issues/331.


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkBatchInsertRecord.java
```diff
@@ -0,0 +1,598 @@
...
+             * SQL Server supports numerous string literal formats for temporal types, hence sending them as varchar with approximate
+             * precision(length) needed to send supported string literals. string literal formats supported by temporal types are available in MSDN
+             * page on data types.
+             */
+            case java.sql.Types.DATE:
+            case java.sql.Types.TIME:
+            case java.sql.Types.TIMESTAMP:
+            case microsoft.sql.Types.DATETIMEOFFSET:
+                // The precision is just a number long enough to hold all types of temporal data, doesn't need to be exact precision.
+                columnMetadata.put(positionInTable, new ColumnMetadata(colName, jdbcType, 50, scale, dateTimeFormatter));
```
rene-ye: Can we make the "50", which I'm assuming is the precision, a const so changing it is easier when necessary.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -5829,6 +5831,39 @@ public void onEviction(Sha1HashKey key, PreparedStatementHandle handle) {
             }
         }
     }
+
+    boolean isAzureDW() throws SQLServerException, SQLException {
+        if (null == isAzureDW) {
+            try (Statement stmt = this.createStatement(); ResultSet rs = stmt.executeQuery("SELECT CAST(SERVERPROPERTY('EngineEdition') as INT)");)
```
rene-ye: We don't seem to catch/handle any exceptions from this try statement, is there a reason why we're using it here.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -5829,6 +5831,39 @@ public void onEviction(Sha1HashKey key, PreparedStatementHandle handle) {
...
+                // 1 = Personal or Desktop Engine (Not available for SQL Server.)
+                // 2 = Standard (This is returned for Standard and Workgroup.) 
+                // 3 = Enterprise (This is returned for Enterprise, Enterprise Evaluation, and Developer.)
+                // 4 = Express (This is returned for Express, Express with Advanced Services, and Windows Embedded SQL.)
+                // 5 = SQL Azure
+                // 6 = SQL Azure DW
+                // Base data type: int
+                final int ENGINE_EDITION_FOR_SQL_AZURE_DW = 6;
+                rs.next();
+                int engineEdition = rs.getInt(1);
```
rene-ye: Change block to
`isAzureDW = rs.getInt(1) == ENGINE_EDITION_FOR_SQL_AZURE_DW`
or even more concise but possibly a little confusing
`return (rs.getInt(1) == ENGINE_EDITION_FOR_SQL_AZURE_DW)`

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -2452,8 +2454,72 @@ public final void addBatch() throws SQLServerException {
...
+                    return updateCounts; 
+                }
+
+                String tableName = parseUserSQLForTableNameDW(false, false);
+                ArrayList<String> columnList = parseUserSQLForColumnListDW();
+                ArrayList<String> valueList = parseUserSQLForValueListDW(false);
+                
+                String destinationTableName = tableName;
+                // Get destination metadata
+                try (SQLServerResultSet rs = ((SQLServerStatement) connection.createStatement())
```
rene-ye: try with resources, but we don't handle any exceptions.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -2554,6 +2620,368 @@ public final void addBatch() throws SQLServerException {
...
+        if (!hasValuesBeenFound) {
+            // look for keyword "VALUES"
+            if (localUserSQL.substring(0, 6).equalsIgnoreCase("VALUES")) {
+                localUserSQL = localUserSQL.substring(6);
+                
+                localUserSQL = localUserSQL.trim();
+                
+                // ignore all comments
+                if (localUserSQL.substring(0, 2).equalsIgnoreCase("/*")) {
+                    int temp = localUserSQL.indexOf("*/") + 2;
```
rene-ye: This logic is done quite a lot. Can we maybe make a private function that takes in a string and returns what we need. So the final result for all these if blocks would be something like
`if(localUserSQL.substring(0, 2).equalsIgnoreCase("/*")) {
    return newPrivateFunction("*/");
}`

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkBatchInsertRecord.java
peterbae: The comment on line 217 explains why this is unnecessary.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
peterbae: Normally, when we create a new Statement or ResultSet object, it's good practice to close it. However, the try-with-resources syntax (which i've used here) allows those objects to get deleted automatically. You can read more on this page: https://blogs.oracle.com/weblogicserver/using-try-with-resources-with-jdbc-objects

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
peterbae: i think right now is fine, i merely adopted the same part of code from our fx framework anyways.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
peterbae: see my comment above

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerStatement.java
```diff
@@ -963,6 +963,28 @@ final void resetForReexecute() throws SQLServerException {
...
+        checkClosed();
+        // Used to check just the first letter which would cause
+        // "Set" commands to return true...
+        String temp = sql.trim();
+        if (temp.substring(0, 2).equalsIgnoreCase("/*")) {
+            int index = temp.indexOf("*/") + 2;
+            return isInsert(temp.substring(index));
+        }
+        char c = temp.charAt(0);
+        if (c != 'i' && c != 'I')
```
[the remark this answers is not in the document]
peterbae: If the user decides to put a use statement, it won't be flagged as an insert statement, and the old implementation for handling batch statements will get executed instead. This isInsert statement is currently only being used for queries that come through as an execute batch statement - we don't expect the user to run thousands of the same batch query with a use statement in front of it.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkBatchInsertRecord.java
rene-ye: What if we create a temporal datatype in the future that accommodates precision > 50? If it's an arbitrarily assigned number than it can be just as arbitrarily changed.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
rene-ye: We should at least remove
`if (boolean condition) { `
`    bool = true`
`} else {`
`   bool = false`
`}`

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -5866,6 +5868,33 @@ public void onEviction(Sha1HashKey key, PreparedStatementHandle handle) {
             }
         }
     }
+
+    boolean isAzureDW() throws SQLServerException, SQLException {
+        if (null == isAzureDW) {
+            try (Statement stmt = this.createStatement(); ResultSet rs = stmt.executeQuery("SELECT CAST(SERVERPROPERTY('EngineEdition') as INT)");)
```
v-mabarw: Is there any way to avoid an extra round trip to the server to get this info?  For example, can we get any useful info from the TDS pre-login or login packets?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -2554,6 +2627,319 @@ public final void addBatch() throws SQLServerException {
...
+                    return sb.toString();
+                }
+            } else {
+                sb.append(localUserSQL.charAt(0));
+                localUserSQL = localUserSQL.substring(1);
+            }
+        }
+        
+        // It shouldn't come here. If we did, something is wrong.
+        throw new IllegalArgumentException("localUserSQL");
```
v-mabarw: Why did you pick throwing an IllegalArgumentException (for this & the other parse routines)?  It's probably better to have a descriptive error message instead.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkBatchInsertRecord.java
v-suhame: Precision here should be the one passed by user. I'm assuming 50 is used as precision based on the BulkCopy CSV implementation, it makes sense in case of CSV, because temporal types can be stored in CSV using any of the supported string literal format (https://docs.microsoft.com/en-us/sql/t-sql/data-types/datetime-transact-sql?view=sql-server-2017#supported-string-literal-formats-for-datetime) and we pass it as varchar/nvarchar to Server.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
peterbae: I don't think we can avoid making a call to get the table metadata, but as it turns out, we were making an exact same FMTONLY call in SQLServerBulkCopy::getDestinationMetadata method (which is called as part of the bulk copy process shortly down the line). So instead of making the FMTONLY call to the database twice in the same fashion, the driver will now store the resultset from the first FMTONLY so it can be re-used in getDestinationMetadata. This means we won't be making any additional trips to the database compared to the way it was working before.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkBatchInsertRecord.java
```diff
@@ -0,0 +1,329 @@
...
+        return columnMetadata.get(column).scale;
+    }
+
+    @Override
+    public boolean isAutoIncrement(int column) {
+        return false;
+    }
+
+    @Override
+    public Object[] getRowData() throws SQLServerException {
```
rene-ye: There are a few things which can be done better here.

1. Currently loops twice for, this is actually unnecessary, we can do everything here in 1 loop.
2. Stores a lot of one time use information in arrays. Very performance unfriendly if our row/column count is huge.
3. Very very hard to read. Between the massive switch, parsing logic, and referencing 2 arrays and a set; it's difficult to figure out what it's doing.

Proposed Changes:

1. Move the massive switch statement to it's own function.
2. Move the parsing to its own function (this one might be difficult with how `valueIndex` is updated)
3. Change the iteration logic
3a. Create an Object[] to return
3b. Iterate over the set with foreach
3c. Create an Object to be inserted into our Object[]; use the Set as index
3d. Parse the string from valueList with this index
3e. Take the string from the above step and convert it to a desired value through the switch statement
3f. Store this value into our Object, and then add our Object into our Object[].

I've done a "prototype" implementation here: [gist link](https://gist.github.com/rene-ye/b9f0424e90aa6ff73f7c4261cf7162a0). Please take a look and adjust as neccessary.

src/test/java/com/microsoft/sqlserver/jdbc/preparedStatement/BatchExecutionWithBulkCopyParseTest.java
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    @BeforeEach
+    public void testSetup() throws TestAbortedException, Exception {
+        assumeTrue(13 <= new DBConnection(connectionString).getServerVersion(),
+                "Aborting test case as SQL Server version is not compatible with Always encrypted ");
+
+        connection = DriverManager.getConnection(connectionString);
+        SQLServerStatement stmt = (SQLServerStatement) connection.createStatement();
+        Utils.dropTableIfExists("esimple", stmt);
```
rene-ye: Consider a UID or something, we might be dropping user table here. Also, doesn't the test use Peter_Table, why is it `esimple` here.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
peterbae: I decided to throw an IllegalArgumentException because I thought it best described the symptom. It would be easy to throw an exception detailing at which index the parsing failed if the parsing actually failed at some point, but for these instances where the parsing fails due to getting to the end of the parsing method (which it shouldn't - this means the user's SQL query was incorrect somehow), all we know is that the user's argument was incorrect. I do agree that I could give a better description, though - I'll add that in.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkBatchInsertRecord.java
peterbae: Thanks. I think your implementation is good - I slapped my code onto the existing implementation for the getRowData() method from SQLServerBulkCSVFileRecord, and I didn't double check to see if I can refactor some of the code.

By the way, I decided not to merge the two getRowData() methods together (from SQLServerBulkCSVFileRecord and this file), since the CSV one assumes the data is always String, whereas this one doesn't (which could imply the implementation might change in the future). I'm open to suggestions, though.

src/test/java/com/microsoft/sqlserver/jdbc/preparedStatement/BatchExecutionWithBulkCopyParseTest.java
```diff
@@ -0,0 +1,161 @@
...
+@RunWith(JUnitPlatform.class)
+public class BatchExecutionWithBulkCopyParseTest extends AbstractTest {
+
+    static SQLServerPreparedStatement pstmt = null;
+    static Statement stmt = null;
+    static Connection connection = null;
+
+    @Test
+    public void testIsInsert() throws SQLException, NoSuchMethodException, SecurityException, IllegalAccessException, IllegalArgumentException, InvocationTargetException {
+        String valid1 = "INSERT INTO PeterTable values (1, 2)";
```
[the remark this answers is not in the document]
peterbae: This is parsing test. We don't communicate with the database here.

src/test/java/com/microsoft/sqlserver/jdbc/preparedStatement/BatchExecutionWithBulkCopyParseTest.java
peterbae: Same as above. I removed the esimple part, that was unnecessary.

rene-ye: We can remove this connection stuff altogether if we don't need a database for these tests.

peterbae: We need to initialize the connection object (and remove the rest, which I did), since we need the connection object to create the statement.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkBatchInsertRecord.java
```diff
@@ -0,0 +1,322 @@
...
+    }
+
+    private Object convertValue(ColumnMetadata cm,
+            Object data) throws SQLServerException {
+        switch (cm.columnType) {
+            /*
+             * Both BCP and BULK INSERT considers double quotes as part of the data and throws error if any data (say "10") is to be inserted into an
+             * numeric column. Our implementation does the same.
+             */
+            case Types.INTEGER: {
```
ulvii: https://github.com/Microsoft/mssql-jdbc/issues/531
https://github.com/Microsoft/mssql-jdbc/pull/538/

Do we have to convert everything to String?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkBatchInsertRecord.java
```diff
@@ -0,0 +1,322 @@
...
+                    return binData.substring(2);
+                }
+                else {
+                    return binData;
+                }
+            }
+
+            case 2013:    // java.sql.Types.TIME_WITH_TIMEZONE
+            {
+                DriverJDBCVersion.checkSupportsJDBC42();
```
ulvii: `DriverJDBCVersion.checkSupportsJDBC42();` should be removed

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkBatchInsertRecord.java
```diff
@@ -0,0 +1,322 @@
...
+                    offsetTimeValue = OffsetTime.parse(data.toString(), timeFormatter);
+                else
+                    offsetTimeValue = OffsetTime.parse(data.toString());
+
+                return offsetTimeValue;
+            }
+
+            case 2014: // java.sql.Types.TIMESTAMP_WITH_TIMEZONE
+            {
+                DriverJDBCVersion.checkSupportsJDBC42();
```
ulvii: Same here.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkBatchInsertRecord.java
```diff
@@ -0,0 +1,322 @@
...
+    public Object[] getRowData() throws SQLServerException {
+        Object[] data = new Object[columnMetadata.size()];
+        int valueIndex = 0;
+        String valueData;
+        Object rowData;
+        int columnListIndex = 0;
+        
+        // check if the size of the list of values = size of the list of columns (which is optional)
+        if (null != columnList && columnList.size() != valueList.size()) {
+            MessageFormat form = new MessageFormat(SQLServerException.getErrString("R_CSVDataSchemaMismatch"));
```
ulvii: Should we rename `R_CSVDataSchemaMismatch`?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkBatchInsertRecord.java
```diff
@@ -0,0 +1,322 @@
...
+                // if the user has provided a wildcard for this column, fetch the set value from the batchParam.
+                if (valueData.equalsIgnoreCase("?")) {
+                    rowData = batchParam.get(batchParamIndex)[valueIndex++].getSetterValue();
+                }
+                else if (valueData.equalsIgnoreCase("null")) {
+                    rowData = null;
+                }
+                // if the user has provided a hardcoded value for this column, rowData is simply set to the hardcoded value.
+                else {
+                    rowData = removeSingleQuote(valueData);
```
ulvii: Please test `insert into t values ('''')`

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkBatchInsertRecord.java
```diff
@@ -0,0 +1,322 @@
...
+
+    private String removeSingleQuote(String s) {
+        int len = s.length();
+        return (s.charAt(0) == '\'' && s.charAt(len - 1) == '\'') ? s.substring(1, len - 1) : s;
+    }
+
+    @Override
+    public Object[] getRowData() throws SQLServerException {
+        Object[] data = new Object[columnMetadata.size()];
+        int valueIndex = 0;
```
ulvii: Where is `valueIndex` used?

sadikovi:
Thanks for working on this! Would it be possible to post some benchmark numbers (before and after this patch) or just get a sense of how much performance improvement this patch adds?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -2548,6 +2647,326 @@ public final void addBatch() throws SQLServerException {
...
+        if (checkAndRemoveComments()) {
+            return parseUserSQLForTableNameDW(hasInsertBeenFound, hasIntoBeenFound);
+        }
+        
+        if (localUserSQL.substring(0, 6).equalsIgnoreCase("insert") && !hasInsertBeenFound) {
+            localUserSQL = localUserSQL.substring(6);
+            return parseUserSQLForTableNameDW(true, hasIntoBeenFound);
+        }
+        
+        if (localUserSQL.substring(0, 4).equalsIgnoreCase("into") && !hasIntoBeenFound) {
```
ulvii: Why do we still need `hasIntoBeenFound`, if `INTO` is optional ?

peterbae: because otherwise parsing would be incorrect in case where the user skips the "into" keyword, and decides to have a table name that starts with "into..."

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkBatchInsertRecord.java
peterbae: I don't think we need to. I'm following the implementation of bulk copy from SQLServerBulkCSVFileRecord. The PRs you linked are not related with these files - we can look into optimizing the bulk copy process in another PR.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkBatchInsertRecord.java
peterbae: It's not a big deal. I'd rather not change an existing entry in the resource file for this.

ulvii: Why wouldn't  you want to change it?

peterbae: This will affect other parts of the program too, and this isn't the only place where the naming change should occur. I'd rather handle these clean-up issues in a separate PR.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkBatchInsertRecord.java
peterbae: It works. I've added this scenario as a test case as well.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkBatchInsertRecord.java
peterbae: Turns out the change was pretty simple. I'll make these changes in the next commit.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkBatchInsertRecord.java
peterbae: It turns out we also need to do this, actually. The alternative of using instanceof would be messy here, since the user can provide any type of numeric value through setObject (they could do 123.45f or 12345L to instantiate those numbers as float, double or long, etc), and we would need an if statement for each of those types.

That being said, I found an issue with the existing implementation - it was actually not flooring the data. The number 123.56 would be inserted as 124 into the column, whereas it should've been 123. I've fixed this issue.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCommon.java
```diff
@@ -0,0 +1,299 @@
...
+             * SQL Server supports numerous string literal formats for temporal types, hence sending them as varchar with approximate
+             * precision(length) needed to send supported string literals. string literal formats supported by temporal types are available in MSDN
+             * page on data types.
+             */
+            case java.sql.Types.DATE:
+            case java.sql.Types.TIME:
+            case java.sql.Types.TIMESTAMP:
+            case microsoft.sql.Types.DATETIMEOFFSET:
+                if (this instanceof SQLServerBulkCSVFileRecord) {
+                    columnMetadata.put(positionInSource, new ColumnMetadata(colName, jdbcType, 50, scale, dateTimeFormatter));
```
ulvii: Both `if` and `else` blocks call `columnMetadata.put(positionInSource, new ColumnMetadata(colName, jdbcType, 50, scale, dateTimeFormatter));`. Also why do we have 50 as precision?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCommon.java
```diff
@@ -0,0 +1,299 @@
...
+    }
+
+    /**
+     * Set the format for reading in dates from the batch/file.
+     * 
+     * @param dateTimeFormat
+     *            format to parse data sent as java.sql.Types.TIMESTAMP_WITH_TIMEZONE
+     */
+    public void setTimestampWithTimezoneFormat(String dateTimeFormat) {
+        DriverJDBCVersion.checkSupportsJDBC42();
```
ulvii: No need for `DriverJDBCVersion.checkSupportsJDBC42();`

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCommon.java
```diff
@@ -0,0 +1,299 @@
...
+    }
+
+    /**
+     * Set the format for reading in dates from the batch/file.
+     * 
+     * @param timeFormat
+     *            format to parse data sent as java.sql.Types.TIME_WITH_TIMEZONE
+     */
+    public void setTimeWithTimezoneFormat(String timeFormat) {
+        DriverJDBCVersion.checkSupportsJDBC42();
```
ulvii: No need for DriverJDBCVersion.checkSupportsJDBC42();

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDataSource.java
```diff
@@ -821,6 +821,26 @@ public int getSocketTimeout() {
         int defaultTimeOut = SQLServerDriverIntProperty.SOCKET_TIMEOUT.getDefaultValue();
         return getIntProperty(connectionProps, SQLServerDriverIntProperty.SOCKET_TIMEOUT.toString(), defaultTimeOut);
     }
+    
+    /**
+     * Setting the use Bulk Copy API for Batch Insert on Azure Data Warehouse boolean
```
ulvii: Do we keep Azure Data Warehouse?

peterbae: Nope, we can do away with this comment.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCommon.java
peterbae: done.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCommon.java
peterbae: done.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCommon.java
peterbae: Sure. I agree that the if and else statement having the same is redundant. The 50 precision is used for CSV here since we can't rely on the precision from the SQL server - temporal types can be stored in CSV using any of the supported string literal format, and we pass it as varchar/nvarchar to server.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -106,6 +106,31 @@
     private void setPreparedStatementHandle(int handle) {
         this.prepStmtHandle = handle;
     }
+    
+    /**
+     * boolean value for deciding if the driver should use bulk copy API for batch inserts
+     */
+    private boolean useBulkCopyForBatchInsert;
+    
+    /** Sets the prepared statement's useBulkCopyForBatchInsert value.
```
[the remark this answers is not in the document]
ulvii: Can you also keep the wording consistent? For example `getUseBulkCopyForBatchInsert` starts with "Getting", `getUseBulkCopyForBatchInsert` starts with "Fethces".

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerStatement.java
```diff
@@ -977,6 +980,31 @@ final void resetForReexecute() throws SQLServerException {
             return false;
         return temp.substring(0, 6).equalsIgnoreCase("select");
     }
+    
+    /**
+     * Determine if the SQL is a INSERT.
+     * 
+     * @param sql
+     *            The statment SQL.
+     * @return True is the statement is a select.
```
ulvii: "True is the statement is a select."?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerStatement.java
```diff
@@ -977,6 +980,31 @@ final void resetForReexecute() throws SQLServerException {
...
+     */
+    /* L0 */ final boolean isInsert(String sql) throws SQLServerException {
+        checkClosed();
+        // Used to check just the first letter which would cause
+        // "Set" commands to return true...
+        String temp = sql.trim();
+        if (null == sql || sql.length() < 3) {
```
ulvii: `sql.length() < 6` makes more sense.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerStatement.java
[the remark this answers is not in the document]
ulvii: These lines are unnecessary. 
```
        char c = temp.charAt(0);
        if (c != 'i' && c != 'I')
            return false;
```

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerStatement.java
```diff
@@ -977,6 +980,31 @@ final void resetForReexecute() throws SQLServerException {
...
+            return false;
+        }
+        if (temp.substring(0, 2).equalsIgnoreCase("/*")) {
```
ulvii: Can't we call `checkAndRemoveComments()`  here?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -2548,6 +2647,326 @@ public final void addBatch() throws SQLServerException {
...
     }
+    
+    
+    private String parseUserSQLForTableNameDW(boolean hasInsertBeenFound, boolean hasIntoBeenFound) {
+        // As far as finding the table name goes, There are two cases:
+        // Insert into <tableName> and Insert <tableName>
+        // And there could be in-line comments (with /* and */) in between.
+        // This method assumes the localUserSQL string starts with "insert".
+        localUserSQL = localUserSQL.trim();
+        if (checkAndRemoveComments()) {
```
ulvii: Can't we just `checkAndRemoveComments` from the original query at the beginning?  Why do we have to do this inside every method?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -2548,6 +2647,326 @@ public final void addBatch() throws SQLServerException {
...
+        
+        if (localUserSQL.substring(0, 6).equalsIgnoreCase("insert") && !hasInsertBeenFound) {
+            localUserSQL = localUserSQL.substring(6);
+            return parseUserSQLForTableNameDW(true, hasIntoBeenFound);
+        }
+        
+        if (localUserSQL.substring(0, 4).equalsIgnoreCase("into") && !hasIntoBeenFound) {
+            // is it really "into"?
+            // if the "into" is followed by a blank space or /*, then yes.
+            if (localUserSQL.charAt(4) == ' ' || localUserSQL.charAt(4) == '\t' ||
```
ulvii: This is not accurate. The following query would fail.
```
insert into
table values(1)
```

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
ulvii: Don't you think there are too many unnecessary calls(these methods are all recursive) to `checkAndRemoveComments()` assuming most of the queries will contain no comments at all? Would it be easier/faster just to traverse the string at the very beginning and remove all comments?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -2548,6 +2647,326 @@ public final void addBatch() throws SQLServerException {
...
+            int tempint = localUserSQL.indexOf("]", 1);
+            
+            // keep checking if it's escaped
+            while (localUserSQL.charAt(tempint + 1) == ']') {
+                localUserSQL = localUserSQL.substring(0, tempint) + localUserSQL.substring(tempint + 1);
+                tempint = localUserSQL.indexOf("]", tempint + 1);
+            }
+            
+            // we've found a ] that is actually trying to close the square bracket.
+            // If it's followed by a dot, then it's a database.
```
ulvii: This assumption is wrong. You can also specify schema between database and table: `[database].[schema].[table]`.
You can even do a cross server query by adding a linked server. In that case `[server].[database].[schema].[table]` is possible too.

ulvii: I think your parsing logic below should still work, but please consider the scenarios I mentioned too.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerStatement.java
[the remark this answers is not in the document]
peterbae: Insert isn't the only way to use batch - there's many ways we would be calling isInsert on a non-insert query. I don't really think this fast elimination is worth the space, so i'm going to remove it anyways though.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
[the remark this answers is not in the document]
peterbae: Done.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerStatement.java
peterbae: Done. Also changed the javadocs for isSelect (which my javadoc came from).

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerStatement.java
peterbae: Agreed. fixed!

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
peterbae: Thanks. This was a pretty important catch. I've made the change accordingly.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -2548,6 +2647,326 @@ public final void addBatch() throws SQLServerException {
...
+        
+        // At this point, the next token has to be the table name.
+        // It could be encapsulated in [], "", or have a database name preceding the table name.
+        // If it's encapsulated in [] or "", we need be more careful with parsing as anything could go into []/"".
+        // For ] or ", they can be escaped by ]] or "", watch out for this too.
+        if (localUserSQL.substring(0, 1).equalsIgnoreCase("[")) {
+            int tempint = localUserSQL.indexOf("]", 1);
+            
+            // keep checking if it's escaped
+            while (localUserSQL.charAt(tempint + 1) == ']') {
```
ulvii: Wouldn't the parsing fail for `insert into [aaaa]]]]bbbbb] values (5)`?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
peterbae: Could be, but SQL Server treats the comment as a blank space. For example, the query below:

insert into t1/*this is a comment*/values (1);

will execute normally as if there was a space in between 't1' and 'values'. This means that if we were to sweep over the user's SQL query and remove all the comments, we would have to replace all the comments with a space, which would alter the user's SQL (that we will use to parse for the table names and such), which I don't think is better than the current approach.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
peterbae: I mentioned only the "database" in the comment, but I've already considered the scenario where the user chains multiple additional schemas/databases and such in fromt of the table name. the query will still work as intended here.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -2548,6 +2647,326 @@ public final void addBatch() throws SQLServerException {
...
+                localUserSQL = localUserSQL.substring(0, tempint) + localUserSQL.substring(tempint + 1);
+                tempint = localUserSQL.indexOf("]", tempint + 1);
+            }
+            
+            // we've found a ] that is actually trying to close the square bracket.
+            // If it's followed by a dot, then it's a database.
+            // Otherwise, it's the table.
+            if (localUserSQL.charAt(tempint + 1) == '.') {
```
ulvii: This is a valid query, notice the whitespaces  `insert into [master]     .               dbo            .         tablename values(1)`, which would probably fail here.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -2548,6 +2647,326 @@ public final void addBatch() throws SQLServerException {
...
+        }
+        
+        // At this point, the next chunk of string is the table name (could have database name), without starting with [ or ".
+        StringBuilder sb = new StringBuilder();
+        while (localUserSQL.length() > 0) {
+            if (localUserSQL.charAt(0) == '.' || localUserSQL.charAt(0) == ' ' || localUserSQL.charAt(0) == '\t'
+                    || localUserSQL.charAt(0) == '(') {
+                if (localUserSQL.charAt(0) == '.') {
+                    localUserSQL = localUserSQL.substring(1);
+                    return sb.toString() + "." + parseUserSQLForTableNameDW(true, true);
```
ulvii: What if table name contains a `.`?

ulvii: Same whitespace check failure will happen here too.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -2548,6 +2647,326 @@ public final void addBatch() throws SQLServerException {
...
+        
+        if (localUserSQL.charAt(0) == ',') {
+            localUserSQL = localUserSQL.substring(1);
+            return parseUserSQLForColumnListDWHelper(listOfColumns);
+        }
+        
+        if (localUserSQL.charAt(0) == '[') {
+            int tempint = localUserSQL.indexOf("]", 1);
+            
+            // keep checking if it's escaped
```
ulvii: I guess the parsing would fail if we have `[aaaa]]]]bbbbb]` as a column name too?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -2548,6 +2647,326 @@ public final void addBatch() throws SQLServerException {
...
+            } else {
+                // return tablename
+                String tempstr = localUserSQL.substring(1, tempint);
+                localUserSQL = localUserSQL.substring(tempint + 1);
+                return tempstr;
+            }
+        }
+        
+        // do the same for ""
+        if (localUserSQL.substring(0, 1).equalsIgnoreCase("\"")) {
```
ulvii: Same applies to `insert into "aaaa""""bbbbb" values (5)`

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -2548,6 +2647,326 @@ public final void addBatch() throws SQLServerException {
...
+            // we've found a " that is actually trying to close the quote.
+            String tempstr = localUserSQL.substring(1, tempint);
+            localUserSQL = localUserSQL.substring(tempint + 1);
+            listOfColumns.add(tempstr);
+            return parseUserSQLForColumnListDWHelper(listOfColumns);
+        }
+
+        // At this point, the next chunk of string is the column name, without starting with [ or ".
+        StringBuilder sb = new StringBuilder();
+        while (localUserSQL.length() > 0) {
```
ulvii: No need for nested if blocks here, this is simpler. 

```
if (localUserSQL.charAt(0) == ',') {
}
else if (localUserSQL.charAt(0) == ')'){
}
} else if (checkAndRemoveComments()) {
}
else{
}

```

src/test/java/com/microsoft/sqlserver/jdbc/preparedStatement/BatchExecutionWithBulkCopyTest.java
```diff
@@ -0,0 +1,359 @@
...
+        Method method = stmt.getClass().getDeclaredMethod("isInsert", String.class);
+        method.setAccessible(true);
+        assertTrue((boolean) method.invoke(stmt, valid1));
+        assertTrue((boolean) method.invoke(stmt, valid2));
+        assertTrue((boolean) method.invoke(stmt, valid3));
+        assertFalse((boolean) method.invoke(stmt, invalid));
+    }
+
+    @Test
+    public void testComments() throws SQLException, NoSuchFieldException, SecurityException, IllegalArgumentException, IllegalAccessException, NoSuchMethodException, InvocationTargetException {
```
ulvii: Just throw `Exception`.

src/test/java/com/microsoft/sqlserver/jdbc/preparedStatement/BatchExecutionWithBulkCopyTest.java
```diff
@@ -0,0 +1,359 @@
...
+        
+        String valid = "INSERT INTO " + tableName + " (c1, c3, c5, c8) values "
+                + "("
+                + "?, "
+                + "?, "
+                + "?, "
+                + "?, "
+                + ")";
+        
+        SQLServerPreparedStatement pstmt = (SQLServerPreparedStatement) connection.prepareStatement(valid);
```
ulvii: local `pstmt` and `stmt` are never closed.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
peterbae: It passes. It's part of the test suite.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
peterbae: Yes. I had to change some of the logic for parsing table names, including adding a couple more flags. Thanks for the catch!

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
peterbae: Table name containing . is handles. I made changes for the whitespace check.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
peterbae: the column name parsing is done correctly, but I had to make some changes for it down the line (which also applies to bulk copy from CSV).

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
peterbae: same as above.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
peterbae: done.

src/test/java/com/microsoft/sqlserver/jdbc/preparedStatement/BatchExecutionWithBulkCopyTest.java
peterbae: sure.

src/test/java/com/microsoft/sqlserver/jdbc/preparedStatement/BatchExecutionWithBulkCopyTest.java
peterbae: Sure. I made them all re-use the member variable stmt and pstmt instead.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkBatchInsertRecord.java
```diff
@@ -0,0 +1,319 @@
...
+                String binData = data.toString().trim();
+                if (binData.startsWith("0x") || binData.startsWith("0X")) {
+                    return binData.substring(2);
+                }
+                else {
+                    return binData;
+                }
+            }
+
+            case 2013:    // java.sql.Types.TIME_WITH_TIMEZONE
```
lilgreenbird: why do we have to use hardcoded 2013 instead of Types. TIME_WITH_TIMEZONE here? same with 2014 and TIMESTAMP_WITH_TIMEZONE?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCommon.java
```diff
@@ -0,0 +1,294 @@
...
+            Object[] msgArgs = {positionInSource};
+            throw new SQLServerException(form.format(msgArgs), SQLState.COL_NOT_FOUND, DriverError.NOT_SET, null);
+        }
+
+        if (null != name)
+            colName = name.trim();
+        else if ((columnNames != null) && (columnNames.length >= positionInSource))
+            colName = columnNames[positionInSource - 1];
+
+        if ((columnNames != null) && (positionInSource > columnNames.length)) {
```
lilgreenbird: just nitpicking but seems a bit inconsistent here  as other places we usually do (null != blah)
same for lines 157, 159

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkBatchInsertRecord.java
peterbae: Looking at the code for java.sql.Types, looks like TIME_WITH_TIMEZONE and TIMESTAMP_WITH_TIMEZONE was added later on in JDBC 4.2. They probably wrote this code before those values were added to the Types file, but knew the number code beforehand. I've made the changes to use the actual name.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCommon.java
peterbae: Sure. fixed.

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/BatchExecutionTest.java
```diff
@@ -167,6 +230,61 @@ public void testExecuteBatch1() {
             fail(TestResource.getResource("R_executeBatchFailed") + ": " + e.getMessage());
         }
     }
+    
+    public void testExecuteBatch1UseBulkCopyAPI() {
```
ulvii: This applies to all the tests, don't you need to set `useBulkCopyForBatchInsert` property to `true`?

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/PreparedStatementTest.java
```diff
@@ -295,6 +296,183 @@ public void testStatementPooling() throws SQLException {
...
+     * 
+     * @throws SQLException
+     * @throws SecurityException 
+     * @throws NoSuchFieldException 
+     * @throws IllegalAccessException 
+     * @throws IllegalArgumentException 
+     */
+    @Test
+    @Tag("slow")
+    public void testStatementPoolingUseBulkCopyAPI() throws SQLException, NoSuchFieldException, SecurityException,
```
ulvii: A lot of duplication here. Can't we refactor this?

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/BatchExecutionTest.java
```diff
@@ -167,6 +230,61 @@ public void testExecuteBatch1() {
             fail(TestResource.getResource("R_executeBatchFailed") + ": " + e.getMessage());
         }
     }
+
```
ulvii: This test is almost the exact copy of `testExecuteBatch1()`. Please refactor.

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/BatchExecutionTest.java
```diff
@@ -167,6 +230,61 @@ public void testExecuteBatch1() {
             fail(TestResource.getResource("R_executeBatchFailed") + ": " + e.getMessage());
         }
     }
+    
+    public void testExecuteBatch1UseBulkCopyAPI() {
+        int i = 0;
+        int retValue[] = {0, 0, 0};
+        int updCountLength = 0;
+        try {
+            String sPrepStmt = "update ctstable2 set PRICE=PRICE*20 where TYPE_ID=?";
```
ulvii: `sPrepStmt` is never closed.

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/BatchExecutionTest.java
```diff
@@ -167,6 +230,61 @@ public void testExecuteBatch1() {
...
+            pstmt1 = connection.prepareStatement(sPrepStmt1);
+
+            for (int n = 1; n <= 3; n++) {
+                pstmt1.setInt(1, n);
+                rs = pstmt1.executeQuery();
+                rs.next();
+                retValue[i++] = rs.getInt(1);
+            }
+
+            pstmt1.close();
```
ulvii: Better move this into a try-with-resources block

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/BatchExecutionTest.java
```diff
@@ -167,6 +230,61 @@ public void testExecuteBatch1() {
...
+
+            for (int j = 0; j < updateCount.length; j++) {
+                if (updateCount[j] != retValue[j] && updateCount[j] != Statement.SUCCESS_NO_INFO) {
+                    fail("executeBatch does not execute the Batch of SQL statements, Call to executeBatch is Failed!");
+                }
+            }
+        }
+        catch (BatchUpdateException b) {
+            fail("BatchUpdateException :  Call to executeBatch is Failed!");
```
ulvii: Call to executeBatch Failed!

ulvii: Same below

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/BatchExecutionTest.java
```diff
@@ -167,6 +230,61 @@ public void testExecuteBatch1() {
...
+        }
+        catch (SQLException sqle) {
+            fail("Call to executeBatch is Failed!");
```
ulvii: These catch blocks make no sense.

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/BatchExecutionTest.java
```diff
@@ -111,6 +115,65 @@ public void testAddBatch1() {
             fail(TestResource.getResource("R_addBatchFailed") + ": " + e.getMessage());
         }
     }
+
```
ulvii: The comments `testExecuteBatch1UseBulkCopyAPI()` apply to this test too.

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/BatchExecuteWithErrorsTest.java
```diff
@@ -473,4 +695,208 @@ public void Repro47239large() throws Exception {
         stmt.close();
         conn.close();
     }
+    
+    /**
+     * Tests large methods, supported in 42
+     * 
+     * @throws Exception
+     */
+    @Test
```
ulvii: Duplicate of `Repro47239large`. Please refactor.

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/BatchExecuteWithErrorsTest.java
```diff
@@ -473,4 +695,208 @@ public void Repro47239large() throws Exception {
...
+    @DisplayName("Regression test for using 'large' methods")
+    public void Repro47239largeUseBulkCopyAPI() throws Exception {
+
+        assumeTrue("JDBC42".equals(Utils.getConfiguredProperty("JDBC_Version")), TestResource.getResource("R_incompatJDBC"));
```
ulvii: This can be removed.

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/BatchExecuteWithErrorsTest.java
```diff
@@ -473,4 +695,208 @@ public void Repro47239large() throws Exception {
...
+            // So we have to skip the last scenario of this test case, i.e. "Test Severe (connection-closing) errors"
+            // It is worthwhile to still execute the first 5 test scenarios of this test case, in order to have best test coverage.
+            severe = "--Not executed when testing against SQL Azure";  // this is a dummy statement that never being executed on SQL Azure
+        }
+        else {
+            warning = "RAISERROR ('raiserror level 4',4,1) WITH LOG";
+            error = "RAISERROR ('raiserror level 11',11,1) WITH LOG";
+            severe = "RAISERROR ('raiserror level 20',20,1) WITH LOG";
+        }
+        con.close();
```
ulvii: Use  try-with-resources

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/BatchExecuteWithErrorsTest.java
```diff
@@ -473,4 +695,208 @@ public void Repro47239large() throws Exception {
...
+                }
+            }
+        }
+
+        try {
+            stmt.executeLargeUpdate("drop table " + tableName);
+        }
+        catch (Exception ignored) {
+        }
+        stmt.close();
```
ulvii: Try-with-resources.

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/BatchExecuteWithErrorsTest.java
```diff
@@ -473,4 +695,208 @@ public void Repro47239large() throws Exception {
...
+                }
+                else {
+                    assertTrue(actualExceptionText.equalsIgnoreCase("raiserror level 20"), TestResource.getResource("R_unexpectedExceptionContent") + ": " + actualExceptionText);
+
+                }
+            }
+        }
+
+        try {
+            stmt.executeLargeUpdate("drop table " + tableName);
```
ulvii: Please use `Utils.dropTableIfExists()`

src/test/java/com/microsoft/sqlserver/jdbc/preparedStatement/RegressionTest.java
```diff
@@ -313,7 +317,116 @@ public void batchWithLargeStringTest() throws SQLException {
         }
 
     }
+
```
ulvii: Duplicate of `batchWithLargeStringTest()`. Please refactor.

src/test/java/com/microsoft/sqlserver/jdbc/preparedStatement/RegressionTest.java
```diff
@@ -313,7 +317,116 @@ public void batchWithLargeStringTest() throws SQLException {
         }
 
     }
+    
+    @Test
+    public void batchWithLargeStringTestUseBulkCopyAPI() throws SQLException {
+        Connection con = DriverManager.getConnection(connectionString + ";useBulkCopyForBatchInsert=true;");
```
ulvii: Connection is never closed.

src/test/java/com/microsoft/sqlserver/jdbc/preparedStatement/RegressionTest.java
```diff
@@ -313,7 +317,116 @@ public void batchWithLargeStringTest() throws SQLException {
...
+        ResultSet rs = null;
+        String testTable = "TEST_TABLE_BULK_COPY";
+        Utils.dropTableIfExists(testTable, stmt);
+
+        con.setAutoCommit(false);
+
+        // create a table with two columns
+        boolean createPrimaryKey = false;
+        try {
+            stmt.execute("if object_id('" + testTable + "', 'U') is not null\ndrop table " + testTable + ";");
```
ulvii: Please use `Utils.dropTableIfExists()`

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/BatchExecuteWithErrorsTest.java
```diff
@@ -267,7 +269,227 @@ public void Repro47239() throws SQLException {
...
+                    assertTrue(actualExceptionText.equalsIgnoreCase("Connection reset"), TestResource.getResource("R_unexpectedExceptionContent") + ": " + actualExceptionText);
+                }
+                else {
+                    assertTrue(actualExceptionText.equalsIgnoreCase("raiserror level 20"), TestResource.getResource("R_unexpectedExceptionContent") + ": " + actualExceptionText);
+                }
+            }
+        }
+
+        try {
+            stmt.executeUpdate("drop table " + tableName);
```
ulvii: Please use the method from Utils class.

src/test/java/com/microsoft/sqlserver/jdbc/preparedStatement/BatchExecutionWithBulkCopyTest.java
```diff
@@ -0,0 +1,571 @@
...
+        rs.next();
+        for (int i=0; i < expected.length; i++) {
+            assertEquals(rs.getObject(i + 1).toString(), expected[i].toString());
+        }
+    }
+    
+    @BeforeEach
+    public void testSetup() throws TestAbortedException, Exception {
+        connection = DriverManager.getConnection(connectionString + ";useBulkCopyForBatchInsert=true;");
+        stmt = (SQLServerStatement) connection.createStatement();
```
ulvii: Better move this into a try-with-resources block

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/BatchExecutionTest.java
peterbae: Done.

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/PreparedStatementTest.java
peterbae: Done.

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/BatchExecutionTest.java
peterbae: It's closed.

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/BatchExecutionTest.java
peterbae: It's handled at the end of the test anyhoo.

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/BatchExecutionTest.java
peterbae: I agree that they could handle exception better here, but this isn't my code. I'm just gonna fix it anyways though.

src/test/java/com/microsoft/sqlserver/jdbc/preparedStatement/BatchExecutionWithBulkCopyTest.java
peterbae: the stmt gets cleaned up during the cleanup stage - i think that's okay?

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/BatchExecuteWithErrorsTest.java
peterbae: Done.

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/BatchExecutionTest.java
cheenamalhotra: The connection itself is available in the class, they're all static objects (don't know why), but why create a new one anywhere else in the class?

peterbae: the modifyConnectionForBulkCopyAPI method modifies the connection object. If we don't create a new connection object (with the try block), the tests that come after this one will be affected by the change I made to the connection (since the only other connection object available is static). Therefore, to prevent that happening, I need to make a new connection object just to test my PR.

Also, they connection object is static because originally the test could just re-use the same connection object that doesn't need to change, but now we need to change our connection object accordingly.

ulvii: > It's handled at the end of the test anyhoo.

 - Not if `executeQuery` fails.

peterbae: i think it'll be closed in the terminateVariation() (this is what I meant by end of the test)

Issue #493, opened 2017-09-15
PreparedStatement.getParameterMetaData has problems with encyrpted connections


gs-rezaem:
## Driver version or jar name
HEAD

## SQL Server version
2016

## Client operating system
Win 7

## Java/JVM version
1.8.0_xxx

## Table schema
from the driver test cases. see code below.

## Problem description
calling PreparedStatment.getParameterMetaData throws this exception:
com.microsoft.sqlserver.jdbc.SQLServerException: Operand type clash: void type is incompatible with date encrypted with (encryption_type = 'RANDOMIZED', encryption_algorithm_name = 'AEAD_AES_256_CBC_HMAC_SHA_256', column_encryption_key_name = 'JDBC_CEK', column_encryption_key_database_name = 'XYZ')

## Expected behavior and actual behavior
no exception should be thrown

## Repro code
add one line to the existing jdbc test cases. In AESetup.java, add the last line here:

```
    protected static void populateDateNormalCase(LinkedList<Object> dateValues) throws SQLException {
        String sql = "insert into " + dateTable + " values( " + "?,?,?," + "?,?,?," + "?,?,?," + "?,?,?," + "?,?,?," + "?,?,?" + ")";

        pstmt = (SQLServerPreparedStatement) Util.getPreparedStmt(con, sql, stmtColEncSetting);
        ParameterMetaData parameterMetaData = pstmt.getParameterMetaData();
```

v-afrafi:
Hi @gs-rezaem. Thanks for filing out this issue. As of now, Microsoft JDBC driver does not support retrieval of parameter metadata for encrypted columns. Please check the documentation [here](https://docs.microsoft.com/en-us/sql/connect/jdbc/using-parameter-metadata). 

We will provide the support in near future. Will update you on this github issue once completed.  Thanks!

cheenamalhotra:
Related Issue #887 
'sp_describe_undeclared_parameters' doesn't support temp tables.

_Proposed fix to fix both issues:_ 
Provide an option to switch to SET FMTONLY to gather Parameter MetaData since 'sp_describe_undeclared_parameters' has known limitations and backward compatibility issues.

peterbae:
Issue is being tracked in #927, closing this issue.

PR #687, opened 2018-04-26
Fix PreparedStatement.getParameterMetaData failing against encrypted columns

peterbae:
Fixes issue https://github.com/Microsoft/mssql-jdbc/issues/493

I purposely commented out the code instead of deleting it, since we might want to refer to this part of code in the future.


peterbae:
Turns out we won't be able to use the approach used here, due to the limitations of the implementation that relies on FMTONLY. We will have to search for an alternative - closing this PR in the meantime.

Issue #600, opened 2018-01-07
Computed columns shold be reported as IS_GENERATEDCOLUMN in DatabaseMetaData.getColumns(...)


schemacrawler:
## Driver version or jar name
Please tell us what the JDBC driver version or jar name is.   
*6.3.6.jre8-preview*

## SQL Server version
Please tell us what the SQL Server version is.   
*SQL Server 2017 14.00.1000.169.v1*

## Client operating system
Please tell us what oprating system the client program is running on.    
*Windows 10 Professional*

## Java/JVM version
Example: java version "1.8.0", IBM J9 VM   
*1.8.0_121-b13*

## Table schema
Please tell us the table schema   

```sql
CREATE TABLE X_EMPLOYEES
( 
  EMPLOYEEID INTEGER NOT NULL,
  EMPLOYEE_NAME VARCHAR(30), 
  START_DATE DATE, 
  END_DATE DATE, 
  ANNUAL_SALARY INTEGER,
  HOURLY_RATE AS (ANNUAL_SALARY/2080),
  ACTIVE AS (CASE WHEN END_DATE IS NULL THEN 'Y' ELSE 'N' END)
); 
```

## Problem description
Please share more details with us.

[`DatabaseMetaData.getColumns(...)`](https://docs.oracle.com/javase/8/docs/api/java/sql/DatabaseMetaData.html#getColumns-java.lang.String-java.lang.String-java.lang.String-java.lang.String-) should return a value of `YES` in `IS_GENERATEDCOLUMN` for columns `HOURLY_RATE` and `ACTIVE` in the table `X_EMPLOYEES` above.

## Expected behavior and actual behavior
Please tell us what should happen and what happened instead    

Please see the problem description above.

## Repro code
Please share repro code with us, or tell us how to reproduce the issue.

Create the table using the SQL above, and run [`DatabaseMetaData.getColumns(...)`](https://docs.oracle.com/javase/8/docs/api/java/sql/DatabaseMetaData.html#getColumns-java.lang.String-java.lang.String-java.lang.String-java.lang.String-), and check the value of `IS_GENERATEDCOLUMN` for columns `HOURLY_RATE` and `ACTIVE` in the table `X_EMPLOYEES` above.

ulvii:
Hi @schemacrawler ,

Thank you for reporting the issue. [The fix ](https://github.com/Microsoft/mssql-jdbc/pull/695)will be included in one of the upcoming preview releases.

[...]

PR #695, opened 2018-05-09
Mark computed columns as IS_GENERATEDCOLUMN in the result set returned by getColumns()

ulvii:
Currently the driver marks computed columns as `SS_IS_COMPUTED`, which violates JDBC spec. Renaming the field to `IS_GENERATEDCOLUMN`.  Fix for #600.


sualeh:
Thanks

PR #708, opened 2018-05-30
Request Boundary methods - beginRequest()/endRequest() implementation

ulvii:
Implementation and a test for Request Boundary methods that were introduced with JAVA 9.

JDBC spec:
https://docs.oracle.com/javase/9/docs/api/java/sql/Connection.html#beginRequest--
https://docs.oracle.com/javase/9/docs/api/java/sql/Connection.html#endRequest--

`beginRequest()`:

- Backs up the following connection properties that are modifiable through public methods - `databaseAutoCommitMode`, `transactionIsolationLevel`, `networkTimeout`, `holdability`, `sendTimeAsDatetime`, `statementPoolingCacheSize`, `disableStatementPooling`, `serverPreparedStatementDiscardThreshold`, `enablePrepareOnFirstPreparedStatementCall`, `sCatalog`, `sqlWarnings`.

- Marks the beginning of the work unit.

- No-op if a work unit has already started. 

`endRequest()`:

- Rolls back the open transactions. 

- Clears/closes the statements created inside the work unit.

- Reverts the modifiable connection properties mentioned above back to their original values and calls their setters.

- Marks the end of the work unit. 

- No-op without an intervening call to `beginRequest()`.

**Remarks**:
The the methods do not keep track of `columnEncryptionTrustedMasterKeyPaths`, because this is a static field and shared among the connections.


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -5300,14 +5317,99 @@ public boolean isWrapperFor(Class<?> iface) throws SQLException {
...
+                originalNetworkTimeout = getNetworkTimeout();
+                originalHoldability = holdability;
+                originalSendTimeAsDatetime = sendTimeAsDatetime;
+                originalStatementPoolingCacheSize = statementPoolingCacheSize;
+                originalDisableStatementPooling = disableStatementPooling;
+                originalServerPreparedStatementDiscardThreshold = serverPreparedStatementDiscardThreshold;
+                originalEnablePrepareOnFirstPreparedStatementCall = enablePrepareOnFirstPreparedStatementCall;
+                originalSCatalog = sCatalog;
+                originalSqlWarnings = sqlWarnings;
+                openStatements = new ArrayList<Statement>();
```
rene-ye: There's a lot more adding/removing compared to getting/setting. Using a LinkedList would yield better performance? The try-with-resources list removal behavior will still remain.

src/test/java/com/microsoft/sqlserver/jdbc/requestboundary/RequestBoundaryMethodsTest.java
```diff
@@ -0,0 +1,421 @@
...
+            }
+            if (null != ps) {
+                ps.close();
+            }
+            if (null != cs) {
+                cs.close();
+            }
+            if (null != con) {
+                con.close();
+            }
```
cheenamalhotra: Please use try-with-resources in all tests.

src/test/java/com/microsoft/sqlserver/jdbc/requestboundary/RequestBoundaryMethodsTest.java
```diff
@@ -0,0 +1,421 @@
...
+
+                    }
+                };
+
+                int originalNetworkTimeout = con.getNetworkTimeout();
+                int originalHoldability = con.getHoldability();
+                con.beginRequest();
+                thread1.start();
+                thread2.start();
+                thread3.start();
```
cheenamalhotra: You could use `CountDownLatch` for triggering all threads at exact same time, I guess that is what we should be testing here.

src/test/java/com/microsoft/sqlserver/jdbc/requestboundary/RequestBoundaryMethodsTest.java
```diff
@@ -0,0 +1,421 @@
...
+                try {
+                    // Wait for threads to complete
+                    Thread.sleep(3000);
```
cheenamalhotra: Again, Thread.sleep is not required. Java Concurrency has `CountDownLatch.await()` or `ExecutorService.awaitTermination()` APIs that would wait for all threads to finish and then proceed further.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerStatement.java
```diff
@@ -631,7 +631,9 @@ void closeInternal() {
         // Regardless what happens when cleaning up,
         // the statement is considered closed.
         assert !bIsClosed;
-
+        
+        connection.removeOpenStatement(this);
```
cheenamalhotra: This should be done right at the end of `Statement.close()` just before logging to avoid losing references of statements if they were not closed properly (failure possible in `discardLastExecutionResults`) due to any reason in a pooled connection scenario. It could lead into losing references/memory leakage of unclosed statements/resultsets.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -5300,14 +5317,99 @@ public boolean isWrapperFor(Class<?> iface) throws SQLException {
...
+    private int originalStatementPoolingCacheSize;
+    private boolean originalDisableStatementPooling;
+    private int originalServerPreparedStatementDiscardThreshold;
+    private Boolean originalEnablePrepareOnFirstPreparedStatementCall;
+    private String originalSCatalog = null;
+    private volatile SQLWarning originalSqlWarnings = null;
+    private List<Statement> openStatements = null;
+
+    protected void beginRequestInternal() throws SQLException {
         DriverJDBCVersion.checkSupportsJDBC43();
```
cheenamalhotra: Please remove unnecessary checks in both internal methods - need not check for checksSupportsJDBC43 anymore since the public APIs will be available only with SQLServerConnection43 class object.

src/test/java/com/microsoft/sqlserver/jdbc/requestboundary/RequestBoundaryMethodsTest.java
```diff
@@ -0,0 +1,421 @@
...
+        boolean autoCommitMode2 = false;
+        int transactionIsolationLevel2 = SQLServerConnection.TRANSACTION_SERIALIZABLE;
+        int networkTimeout2 = 10000;
+        int holdability2 = ResultSet.CLOSE_CURSORS_AT_COMMIT;
+        boolean sendTimeAsDatetime2 = false;
+        int statementPoolingCacheSize2 = 10;
+        boolean disableStatementPooling2 = false;
+        int serverPreparedStatementDiscardThreshold2 = 100;
+        boolean enablePrepareOnFirstPreparedStatementCall2 = true;
+        String sCatalog2 = "tempdb";
```
cheenamalhotra: Why are we hardcoding database names (model & tempdb) here? This test would fail if attempted to run with Azure DB specified in connection String.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -5300,14 +5317,99 @@ public boolean isWrapperFor(Class<?> iface) throws SQLException {
...
+                    if (0 > originalServerPreparedStatementDiscardThreshold) {
+                        setServerPreparedStatementDiscardThreshold(DEFAULT_SERVER_PREPARED_STATEMENT_DISCARD_THRESHOLD);
+                    }
+                    else {
+                        setServerPreparedStatementDiscardThreshold(originalServerPreparedStatementDiscardThreshold);
+                    }
+                }
+                if (enablePrepareOnFirstPreparedStatementCall != originalEnablePrepareOnFirstPreparedStatementCall) {
+                    if (null == originalEnablePrepareOnFirstPreparedStatementCall) {
+                        setEnablePrepareOnFirstPreparedStatementCall(DEFAULT_ENABLE_PREPARE_ON_FIRST_PREPARED_STATEMENT_CALL);
```
cheenamalhotra: Can we ignore this null check and use getter directly?

src/test/java/com/microsoft/sqlserver/jdbc/requestboundary/RequestBoundaryMethodsTest.java
ulvii: The method tests whether statements are closed after `endRequest()`. This should be fine.

src/test/java/com/microsoft/sqlserver/jdbc/requestboundary/RequestBoundaryMethodsTest.java
ulvii: No need to trigger threads at the same time, that's not the point of the test. We just need to make sure connection is back to its original state once the threads finished execution.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
ulvii: Agreed.

src/test/java/com/microsoft/sqlserver/jdbc/requestboundary/RequestBoundaryMethodsTest.java
ulvii: Agreed.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerStatement.java
ulvii: Agreed.

src/test/java/com/microsoft/sqlserver/jdbc/requestboundary/RequestBoundaryMethodsTest.java
ulvii: Agreed.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
cheenamalhotra: Well it's null by default, I guess that's why you're checking for null for original value can be null, but if you always use getter() for original value - it will never be null and this check won't be needed. Also we don't even need to reset it explicitly to its default value - just store Original back if new value is different.

src/test/java/com/microsoft/sqlserver/jdbc/requestboundary/RequestBoundaryMethodsTest.java
cheenamalhotra: I think if you're testing statments closed or not, you can use _asserts_ after endRequest is called. 
`Assert.assertTrue(closeable.isClosed());`

And connecion is not part of test, so it should be always in try-with-resources block. Basically, any closeable object created in tests must always be created in try-with-resources blocks to avoid resource leaks if tests fail due to any reason.

ulvii: Asserts are already there.

I agree that we should use try-with-resources when possible, but for this scenario I will keep the finally block instead, because the method tests multiple statements between `beginRequest()`/ `endRequest()` and we would need to add multiple try blocks too. Which in my opinion will make the code less readable.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
ulvii: True.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -5296,14 +5313,87 @@ public boolean isWrapperFor(Class<?> iface) throws SQLException {
...
+                if (getEnablePrepareOnFirstPreparedStatementCall() != originalEnablePrepareOnFirstPreparedStatementCall) {
+                    setEnablePrepareOnFirstPreparedStatementCall(originalEnablePrepareOnFirstPreparedStatementCall);
+                }
+                if (!sCatalog.equals(originalSCatalog)) {
+                    setCatalog(originalSCatalog);
+                }
+                sqlWarnings = originalSqlWarnings;
+                if (null != openStatements) {
+                    while (!openStatements.isEmpty()) {
+                        try (Statement st = openStatements.get(0)) {
```
peterbae: Check this logic again. This might result in infinite loop.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -5296,14 +5313,87 @@ public boolean isWrapperFor(Class<?> iface) throws SQLException {
...
+    private int originalNetworkTimeout;
+    private int originalHoldability;
+    private boolean originalSendTimeAsDatetime;
+    private int originalStatementPoolingCacheSize;
+    private boolean originalDisableStatementPooling;
+    private int originalServerPreparedStatementDiscardThreshold;
+    private Boolean originalEnablePrepareOnFirstPreparedStatementCall;
+    private String originalSCatalog = null;
+    private volatile SQLWarning originalSqlWarnings = null;
+    private List<Statement> openStatements = null;
```
peterbae: I don't think we need to assign false or null randomly to these variables - primitives like boolean will initialize to false by default, and objects will initialize to null by default. I'm not sure if this was intentional, but we could probably clean this up.

peterbae: i like how you put the variables here (member variables usually go to the top, but I see that this class already contains instances where they put the relevant member variables near the relevant methods, so this is probably the right way to organize things. I'm assuming this was intentional)

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -5296,14 +5313,87 @@ public boolean isWrapperFor(Class<?> iface) throws SQLException {
...
     }
 
-    public void endRequest() throws SQLFeatureNotSupportedException {
-        DriverJDBCVersion.checkSupportsJDBC43();
-        throw new SQLFeatureNotSupportedException("endRequest not implemented");
+    protected void endRequestInternal() throws SQLException {
+        synchronized (this) {
+            if (requestStarted) {
+                if (!databaseAutoCommitMode) {
+                    rollback();
```
peterbae: Looks like you decided to do a rollback here. If database is not in autocommit mode, at least the rollback will always succeed. However, it looks like Oracle JDBC spec for endRequest suggests that if the pooling manager calls endRequest while there is a transaction going on, the driver throw a SQLException (https://docs.oracle.com/javase/9/docs/api/java/sql/Connection.html#endRequest--). Have you considered why we're not throwing an exception if there's a transaction?

ulvii: Most of the 3rd party pool managers do a rollback before returning connections to the pool, I decided to keep it the same way.

peterbae: cool, I think that's fair.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -5296,14 +5313,87 @@ public boolean isWrapperFor(Class<?> iface) throws SQLException {
...
+    private boolean requestStarted = false;
+    private boolean originalDatabaseAutoCommitMode;
+    private int originalTransactionIsolationLevel;
+    private int originalNetworkTimeout;
+    private int originalHoldability;
+    private boolean originalSendTimeAsDatetime;
+    private int originalStatementPoolingCacheSize;
+    private boolean originalDisableStatementPooling;
+    private int originalServerPreparedStatementDiscardThreshold;
+    private Boolean originalEnablePrepareOnFirstPreparedStatementCall;
```
peterbae: oh also this is the wrapper Boolean, this would be null by default. Change this to "boolean" if you want to avoid weird confusions in the future.

peterbae: huh. you know what, I saw that getEnablePrepareOnFirstPreparedStatementCall() returns boolean and assumed enablePrepareOnFirstPreparedStatementCall was boolean too. Looking at the code, the Boolean was intentional (it makes use of the null state), so keeping the originalEnablePrepareOnFirstPreparedStatementCall as Boolean here is the right choice too.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
peterbae: Actually, after discussing with Ulvi it turns out the try block calls close() which calls removeOpenStatement, so this while loop does eliminate elements from the list successfully. no change needed here.

PR #709, opened 2018-05-31
Introduce support for "Data Classification Specifications" on fetched resultsets

cheenamalhotra:
This PR introduces new APIs that provide support and read Data Sensitivity Classification information from SQL Server.

This feature will be available/functional from new versions of SQL Servers (from 2018). For older versions of SQL Server, request for Data Classification Feature-Ack will be a No-Op.


src/main/java/com/microsoft/sqlserver/jdbc/InformationType.java
```diff
@@ -0,0 +1,20 @@
+package com.microsoft.sqlserver.jdbc;
```
ulvii: `InformationType` and `Label` classes are exactly the same. Please also make sure the class names are descriptive.

src/main/java/com/microsoft/sqlserver/jdbc/InformationType.java
```diff
@@ -0,0 +1,20 @@
...
+    }
+    
+    public String getName() {
+		return name;
+	}
+
+	public String getId() {
+		return id;
+	}
+}
```
ulvii: Please add a new line to the end of the file.

src/main/java/com/microsoft/sqlserver/jdbc/InformationType.java
```diff
@@ -0,0 +1,20 @@
...
+    public String id;
+
+    public InformationType(String name, String id)
+    {
+        this.name = name;
+        this.id = id;
+    }
+    
+    public String getName() {
+		return name;
```
ulvii: Please apply the formatter.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -3421,7 +3426,20 @@ int writeFedAuthFeatureRequest(boolean write,
...
         return totalLen;
     }
+    
+    int writeDataClassificationFeatureRequest (boolean write /* if false just calculates the length */,
+    		TDSWriter tdsWriter) throws SQLServerException {
+        int len = 6; // 1byte = featureID, 4bytes = featureData length, 1 bytes = Version
 
+        if (write) {
+            // Write Feature ID, length of the version# field and Sensitivity Classification Version#
+        	tdsWriter.writeByte((byte) TDS.TDS_FEATUREEXT_DATACLASSIFICATION);
```
ulvii: Let's declare `TDS_FEATUREEXT_DATACLASSIFICATION` as a `byte` instead.

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -112,9 +115,17 @@
...
 
     // AE constants
+    // 0x03 is for x_eFeatureExtensionId_Rcs
     static final int TDS_FEATURE_EXT_AE = 0x04;
     static final int MAX_SUPPORTED_TCE_VERSION = 0x01; // max version
     static final int CUSTOM_CIPHER_ALGORITHM_ID = 0; // max version
+    // 0x06 is for x_eFeatureExtensionId_LoginToken 
+    // 0x07 is for x_eFeatureExtensionId_ClientSideTelemetry 
+    // Data Classification constants
+    static final int TDS_FEATUREEXT_DATACLASSIFICATION = 0x09;
```
ulvii: Can we keep/make the naming consistent? There are 3 more similar constants:
`TDS_FEATURE_EXT_AE`, `TDS_FEATURE_EXT_FEDAUTH`, `TDS_FEATURE_EXTSION_ACK`. I would suggest using `TDS_FEATURE_EXT` for all of them.

src/main/java/com/microsoft/sqlserver/jdbc/InformationType.java
cheenamalhotra: We cannot rename classes, as they are as per design specs.

On the other hand, I think it would be better to move all these new classes to `dataclassification` package so that customers would relate `Label` and` InformationType` classes to data classification information and not mix with other driver classes.

Other drivers have implemented differently with all classes defined in a common class `DataClassification `, but in Java we shall do in separate individual classes as they are public APIs.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
cheenamalhotra: This cannot be done, as when this token is received from SQL Server, it is received as `int` as you can see [here](https://github.com/Microsoft/mssql-jdbc/pull/709/files#diff-0a4455f5ffd8e31e9e1badaa7822375eR4067) that is the reason all Feature Ext tokens in the driver are of `int` type.

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
cheenamalhotra: I would not rename others, but would make the one I added `TDS_FEATURE_EXT_DATACLASSIFICATION`

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
cheenamalhotra: Cannot change it right away, as while reading token, featureId is compared in `int` type, as you can see [here](https://github.com/Microsoft/mssql-jdbc/pull/709/files#diff-0a4455f5ffd8e31e9e1badaa7822375eR4067). If it will be changed for one, it will be changed for others too and need to see its impact on the driver behavior.

ulvii: featureID is a `byte` too, it should not be `int` at the first place.

src/test/java/com/microsoft/sqlserver/jdbc/resultset/DataClassificationTest.java
```diff
@@ -0,0 +1,158 @@
...
+        }
+    }
+    
+    /**
+     * Drops table from the database
+     * @param stmt
+     * @throws SQLException
+     */
+    private void dropTable(Statement stmt) throws SQLException {
+        stmt.execute("DROP TABLE " + tableName);
```
ulvii: Please `Utils.dropTableIfExists()` instead.

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -6376,9 +6391,13 @@ final SQLServerConnection getConnection() {
     private boolean isStreaming = true;
     private boolean useColumnEncryption = false;
     private boolean serverSupportsColumnEncryption = false;
+    private boolean serverSupportsDataClassification = false;
 
     private final byte valueBytes[] = new byte[256];
-    private static final AtomicInteger lastReaderID = new AtomicInteger(0);
+    
+	protected SensitivityClassification sensitivityClassification;
```
peterbae: formatting

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -7114,6 +7138,13 @@ final void TryProcessFeatureExtAck(boolean featureExtAckReceived) throws SQLServ
         if (isColumnEncryptionSettingEnabled() && !featureExtAckReceived)
             throw new SQLServerException(this, SQLServerException.getErrString("R_AE_NotSupportedByServer"), null, 0, false);
     }
+    
+    boolean TrySetSensitivityClassification(SensitivityClassification sensitivityClassification) {
```
peterbae: method names should start with a lowercase

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -3421,7 +3426,20 @@ int writeFedAuthFeatureRequest(boolean write,
         }
         return totalLen;
     }
+    
+    int writeDataClassificationFeatureRequest (boolean write /* if false just calculates the length */,
```
peterbae: I'd put the protected keyword here explicitly to maintain consistency throughout the file.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -4116,11 +4134,40 @@ private void onFeatureExtAck(int featureId,
...
+                
+                byte supportedDataClassificationVersion = data[0];
+                if ((0 == supportedDataClassificationVersion) || (supportedDataClassificationVersion > TDS.MAX_SUPPORTED_DATA_CLASSIFICATION_VERSION)) {
+                	if (connectionlogger.isLoggable(Level.SEVERE)) {
+                        connectionlogger.severe(toString() + " Invalid version number for Data Classification");
+                    }
+                    throw new SQLServerException(SQLServerException.getErrString("R_InvalidDataClsTokenNumber"), null);
+                }
 
+                if (data.length != 2) {
```
peterbae: Do we need to check 1 > data.length and data.length != 2 separately?

src/test/java/com/microsoft/sqlserver/jdbc/resultset/DataClassificationTest.java
```diff
@@ -0,0 +1,151 @@
...
+        }
+    }
+
+    /**
+     * Checks if object SYS.SENSITIVITY_CLASSIFICATIONS exists in SQL Server
+     * 
+     * @param Statement
+     * @return boolean
+     */
+    private boolean serverSupportsDataClassification(Statement stmt) {
```
ulvii: Can we move this to `Utils` class?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
cheenamalhotra: Didn't want to change the way/sequence we check for this token, exact same way is done for AE feature as well as in .NET driver.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResource.java
```diff
@@ -393,5 +393,7 @@ static String getResource(String key) {
 				{"R_invalidSSLProtocol", "SSL Protocol {0} label is not valid. Only TLS, TLSv1, TLSv1.1, and TLSv1.2 are supported."},
 				{"R_cancelQueryTimeoutPropertyDescription", "The number of seconds to wait to cancel sending a query timeout."},
 				{"R_invalidCancelQueryTimeout", "The cancel timeout value {0} is not valid."},
+				{"R_UnknownDataClsTokenNumber","Unknown token for Data Classification."},	// From Server
+				{"R_InvalidDataClsVersionNumber","Invalid version number {0} for Data Classification."},	// From Server
```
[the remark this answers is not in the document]
cheenamalhotra: Should be used [here](https://github.com/Microsoft/mssql-jdbc/pull/709/files/0224a03e9a09585b630ae3d9f50705d3ae7bb122#diff-0a4455f5ffd8e31e9e1badaa7822375eR4157). Good catch!

src/test/java/com/microsoft/sqlserver/jdbc/resultset/DataClassificationTest.java
cheenamalhotra: Will do.

src/main/java/com/microsoft/sqlserver/jdbc/StreamColumns.java
```diff
@@ -210,6 +220,97 @@ void setFromTDS(TDSReader tdsReader) throws SQLServerException {
...
+        List<Label> labels = new ArrayList<Label>(numLabels);
+
+        for (int i = 0; i < numLabels; i++) {
+            labels.add(readSensitivityLabel(tdsReader));
+        }
+
+        // get the information type count
+        int numInformationTypes = tdsReader.readUnsignedShort();
+
+        List<InformationType> informationTypes = new ArrayList<InformationType>(numInformationTypes);
```
rene-ye: I think we should generally use LinkedLists over ArrayLists, seeing how little we do random access compared to iteration. Also `new ArrayList<InformationType>` is repetitive, the compiler will infer the template type if we just use `new ArrayList<>`.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -4126,11 +4144,40 @@ private void onFeatureExtAck(int featureId,
...
+                	if (connectionlogger.isLoggable(Level.SEVERE)) {
+                        connectionlogger.severe(toString() + " Unknown token for Data Classification");
+                    }
+                    throw new SQLServerException(SQLServerException.getErrString("R_UnknownDataClsTokenNumber"), null);
+                }
+                
+                byte enabled = data[1];
+                serverSupportsDataClassification = (enabled == 0) ? false : true;
+                break;
+            }
```
lilgreenbird: is it just me or the spacing is all off in this method?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -4126,11 +4144,40 @@ private void onFeatureExtAck(int featureId,
...
+                
+                byte supportedDataClassificationVersion = data[0];
+                if ((0 == supportedDataClassificationVersion) || (supportedDataClassificationVersion > TDS.MAX_SUPPORTED_DATA_CLASSIFICATION_VERSION)) {
+                	if (connectionlogger.isLoggable(Level.SEVERE)) {
+                        connectionlogger.severe(toString() + " Invalid version number for Data Classification");
+                    }
+                    throw new SQLServerException(SQLServerException.getErrString("R_InvalidDataClsVersionNumber"), null);
+                }
 
+                if (data.length != 2) {
```
lilgreenbird: can we keep this consistent? above is (1> data.length) etc..

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -4126,11 +4144,40 @@ private void onFeatureExtAck(int featureId,
...
+                	if (connectionlogger.isLoggable(Level.SEVERE)) {
+                        connectionlogger.severe(toString() + " Unknown token for Data Classification");
+                    }
+                    throw new SQLServerException(SQLServerException.getErrString("R_UnknownDataClsTokenNumber"), null);
+                }
+                
+                byte enabled = data[1];
```
lilgreenbird: this variable seem unnecessary?

src/test/java/com/microsoft/sqlserver/testframework/util/Util.java
```diff
@@ -301,6 +301,25 @@ public static boolean supportJDBC43(Connection con) throws SQLException {
...
+     */
+    public static boolean serverSupportsDataClassification(Statement stmt) {
+        try {
+            stmt.execute("SELECT * FROM SYS.SENSITIVITY_CLASSIFICATIONS");
+        }
+        catch (SQLException e) {
+            // Check for Error 208: Invalid Object Name
+            if (e.getErrorCode() == 208) {
+                return false;
+            }
```
lilgreenbird: what happens if this fails and it's not error 208?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -4416,9 +4463,12 @@ else if (serverMajorVersion >= 9) // Yukon (9.0) --> TDS 7.2 // Prelogin disconn
         if (federatedAuthenticationInfoRequested || federatedAuthenticationRequested) {
             len2 = len2 + writeFedAuthFeatureRequest(false, tdsWriter, fedAuthFeatureExtensionData);
         }
-
-        len2 = len2 + 1; // add 1 to length becaue of FeatureEx terminator
-
+        //Data Classification is always enabled (by default)
+        
+        len2 += writeDataClassificationFeatureRequest(false, tdsWriter);
```
lilgreenbird: hmm why len2? (there is no len1), perhaps a more meaningful name?

src/test/java/com/microsoft/sqlserver/testframework/util/Util.java
cheenamalhotra: It would return true

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
cheenamalhotra: used in the next line

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
cheenamalhotra: keeping it the same way as before, not changing existing variables unless required.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
lilgreenbird: I mean, it's only used once in the next line, seems unnecessary to define a variable for it...no biggie tho..

src/test/java/com/microsoft/sqlserver/testframework/util/Util.java
lilgreenbird: right, missed that ;)

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -100,9 +103,11 @@
     static final int TDS_DONEPROC = 0xFE;
     static final int TDS_DONEINPROC = 0xFF;
     static final int TDS_FEDAUTHINFO = 0xEE;
+    static final int TDS_SQLRESCOLSRCS = 0xa2;
```
ulvii: `TDS_SQLRESCOLSRCS` is not used anywhere

src/test/java/com/microsoft/sqlserver/jdbc/resultset/DataClassificationTest.java
```diff
@@ -0,0 +1,134 @@
+package com.microsoft.sqlserver.jdbc.resultset;
```
ulvii: Please add the license header.

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -112,9 +117,17 @@
...
-    static final int MAX_SUPPORTED_TCE_VERSION = 0x01; // max version
+    // 0x03 is for x_eFeatureExtensionId_Rcs
+    static final byte TDS_FEATURE_EXT_AE = 0x04;
+    static final byte MAX_SUPPORTED_TCE_VERSION = 0x01; // max version
     static final int CUSTOM_CIPHER_ALGORITHM_ID = 0; // max version
+    // 0x06 is for x_eFeatureExtensionId_LoginToken 
+    // 0x07 is for x_eFeatureExtensionId_ClientSideTelemetry 
+    // Data Classification constants
+    static final byte TDS_FEATURE_EXT_DATACLASSIFICATION = 0x09;
+    static final byte DATA_CLASSIFICATION_NOT_ENABLED = 0x00;
```
ulvii: `DATA_CLASSIFICATION_NOT_ENABLED` is not used anywhere.

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -6376,9 +6391,13 @@ final SQLServerConnection getConnection() {
...
     private boolean useColumnEncryption = false;
     private boolean serverSupportsColumnEncryption = false;
+    private boolean serverSupportsDataClassification = false;
 
     private final byte valueBytes[] = new byte[256];
-    private static final AtomicInteger lastReaderID = new AtomicInteger(0);
+    
+    protected SensitivityClassification sensitivityClassification;
+    
+	private static final AtomicInteger lastReaderID = new AtomicInteger(0);
```
ulvii: Please apply the formatter to your changes.

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -7114,6 +7138,11 @@ final void TryProcessFeatureExtAck(boolean featureExtAckReceived) throws SQLServ
         if (isColumnEncryptionSettingEnabled() && !featureExtAckReceived)
             throw new SQLServerException(this, SQLServerException.getErrString("R_AE_NotSupportedByServer"), null, 0, false);
     }
+
+    boolean trySetSensitivityClassification(SensitivityClassification sensitivityClassification) {
+        this.sensitivityClassification = sensitivityClassification;
+        return true;
```
ulvii: Why does this method return `boolean`? `return true` makes no sense.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -4126,11 +4144,40 @@ private void onFeatureExtAck(int featureId,
...
-                assert supportedTceVersion == TDS.MAX_SUPPORTED_TCE_VERSION; // Client support TCE version 1
                 serverSupportsColumnEncryption = true;
                 break;
             }
+            case TDS.TDS_FEATURE_EXT_DATACLASSIFICATION: {
+            	if (connectionlogger.isLoggable(Level.FINER)) {
+                    connectionlogger.fine(toString() + " Received feature extension acknowledgement for Data Classification.");
+                }
+
+                if (1 > data.length) {
```
ulvii: Please remove this block and move `if (data.length != 2)` before `if ((0 == supportedDataClassificationVersion) || (supportedDataClassificationVersion > TDS.MAX_SUPPORTED_DATA_CLASSIFICATION_VERSION))`.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerException.java
```diff
@@ -72,6 +72,11 @@ final int getErrorCode() {
     static final int DRIVER_ERROR_INTERMITTENT_TLS_FAILED = 7;
     static final int ERROR_SOCKET_TIMEOUT = 8;
     static final int ERROR_QUERY_TIMEOUT = 9;
+    static final int DataClassificationInvalidVersion = 24;
```
ulvii: Is there a reason why you specified the error codes but didn't use any of them? You could use `DataClassificationInvalidVersion` with  `throw new SQLServerException(SQLServerException.getErrString("R_InvalidDataClsVersionNumber"), null);` for example.

I think there is actually no need for these error codes in JDBC driver.

src/main/java/com/microsoft/sqlserver/jdbc/dataclassification/ColumnSensitivity.java
```diff
@@ -0,0 +1,16 @@
+package com.microsoft.sqlserver.jdbc.dataclassification;
```
ulvii: Please add license headers too all new files.

src/main/java/com/microsoft/sqlserver/jdbc/tdsparser.java
```diff
@@ -167,7 +166,7 @@ final StreamError getDatabaseError() {
         return databaseError;
     }
 
-    TDSTokenHandler(String logContext) {
+	TDSTokenHandler(String logContext) {
```
ulvii: Formatting.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerException.java
cheenamalhotra: The code is ported from .NET driver as such, tokens and error codes not used here are not used in ADO driver too, but may be used in future expansion of this feature. I let them stay as such as they will be useful to track changes in future in comparison to ADO changes.

ulvii: Does 24 have a special meaning or this was just 24th error code in .Net driver? If that's the case, we shouldn't jump from 9 to 24. 
Please also keep the naming consistent with other error codes.

src/test/java/com/microsoft/sqlserver/jdbc/resultset/DataClassificationTest.java
```diff
@@ -0,0 +1,134 @@
...
+            }
+        }
+    }
+
+    /**
+     * Creates a new table in database with data classification column tags Inserts rows of data in the table
+     * 
+     * @param connection
+     * @param stmt
+     * @param tableName
```
ulvii: tableName is not a `@param`

src/test/java/com/microsoft/sqlserver/jdbc/resultset/DataClassificationTest.java
```diff
@@ -0,0 +1,134 @@
+package com.microsoft.sqlserver.jdbc.resultset;
+
+import java.sql.Connection;
```
ulvii: Apply the formatter please

src/test/java/com/microsoft/sqlserver/jdbc/resultset/DataClassificationTest.java
```diff
@@ -0,0 +1,134 @@
...
+            ps.setString(7, "(313) 555-3349");
+            ps.execute();
+        }
+    }
+
+    /**
+     * Selects data from the table and triggers verifySensitivityClassification method
+     * 
+     * @param stmt
+     * @param queries
```
ulvii: queries is not a `@param`

src/test/java/com/microsoft/sqlserver/jdbc/resultset/DataClassificationTest.java
cheenamalhotra: Already applied - I don't see any change on doing again

src/test/java/com/microsoft/sqlserver/jdbc/resultset/DataClassificationTest.java
cheenamalhotra: Removed

src/main/java/com/microsoft/sqlserver/jdbc/dataclassification/ColumnSensitivity.java
cheenamalhotra: Agreed

Issue #716, opened 2018-06-05
Performance problems with statement caching due to SHA1 hash


RDearnaley:
## Driver version or jar name
mssql-jdbc version 6.4.0.jre8

## SQL Server version
We support MS SQL Server versions back to 2012, and have been testing on version 11.0.5058.0

## Client operating system
We support Windows/Linux/Mac/various other flavors of UNIX, have been testing with the client+ JDBC driver on Linux (with the SQL Server on Windows, obviously)

## Java/JVM version
1.8

## Table schema
(large, complex, and irrelevant to this issue)

## Problem description
We recently switched our code from using the old sqljdbc42 4.2.6420.100 MS SQL Server JDBC driver to using mssql-jdbc version 6.4.0.jre8 -- we encountered a significant (O(20-35%)) performance hit when doing so, which we have been investigating. Our performance test setup uses a repeating pattern of O(100) distinct read queries of large SQL text length (up to around 16kb) that were generated by Hibernate with some queries repeated many times withing each overall repeat pattern, each individual query returning only a small amount of data, with the client and database on two different machines in the same datacenter. Thus prepared statement caching is a big win-- the connection pool we are using no longer provides this, which is why we switched to mssql-jdbc version 6.4.0.jre8.

Using the YourKit Java profiting tool, I tracked down part of the problem to the hashing code for the JDBC driver's prepared statement cache, which is using SHA-1 hashing (which it delegates to the Java security system, which delegates it to our chosen security provided library). For this test setup 10-15% of the total test time was being spent performing SHA-1 hashing in order to check the prepared statement cache -- quite unacceptable overhead for a JDBC driver! (We experimented with simply turning the caching off, but alas that made things even slower -- despite its current heavy hashing overhead, the prepared statement cache is still 5-10%faster on average than repreparing every statement.)

Here is the relevant section of the stack trace (in caller-first order) when this SHA-1 behavior was happening:

... (Hikari connection pool library) ...
ProxyPreparedStatement.java:52 com.microsoft.sqlserver.jdbc.SQLServerPreparedStatement.executeQuery()
SQLServerPreparedStatement.java:401 com.microsoft.sqlserver.jdbc.SQLServerStatement.executeStatement(TDSCommand)
SQLServerStatement.java:204 com.microsoft.sqlserver.jdbc.SQLServerStatement.executeCommand(TDSCommand)
[...]

PR #717, opened 2018-06-06
Change Sha1HashKey to CityHash128Key for generating PreparedStatement handle and metadata cache keys

cheenamalhotra:
Fixes issue #716


src/main/java/com/microsoft/sqlserver/jdbc/Util.java
```diff
@@ -1096,3 +1096,364 @@ else if (databaseName.length() > 0)
         return fullName.toString();
     }
 }
+
```
ulvii: Maybe a separate file for this?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -119,37 +117,31 @@
...
-        Sha1HashKey(String sql,
+        CityHash128Key(String sql,
                 String parametersDefinition) {
             this(String.format("%s%s", sql, parametersDefinition));
         }
 
-        Sha1HashKey(String s) {
-            bytes = getSha1Digest().digest(s.getBytes());
+        CityHash128Key(String s) {
+            segments = CityHash.cityHash128(s.getBytes(), 0, s.length());
```
brettwooldridge: Performance of this can be improved, especially in the case of large SQL strings.  ``s.getBytes()`` ends up calling ``StringCoding.encode(...)``, which converts the internal String chars to bytes using the platform default encoding.

I recommend using ``s.getBytes(int srcBegin, int srcEnd, byte dst[], int dstBegin)``, which, while deprecated (but will never be removed), has *substantially* higher performance.
```java
CityHash128Key(String s) {
   byte[] bytes = byte[s.length()];
   s.getBytes(0, s.length(), bytes, 0);
   segments = CityHash.cityHash128(bytes, 0, s.length());
}
```

RDearnaley: s.getBytes(int srcBegin, int srcEnd, byte dst[], int dstBegin) truncates the upper 8 bits of Java's internal Unicode-16 representation. Since this is SQL before inserting parameter values, it should be mostly 7-bit ASCII -- the only exception would be database names/table names/column names, at least some of which can be full Unicode. For European languages, their Unicode-16 spaces tend to be 256 characters wide and the odds of this causing a spurious collision seem very small -- but I'd be less confident if the language in question was, say, Chinese, that you couldn't have two SQLs that differed only by one ideogram in a table/column name, and the two ideograms in question happened to have the same lower 8 bits. It's not very likely, but it's significantly less astronomically unlikely than a 128-bit hash collision.

RDearnaley: So yes, we need to avoid character reencoding, but I think we also need to avoid throwing away hash entropy while we do so. I was wondering about something like getBytes(Characterset.UTF_16BE) -- if it matches the internal representation used by String, will it skip reencoding?

brettwooldridge: @RDearnaley It appears only option would be a custom ``Charset`` and ``CharsetEncoder``.  It is *possible*, as there is a ``CharsetProvider`` SPI available.  The "raw encoder" would simply "encode" chars as pairs of bytes (upper 8-bits, lower 8-bits).  The cost is a 2x sized byte array due to the naive approach, but possibly worth it as the byte array is quickly garbage collected, or could potentially be cached as a ``WeakReference`` in a ``ThreadLocal`` inside of the ``CityHash128Key`` class.

RDearnaley: How strange -- yes, taking a look in the Java code for them, it appears that none of UTF_16BE, UTF_16 etc make use of the obvious optimization that one of them has to be a no-op. I'm wondering if someone might have already created a raw double byte encoding to speed this up -- it seems a fairly obvious performance optimization for cases like this where you care about the entropy content rather than the specific values. If not, it might make a good addition to the Java version of CityHash

RDearnaley: There is another possible approach whose speed would be worth testing:

    private static byte[] toBytes( final String str )
    {
        final byte[] buf= new byte[ 2 * str.length() ];
        for (int i = 0, j = 0; i < str.length(); i++) {
            char c = str.charAt( i );
            buf[j++] = (byte) c;
            buf[j++] = (byte) (c >>> 8);  // Or >> might be faster?
        }
        return buf;
    }

or just inline this since we only use it once.

brettwooldridge: Of course it would work, but getChars() is going to create an additional copy, incurring a CPU hit as well as increasing GC pressure.  Regarding inlining, probably unnecessary as the JIT will inline it when it gets hot.

I think a custom encoder would offer better performance, as the internal char array is passed without copy.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -119,37 +117,31 @@
...
-    static class Sha1HashKey {
-        private byte[] bytes;
+    static class CityHash128Key {
+        private long[] segments;
+        private int hashCode;
 
-        Sha1HashKey(String sql,
+        CityHash128Key(String sql,
                 String parametersDefinition) {
             this(String.format("%s%s", sql, parametersDefinition));
```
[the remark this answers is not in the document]
RDearnaley: Having taken another look, yes, this showed up on my YourKit run -- in fact I saw it roughly as much as the character encoding issue we've been discussing above.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
RDearnaley: Would the str.charAt(i) also be making a copy, or would that just go on the stack/in a register?

brettwooldridge: str.charAt(i) is just a simple array access, with the value passed on the stack.  If the compiler inlines the call, it could be a viable option.  You'd have to find it in the output of ``-XX:+UnlockDiagnosticVMOptions -XX:+LogCompilation -XX:+PrintCompilation``.

RDearnaley:
@cheenamalhotra I have sent you a pull request https://github.com/cheenamalhotra/mssql-jdbc/pull/11 with many of the additional speedups we've discussed above

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -258,28 +255,29 @@ void removeReference() {
...
         JDBCSyntaxTranslator translator = new JDBCSyntaxTranslator();
 
         String parsedSql = translator.translate(sql);
         String procName = translator.getProcedureName(); // may return null
         boolean returnValueSyntax = translator.hasReturnValueSyntax();
-        int paramCount = countParams(parsedSql);
+        int[] parameterPositions = locateParams(parsedSql);
 
-        ParsedSQLCacheItem cacheItem = new ParsedSQLCacheItem(parsedSql, paramCount, procName, returnValueSyntax);
+        ParsedSQLCacheItem  cacheItem = new ParsedSQLCacheItem (parsedSql, parameterPositions, procName, returnValueSyntax);
```
rene-ye: Seems to have an extra space? Might need to apply formatter. Same line 267.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -291,30 +289,36 @@ static ParsedSQLCacheItem parseAndCacheSQL(Sha1HashKey key,
...
-    private static int countParams(String sql) {
-        int nParams = 0;
+     /**
+      * Locate statement parameters.
+      * 
+      * @param sql
+      *          SQL text to parse for positions of parameters to intialize.
+      */
+    private static int[] locateParams(String sql) {
+        List<Integer> parameterPositions = new ArrayList<Integer>(0);
```
rene-ye: Template type inferred from LHS, not necessary to declare it a second time.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -291,30 +289,36 @@ static ParsedSQLCacheItem parseAndCacheSQL(Sha1HashKey key,
...
+            parameterPositions.add(offset);
+        }
 
-        return nParams;
+        // Convert to int[]
+        int[] result = new int[parameterPositions.size()];
+        int i = 0;
+        for (Integer parameterPosition : parameterPositions) {
+            result[i++] = parameterPosition;
+        }
```
rene-ye: Use a LinkedList to make adding/iterating faster OR use a parralell stream with ArrayList. Not necessary to keep `int i`, replace lines 317 - 320 with the following:
`parameterPositions.stream().parallel().forEach(counter -> result[counter] = parameterPositions.get(counter));`

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -5379,7 +5384,13 @@ String replaceParameterMarkers(String sqlSrc,
...
         int paramIndex = 0;
         while (true) {
-            int srcEnd = ParameterUtils.scanSQLForChar('?', sqlSrc, srcBegin);
+            int srcEnd;
+            if (paramIndex >= paramPositions.length) {
+                srcEnd = sqlSrc.length();
+            }
+            else {
+                srcEnd = paramPositions[paramIndex];
+            }
```
rene-ye: 5388 - 5393 can be replaced with
`srcEnd = (paramIndex >= paramPositions.length) ? sqlSrc.length() : paramPositions[paramIndex];`

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -122,42 +123,41 @@
...
-            this(String.format("%s%s", sql, parametersDefinition));
+            this(sql + parametersDefinition);
         }
 
-        Sha1HashKey(String s) {
-            bytes = getSha1Digest().digest(s.getBytes());
+        CityHash128Key(String s) {
+            unhashedString = s;
+            byte[] bytes = new byte[s.length()];
+            s.getBytes(0, s.length(), bytes, 0);
```
peterbae: getBytes is a deprecated method - suppress warning or replace?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
peterbae: List<Integer> parameterPositions = new LinkedList<>(); would be better imo

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
RDearnaley: Suppress the warning -- there is an excellent performance reason why we're using it. And as Brett has pointed out, it's never going away, because it's sometimes the right thing to do.

PR #720, opened 2018-06-12
enabled verifyData in Junit tests

rene-ye:
fixed issues with tests which were the reason why verifyresults was commented out. Also applied formatter to file.


rene-ye:
temporarily closing issue, requires more investigation for datetime/datetime2 precision.

PR #722, opened 2018-06-13
Add support for UTF-8 feature extension.

ulvii:
Support for UTF-8 feature extension that will be introduced in newer versions of SQL Server.


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -4419,6 +4441,8 @@ else if (serverMajorVersion >= 9) // Yukon (9.0) --> TDS 7.2 // Prelogin disconn
 
         len2 = len2 + 1; // add 1 to length becaue of FeatureEx terminator
 
+        len2 = len2 + writeUTF8SupportFeatureRequest(false, tdsWriter);
```
rene-ye: This is a little convoluted, the function basically always returns "5". It also has a Boolean to turn on/off the actual writing. Shouldn't we remove the length calculation from this function and just have a static variable which equals 5? Can also remove the Boolean parameter.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -4598,6 +4622,8 @@ else if (serverMajorVersion >= 9) // Yukon (9.0) --> TDS 7.2 // Prelogin disconn
             writeFedAuthFeatureRequest(true, tdsWriter, fedAuthFeatureExtensionData);
         }
 
+        writeUTF8SupportFeatureRequest(true, tdsWriter);
```
rene-ye: Length is being "calculated" and returned, but we don't use the value. Seems like we're always using half of the functionalities.

src/test/java/com/microsoft/sqlserver/jdbc/unit/UTF8SupportTest.java
```diff
@@ -0,0 +1,159 @@
...
+        validate(String.join("", Collections.nCopies(800, "teststring")));
+        validate(String.join("", Collections.nCopies(800, "тест12")));
+        validate(String.join("", Collections.nCopies(800, "╡╤∞-")));
+    }
+
+    @BeforeAll
+    public static void setUp() throws ClassNotFoundException, SQLException {
+        databaseName = RandomUtil.getIdentifier("UTF8Database");
+        tableName = AbstractSQLGenerator.escapeIdentifier(RandomUtil.getIdentifier("RequestBoundaryTable"));
+        connection = PrepUtil.getConnection(getConfiguredProperty("mssql_jdbc_test_connection_properties"));
```
cheenamalhotra: who closes this connection?

src/test/java/com/microsoft/sqlserver/jdbc/unit/UTF8SupportTest.java
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    private static void createTable(String columnType) throws SQLException {
+        try (Statement stmt = connection.createStatement();) {
+            Utils.dropTableIfExists(tableName, stmt);
+            stmt.executeUpdate("CREATE TABLE " + tableName + " (c " + columnType + ")");
+        }
+    }
+
+    public void clearTable() throws SQLException {
```
cheenamalhotra: Instead of this new function, you could use Utils.dropTableIfExists(..)

src/test/java/com/microsoft/sqlserver/jdbc/unit/UTF8SupportTest.java
```diff
@@ -0,0 +1,162 @@
...
+        validate(String.join("", Collections.nCopies(400, "teststring")) + "1");
+        validate(String.join("", Collections.nCopies(400, "тест12")) + "1");
+        validate(String.join("", Collections.nCopies(400, "╡╤∞-")) + "1");
+
+        createTable("varchar(8000)");
+        validate(String.join("", Collections.nCopies(800, "teststring")));
+        validate(String.join("", Collections.nCopies(800, "тест12")));
+        validate(String.join("", Collections.nCopies(800, "╡╤∞-")));
+
+        createTable("varchar(MAX)");
```
cheenamalhotra: Please don't use hardcoded table names

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
ulvii: This is done to keep the same structure for all feature extension requests, in case we calculate the length differently in the future. Take a look at `writeFedAuthFeatureRequest`, `writeAEFeatureRequest`, `writeDataClassificationFeatureRequest`, The methods are implemented the same way in .NET driver too.

rene-ye: It makes sense with fedAuth because the length is actually variable. For the others, it seems there's a set value(6 for AE, 5 for UTF8, 6 for data classification). I guess it's fine if we're consistent, but I'm just saying that other than the fedAuth function, the others don't really make a lot of sense.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -4130,7 +4140,19 @@ private void onFeatureExtAck(int featureId,
...
                 break;
             }
+            case TDS.TDS_FEATURE_EXT_UTF8SUPPORT: {
+                if (connectionlogger.isLoggable(Level.FINER)) {
+                    connectionlogger.fine(toString() + " Received feature extension acknowledgement for UTF8 support.");
+                }
 
+                if (1 > data.length) {
+                    if (connectionlogger.isLoggable(Level.SEVERE)) {
+                        connectionlogger.severe(toString() + " Unknown value for UTF8 support.");
```
lilgreenbird: use R_unknownUTF8SupportValue here too?

ulvii: no, we don't localize the logger messages.

PR #726, opened 2018-06-25
Removing dependencies from tests that rely on other sources

peterbae:
Removing dependencies and replacing them with equivalent functions in tests.


peterbae:
Unsure why the two files look like they've been modified everywhere, will open a new one.

PR #729, opened 2018-06-26
Remove dependencies from tests that require additional libraries

peterbae:
Remove dependencies from tests that are from outside required libraries.

Replaced:

org.apache.commons.lang3.RandomStringUtils
org.apache.commons.codec.binary.Hex
org.apache.commons.codec.DecoderException;

And replaced with with other methods / java.util functions.


src/test/java/com/microsoft/sqlserver/testframework/sqlType/SqlDateTime2.java
```diff
@@ -48,7 +47,19 @@ public SqlDateTime2() {
...
+        
+        Random rnd = new Random();
+        // this will generate a number that is this.precision number of digits.
+        int randomNumeric;
+        if (this.precision <= 0) {
+            randomNumeric = 0;
+        } else if (this.precision == 1) {
+            randomNumeric = rnd.nextInt(10);
+        } else {
+            randomNumeric = (int) Math.pow(10, this.precision - 1) + rnd.nextInt((int) (9 * Math.pow(10, this.precision - 1)));
```
lilgreenbird: can we save this to a var so we don't have to call Math.pow(10, this.precision - 1) twice?

src/test/java/com/microsoft/sqlserver/testframework/sqlType/SqlTime.java
```diff
@@ -50,7 +49,18 @@ public SqlTime() {
...
+        // this will generate a number that is this.scale number of digits.
+        int randomNumeric;
+        if (this.scale <= 0) {
+            randomNumeric = 0;
+        } else if (this.scale == 1) {
+            randomNumeric = rnd.nextInt(10);
+        } else {
+            randomNumeric = (int) Math.pow(10, this.scale - 1) + rnd.nextInt((int) (9 * Math.pow(10, this.scale - 1)));
+        }
+
```
lilgreenbird: this is the same as above can we move this out to a method which generates n random digits?

src/test/java/com/microsoft/sqlserver/testframework/DBTable.java
```diff
@@ -474,4 +472,11 @@ boolean passDataAsHex(int colNum) {
         return (JDBCType.BINARY == getColumn(colNum).getJdbctype() || JDBCType.VARBINARY == getColumn(colNum).getJdbctype()
                 || JDBCType.LONGVARBINARY == getColumn(colNum).getJdbctype());
```
lilgreenbird: I know this didn't get changed, but would be better if we just do getColumn(colNum).getJdbctype() once

src/test/java/com/microsoft/sqlserver/testframework/sqlType/SqlDateTime.java
```diff
@@ -41,4 +42,22 @@ public SqlDateTime() {
     public Object createdata() {
         return new Timestamp(ThreadLocalRandom.current().nextLong(((Timestamp) minvalue).getTime(), ((Timestamp) maxvalue).getTime()));
     }
+    
+    protected int generateRandomInt(int length) {
```
lilgreenbird: comment header?

src/test/java/com/microsoft/sqlserver/testframework/sqlType/SqlDateTime.java
```diff
@@ -41,4 +42,22 @@ public SqlDateTime() {
...
+        Random rnd = new Random();
+        int power = (int) Math.pow(10, length - 1);
+        
+        //Example of how this works:
+        // if length is 3, then we add 100 + random number between 0~899, so that we get a random number between 100~999.
+        int randomNumeric;
+        if (length <= 0) {
```
lilgreenbird: nitpicking....
we normally do (0 >= blah) etc...

src/test/java/com/microsoft/sqlserver/testframework/sqlType/SqlDateTime.java
peterbae: nah we don't need it

src/test/java/com/microsoft/sqlserver/testframework/DBTable.java
peterbae: cool, done

src/test/java/com/microsoft/sqlserver/testframework/sqlType/SqlDateTime.java
peterbae: i think that's only for comparing to null (for good reasons)

PR #732, opened 2018-06-29
Export driver in automatic module

cheenamalhotra:
- Adds automatic-module-name to Manifest
- All JUnit Tests made public (to support module architecture)


src/test/java/com/microsoft/sqlserver/jdbc/tvp/TVPTypesTest.java
```diff
@@ -1,589 +1,589 @@
...
-        }
-        if (null != rs) {
-            rs.close();
-        }
-        if (null != tvp) {
-            tvp.clear();
-        }
-    }
-
+/*
```
lilgreenbird: why did this whole file change? is it formatting only?

cheenamalhotra: I'll revert the formatting - unnecessary here

cheenamalhotra: Somehow GitHub is doing unnecessary comparisons, I compared the file to dev branch's and there's no formatting change actually. The only change is in changing private Junit test functions to public.

PR #746, opened 2018-07-12
Adding a new test for beginRequest()/endRequest()

ulvii:
The goal is to make the build fail if there are new public non-static methods in SQLServerConnection and notify the developer to decide whether it needs to be handled by beginRequest()/endRequest().

### DIFF
--- a/src/test/java/com/microsoft/sqlserver/jdbc/connection/RequestBoundaryMethodsTest.java
+++ b/src/test/java/com/microsoft/sqlserver/jdbc/connection/RequestBoundaryMethodsTest.java
@@ -374,7 +374,7 @@ public class RequestBoundaryMethodsTest extends AbstractTest {
     public void testNewMethods() {
         Method[] methods = SQLServerConnection.class.getDeclaredMethods();
         for (Method method : methods) {
-            assertTrue(isVerfied(method),
+            assertTrue(isVerified(method),
                     "A failure is expected if you are adding a new public non-static method to SQLServerConnection."
                             + " See the test for instructions on how to fix the failure. ");
         }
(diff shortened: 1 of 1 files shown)


src/test/java/com/microsoft/sqlserver/jdbc/connection/RequestBoundaryMethodsTest.java
```diff
@@ -412,4 +436,56 @@ private void compareValuesAgainstConnection(SQLServerConnection con, boolean aut
     private void generateWarning(SQLServerConnection con) throws SQLException {
         con.setClientInfo("name", "value");
     }
+
+    private boolean isVerified(Method method) {
+        return (!Modifier.isPublic(method.getModifiers()) || Modifier.isStatic(method.getModifiers())
+                || method.getName().startsWith("get") || getVerifiedMethodNames().contains(method.getName()));
```
peterbae: why do we allow any method that starts with "get" to be verified? if I added a new connection property and added a public getter for it, isn't it something that the boundary request method needs to handle? (it would likely get caught by the setter that comes with it, though)

rene-ye: The point of beginRequest() is to "memorize" the connection state and its properties. Getting a connection property doesn't affect the properties so it shouldn't matter. On the other hand, it is kind of hacky and I wouldn't be opposed to triggering this error on all newly added public methods, just to be safe.

ulvii: Well, a getter won't change the connection local state. And I didn't want to hard-code all of them in the test.

Issue #744, opened 2018-07-11
add support for getObject(n, java.time.LocalDateTime.class)


gordthompson:
(Continuation of #339. Also prompted by [this Stack Overflow question](https://stackoverflow.com/q/51273979/2144390).)

As of `mssql-jdbc-6.5.4.jre8-preview.jar` the driver still does not support the following statement when retrieving a value from a `datetime2` column

```java
java.time.LocalDateTime x = rs.getObject(1, java.time.LocalDateTime.class);
```

failing with

> com.microsoft.sqlserver.jdbc.SQLServerException: The conversion to class java.time.LocalDateTime is unsupported.

If the objective is to support Java 8 then the driver really should be able to handle this.

PR #749, opened 2018-07-16
Add support for LocalDate, LocalTime and LocalDateTime to be passed as 'type' in ResultSet.getObject()

gordthompson:
Issue #744 - support for rs.getObject(n, java.time.LocalDateTime.class)

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResultSet.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResultSet.java
@@ -2381,15 +2381,14 @@ public class SQLServerResultSet implements ISQLServerResultSet, java.io.Serializ
             if (ts == null) {
                 returnValue = null;
             } else {
+                java.time.LocalDateTime ldt = java.time.LocalDateTime
+                        .ofInstant(ts.toInstant(), java.time.ZoneId.of("UTC")); 
                 if (type == java.time.LocalDateTime.class) {
-                    returnValue = java.time.LocalDateTime.ofInstant(ts.toInstant(), 
-                            java.time.ZoneId.of("UTC"));
+                    returnValue = ldt;
                 } else if (type == java.time.LocalDate.class) {
-                    returnValue = java.time.LocalDate.ofInstant(ts.toInstant(), 
-                            java.time.ZoneId.of("UTC"));
+                    returnValue = ldt.toLocalDate();
                 } else {
-                    returnValue = java.time.LocalTime.ofInstant(ts.toInstant(), 
-                            java.time.ZoneId.of("UTC"));
+                    returnValue = ldt.toLocalTime();
                 }
             }
         } else if (type == microsoft.sql.DateTimeOffset.class) {
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResultSet.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResultSet.java
@@ -2373,13 +2373,24 @@ public class SQLServerResultSet implements ISQLServerResultSet, java.io.Serializ
             returnValue = getTime(columnIndex);
         } else if (type == java.sql.Timestamp.class) {
             returnValue = getTimestamp(columnIndex);
-        } else if (type == java.time.LocalDateTime.class) {
-            returnValue = null;
+        } else if (type == java.time.LocalDateTime.class 
+                || type == java.time.LocalDate.class 
+                || type == java.time.LocalTime.class) {
             java.sql.Timestamp ts = getTimestamp(columnIndex,
                     Calendar.getInstance(java.util.TimeZone.getTimeZone("UTC")));
-            if (ts != null) {
-                returnValue = java.time.LocalDateTime.ofInstant(ts.toInstant(), 
-                        java.time.ZoneId.of("UTC"));
+            if (ts == null) {
+                returnValue = null;
+            } else {
+                if (type == java.time.LocalDateTime.class) {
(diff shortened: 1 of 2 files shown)


cheenamalhotra:
Hi @gordthompson 

Thanks for coming up with PR.

We should also add related API for `SQLServerResultSet.getLocalDateTime(...)` [also defined in `ISQLServerResultSet `interface and overridden].

I would also like to know your thoughts about providing `LocalDateTime` support in `PreparedStatement.setLocalDateTime()` as well? Although internally the driver would process it same way, but would it make sense for the sake of `LocalDateTime` datatype support?

mrotteveel:
@cheenamalhotra No such method exists in the JDBC (`java.sql`) API itself, so adding such methods is not very effective, as most people will only be using methods exposed by the JDBC API itself (casting or unwrapping has a lot of caveats when using connection pools for example). I'd urge you to not add them, because it will be simpler and less confusing.

When support for JSR-310 (`java.time`) objects like `LocalDateTime` (and others) in was added in JDBC 4.2, the JDBC expert group decided that adding specific `getXXX`/`setXXX`/`updateXXX` methods for these types was not necessary. 

See this recent comment on the jdbc-spec-discuss mailing list: http://mail.openjdk.java.net/pipermail/jdbc-spec-discuss/2018-July/000361.html by Lance Andersen, the JDBC spec-lead.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResultSet.java
```diff
@@ -2373,6 +2373,16 @@ public Object getObject(int columnIndex) throws SQLServerException {
             returnValue = getTime(columnIndex);
         } else if (type == java.sql.Timestamp.class) {
             returnValue = getTimestamp(columnIndex);
+        } else if (type == java.time.LocalDateTime.class) {
+            returnValue = null;
+            java.sql.Timestamp ts = getTimestamp(columnIndex,
+                    Calendar.getInstance(java.util.TimeZone.getTimeZone("UTC")));
```
mrotteveel: I'm not sure if this is correct given JDBC specifies that the time should be rendered in the JVM default time zone.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResultSet.java
```diff
@@ -2373,6 +2373,16 @@ public Object getObject(int columnIndex) throws SQLServerException {
...
+            if (ts != null) {
+                java.time.format.DateTimeFormatter dtf = java.time.format.DateTimeFormatter
```
mrotteveel: Parsing to string and back is inefficient. Consider using `Timestamp.toLocalDateTime()` (assuming my previous comment is addressed), or `LocalDateTime.ofInstant(ts.getInstant(), <zone offset you want>)`.

mrotteveel:
@gordthompson I left two comments regarding the conversion. I also wonder if you considered adding support for `LocalTime`, `OffsetDateTime` and `OffsetTime` which were also added in JDBC 4.2.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResultSet.java
gordthompson: @mrotteveel The Timestamp needs to be retrieved as UTC to prevent certain date/time values from being corrupted if they are in the "lost hour" when switching from Standard Time to Daylight Time (a.k.a. Summer Time). For example, in my time zone (America/Edmonton), the datetime2 value `2018-03-11T02:00:00` will magically change to `2018-03-11T03:00:00` if we try to retrieve it in the default time zone since there was no `02:00` that morning; the time changed from `01:59` to `03:00`.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResultSet.java
gordthompson: @mrotteveel Thanks. I went with the latter.

gordthompson:
@mrotteveel I focused on LocalDateTime because that was where people could get stung by the "lost hour" issue (e.g., [this SO answer](https://stackoverflow.com/a/51276605/2144390)). SQL Server `datetime2` really *is* a Java `LocalDateTime` (since datetime2 has no time zone component) so I thought it was most important to support that conversion directly.

gordthompson:
@cheenamalhotra The latest [Travis CI failure](https://travis-ci.org/Microsoft/mssql-jdbc/builds/405006021?utm_source=github_status&utm_medium=notification) appears to be related to Travis CI itself; it never got as far as testing the code.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResultSet.java
mrotteveel: Clear. You may want to consider skipping timestamp retrieval and instead convert directly from the datetime value to `LocalDateTime`; although that might also be considered as a future improvement.

gordthompson:
@mrotteveel For (more) completeness I also added support for LocalDate and LocalTime, also achieved by first retrieving a Timestamp as UTC and then converting. I agree that it might be more efficient to bypass the Timestamp and retrieve the values directly from the TDS payload (as I did for a pyodbc Output Converter Function [here](https://github.com/mkleehammer/pyodbc/wiki/Using-an-Output-Converter-function)) , but I strongly suspect that the Java code would then have to do significantly different things depending on the column type from which the value was retrieved. That's a can of worms that I don't have time to open right now, but perhaps it should be noted as a possible future enhancement.

PR #756, opened 2018-07-23
Disallow non-parameterized queries for Bulk Copy API for batch insert

peterbae:
As for the error messages, I am throwing an illegalArgumentException but it gets caught and gets turned into a logging message, so this part doesn't need translation into other languages.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
@@ -1975,58 +1975,57 @@ public class SQLServerPreparedStatement extends SQLServerStatement implements IS
                 checkAdditionalQuery();
 
                 String destinationTableName = tableName;
-                SQLServerStatement stmt = (SQLServerStatement) connection.createStatement(ResultSet.TYPE_FORWARD_ONLY,
-                        ResultSet.CONCUR_READ_ONLY, connection.getHoldability(), stmtColumnEncriptionSetting);
+                try (SQLServerStatement stmt = (SQLServerStatement) connection.createStatement(
+                        ResultSet.TYPE_FORWARD_ONLY, ResultSet.CONCUR_READ_ONLY, connection.getHoldability(),
+                        stmtColumnEncriptionSetting)) {
+                    // Get destination metadata
+                    try (SQLServerResultSet rs = stmt.executeQueryInternal(
+                            "sp_executesql N'SET FMTONLY ON SELECT * FROM " + destinationTableName + " '");) {
 
-                // Get destination metadata
-                try (SQLServerResultSet rs = stmt.executeQueryInternal(
-                        "sp_executesql N'SET FMTONLY ON SELECT * FROM " + destinationTableName + " '");) {
-
-                    if (null != columnList && columnList.size() > 0) {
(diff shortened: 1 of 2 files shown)


src/test/java/com/microsoft/sqlserver/jdbc/preparedStatement/BatchExecutionWithBulkCopyTest.java
```diff
@@ -500,6 +502,96 @@ public void testAlColumnsLargeBatch() throws Exception {
...
+            f1.set(connection, true);
+
+            pstmt.setInt(1, 1);
+            pstmt.setInt(2, 1);
+            pstmt.setInt(3, 1);
+            pstmt.setInt(4, 1);
+            pstmt.addBatch();
+
+            pstmt.executeBatch();
+            throw new Exception("shouldn't come here.");
```
rene-ye: Can we throw "Test did not throw an exception when it was Expected" instead? I think there's existing tests that do this and we should be consistent.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -1964,6 +1964,14 @@ public final void clearBatch() throws SQLServerException {
                 ArrayList<String> columnList = parseUserSQLForColumnListDW();
                 ArrayList<String> valueList = parseUserSQLForValueListDW(false);
 
+                for (int i = 0; i < valueList.size(); i++) {
+                    if (!valueList.get(i).equals("?")) {
+                        // throw IllegalArgumentException and fallback to original logic for batch insert
+                        throw new IllegalArgumentException(
+                                "Only fully parameterized queries are allowed for using Bulk Copy API for batch insert at the moment.");
```
cheenamalhotra: Why don't we fallback to batch insert Original implementation instead of throwing exception here? It may not be a query issue, its just the 'useBulkCopyWithBatchInserts' API will not parse partial parameterized queries. Applies to all other exceptions.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -1964,6 +1964,16 @@ public final void clearBatch() throws SQLServerException {
...
+                    }
+                }
+
+                checkAdditionalQuery();
+
                 String destinationTableName = tableName;
                 SQLServerStatement stmt = (SQLServerStatement) connection.createStatement(ResultSet.TYPE_FORWARD_ONLY,
```
cheenamalhotra: Statement is created here, and closed in next try's finally block?

peterbae: that try block is closing the resultset. this statement is closed during finally.

ulvii: @peterbae , you should move statement creation to try-with resources.

ulvii: @peterbae , Why don't you just move statement creation to try-with-resources?

peterbae: moved.

PR #763, opened 2018-07-30
Improve performance of readLong function unrolling loop and using bitwise operators instead of additions

robertonr:
Hello, I'm using mssql-jdbc driver for long time and profiling my application for 20 minutes I found that TDSChannel.read() is taking around 548297 milliseconds of CPU (I guess it's due to synchronization). However improving readLong method I reduced around 30000 milliseconds of pure CPU usage.

Could you merge this small change, please?


rene-ye:
Hi @robertonr, thank you for your contribution to the mssql-jdbc driver. Please restrict the visibility of `writeLong` as to not introduce any new public APIs. If the concern is allowing the function to be tested, the team can do that. Other than that the changes look good.

robertonr:
Hello, the chage of visibility is done, thank you!

PR #773, opened 2018-08-03
Degradation fix for column encryption enabled (AE)

rene-ye:
Fixed an issue where the logger was needlessly traversing the stack to obtain method name.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerAeadAes256CbcHmac256Algorithm.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerAeadAes256CbcHmac256Algorithm.java
@@ -34,7 +34,6 @@ class SQLServerAeadAes256CbcHmac256Algorithm extends SQLServerEncryptionAlgorith
             .getLogger("com.microsoft.sqlserver.jdbc.SQLServerAeadAes256CbcHmac256Algorithm");
 
     final static String algorithmName = "AEAD_AES_256_CBC_HMAC_SHA256";
-    static Cipher decryptCipher = null;
 
     // Stores column encryption key which includes root key and derived keys
     private SQLServerAeadAes256CbcHmac256EncryptionKey columnEncryptionkey;
@@ -48,6 +47,8 @@ class SQLServerAeadAes256CbcHmac256Algorithm extends SQLServerEncryptionAlgorith
     private byte[] version = new byte[] {0x01};
     // Added so that java hashing algorithm is similar to c#
     private byte[] versionSize = new byte[] {1};
+    // Cipher used to decrypt
+    private Cipher decryptCipher = null;
 
     /*
      * Minimum Length of cipherText without authentication tag. This value is 1 (version byte) + 16 (IV) + 16 (minimum
@@ -107,8 +108,7 @@ class SQLServerAeadAes256CbcHmac256Algorithm extends SQLServerEncryptionAlgorith
      * @throws SQLServerException
      */
     protected byte[] encryptData(byte[] plainText, boolean hasAuthenticationTag) throws SQLServerException {
-        aeLogger.entering(SQLServerAeadAes256CbcHmac256Algorithm.class.getName(),
-                "encryptData", "Encrypting data.");
+        aeLogger.entering(SQLServerAeadAes256CbcHmac256Algorithm.class.getName(), "encryptData", "Encrypting data.");
         // we will generate this initialization vector based whether
         // this encryption type is deterministic
         assert (plainText != null);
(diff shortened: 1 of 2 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerAeadAes256CbcHmac256Algorithm.java
```diff
@@ -290,31 +298,30 @@
...
         byte[] plainText = null;
         // key to be used for decryption
         SecretKeySpec skeySpec = new SecretKeySpec(columnEncryptionkey.getEncryptionKey(), "AES");
         IvParameterSpec ivector = new IvParameterSpec(iv);
-        Cipher decryptCipher;
         try {
             // AES encryption CBC mode and PKCS5 padding
-            decryptCipher = Cipher.getInstance("AES/CBC/PKCS5Padding");
-            decryptCipher.init(Cipher.DECRYPT_MODE, skeySpec, ivector);
+            if (decryptCipher != null) {
```
saurabh500: Why not instantiate and initialize the decryptCipher here itself? Why is a static member to this class needed when the `Cipher.getInstance(...)` itself is static?

saurabh500:
What is the cause of degradation? The logger traversing the stack or lack of initialization of cipher?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerAeadAes256CbcHmac256Algorithm.java
saurabh500: Looks like the code earlier was doing what I am talking about, above. Why is this change for declaring `decryptCipher` as a static class member needed?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerAeadAes256CbcHmac256Algorithm.java
```diff
@@ -80,6 +82,12 @@
         }
         this.algorithmVersion = algorithmVersion;
         version[0] = algorithmVersion;
+
+        try {
+            decryptCipher = Cipher.getInstance("AES/CBC/PKCS5Padding");
+        } catch (NoSuchAlgorithmException | NoSuchPaddingException e) {
+            // not possible to get here
```
SQLFang: From the comment, it sounds like these exceptions is not expected to ever happen. How come we are try-catching and then swallowing these exceptions then? Shouldn't we let these exceptions get thrown if something unexpected occurs?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerAeadAes256CbcHmac256Algorithm.java
rene-ye: There is no reason for the decryptCipher variable to be static, I will change this. However, I believe it should be done once in the constructor as there isn't a point to get an instance for every decryptData(byte,byte,int,int) call.

saurabh500: The issue I see with having this in the constructor is that you are replicating the behavior of Cipher.getInstance() which is static itself and adding this additional layer of caching and making assumption about the API that it would be safe to use the same object when the API itself doesn't make any such guarantees in the documentation. 
A static is good if you want to cache the object at your end and it is safe to do so, but since Cipher.getInstance(...) is static itself, it should decide whether it should give a new object everytime or return a cached object. 

In case the behavior of Cipher.getInstance() changes for any reason across JDK versions, then this driver will have to comb through this caching. Such bugs are hard to detect. Such layers of caching of objects returned from static call is not a good practice.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerAeadAes256CbcHmac256Algorithm.java
saurabh500: These exceptions must be logged. I can see the source of OpenJDK where this exception can be thrown. In case the Cipher is not found in the security providers, then an exception can definitely surface.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerAeadAes256CbcHmac256Algorithm.java
rene-ye: Agreed, I will revert the changes.

rene-ye:
Uploading jars. These jars contain the final changes that will be going into the driver.
[AE Jars.zip](https://github.com/Microsoft/mssql-jdbc/files/2267978/AE.Jars.zip)

Issue #775, opened 2018-08-06
SQLServerBulkCopy / getDestinationMetadata fails when there's an apostrophe in the table name


Gaff:
## Driver version or jar name
mssql-jdbc:6.4.0-jre8

## Problem description
When attempting to bulk upload to a table with `'` in the name the upload fails with:

    com.microsoft.sqlserver.jdbc.SQLServerException: Unable to retrieve column metadata.
	at 
 com.microsoft.sqlserver.jdbc.SQLServerBulkCopy.getDestinationMetadata(SQLServerBulkCopy.java:1781)

The code does queries such as:

    select collation_name from sys.columns where " + "object_id=OBJECT_ID('" + destinationTableName + "') 

Which fails because `destinationTableName` contains unescaped `'`

cheenamalhotra:
Hi @Gaff 

Thanks for bringing forward this issue with us. I've created a [PR](https://github.com/Microsoft/mssql-jdbc/pull/780) that fixes this issue in the driver. Please take a look and let me know for any changes needed.

Gaff:
Wow thanks for the quick turnaround! Looks good (in so much as I can tell, I'm not going to claim to be an expert).

cheenamalhotra:
Hi @Gaff 

PR #780 has been merged and will be available since v7.1.1 release version. Closing issue.

Issue #778, opened 2018-08-08
Executing a PreparedStatement with SET SHOWPLAN_ALL ON or SET SHOWPLAN_XML ON does not return a ResultSet.


rtwruck:
## Driver version or jar name
mssql-jdbc-7.0.0.jre8.jar

## SQL Server version
12.00.5571

## Client operating system
Linux

## Java/JVM version
Java(TM) SE Runtime Environment (build 1.8.0_144-b01)
Java HotSpot(TM) 64-Bit Server VM (build 25.144-b01, mixed mode)

## Table schema
any

## Problem description
Executing a PreparedStatement with SET SHOWPLAN_ALL ON or SET SHOWPLAN_XML ON does not return a ResultSet.
It makes no difference whether prepareStatement is called before or after issuing the SET statement.
A ResultSet with the expected plan is returned when using createStatement().execute(sqlText) - but that obviously does not work when the statement takes parameters.
In addition, there's a NullPointerException when you call PreparedStatement.getParameterMetaData() with SET SHOWPLAN in effect, presumably because the returned ResultSet contains the plan:
	at com.microsoft.sqlserver.jdbc.SQLServerParameterMetaData.parseQueryMeta(SQLServerParameterMetaData.java:211)
	at com.microsoft.sqlserver.jdbc.SQLServerParameterMetaData.<init>(SQLServerParameterMetaData.java:602)
	at com.microsoft.sqlserver.jdbc.SQLServerPreparedStatement.getParameterMetaData(SQLServerPreparedStatement.java:3057)
	at com.microsoft.sqlserver.jdbc.SQLServerPreparedStatement.getParameterMetaData(SQLServerPreparedStatement.java:3068)

## Expected behavior and actual behavior
Expected:
connection.createStatement().execute("SET SHOWPLAN_ALL ON") == false &&
connection.prepareStatement(sqlText).execute() == true
Actual:
connection.createStatement().execute("SET SHOWPLAN_ALL ON") == false &&
connection.prepareStatement(sqlText).execute() == false

## Repro code
See above.

rene-ye:
[...]

PR #780, opened 2018-08-08
Fix issues with apostrophe being passed in table name + Improvements in SQLServerParameterMetadata

cheenamalhotra:
- Fixes issue #775 
- Fixes Null Pointer Exception reported in #778
- Updates Junit tests to test apostrophe(') in randomized table name


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -2123,8 +2124,9 @@ public final void clearBatch() throws SQLServerException {
                 try (SQLServerStatement stmt = (SQLServerStatement) connection.createStatement(
                         ResultSet.TYPE_FORWARD_ONLY, ResultSet.CONCUR_READ_ONLY, connection.getHoldability(),
                         stmtColumnEncriptionSetting);
-                        SQLServerResultSet rs = stmt.executeQueryInternal(
-                                "sp_executesql N'SET FMTONLY ON SELECT * FROM " + tableName + " '");) {
+                        SQLServerResultSet rs = stmt
```
lilgreenbird: rs not closed

cheenamalhotra: Its inside try-w-resource try block - hence does get closed

lilgreenbird: oops missed that

src/main/java/com/microsoft/sqlserver/jdbc/Util.java
```diff
@@ -985,6 +985,17 @@ static synchronized boolean checkIfNeedNewAccessToken(SQLServerConnection connec
...
     }
+
+    /**
+     * Escapes single quotes (') in object name to convert and pass it as String safely.
+     * 
+     * @param name
+     *        Object name to be passed as String
+     * @return Converted object name
+     */
+    static String escapeQuotes(String name) {
```
ulvii: Can we rename this method to `escapeSingleQuotes`?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerParameterMetaData.java
```diff
@@ -208,91 +206,98 @@ private String parseInsertColumns(String sql, String columnMarker) throws SQLSer
...
-                } else {
-                    qm.precision = rsQueryMeta.getInt("suggested_precision");
-                    qm.scale = rsQueryMeta.getInt("suggested_scale");
-
-                    Matcher matcher = datatypePattern.matcher(typename);
-                    if (matcher.matches()) {
-                        // the datatype has some precision/scale defined explicitly.
-                        ssType = SSType.of(matcher.group(1));
-                        if (typename.equalsIgnoreCase("varchar(max)") || typename.equalsIgnoreCase("varbinary(max)")) {
+            if (null != rsQueryMeta) {
```
ulvii: Don't we need the same check for `parseQueryMetaFor2008`?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerParameterMetaData.java
```diff
@@ -557,28 +561,47 @@ private void checkClosed() throws SQLServerException {
...
-                }
-                rsProcedureMeta.beforeFirst();
-
-                // Sixth is DATA_TYPE
-                rsProcedureMeta.getColumn(6).setFilter(new DataTypeFilter());
-                if (con.isKatmaiOrLater()) {
-                    rsProcedureMeta.getColumn(8).setFilter(new ZeroFixupFilter());
-                    rsProcedureMeta.getColumn(9).setFilter(new ZeroFixupFilter());
-                    rsProcedureMeta.getColumn(17).setFilter(new ZeroFixupFilter());
+                // DO NOT close the Statement as resultSet 'rsProcedureMeta' is required by other APIs
```
ulvii: Try-with-resources will close the statement though. Is this comment still accurate?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerParameterMetaData.java
```diff
@@ -685,25 +704,26 @@ public boolean isWrapperFor(Class<?> iface) throws SQLException {
...
 
-    private void verifyParameterPosition(int param) throws SQLServerException {
+    private Map<String, Object> getParameterInfo(int param) throws SQLServerException {
         boolean bFound = false;
-        try {
-            if (((SQLServerPreparedStatement) stmtParent).bReturnValueSyntax && isTVP) {
-                bFound = rsProcedureMeta.absolute(param);
-            } else {
-                bFound = rsProcedureMeta.absolute(param + 1); // Note row 1 is the 'return value' meta data
+        if ((stmtParent).bReturnValueSyntax && isTVP) {
```
ulvii: I can't figure out why the variable would be named `bFound`. Please rename it to a descriptive one.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerParameterMetaData.java
```diff
@@ -557,28 +561,47 @@ private void checkClosed() throws SQLServerException {
...
+                        rsProcedureMeta.getColumn(17).setFilter(new ZeroFixupFilter());
+                    }
+
+                    procMetadata = new ArrayList<>();
+
+                    // Process ResultSet Procedure Metadata for API usage
+                    while (rsProcedureMeta.next()) {
+                        procMetadata.add(new HashMap<String, Object>() {
+                            {
+                                put("DATA_TYPE", rsProcedureMeta.getShort("DATA_TYPE"));
```
ulvii: I think we should have some kind of an exception handling here. 
If one of the `ResultSet` getters fails here, the exception message will be misleading. The `catch` block you removed below can be moved up here.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerParameterMetaData.java
```diff
@@ -735,34 +754,25 @@ public String getParameterClassName(int param) throws SQLServerException {
...
-            }
-        } catch (SQLException e) {
-            SQLServerException.makeFromDriverError(con, stmtParent, e.toString(), null, false);
-            return 0;
+        if (procMetadata == null) {
+            // PreparedStatement
+            return queryMetaMap.size();
+        } else {
+            // Row 1 is Return Type metadata
+            return procMetadata.size() - 1;
```
ulvii: what if the size is 0?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerParameterMetaData.java
cheenamalhotra: There is a catch block in this method at the end. The exception handling that was done for these getters is done same way here.

ulvii: Not really, the catch block will just re-throw the exception. 
```
catch (SQLServerException e) {
    throw e;
}
```

cheenamalhotra: So what's the problem with that? We don't need to wrap SQLServerException again!

Issue #755, opened 2018-07-21
TimeoutTimer thread pool max


shayaantx:
Hello,

I noticed the TimeoutTimer in older versions started a thread for every query that had a query timeout greater than 0. Then I saw in https://github.com/Microsoft/mssql-jdbc/issues/150 someone had reported it. https://github.com/Microsoft/mssql-jdbc/pull/162 fixed it.

Not sure if I'm missing something, but how is this problem fixed? In TimeoutTimer the executor service is using a cached thread pool via java's helper methods in ExecutorService. This creates a thread pool with a core pool size of 0 and a max pool size of Integer.MAX. If I'm reading that correctly, if I spawn 1k queries at once with timeouts, I will have 1k additional timeout threads (since no max on the thread pool exists). Seems like the problem still exists, you've just capped the max number of threads that could get created from unlimited to Integer.MAX by using a thread pool.

How come we can't do something like what the oracle jdbc driver does? They have one thread that all queries (that have a timeout) register with and that thread polls every second checking if they should interrupt each thread running the query?

Forgive me if I'm missing something with the original proposed fix

ulvii:
Hi @shayaantx ,

Please provide a repro code, we will investigate and get back to you.

shayaantx:
I can attempt to write some reproducible code, but I mean I think the latest code highlights the issue I described.

In the TDSCommand and TDSReader objects, you guys use the class TimeoutTimer (if users of the driver set a query timeout greater than 0). This class has two methods that use a static final ExecutorService that is created once via Executors.newCachedThreadPool(new ThreadFactory()). That specific implementation makes a thread pool that has a max size of Integer.MAX_VALUE, this type of thread pool will reuse threads if they are available, but since the max is such a large number you can run the pool up as many threads as you need.

So if I run 5k queries at once with 5k timeouts configured, depending on how fast each query executes, you could have 10k threads in use, instead of 5k. Obviously this is an extreme case, but the same problem would appear with 500 simultaneous queries configured with timeouts (which in my use cases, is normal amount of queries to run in parallel).

The design here of the query timeout logic seems inefficient. You are creating one timer thread for every query being run at any given time with a query timeout. Oracle doesn't have this problem, mysql has the same issue (https://github.com/mysql/mysql-connector-j/blob/1f61b0b0270d9844b006572ba4e77f19c0f230d4/src/com/mysql/jdbc/StatementImpl.java#L67).

[...]

PR #782, opened 2018-08-11
TimeoutTimer to TimeoutPoller performance enhancement

shayaantx:
Modified the TimeoutTimer to not be a unlimited thread pool, instead its one thread that lets TDS commands register/deregister with it, when a timeout is configured, and that thread will timeout the command if necessary. Included test cases

See https://github.com/Microsoft/mssql-jdbc/issues/755 for more context

Please let me know of any necessary changes or requests, I'm not sure why git is showing all these white space changes, but in the git history in my ide it doesn't show...


shayaantx:
no idea why my change would cause the sql2017 build to fail for below results, I'll run them later locally and look into it

[ERROR] Failures: 
[ERROR]   bvtTest.testTwoResultsetsDifferentStmt:374 Unexpected bigint value, expected: 5869790151273323877 .Retrieved: -1985958062194855290 ==> expected: <true> but was: <false>
[ERROR]   bvtTest.testTwoResultsetsSameStmt:409 Unexpected bigint value, expected: 5869790151273323877 .Retrieved: -1985958062194855290 ==> expected: <true> but was: <false>

cheenamalhotra:
@shayaantx Yes, we are aware of the intermittent test failures, and have a fix coming soon with PR #762. Sorry for any inconvenience.

cheenamalhotra:
Hi @shayaantx 

About whitespace changes, please make sure your code is formatted with [mssql-jdbc-formatter](https://github.com/Microsoft/mssql-jdbc/blob/dev/mssql-jdbc_formatter.xml).

shayaantx:
@cheenamalhotra I fixed the white space changes, please let me know if anything else is necessary

shayaantx:
@cheenamalhotra anything else required to get this submitted?

ulvii:
Hi @shayaantx ,

The team will review the changes for the next release, please stay tuned.

cheenamalhotra:
Hi @shayaantx 

Please merge PR [#1](https://github.com/shayaantx/mssql-jdbc/pull/1) in your forked repository so we can proceed with testing this PR in internal test lab thoroughly.

shayaantx:
@cheenamalhotra done

Please let me know of any issues and I can address them

cheenamalhotra:
Thanks @shayaantx for the quick turnaround!

We reviewed the Pull Request, and were wondering if you'd consider extending the implementation to [`BulkTimeoutTimer`](https://github.com/Microsoft/mssql-jdbc/blob/0a1759f7b95ab38f45dbefec370e0eadf4776e9d/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java#L253) as well. In this case, it would also be neater to maintain a separate Java class e.g. `TimeoutHandler` (or anything appropriate) for holding `TimeoutPoller` and `TimeoutCommand` classes instead of keeping them in `IOBuffer`.

shayaantx:
@cheenamalhotra I've made the changes to use TimeoutPoller in the bulk copy mechanism.

I was looking at BulkCopyTimeoutTest#testZeroTimeOut and I don't think this test actually will work with my changes and I don't think it worked before my changes based on how I think it was initially designed.

The test bulk copy takes like 100ms locally, so a zero timeout, should trigger a timeout exception, however this test case is **not** designed to expect a exception thrown, so it doesn't actually validate a timeout is thrown.

I'd like to propose adding jmockit to pom.xml to include the library to easily mock parts of the bulk copy (without large refactoring) to make this test actually work (basically mock core parts of the bulk copy to introduce an actual way to timeout the copy, simulating the timeout).  Please let me know if you'd like to see an example in my fork and I'll put it up

cheenamalhotra:
Thanks for the changes @shayaantx . Very much appreciated.

JMockit would be worth looking into, please feel free to setup a PR sometime that demonstrates your example, we'll review and take it from there and will try scaling it up if it proves useful :)

ulvii:
Hi @shayaantx ,

Great work, we appreciate your efforts. 👍 

I took a quick look at the PR and wanted to mention a few things before the team proceeds with the detailed review. 

1. _The test bulk copy takes like 100ms locally, so a zero timeout, should trigger a timeout exception._
-Note that 0 timeout indicates no limit; the driver waits indefinitely to complete the bulk copy operation. See `BulkCopyTimeout` [here](https://docs.microsoft.com/en-us/sql/connect/jdbc/using-bulk-copy-with-the-jdbc-driver?view=sql-server-2017#sqlserverbulkcopyoptions). **Do you still think JMockit is required?** 
2. I also noticed you added `-javaagent:${settings.localRepository}/org/jmockit/jmockit/${jmockit.version}/jmockit-${jmockit.version}.jar` to JVM arguments in pom.xml to get JMockit working. This means if someone wants to run the tests without using Maven, on Eclipse for example, they have to manually provide `javaagent` argument. I will wait for the team to decide on this. 
3. _I also moved TDSWriter/TDSCommand out of IOBuffer, I really wasn't sure if you guys/gals would approve such a change, but I need these classes accessible to do some mocking._
-TDSWriter/TDSCommand should not be exposed to applications, they should remain package-private. 
4. Please also apply the [formatter](https://github.com/Microsoft/mssql-jdbc/blob/dev/mssql-jdbc_formatter.xml) to your changes, remove unused imports, add license header to the new files. 
5. Please also remove ErrorConstants.java and directly use `R_queryTimedOut` and `R_connectionIsClosed`, to keep it consistent in the driver. 
5. _I noticed that if I timeout a bulk copy at certain times, the driver will correctly throw the timeout exception, however if we try to do anything else with the statement, sql server throws "Bulk load data was expected but not sent"._
-This should not really happen, statement has to be available even if bulk copy operation failed with an exception. I tried a few test cases, also BulkCopyTimeoutTest, but could not reproduce the issue. We recently merged a [PR](https://github.com/Microsoft/mssql-jdbc/pull/801) which was meant to fix the issues with bulk copy error handling. If you could provide me a simple repro code, I can take a look if it is an actual driver issue.

shayaantx:
@ulvii 

Thanks for the response

1. That comment by me was invalid since a zero timeout like you mentioned doesn't trigger a timeout, Something like* jmockit is helpful for a test like testRealTimeout() where we want to test bulk copy functionality and the timeout functionality (since thats what I changed)

2. Yes, you would need to add that or we would need to confirm the junit tests differently. It is definitely possible to configure your junit runner/suite to not have this required, but I'd need to spend some time investigating it.

3. Yea I figured that wouldn't be ideal. If using the mocking framework is desired, then tests for package private classes would need to be in same package in the test src (so in this case this one bulk copy timeout test would need to be moved from com.microsoft.sqlserver.jdbc.bulkCopy to com.microsoft.sqlserver.jdbc.

4. I've been applying the formatter with eclipse, which file was missing the formatting?

5. Can do

6. Replace the below block of code in BulkCopyTestUtil#performBulkCopyWithoutErrorHandling (line#370)
`
		// we drop the destination table in a separate connection because we could of
		// timed out before sending
		// a batch so then any subsequent operations in sql server will result in an
		// error "Bulk load data was expected but not sent"
		dropDestinationTable(wrapper, destinationTable);`

with

`finally {
                if (dropDest && null != destinationTable) {
                    stmt.dropTable(destinationTable);
                }
            }`

shayaantx:
If it is simpler I can go back to my changes before the jmockit pr, but someone on your end should confirm my changes are valid (I tested them and they seem fine). It's just weird for me to commit any code without some test case covering most of the edge cases

Please let me know what your preference is and I'll get it done

cheenamalhotra:
Hi @shayaantx 

Please make changes as suggested by @ulvii above. The file changes are making difficult to review specific changes requested. For formatting code, please use [mssql-jdbc-formatter](https://github.com/Microsoft/mssql-jdbc/blob/dev/mssql-jdbc_formatter.xml) from this repository in eclipse.

Also, JMockit should not be part of this PR, another PR would be helpful if we need to add that to tests.

shayaantx:
my fork turned into a mess during the revert, I'm going to just reimplement again in separate pull :|

PR #785, opened 2018-08-23
Changes to read SQL Warnings after ResultSet is read completely

cheenamalhotra:
As of now JDBC driver does not capture any warnings after ResultSet is fully consumed and the TDS_MSG tokens are not consumed but are left unused in Stream.

Use Case:
Capturing warnings at the end of ResultSet is helpful for PolyBase External Data tables where rows could get rejected due to bad column values. This information is sent by SQL Server after ResultSet data.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerStatement.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerStatement.java
@@ -1366,6 +1366,10 @@ public class SQLServerStatement implements ISQLServerStatement {
     /**
      * Returns the next result in the TDS response token stream, which may be a result set, update count or exception.
      *
+     * @param clearFlag
+     *        Boolean Flag if set to true, clears the last stored results in ResultSet. If set to false, does not clear
+     *        ResultSet and continues processing TDS Stream for more results.
+     * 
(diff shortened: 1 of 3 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerStatement.java
```diff
@@ -1591,8 +1591,9 @@ boolean onInfo(TDSReader tdsReader) throws SQLServerException {
             return false;
         }
 
-        // Clear out previous results
-        clearLastResult();
+        // Clear out previous results only when clearFlag = true
+        if (clearFlag)
```
peterbae: add brackets even if there's only one line in the if statement

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerStatement.java
```diff
@@ -1368,7 +1368,7 @@ final void clearLastResult() {
      *
```
peterbae: Modify javadoc now that clearFlag is a parameter

cheenamalhotra: Done

Issue #787, opened 2018-08-27
Unable to insert a BigDecimal with high scale


jan-krakora:
## Driver version or jar name
mssql-jdbc-7.0.0.jre8.jar

## SQL Server version
SQL Server 2017

## Client operating system
Windows 10 x64

## Java/JVM version
Oracle JDK 9.0.4

## Problem description
When trying to insert a BigDecimal value parsed from text like "9.99E+99" I've got an error
com.microsoft.sqlserver.jdbc.SQLServerException: Error converting data type nvarchar to decimal.

jan-krakora:
Hi @cheenamalhotra ,

thank you for your response. My column type is NUMERIC. I would expect an error like "Too high value precision" then. Message "Error converting data type nvarchar to decimal" is IMHO very misleading.

cheenamalhotra:
Hi @jan-krakora ,

Please test PR #796 changes for Out of Range Exception to be thrown instead.

jan-krakora:
How can I test it? Is there a way to get a PR as a branch and build the jdbc driver? Looking at your commit it  seems fine to me.

cheenamalhotra:
Hi @jan-krakora 

Please find the Jars for testing: [mssql-jdbc-PR796.zip](https://github.com/Microsoft/mssql-jdbc/files/2337818/mssql-jdbc-PR796.zip)

jan-krakora:
Hi @cheenamalhotra,

I can confirm the exception with 'out of range' message is thrown. Thank you.

cheenamalhotra:
Hi @jan-krakora 

PR #796 has been merged and changes will be available since v7.1.1 release. Closing issue.

PR #796, opened 2018-08-30
Changes error message to be thrown when data out of range for DECIMAL/NUMERIC types

cheenamalhotra:
Issue #787 

This change now throws SQLServerException if data passed is out of range for DECIMAL/NUMERIC types.


ulvii:
Please, add a test.

cheenamalhotra:
@ulvii As discussed, test will be added later (added in Backlog) with a few changes in existing Junit tests.

src/main/java/com/microsoft/sqlserver/jdbc/dtv.java
```diff
@@ -1083,8 +1083,16 @@ void execute(DTV dtv, Double doubleValue) throws SQLServerException {
 
         void execute(DTV dtv, BigDecimal bigDecimalValue) throws SQLServerException {
             if (DDC.exceedsMaxRPCDecimalPrecisionOrScale(bigDecimalValue)) {
-                String strValue = bigDecimalValue.toString();
-                tdsWriter.writeRPCStringUnicode(name, strValue, isOutParam, collation);
+                if (JDBCType.DECIMAL == dtv.getJdbcType() || JDBCType.NUMERIC == dtv.getJdbcType()) {
+                    // Throw exception for DECIMAL and NUMERIC Datatypes
+                    MessageFormat form = new MessageFormat(SQLServerException.getErrString("R_valueOutOfRangeSQLType"));
+                    Object[] msgArgs = {JDBCType.DECIMAL == dtv.getJdbcType() ? SSType.DECIMAL : SSType.NUMERIC};
```
ulvii: Please also change
`Object[] msgArgs = {JDBCType.DECIMAL == dtv.getJdbcType() ? SSType.DECIMAL : SSType.NUMERIC};`
to 
`Object[] msgArgs = {JDBCType.DECIMAL == dtv.getJdbcType() ? JDBCType.DECIMAL : JDBCType.NUMERIC};`

ulvii: okay, I think this line can just be:
`Object[] msgArgs = {dtv.getJdbcType()};`

ulvii: This line just just be:
`Object[] msgArgs = {dtv.getJdbcType};`

cheenamalhotra: `dtv.getJdbcType()` will return `"java.math.BigDecimal"` for both `JDBCType.DECIMAL` and `JDBCType.NUMERIC` - which is not what we want to throw in error message. We are trying to specify SQLType used for parameter setter.

ulvii: Does it really return `"java.math.BigDecimal"`? Please give it a try.

*We are trying to specify SQLType used for parameter setter.* 
That is my point. The current implementation uses SQL Server types(`SSType.DECIMAL` ,`SSType.NUMERIC`) when throwing `R_valueOutOfRangeSQLType`. Using `dtv.getJdbcType()` will make sure we are using Java SQL types instead. 
Also notice that there is a slight difference too: Java SQL Type is `DECIMAL` whereas SQL Server type is `decimal`.

PR #797, opened 2018-08-30
Fix closeable resource leaks in Junit tests

lilgreenbird:
1st part of Junit test cleanup:
-  use try-with-resources on closeables and fix leaks
- get rid shared variables between tests
- move testframework/util/Util.java to TestUtil (part of work get rid of testframework)


src/test/java/com/microsoft/sqlserver/jdbc/connection/PoolingTest.java
```diff
@@ -81,15 +82,32 @@ public void testPooling() throws SQLException {
...
+                }
+                if (null != con2) {
+                    con2.close();
+                }
+            }
+        } finally {
+            if (null != pc) {
+                pc.close();
+            }
+        }
```
cheenamalhotra: This test can be compressed as well.. with try-with-resources.

src/test/java/com/microsoft/sqlserver/jdbc/connection/TimeoutTest.java
```diff
@@ -193,39 +224,59 @@ public void testCancelQueryTimeoutOnStatement() throws Exception {
...
+                throw new Exception(TestResource.getResource("R_expectedExceptionNotThrown"));
+            } catch (Exception e) {
+                if (!(e instanceof SQLException)) {
+                    throw e;
+                }
+                assertEquals(e.getMessage(), TestResource.getResource("R_readTimedOut"),
+                        TestResource.getResource("R_invalidExceptionMessage"));
+            }
+            try {
+                conn.createStatement().execute("SELECT @@version");
```
cheenamalhotra: Hidden statement here.. use the same 'stmt' instance instead

src/test/java/com/microsoft/sqlserver/jdbc/databasemetadata/DatabaseMetaDataForeignKeyTest.java
```diff
@@ -100,22 +95,40 @@ public static void terminateVariation() throws SQLException {
...
-        } catch (SQLException e) {
-            assertTrue(e.getMessage().startsWith(TestResource.getResource("R_dbNameIsCurrentDB")));
+        SQLServerResultSet rs1 = null;
+        SQLServerResultSet rs2 = null;
+        SQLServerResultSet rs3 = null;
+     
+        try (SQLServerConnection conn = (SQLServerConnection) DriverManager.getConnection(connectionString)) {
+            SQLServerDatabaseMetaData dmd = (SQLServerDatabaseMetaData) conn.getMetaData();
+
+            rs1 = (SQLServerResultSet) dmd.getImportedKeys(null, null, table1);
```
cheenamalhotra: Move all ResultSets to one try w. res. block - validate method calls can happen all at once inside the block..

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/StatementTest.java
```diff
@@ -1866,7 +1866,7 @@ public void testNBCROWWithRandomAccess() throws Exception {
                 con = ds.getConnection();
                 Statement stmt = con.createStatement();
```
cheenamalhotra: Has this class been worked upon too? I see lot of connection/statements not closed properly.
Also `Class.forName("com.microsoft.sqlserver.jdbc.SQLServerDriver");` is unnecessary inside tests.. let's remove them too.

src/test/java/com/microsoft/sqlserver/jdbc/TestUtils.java
```diff
@@ -203,15 +566,15 @@ public static Object roundDatetimeValue(Object value) {
             case 1:
                 ts.setNanos((millis - 1) * 1000000);
                 break;
-
+
```
rene-ye: Some weird spacing issues here.

Issue #559, opened 2017-11-20
Cannot query for time only variables


rPraml:
## Driver version or jar name
latest (6.3.5)

## SQL Server version
SQL Server 14.0.3006

## Client operating system
Windows

## Java/JVM version
Example: java version "1.8.x", Oracle VM

## Table schema
```sql
create table dateandtime (id integer not null, 
 my_date date, 
 my_time time, 
 my_timestamp datetime2 
 constraint pk_esimple primary key (id))
```

## Problem description
A prepared statement with `select * from dateandtime where my_time = ?` and `pstmt.setTime(1,...)` fails with `com.microsoft.sqlserver.jdbc.SQLServerException: The data types time and datetime are incompatible in the equal to operator.`

## Expected behavior and actual behavior
Query with Time should work

## Repro code
See https://github.com/Microsoft/mssql-jdbc/pull/558

v-afrafi:
Hello @rPraml,
Since earlier versions of the SQL Server do not support time datatype, the driver by default sends the time datatype as datetime. In order to send value as time, you need to change the connection property of `sendTimeAsDatetime ` to false. 

You can also read further regarding how time values are sent on [msdn docs](https://docs.microsoft.com/en-us/sql/connect/jdbc/configuring-how-java-sql-time-values-are-sent-to-the-server). Thank you.

rPraml:
Hello @v-afrafi 
applying `sendTimeAsDatetime=false` will solve this issue, but I'm wondering, if there are still more options that are required, so that the (latest) driver works correctly with the latest SQL-Server.
(I understand that there may be options to enable backward compatiblility to servers <= 2005, but 2005 is how old now?)

I also read the msdn docs:
> In SQL Server JDBC Driver 3.0, sendTimeAsDatetime is true by default.
> In a future release, the sendTimeAsDatetime connection property may be set to false by default. 

I would really appreciate if the default value would be changed to false NOW. (or at least soon)
 
My expection is, if I develop a new application with the latest driver and sql-server version, that everything works out of the box and does not require to apply a bunch of compatibility-parameters to the connection strings, that fixes issues that were related to SQL-Server 2005 (which extended support ended at April 12, 2016) 

I also updated the Pull Request with additional tests to show, that the behaviour is a bit inconsistent: If I use TVP & TIME - everything works fine (also with `sendTimeAsDatetime=true`)

cheers
Roland

v-afrafi:
[...]

PR #798, opened 2018-09-01
java.sql.Time Support is still broken

rPraml:
I expect, that this test will fail now.
In #559 it was mentioned, that the `sendTimeAsDateTime` default value should be changed to false. It seems that this did not happen yet.


rPraml:
It seems that the proposed change in #559 was reverted
https://github.com/Microsoft/mssql-jdbc/commit/1142c71056bf36c8d7395dfa4684529d35951bc5
https://github.com/Microsoft/mssql-jdbc/commit/4292be937c4cd85ae17c997108413ee613e947f8

Can anyone tell what is the current state of this option. Has it side effects?

cheenamalhotra:
Hi @rPraml 

Yes this change was reverted as we experienced lots of test failures with this property change, and kept it for later as it will require deeper investigation into the driver as well.

cheenamalhotra:
@rPraml 

I've moved this change in PR #1007, since we're tracking property changes there. Closing this PR.

Issue #788, opened 2018-08-28
NullPointerException after upgrading JDBC-driver from 6.4 to 7.0


christophs78:
## Driver version or jar name
mssql-jdbc-7.0.0.jre8.jar

## SQL Server version
13.0.4001.0

## Client operating system
Windows 10 1709 
(But i think it does not matter.)

## Java/JVM version
Java HotSpot(TM) 64-Bit Server VM, Version 1.8.0_171-b1 (Oracle)
(But i think it does not matter.)

## Table schema
Sorting: Latin1_general_CI_AS
Compatibility: SQL Server 2012 (110)

## Problem description
Caused by: java.lang.NullPointerException
	at com.microsoft.sqlserver.jdbc.SQLServerClobBase.length(SQLServerClob.java:388)
	at com.microsoft.sqlserver.jdbc.SQLServerClob.length(SQLServerClob.java:97)
	at org.hibernate.type.descriptor.java.DataHelper.determineLengthForBufferSizing(DataHelper.java:285)
	at org.hibernate.type.descriptor.java.DataHelper.extractString(DataHelper.java:264)
	...
	at org.hibernate.query.internal.AbstractProducedQuery.doList(AbstractProducedQuery.java:1445)
	at org.hibernate.query.internal.AbstractProducedQuery.list(AbstractProducedQuery.java:1414)
	at org.hibernate.query.Query.getResultList(Query.java:146)
	at org.hibernate.query.criteria.internal.compile.CriteriaQueryTypeQueryAdapter.getResultList(CriteriaQueryTypeQueryAdapter.java:72)
	...
	at org.springframework.aop.framework.JdkDynamicAopProxy.invoke(JdkDynamicAopProxy.java:212)
	at com.sun.proxy.$Proxy146.findAllByOrderByServiceNameAsc(Unknown Source)
	at iete.view.RecordIssueView.init(RecordIssueView.java:45)

There was a change in SQLServerClobBase.length with Version 7.0.
[https://github.com/Microsoft/mssql-jdbc/commit/b60068c1d92e01f7d62658bad36a0ea0acca48d4](https://github.com/Microsoft/mssql-jdbc/commit/b60068c1d92e01f7d62658bad36a0ea0acca48d4)

## Expected behavior and actual behavior
NullPointerException should not be thrown.

## Repro code
Database-View:
![grafik](https://user-images.githubusercontent.com/10461942/44728766-81faed00-aadd-11e8-909c-77309f215513.png)

[...]

PR #799, opened 2018-09-05
Fixed a bug where calling length() after obtaining a stream would close the stream for Clobs/NClobs

rene-ye:
Addresses #788. Also addressed an issue where varchar(max)/Clob objects were always being encoded to UTF-16LE instead of using the Collation specified in the Clob object.

List of Changes (the following applies to NClob as well):
1. calling Clob.length() no longer attempts to load the stream into a string. This fixes a null pointer issue as well as length() closing user streams.
2. added streaming capabilities for Clob.getAsciiStream(). NClobs will always have a non-streaming implementation for getAsciiStream().
3. Clobs no longer default to UTF-16LE. Clobs now respect the collation from the SQL Server. NClobs remain unchanged, will always be UTF-16LE encoding.


src/test/java/com/microsoft/sqlserver/jdbc/unit/lobs/lobsTest.java
```diff
@@ -432,7 +431,6 @@ private void testLobsInsertRetrive(String types[], Class lobClass) throws Except
                 while ((read = stream.read(chunk)) > 0)
                     buffer.write(chunk, 0, read);
                 assertEquals(chunk.length, size);
-
```
peterbae: avoid adding new lines if no changes are added (quoted from our onenote discussion 👌 )

src/test/java/com/microsoft/sqlserver/jdbc/unit/lobs/lobsStreamingTest.java
```diff
@@ -0,0 +1,249 @@
+package com.microsoft.sqlserver.jdbc.unit.lobs;
```
peterbae: File names usually start with uppercase 👌

src/test/java/com/microsoft/sqlserver/jdbc/unit/lobs/lobsStreamingTest.java
```diff
@@ -0,0 +1,249 @@
...
+import org.junit.platform.runner.JUnitPlatform;
+import org.junit.runner.RunWith;
+
+import com.microsoft.sqlserver.testframework.AbstractSQLGenerator;
+import com.microsoft.sqlserver.testframework.AbstractTest;
+import com.microsoft.sqlserver.testframework.util.RandomUtil;
+
+
+@RunWith(JUnitPlatform.class)
+public class lobsStreamingTest extends AbstractTest {
```
peterbae: Can you also add a test case that actually surfaces the original NPE problem? I ran this test file against dev branch, and the test doesn't seem to test the NPE case.

src/test/java/com/microsoft/sqlserver/jdbc/unit/lobs/lobsTest.java
rene-ye: line was removed because I didn't like the spacing. I think this file is going to have merge conflicts with the Junit PR anyways, will address then.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerClob.java
```diff
@@ -301,11 +311,18 @@ public Reader getCharacterStream() throws SQLException {
         Reader getterStream = null;
         if (null == value && !activeStreams.isEmpty()) {
             InputStream inputStream = (InputStream) activeStreams.get(0);
-            getterStream = new BufferedReader(new InputStreamReader(inputStream, UTF_16LE));
+            try {
+                inputStream.reset();
+            } catch (IOException e) {
+                throw new SQLServerException(e.getMessage(), null, 0, e);
```
cheenamalhotra: Please use "SQLServerException.makeFromDriverError" instead.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerClob.java
```diff
@@ -281,9 +280,20 @@ public InputStream getAsciiStream() throws SQLException {
...
-        getStringFromStream();
-        InputStream getterStream = new BufferedInputStream(
-                new ReaderInputStream(new StringReader(value), US_ASCII, value.length()));
+        // Need to use a BufferedInputStream since the stream returned by this method is assumed to support mark/reset
+        InputStream getterStream;
+        if (null == value && !activeStreams.isEmpty()) {
+            InputStream inputStream = (InputStream) activeStreams.get(0);
+            try {
+                inputStream.reset();
+                getterStream = inputStream;
```
cheenamalhotra: Can we close 'inputStream' here?

src/test/java/com/microsoft/sqlserver/jdbc/unit/lobs/LobsStreamingTest.java
```diff
@@ -0,0 +1,282 @@
...
+                while (rs.next()) {
+                    Clob c = rs.getClob(2);
+                    Reader r = c.getCharacterStream();
+                    long clobLength = c.length();
+                    String recieved = getStringFromReader(r,clobLength);// streaming string
+                    assertEquals(lobs.get(rs.getInt(1)), recieved);// compare streamed string to initial string
+                }
+                rs.close();
+            }
+            stmt.execute("DROP TABLE [" + tableName + "]");
```
cheenamalhotra: Drop table in finally block to avoid table leftovers in case exception occurs.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerClob.java
rene-ye: It's the same input stream that the user gets, if we close it, the stream returned to the user gets closed.

src/test/java/com/microsoft/sqlserver/jdbc/unit/lobs/LobsStreamingTest.java
rene-ye: stmt was created in try-catch, cannot access in finally block.

src/test/java/com/microsoft/sqlserver/jdbc/unit/lobs/LobsStreamingTest.java
```diff
@@ -0,0 +1,282 @@
...
+                pstmt.executeBatch();
+
+                ArrayList<NClob> lobsFromServer = new ArrayList<>();
+                ArrayList<String> streamedStrings = new ArrayList<>();
+                ResultSet rs = stmt.executeQuery("SELECT * FROM [" + tableName + "] ORDER BY id ASC");
+                while (rs.next()) {
+                    NClob c = rs.getNClob(2);
+                    assertEquals(c.length(), lobs.get(rs.getInt(1)).length());
+                    lobsFromServer.add(c);
+                    String recieved = getStringFromInputStream(c.getAsciiStream());// streaming string
```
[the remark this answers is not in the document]
rene-ye: Resultsets are automatically closed when statement is closed, and statement is in try-with-resources. Inputstreams are closed when resultset is closed, and I manually call resultset.close() in order to load stream into memory.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerClob.java
```diff
@@ -381,9 +398,8 @@ public String getSubString(long pos, int length) throws SQLException {
      */
     public long length() throws SQLException {
         checkClosed();
-
-        if (value == null && activeStreams.get(0) instanceof PLPInputStream) {
-            return (long) ((PLPInputStream) activeStreams.get(0)).payloadLength / 2;
+        if (value == null && activeStreams.get(0) instanceof BaseInputStream) {
```
lilgreenbird: null == value

src/test/java/com/microsoft/sqlserver/jdbc/unit/lobs/LobsStreamingTest.java
ulvii: 1. There is already `dropTableIfExists()` in Utils.java, consider using it.
2. Move the `drop table` into a `finally` block. You can either  create a new `Statement` for it or just use regular try-catch blocks.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerClob.java
```diff
@@ -381,9 +397,8 @@ public String getSubString(long pos, int length) throws SQLException {
      */
     public long length() throws SQLException {
         checkClosed();
-
-        if (value == null && activeStreams.get(0) instanceof PLPInputStream) {
-            return (long) ((PLPInputStream) activeStreams.get(0)).payloadLength / 2;
+        if (null == value && activeStreams.get(0) instanceof BaseInputStream) {
+            return (long) ((BaseInputStream) activeStreams.get(0)).payloadLength;
         }
         return value.length();
```
lilgreenbird: check for null? value could be null here

rene-ye: If value is null, it means the lob object is being streamed and the if condition above will return the length of the stream. Else, it means the lob object is loaded into memory and value won't be null.

src/test/java/com/microsoft/sqlserver/jdbc/unit/lobs/LobsStreamingTest.java
```diff
@@ -0,0 +1,313 @@
...
+        while (salt.length() < length) {
+            int index = (int) (rnd.nextFloat() * validCharacters.length());
+            salt.append(validCharacters.charAt(index));
+        }
+        String saltStr = salt.toString();
+        return saltStr;
+
+    }
+
+    private String getStringFromInputStream(InputStream is, Charset c) {
```
ulvii: Close the Scanner here.

src/test/java/com/microsoft/sqlserver/jdbc/unit/lobs/LobsStreamingTest.java
```diff
@@ -0,0 +1,313 @@
...
+                        NClob c = rs.getNClob(2);
+                        assertEquals(c.length(), lobs.get(index).length());
+                        lobsFromServer.add(c);
+                        String recieved = getStringFromInputStream(c.getAsciiStream(),
+                                java.nio.charset.StandardCharsets.US_ASCII);// streaming string
+                        streamedStrings.add(recieved);
+                        assertEquals(lobs.get(index), recieved);// compare streamed string to initial string
+                    }
+                    rs.close();
+                    for (int i = 0; i < lobs.size(); i++) {
```
ulvii: This part is unnecessary now.

src/test/java/com/microsoft/sqlserver/jdbc/unit/lobs/LobsStreamingTest.java
rene-ye: Closing scanner closes the Inputstream. The stream is needed at a later part of the test.

src/test/java/com/microsoft/sqlserver/jdbc/unit/lobs/LobsStreamingTest.java
```diff
@@ -0,0 +1,271 @@
...
+import java.sql.NClob;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.util.ArrayList;
+import java.util.Random;
+import java.util.concurrent.ThreadLocalRandom;
+import java.util.stream.IntStream;
+
```
ulvii: Remove unused imports.

src/test/java/com/microsoft/sqlserver/jdbc/unit/lobs/LobsStreamingTest.java
```diff
@@ -0,0 +1,268 @@
...
+    private static final int LOB_ARRAY_SIZE = 500; // number of rows to insert into the table and compare
+    private static final int LOB_LENGTH_MIN = 8000;
+    private static final int LOB_LENGTH_MAX = 32000;
+    private static final String ASCII_CHARACTERS = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890!@#$%^&*()-=_+,./;'[]<>?:{}|`~\"\\";
+    private static final String UNICODE_CHARACTERS = ASCII_CHARACTERS
+            + "Ǥ⚌c♮ƺåYèĢù⚏Ȓ★ǌäõpƸŃōoƝĤßuÙőƆE♹gǇÜŬȺǱ!Û☵ŦãǁĸNQŰǚǻTÖC]ǶýåÉbɉ☩=\\ȍáźŗǃĻýű☓☄¸T☑ö^k☏I:x☑⚀läiȉ☱☚⚅ǸǎãÂ";
+    private static String tableName = null;
+
+    private static enum LoB {
```
lilgreenbird: seem to be odd camel casing B here??

src/test/java/com/microsoft/sqlserver/jdbc/unit/lobs/LobsStreamingTest.java
```diff
@@ -0,0 +1,268 @@
...
+            }
+            pstmt.executeBatch();
+        }
+    }
+
+    @Test
+    @DisplayName("testLengthAfterStream")
+    public void testLengthAfterStream() throws SQLException, IOException {
+        try (Connection conn = DriverManager.getConnection(connectionString);) {
+            try (Statement stmt = conn.createStatement()) {
```
lilgreenbird: again nitpicking...why not put that in the same try block and save an indentation level?

src/test/java/com/microsoft/sqlserver/jdbc/unit/lobs/LobsStreamingTest.java
```diff
@@ -0,0 +1,268 @@
...
+                createLobTable(stmt, tableName, LoB.CLOB);
+                ArrayList<String> lob_data = createRandomStringArray(LoB.CLOB);
+                insertData(conn, tableName, lob_data);
+
+                ResultSet rs = stmt.executeQuery("SELECT * FROM [" + tableName + "] ORDER BY id ASC");
+                while (rs.next()) {
+                    Clob c = rs.getClob(2);
+                    Reader r = c.getCharacterStream();
+                    long clobLength = c.length();
+                    String recieved = getStringFromReader(r, clobLength);// streaming string
```
lilgreenbird: received - "i before e except after c" ;)

src/test/java/com/microsoft/sqlserver/jdbc/unit/lobs/LobsStreamingTest.java
```diff
@@ -0,0 +1,268 @@
...
+                insertData(conn, tableName, lob_data);
+
+                ArrayList<NClob> lobsFromServer = new ArrayList<>();
+                ResultSet rs = stmt.executeQuery("SELECT * FROM [" + tableName + "] ORDER BY id ASC");
+                while (rs.next()) {
+                    int index = rs.getInt(1);
+                    NClob c = rs.getNClob(2);
+                    assertEquals(c.length(), lob_data.get(index).length());
+                    lobsFromServer.add(c);
+                    String recieved = getStringFromInputStream(c.getAsciiStream());// NClob AsciiStream is never
```
lilgreenbird: received

PR #801, opened 2018-09-15
Fixed issues in BulkCopy exception handling

ulvii:
1. Removing the catch block, because the driver should not blindly send attention **packets**. To cancel a query, attention **packets** should only be sent if the client has already sent a request, which has the last packet with EOM bit (0x01) set in status. If a complete request has not been sent to the server, the driver  MUST send the next packet with **both ignore bit (0x02) and EOM bit (0x01)** set in the status to cancel the request (2.2.1.7 of TDS Technical Documentation). This is already properly handled in `IOBuffer.execute()`, which is the only method that calls `doInsertBulk()`. 
2. Adding `this.tdsMessageType == TDS.PKT_BULK`, because prior to `TDS.PKT_BULK`, the driver has already sent `INSERT BULK` SQL statement to the server, which in case of an exception needs to be cancelled too.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
@@ -3132,7 +3132,7 @@ final class TDSWriter {
     // the client MUST send the next packet with both ignore bit (0x02) and EOM bit (0x01)
     // set in the status to cancel the request.
     final boolean ignoreMessage() throws SQLServerException {
-        if (packetNum > 0) {
+        if (packetNum > 0 || this.tdsMessageType == TDS.PKT_BULK) {
             assert !isEOMSent;
 
             if (logger.isLoggable(Level.FINER))
--- a/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
@@ -3132,7 +3132,7 @@ final class TDSWriter {
     // the client MUST send the next packet with both ignore bit (0x02) and EOM bit (0x01)
     // set in the status to cancel the request.
     final boolean ignoreMessage() throws SQLServerException {
-        if (packetNum > 0 || this.tdsMessageType == TDS.PKT_BULK) {
+        if (packetNum > 0 || TDS.PKT_BULK == this.tdsMessageType) {
             assert !isEOMSent;
 
             if (logger.isLoggable(Level.FINER))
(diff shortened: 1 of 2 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -3132,7 +3132,7 @@ final void endMessage() throws SQLServerException {
     // the client MUST send the next packet with both ignore bit (0x02) and EOM bit (0x01)
     // set in the status to cancel the request.
     final boolean ignoreMessage() throws SQLServerException {
-        if (packetNum > 0) {
+        if (packetNum > 0 || this.tdsMessageType == TDS.PKT_BULK) {
```
lilgreenbird: we usually do ( 0 < packetNum || TDS.PKT_BULK == this.tdsMessageType)

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
```diff
@@ -1559,22 +1559,6 @@ private boolean doInsertBulk(TDSCommand command) throws SQLServerException {
...
-
-            // Send Attention packet to interrupt a complete request that was already sent to the server
-            command.startRequest(TDS.PKT_CANCEL_REQ);
-
-            TDSParser.parse(command.startResponse(), command.getLogContext());
-            command.interrupt(ex.getMessage());
-            command.onRequestComplete();
-
-            throw ex;
         } finally {
```
lilgreenbird: why not catch unexpected exception anyway?

ulvii: it is handled in `IOBuffer.execute()`

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
ulvii: This only make sense for `==`, not to confuse it with assignment `=`, so I only updated my change.

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -3132,7 +3132,7 @@ final void endMessage() throws SQLServerException {
     // the client MUST send the next packet with both ignore bit (0x02) and EOM bit (0x01)
     // set in the status to cancel the request.
     final boolean ignoreMessage() throws SQLServerException {
-        if (packetNum > 0) {
+        if (packetNum > 0 || TDS.PKT_BULK == this.tdsMessageType) {
```
peterbae: Do we need this TDS.PKT_BULK == this.tdsMessageType check? In SQLServerBulkCopy::sendBulkCopyCommand, I see that we start the request by sending the TDS.PKT_BULK packet. However, before we get there, we already write PKT_QUERY packet, which increments the packetNum to >0. Doesn't this mean we don't need to change this if statement?

ulvii: `packetNum` is reset every time you start a new request, `TDS.PKT_BULK` in this case, but we still need to cancel `TDS.PKT_QUERY`.

PR #806, opened 2018-09-20
Improvements in Database metadata to prevent Statement leaks and enhance Statement caching

cheenamalhotra:
This PR introduces below changes to SQLServerDatabaseMetadata:

- Improved Statement caching for different catalogs
- Enables Statement caching for 'null' catalog
- Enables closing un-cached Statements on completion of ResultSet _(Prevents Statement Leaks)_


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
```diff
@@ -1768,8 +1766,6 @@ private void getDestinationMetadata() throws SQLServerException {
                 rs.close();
             if (null != stmt)
```
lilgreenbird: why not try-with-resouce for this as well?

cheenamalhotra: This ResultSet 'rs' and Statement 'stmt' objects are created based on IF condition above, where, Statement might not even be created in 1 condition, hence the finally block is kept as it is.

cheenamalhotra:
Jars for testing : [mssql-jdbc-PR806.zip](https://github.com/Microsoft/mssql-jdbc/files/2421400/mssql-jdbc-PR806.zip)

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
```diff
@@ -265,15 +288,24 @@ private CallableStatement getCallableStatementHandle(CallableHandles request,
             String catalog) throws SQLServerException {
         CallableStatement CS = null;
         HandleAssociation hassoc = handleMap.get(request);
-        if (null == hassoc || null == hassoc.databaseName || !hassoc.databaseName.equals(catalog)) {
-            CS = request.prepare(connection);
-            hassoc = new HandleAssociation(catalog, CS);
-            HandleAssociation previous = handleMap.put(request, hassoc);
-            if (null != previous) {
-                previous.close();
+        try {
```
peterbae: I'm having a hard time understanding the logic change behind HandleAssociation. Can you explain to me why we need another map inside HandleAssociation?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
```diff
@@ -265,15 +288,24 @@ private CallableStatement getCallableStatementHandle(CallableHandles request,
...
+            if (null == hassoc) {
+                CS = request.prepare(connection);
```
peterbae: This CallableStatement used to be closed when this method called previous.close() from line 273. However, we removed this part, and this CS will never be closed. The method the calls this method (getResultSetFromStoredProc) also doesn't clean up this returned CS. Is this CS intentionally left unclosed?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
```diff
@@ -1719,29 +1718,28 @@ private void getDestinationMetadata() throws SQLServerException {
...
-            for (int i = 1; i <= destColumnCount; ++i) {
-                if (rsMoreMetaData.next()) {
-                    if (!connection.getServerSupportsColumnEncryption()) {
-                        destColumnMetadata.put(i, new BulkColumnMetaData(rs.getColumn(i),
-                                rsMoreMetaData.getString("collation_name"), null));
+            try (SQLServerStatement statementMoreMetadata = (SQLServerStatement) connection.createStatement();
+                    SQLServerResultSet rsMoreMetaData = (!connection
+                            .getServerSupportsColumnEncryption() ? statementMoreMetadata
+                                    .executeQueryInternal("select collation_name from sys.columns where "
+                                            + "object_id=OBJECT_ID('" + Util.escapeSingleQuotes(destinationTableName) + "') "
```
ulvii: This is less readable to me. Maybe you could only "if else" the query string?

cheenamalhotra: Then we can't place it in try with resources block. Also will need another try block, and finally block to manually close it.
I don't see any issues - its a simple ternary operator.

PR #812, opened 2018-09-28
Added functionality to insert specific column data  in table using writeToServer polymorphic form and Casting fix on Double and Int for nulls 

sumit88:
Added polymorphic forms  ( write to server used one to one mapping of  the columns earlier , with this a user can insert rows of data with a specific set of columns )

  writeToServer(String  columns)
  getDestinationMetadata(String  columns)

Changed casting of Int and Double .
Were getting  type cast errors when data on that column was null .


peterbae:
Hi @sumit88, sorry about this PR getting closed. It looks like this PR was being pointed to an older branch in our repository, and removing the old branch automatically closed this PR. Please feel free to re-open this branch against dev, but first we'll discuss this in the original issue.

PR #817, opened 2018-10-01
Change Bulk Copy API for batch insert to respect the timeout value

peterbae:
Statements  can have timeouts on their query, and currently this isn't being respected by Bulk Copy API for batch insert. This PR fixes that issue.

I've tested this manually and confirmed the queryTimeout value is correctly applied to the Bulk Copy operation.


ulvii:
Please add tests, for default case and when timeout is set.

peterbae:
@ulvii I wanted to write a test as well but that would be difficult. This part of the code is only accessed when the target server is Azure DW.

PR #822, opened 2018-10-04
Refresh Access Token instead of reconnecting on Token Expiry

cheenamalhotra:
Applies to below Authentication modes:

- Active Directory Password
- Active Directory Integrated

Currently driver reconnects when access token is about to expire in 45 mins, leading to a new connection to target Server everytime. With this change, the driver will use Refresh Token received by Azure, to refresh access token in use without the need to reconnect.

>Note: This PR applies only to user applications using ADAL4J library, and *NOT* making use of *sqljdbc_auth.dll* native library to establish Integrated Authentication with Azure databases/warehouses.


cheenamalhotra:
Alternate solution found with no need to refresh/manage access token inside driver.

PR #827, opened 2018-10-09
resolving an intermittent error in junit tests

rene-ye:
removing sql error code check as it's not used consistently in the driver


src/test/java/com/microsoft/sqlserver/jdbc/unit/lobs/lobsTest.java
```diff
@@ -139,7 +139,7 @@ private void testInvalidLobs(Class lobClass, boolean isResultSet) throws SQLExce
                 datasize = packetSize;
                 break;
             case 1:
-                datasize = packetSize + ThreadLocalRandom.current().nextInt(packetSize) + 1;
+                datasize = packetSize + (ThreadLocalRandom.current().nextInt(packetSize/2) * 2);
```
lilgreenbird: can you add a comment here??

PR #830, opened 2018-10-12
Add support for OffsetDateTime to be passed as 'type' in ResultSet.getObject()

harawata:
As `PreparedStatement.setObject(int, Object)` already supports `java.time.OffsetDateTime`, it may make sense.
This was proposed in https://github.com/Microsoft/mssql-jdbc/pull/749#issuecomment-405637052 , but no work has been done, it seems.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResultSet.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResultSet.java
@@ -2398,6 +2398,13 @@ public class SQLServerResultSet implements ISQLServerResultSet, java.io.Serializ
                     returnValue = ldt.toLocalTime();
                 }
             }
+        } else if (type == java.time.OffsetDateTime.class) {
+            microsoft.sql.DateTimeOffset dateTimeOffset = getDateTimeOffset(columnIndex);
+            if (dateTimeOffset == null) {
+                returnValue = null;
+            } else {
+                returnValue = dateTimeOffset.getOffsetDateTime();
+            }
         } else if (type == microsoft.sql.DateTimeOffset.class) {
             returnValue = getDateTimeOffset(columnIndex);
         } else if (type == UUID.class) {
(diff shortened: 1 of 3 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResultSet.java
```diff
@@ -2398,6 +2398,13 @@ public Object getObject(int columnIndex) throws SQLServerException {
...
                 }
             }
+        } else if (type == java.time.OffsetDateTime.class) {
+            microsoft.sql.DateTimeOffset dateTimeOffset = getDateTimeOffset(columnIndex);
+            if (dateTimeOffset == null) {
+                returnValue = null;
+            } else {
+                returnValue = dateTimeOffset.getOffsetDateTime();
+            }
         } else if (type == microsoft.sql.DateTimeOffset.class) {
```
ulvii: Would you add the same support for `java.time.OffsetTime`?

src/main/java/microsoft/sql/DateTimeOffset.java
```diff
@@ -215,6 +215,18 @@ public int hashCode() {
...
     }
 
+    /**
+     * Returns OffsetDateTime equivalent to this DateTimeOffset object.
+     *
+     * @return OffsetDateTime equivalent to this DateTimeOffset object.
+     */
+    public java.time.OffsetDateTime getOffsetDateTime() {
+        java.time.ZoneOffset zoneOffset = java.time.ZoneOffset.ofTotalSeconds(60 * minutesOffset);
+        java.time.LocalDateTime localDaetTime = java.time.LocalDateTime.ofEpochSecond(utcMillis / 1000, nanos,
```
ulvii: `localDaetTime` to `localDateTime`

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResultSet.java
harawata: Thanks for the review, @ulvii !
I'll look into it.

harawata: I have updated the PR.
Let me know if there is anything you want me to change. :)

PR #831, opened 2018-10-14
java.time.OffsetDateTime value sent to the server should not be affected by the default time zone.

harawata:
Although `SQLServerPreparedStatement#setObject(int, Object)` supports `java.time.OffsetDateTime`, the driver sends an incorrect temporal data when the parameter has a different time zone than the default time zone.
Please see the attached test case.

`java.sql.Timestamp#getTime()` used in `DTV#sendTemporal()` subsequently invokes `java.util.Date#normalize()` which takes the default TimeZone into account.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/dtv.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/dtv.java
@@ -678,14 +678,7 @@ final class DTV {
                                                                                                                         * 1000,
                                                                                                                 "");
 
-                        // The behavior is similar to microsoft.sql.DateTimeOffset
-                        // In Timestamp format, only YEAR needs to have 4 digits. The leading zeros for the rest of the
-                        // fields can be omitted.
-                        String offDateTimeStr = String.format("%04d", offsetDateTimeValue.getYear()) + '-'
-                                + offsetDateTimeValue.getMonthValue() + '-' + offsetDateTimeValue.getDayOfMonth() + ' '
-                                + offsetDateTimeValue.getHour() + ':' + offsetDateTimeValue.getMinute() + ':'
-                                + offsetDateTimeValue.getSecond();
-                        utcMillis = Timestamp.valueOf(offDateTimeStr).getTime();
+                        utcMillis = offsetDateTimeValue.toEpochSecond() * 1000;
                         break;
 
                     case DATETIMEOFFSET: {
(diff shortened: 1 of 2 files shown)


petarov:
Any news on this PR? This is really a problem on my side as well.

src/main/java/com/microsoft/sqlserver/jdbc/dtv.java
```diff
@@ -678,14 +678,7 @@ private void sendTemporal(DTV dtv, JavaType javaType, Object value) throws SQLSe
...
 
-                        // The behavior is similar to microsoft.sql.DateTimeOffset
-                        // In Timestamp format, only YEAR needs to have 4 digits. The leading zeros for the rest of the
-                        // fields can be omitted.
-                        String offDateTimeStr = String.format("%04d", offsetDateTimeValue.getYear()) + '-'
-                                + offsetDateTimeValue.getMonthValue() + '-' + offsetDateTimeValue.getDayOfMonth() + ' '
-                                + offsetDateTimeValue.getHour() + ':' + offsetDateTimeValue.getMinute() + ':'
-                                + offsetDateTimeValue.getSecond();
-                        utcMillis = Timestamp.valueOf(offDateTimeStr).getTime();
+                        utcMillis = offsetDateTimeValue.toEpochSecond() * 1000;
```
ulvii: I suppose this change is also required for `OFFSETTIME`.

harawata: Thanks for pointing out! Updated.

src/test/java/com/microsoft/sqlserver/jdbc/preparedStatement/SetObjectTest.java
```diff
@@ -0,0 +1,134 @@
...
+
+    /**
+     * Tests setObject(n, java.time.OffsetTime.class) when 'setSendTimeAsDatetime' connection property is false.
+     * 
+     * @throws SQLException
+     */
+    @Test
+    public void testSetObjectWithOffsetTime_sendTimeAsDatetimeDisabled() throws SQLException {
+        try (Connection con = DriverManager.getConnection(connectionString)) {
+            ((SQLServerConnection) con).setSendTimeAsDatetime(false);
```
cheenamalhotra: Changing 'setSendTimeAsDatetime' connection property has no impact on OffsetDateTime and OffsetTime java types or underlying column DATETIMEOFFSET. Is there a specific reason to test with this property change?

harawata: Thanks for the review, @cheenamalhotra !
Note the difference between `expectedDto`s.
The connection property affects the 'date' part supplemented by the driver when inserting `java.time.OffsetTime` into `DATETIMEOFFSET` column.
I should mention that this behavior is ported from the original code. Please see the following line.
https://github.com/Microsoft/mssql-jdbc/pull/831/files#diff-452e970f57e6bb485626382755aeb835L635

Issue #816, opened 2018-09-29
Driver rejects valid certificate of Azure SQL Database when encrypt=true is specified


jd3nn1s:
## Driver version
6.2.2.jre8, 6.4.0.jre8, master (7.1.2-SNAPSHOT)

## SQL Server version
SQL Azure

## Client Operating System
Windows 10, Windows 2012 R2 and Linux

## JAVA/JVM version
Oracle Java 1.8 and openjdk version "10.0.2" 2018-07-17

## Table schema
N/A

## Problem description
1. Expected behaviour:

Able to connect to SQL Azure using TLS with only the encrypt=true option

2. Actual behaviour:

Connection is not established and an exception is raised

3. Error message/stack trace:

Received an error: "Failed to validate the server name in a certificate during Secure Sockets Layer (SSL) initialization."

4. Any other details that can be helpful:

SQL Azure uses a wildcard subject alternative name of *.database.windows.net in its certificate. While HostNameOverrideX509TrustManager supports subject alternative names (SANs), it does not support checking against wildcard hostnames. Therefore the certificate provided by SQL Azure is rejected by the driver.

I created a PoC where, in the default case, the default Java X509TrustManager is used instead of the driver's own implementation, taking care to set the endpoint identification protocol in the SSLParameters to "HTTPS". This has an implementation of wildcard checking and did not reject the certificate.

## Reproduction code
```
import java.sql.Connection;
import java.sql.DriverManager;

public class Main {

    public static void main(String[] argv) throws Exception {
        String jdbcUrl = "jdbc:sqlserver://xxxxx.database.windows.net:1433;databaseName=jdbctest;username=user1;password=pass1;encrypt=true";
        Connection conn = DriverManager.getConnection(jdbcUrl);
        conn.close();
    }
}

```

jd3nn1s:
Looks like this could be a duplicate of #623.

The issue is not intermittent for me - it happens every time.

peterbae:
Hi @jd3nn1s, thanks for contacting us. Could you try adding "hostNameInCertificate=*.database.windows.net;" to your connection string, and let me know if that solves your problem? The details can be found on this page: https://docs.microsoft.com/en-us/sql/connect/jdbc/setting-the-connection-properties?view=sql-server-2017

jd3nn1s:
Yes this would workaround the issue, however I’m more interested in a fix.

[...]

PR #836, opened 2018-10-18
Make SSL certificate validation respect wildcards

peterbae:
Fixes issue #816.

Previously, we needed to use hostNameInCertificate when attempting to make SSL connections to Azure servers because the * (wildcard) character in the SSL certificate from Azure was being processed literally. This change make it so that the wildcard is actually respected as "any characters".

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
@@ -1412,6 +1412,14 @@ final class TDSChannel {
                 return false;
             }
 
+            // Respect wildcard
+            if (nameInCert.startsWith("*")) {
+                String nameInCertWithoutWildCard = nameInCert.substring(1);
+                if (hostName.endsWith(nameInCertWithoutWildCard)) {
+                    return true;
+                }
+            }
+
             // Verify that the name in certificate matches exactly with the host name
             if (!nameInCert.equals(hostName)) {
                 if (logger.isLoggable(Level.FINER))


src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -1411,6 +1411,35 @@ private boolean validateServerName(String nameInCert) throws CertificateExceptio
...
             }
+            
+            // Respect wildcard
+            if (nameInCert.contains("*")) {
+                // We do not allow wildcards to exist past the first period.
+                if (nameInCert.indexOf("*") > nameInCert.indexOf(".")) {
+                    return false;
+                }
+                String certBeforeWildcard = nameInCert.substring(0, nameInCert.indexOf("*"));
+                int firstPeriodAfterWildcard = nameInCert.indexOf(".", nameInCert.indexOf("*"));
```
rene-ye: Not sure how relevant this is realistically but we should retain the position of "*" and "." since it seems we're using it a fair amount, but traversing the string everytime with indexOf().

Another thing is we check `nameInCert.contains("*")` at the beginning, but then immediately traverse the string to search for "*" again to see if it's before ".". This logic can probably be cleaned up to traverse once as `indexOf()` returns -1 if the character isn't found.

peterbae: Good suggestion. I've reflected this in my newest commit.

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -1412,6 +1412,50 @@ private boolean validateServerName(String nameInCert) throws CertificateExceptio
...
+                if (wildcardIndex > nameInCert.indexOf(".")) {
+                    return false;
+                }
+                
+                // We do not allow wildcards in IDNs.
+                if (nameInCert.startsWith("xn--")) {
+                    return false;
+                }
+                
+                // We do not allow * plus a top-level domain.
```
rene-ye: Please change multi line comments to /* */ format.

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -1412,6 +1412,50 @@ private boolean validateServerName(String nameInCert) throws CertificateExceptio
...
+                if ((nameInCert.length() - nameInCert.replace(".", "").length()) <= 1) {
+                    return false;
+                }
+                
+                String certBeforeWildcard = nameInCert.substring(0, wildcardIndex);
+                int firstPeriodAfterWildcard = nameInCert.indexOf(".", wildcardIndex);
+                String certAfterWildcard;
+
+                if (firstPeriodAfterWildcard < 0) {
+                    // if we get something like peter.database.c*, then make certAfterWildcard empty so that we accept
```
rene-ye: Multi-line comment

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -1412,6 +1412,50 @@ private boolean validateServerName(String nameInCert) throws CertificateExceptio
...
+                
+                // We do not allow wildcards in IDNs.
+                if (nameInCert.startsWith("xn--")) {
+                    return false;
+                }
+                
+                // We do not allow * plus a top-level domain.
+                // This if statement counts the number of .s in the nameInCert. If it's 1 or less, then reject it.
+                // This also catches cases where nameInCert is just *
+                if ((nameInCert.length() - nameInCert.replace(".", "").length()) <= 1) {
```
rene-ye: This probably works fine but is a weird workaround to existing API. The following should work just as well and is more readable/intuitive.
```java
if (nameInCert.chars().filter(ch -> ch == '.').count() <= 1)
```

peterbae: i think this looks more confusing, so i'll leave it as is

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
peterbae: the current comment makes sense grammatically.

src/test/java/com/microsoft/sqlserver/jdbc/SSLCertificateValidation.java
```diff
@@ -0,0 +1,105 @@
+package com.microsoft.sqlserver.jdbc;
```
ulvii: Please add the license header.

src/test/java/com/microsoft/sqlserver/jdbc/SSLCertificateValidation.java
```diff
@@ -0,0 +1,135 @@
...
+        // Server Name = msjdbc.database.windows.net
+        // SAN = *.com
+        // Expected result: false
+        // A cert with * plus a top-level domain is not allowed.
+        assertFalse((boolean) method.invoke(hsoObject, "*.com"));
+
+        // Server Name = msjdbc.database.windows.net
+        // SAN = xn--caf-dma*.com
+        // Expected result: fail
+        assertFalse((boolean) method.invoke(hsoObject, "xn--caf-dma*.com"));
```
rene-ye: Shouldn't this test case be moved below now that we added tests for ""xn--ms.database.windows.net"? The cert and servers don't match here anyways, we're not sure if it's rejecting if it's a wildcard with an IDN or if it's just comparing.

cheenamalhotra:
@peterbae Please pull changes from PR https://github.com/peterbae/mssql-jdbc/pull/7/

src/test/java/com/microsoft/sqlserver/jdbc/SSLCertificateValidation.java
peterbae: I agree that we should cover xn-- case with a wildcard that actually matches the server name (but it should fail regardless since we don't do wildcard matching against xn--). I've made this change.

PR #838, opened 2018-10-19
Feature: Active Directory MSI Authentication support

cheenamalhotra:
Introduces:
- Support for Azure AD MSI Authentication with:
   - System Assigned Managed Identity
   - User Assigned Managed Identity
- New Authentication Mode: "**ActiveDirectoryMSI**"
- New Connection Property: 
_msiClientId_: Accepts Client Id for "User Assigned Managed Identity" to retrieve Access Token

Fixes: 
- Issues with Refresh Token creating new Connection every 15 mins


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResource.java
```diff
@@ -245,6 +245,7 @@ static String getResource(String key) {
             {"R_statementPoolingCacheSizePropertyDescription",
                     "This setting specifies the size of the prepared statement cache for a connection. A value less than 1 means no cache."},
             {"R_gsscredentialPropertyDescription", "Impersonated GSS Credential to access SQL Server."},
+            {"R_msiObjectIdPropertyDescription", "Object Id of User Assigned Managed Identity to be used for generating access token for Azure AD MSI Authentication"},
```
peterbae: Id => ID? in both the entry name + description

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -4003,23 +4027,24 @@ final boolean doExecute() throws SQLServerException {
...
         assert null != fedAuthInfo;
 
         attemptRefreshTokenLocked = true;
-        fedAuthToken = getFedAuthToken(fedAuthInfo);
+        String accessToken = getFedAuthToken(fedAuthInfo);
         attemptRefreshTokenLocked = false;
 
         // fedAuthToken cannot be null.
-        assert null != fedAuthToken;
+        assert null != accessToken;
```
peterbae: change comment to reflect change to accessToken

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -4032,9 +4057,15 @@ private SqlFedAuthToken getFedAuthToken(SqlFedAuthInfo fedAuthInfo) throws SQLSe
 
         while (true) {
             if (authenticationString.trim().equalsIgnoreCase(SqlAuthentication.ActiveDirectoryPassword.toString())) {
-                fedAuthToken = SQLServerADAL4JUtils.getSqlFedAuthToken(fedAuthInfo, user, password,
+                accessToken = SQLServerADAL4JUtils.getSqlFedAuthToken(fedAuthInfo, user, password,
```
peterbae: Can we turn the three else if statements (along with the trim + string comparison) into one switch statement?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerADAL4JUtils.java
```diff
@@ -28,7 +28,7 @@
     static final private java.util.logging.Logger adal4jLogger = java.util.logging.Logger
             .getLogger("com.microsoft.sqlserver.jdbc.internals.SQLServerADAL4JUtils");
 
-    static SqlFedAuthToken getSqlFedAuthToken(SqlFedAuthInfo fedAuthInfo, String user, String password,
+    static String getSqlFedAuthToken(SqlFedAuthInfo fedAuthInfo, String user, String password,
```
peterbae: I see that you've changed all of our current methods that return a SqlFedAuthToken to return string (the string representing the access token). I understand that MSI access tokens last 8 hours instead of the usual 1 hour ones from Azure, but we use these methods for getting the access tokens from Azure as well. If we want to add token refresh logic to our driver in the future, we're going to have to revert these methods so that we retrieve both the refresh token and access token (and store it as a SqlFedAuthToken, but this PR removes that class). Did I miss something regarding how our driver will handle refresh tokens in the future, or have you already considered this case?

peterbae:
Uploading the jre 8 jar created from this PR.

[mssql-jdbc-msi-support-PR838.jre8.jar.zip](https://github.com/Microsoft/mssql-jdbc/files/2549989/mssql-jdbc-msi-support-PR838.jre8.jar.zip)

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
cheenamalhotra: Code changed back to using FedAuthToken

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResource.java
cheenamalhotra: I would keep it camel case (msiClientId) to avoid confusions with other property cases.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -4125,6 +4177,161 @@ private SqlFedAuthToken getFedAuthToken(SqlFedAuthInfo fedAuthInfo) throws SQLSe
...
+                    // Do not retry if maxRetry limit has been reached.
+                    break;
+                } else {
+                    try {
+                        int responseCode = connection.getResponseCode();
+                        // Check Error Response Code from Connection
+                        if (responseCode == 410 || responseCode == 429 || responseCode == 404
+                                || (responseCode >= 500 && responseCode <= 599)) {
+                            try {
+                                int retryTimeoutInMs = retrySlots.get(new Random().nextInt(retry - 1));
```
rene-ye: Use ThreadLocalRandom.current().nextInt(min,max) instead. Also remove the import from `utils`.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -4125,6 +4177,161 @@ private SqlFedAuthToken getFedAuthToken(SqlFedAuthInfo fedAuthInfo) throws SQLSe
...
+                        connectionlogger.finer(toString() + " Using Azure MSI auth: " + urlString);
+                    }
+                }
+
+                connection.connect();
+
+                try (InputStream stream = connection.getInputStream()) {
+
+                    BufferedReader reader = new BufferedReader(new InputStreamReader(stream, UTF_8), 100);
+                    String result = reader.readLine();
```
ulvii: Is it guaranteed that the response is only one line?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -4125,6 +4177,161 @@ private SqlFedAuthToken getFedAuthToken(SqlFedAuthInfo fedAuthInfo) throws SQLSe
...
+            } catch (Exception e) {
+                retry++;
+                if (retry > maxRetry) {
+                    // Do not retry if maxRetry limit has been reached.
+                    break;
+                } else {
+                    try {
+                        int responseCode = connection.getResponseCode();
+                        // Check Error Response Code from Connection
+                        if (responseCode == 410 || responseCode == 429 || responseCode == 404
```
ulvii: Please switch to `410 == responseCode ||  429 == responseCode || 404 == responseCode` for consistency.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
cheenamalhotra: Yes.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResource.java
```diff
@@ -534,5 +538,12 @@ static String getResource(String key) {
             {"R_unknownUTF8SupportValue", "Unknown value for UTF8 support."},
             {"R_illegalWKT", "Illegal Well-Known text. Please make sure Well-Known text is valid."},
             {"R_illegalTypeForGeometry", "{0} is not supported for Geometry."},
-            {"R_illegalWKTposition", "Illegal character in Well-Known text at position {0}."},};
+            {"R_illegalWKTposition", "Illegal character in Well-Known text at position {0}."},
+            {"R_ADALMissing", "Failed to load ADAL4J Java library for performing {0} authentication."},
+            {"R_DLLandADALMissing",
+                    "Failed to load both sqljdbc_auth.dll and ADAL4J Java library for performing {0} authentication. Please install one of them to proceed."},
+            {"R_MSITokenFailureClientId", "Couldn't acquire access token from IMDS, verify your clientId."},
```
ulvii: Please make sure the error messages are consistent and also include MSI. Also, do you think we should be using abbreviation MSI or full term Managed Service Identity in the error message?

PR #839, opened 2018-10-20
Removed illegal reflection access in kerberos authentication

lilgreenbird:
1. Removed illegal reflection access, DNSKerberosLocator will handle validation
2. removed unused test


src/main/java/com/microsoft/sqlserver/jdbc/KerbAuthentication.java
```diff
@@ -350,8 +315,8 @@ private String findRealmFromHostname(RealmValidator realmValidator, String hostn
         int index = 0;
         while (index != -1 && index < hostname.length() - 2) {
             String realm = hostname.substring(index);
-            if (authLogger.isLoggable(Level.FINEST)) {
-                authLogger.finest(toString() + " looking up REALM candidate " + realm);
+            if (authLogger.isLoggable(Level.FINER)) {
```
ulvii: Why was this necessary?

src/main/java/com/microsoft/sqlserver/jdbc/dns/DNSKerberosLocator.java
```diff
@@ -18,7 +18,9 @@
     private DNSKerberosLocator() {}
 
     /**
-     * Returns whether a realm is valid.
+     * Returns whether a realm is valid by retrieving the KDC list in DNS SRV records.
+     * This will only work if DNS lookup is setup properly or the realms are properly defined in krb5 config file.
+     * Otherwise this will fail since the realm can not be found.
```
ulvii: cannot

src/main/java/com/microsoft/sqlserver/jdbc/dns/DNSKerberosLocator.java
```diff
@@ -30,13 +32,15 @@ public static boolean isRealmValid(String realmName) throws NamingException {
...
         }
+        
         if (realmName.startsWith(".")) {
             realmName = realmName.substring(1);
         }
         try {
             Set<DNSRecordSRV> records = DNSUtilities.findSrvRecords("_kerberos._udp." + realmName);
             return !records.isEmpty();
         } catch (NameNotFoundException wrongDomainException) {
+            // config error - domain controller can not be located via DNS
```
ulvii: cannot

src/main/java/com/microsoft/sqlserver/jdbc/dns/DNSKerberosLocator.java
```diff
@@ -30,13 +32,15 @@ public static boolean isRealmValid(String realmName) throws NamingException {
         if (realmName == null || realmName.length() < 2) {
             return false;
         }
+
```
ulvii: Please remove this

src/main/java/com/microsoft/sqlserver/jdbc/KerbAuthentication.java
```diff
@@ -251,7 +251,7 @@ private String enrichSpnWithRealm(String spn, boolean allowHostnameCanonicalizat
         }
         String dnsName = m.group(1);
         String portOrInstance = m.group(2);
-        RealmValidator realmValidator = getRealmValidator(dnsName);
+        RealmValidator realmValidator = getRealmValidator();
```
ulvii: Please remove unused imports.

src/main/java/com/microsoft/sqlserver/jdbc/KerbAuthentication.java
lilgreenbird: to make it consistent with the rest of the logging in this file

pierresouchay:
This code removal will break Kerberos cross DC authentication on many JVMs, most notably on Windows

ulvii:
Hi @pierresouchay,

Starting from JDK 9, users get a massive warning in their applications complaining about illegal reflective access when doing Kerberos authentication. Unfortunately, it seemed like there was no alternative that would let us preserve the same functionality and we removed the piece of code that contained the reflection. It is, however in our backlog to revisit PR #40 in the future and we would appreciate if you could give us suggestions on how to achieve the same functionality without illegal reflective access. 

I should also mention that, the removal of reflection did not break any tests in our test lab and it would be really helpful if you guided us to reproduce the regression with exact test scenario and environment details.

pierresouchay:
@ulvii Yes, I explained a bit the reason, but it is clear it will break the feature. I will blacklist next versions of the driver in our company then. See my comments here: https://github.com/Microsoft/mssql-jdbc/issues/918#issuecomment-458765603

Issue #848, opened 2018-10-24
A processing error "Incorrect syntax near '?'." occurred. in buildExecuteMetadata


and-costea:
## Driver version
7.0.0

## SQL Server version
Microsoft SQL Server 2017 (RTM-CU10) (KB4342123) - 14.0.3037.1 (X64)
        Jul 27 2018 09:40:27
        Copyright (C) 2017 Microsoft Corporation
        Express Edition (64-bit) on Linux (Ubuntu 16.04.5 LTS)

## Client Operating System
Fedora 28

## JAVA/JVM version
openjdk version "1.8.0_181"

## Table schema
No schema is necessary.

## Problem description
Incorrect syntax error is thrown when getting the metadata.

1. Expected behaviour: query should work
2. Actual behaviour: exception is thrown.
3. Error message/stack trace:

```
A processing error "Incorrect syntax near '?'." occurred.
com.microsoft.sqlserver.jdbc.SQLServerException.makeFromDriverError(SQLServerException.java:226) 
com.microsoft.sqlserver.jdbc.SQLServerPreparedStatement.buildExecuteMetaData(SQLServerPreparedStatement.java:1089) 
com.microsoft.sqlserver.jdbc.SQLServerPreparedStatement.getMetaData(SQLServerPreparedStatement.java:1058)
...
```
4. Any other details that can be helpful: see miinimized test case

## JDBC trace logs
don't think they're necessary.

## Reproduction code
Minimized test case of the function `com.microsoft.sqlserver.jdbc.SQLServerPreparedStatement.replaceMarkerWithNull`:

```java
package com.microsoft.sqlserver.jdbc;

import static com.microsoft.sqlserver.jdbc.SQLServerPreparedStatement.replaceMarkerWithNull;

public class Something {
  public static void main(String[] args) {
    String x = "select (case when 1 = ? then 1 else 0 end) as [a '], (case when 1 = ? then 1 else 0 end) as [b ']";
    System.out.println(replaceMarkerWithNull(x));
  }
}
```
output: 

```
select (case when 1 = null then 1 else 0 end) as [a '], (case when 1 = ? then 1 else 0 end) as [b ']
```
expected (I assume):
```
select (case when 1 = null then 1 else 0 end) as [a '], (case when 1 = null then 1 else 0 end) as [b ']
```

When we change the query to:
```
select (case when 1 = ? then 1 else 0 end) as [a ?], (case when 1 = ? then 1 else 0 end) as [b ']
```
[...]

PR #850, opened 2018-10-24
Ignore Escaped Single Quotes in query when replacing markers with Null

cheenamalhotra:
Fixes Issue #848

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerStatement.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerStatement.java
@@ -1011,7 +1011,7 @@ public class SQLServerStatement implements ISQLServerStatement {
      * @return the result
      */
     static String replaceMarkerWithNull(String sql) {
-        if (!sql.contains("'")) {
+        if (!containsUnEscapedQuotes(sql)) {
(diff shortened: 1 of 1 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerStatement.java
```diff
@@ -1038,6 +1038,16 @@ static String replaceMarkerWithNull(String sql) {
         }
     }
 
+    static boolean containsUnEscapedQuotes(String sql) {
+        // Look for bracketed strings to be ignored
+        Pattern p = Pattern.compile("\\[(.*?)\\]");
```
and-costea: I don't think this covers the other quoting character allowed for identifiers, i.e. the double quote (") character.

Also, don't think it covers identifiers like [a]]'].

cheenamalhotra: Hi @and-costea 

I will close the PR for now as it would require a strong parser implementation to validate all cases. We have similar parsing logic in the driver but it isn't generalized to be re-used in this case. I will keep the issue open to bring up that parser in effect in this case.

Issue #851, opened 2018-10-25
Geography object latitude/longitude swapped compared to using column.Lat and column.Long in SQL query


prunge-helix:
## Driver version
7.0.0 and 7.1.1

## SQL Server version
Microsoft SQL Server 2017 (RTM) - 14.0.1000.169 (X64) 
	Aug 22 2017 17:04:49 
	Copyright (C) 2017 Microsoft Corporation
	Developer Edition (64-bit) on Linux (Ubuntu 16.04.3 LTS)

## Client Operating System
SQL Server for Linux docker image (2017-latest): https://hub.docker.com/r/microsoft/mssql-server-linux/
Docker version 18.06.1-ce, build e68fc7a
Host OS: Ubunutu 17.10

## JAVA/JVM version
Java(TM) SE Runtime Environment (build 1.8.0_181-b13)
and
java version "10.0.2" 2018-07-17

## Table schema
create table geotest (id integer not null primary key, location geography not null)

## Problem description
This might be just my misunderstanding of how to use geography types, but there seems to be a discrepancy between querying the database using `Geography` objects versus using `Lat` and `Long` attributes on columns.

Suppose the following code is used to insert data:

```
try (ISQLServerPreparedStatement stat = con.prepareStatement("insert into geotest (id, location) values (?, ?)").unwrap(ISQLServerPreparedStatement.class))
{
    stat.setInt(1, 1);
    stat.setGeography(2, Geography.point(-30, 50, 4326));
    stat.executeUpdate();
}
```

Reading using `Geography` objects is fine:

```
try (ISQLServerPreparedStatement stat = con.prepareStatement("select location from geotest").unwrap(ISQLServerPreparedStatement.class);
             ISQLServerResultSet rs = stat.executeQuery().unwrap(ISQLServerResultSet.class))
{
    rs.next();
    Geography g = rs.getGeography(1);
    System.out.println("Select with geography object: latitude=" + g.getLatitude() + ", longitude=" + g.getLongitude());
}
```

where the result lat/long is expected.  But this code that uses `Lat` and `Long` in the query itself:

```
try (ISQLServerPreparedStatement stat = con.prepareStatement("select location.Lat, location.Long from geotest").unwrap(ISQLServerPreparedStatement.class);
             ISQLServerResultSet rs = stat.executeQuery().unwrap(ISQLServerResultSet.class))
{
    rs.next();
    double latitude = rs.getDouble(1);
    double longitude = rs.getDouble(2);
[...]

PR #853, opened 2018-10-25
Fix Geometry.point coordinates

peterbae:
Fixes issue #851.

According to [this](https://docs.microsoft.com/en-us/sql/t-sql/spatial-geography/point-geography-data-type?view=sql-server-2017) page, Geography::point method has the coordinates reversed (lat, long instead of the usual long, lat) compared to the other methods of inserting a Geography point. This PR aligns the driver behavior with the expected behavior.

### DIFF
--- a/src/test/java/com/microsoft/sqlserver/jdbc/datatypes/SQLServerSpatialDatatypeTest.java
+++ b/src/test/java/com/microsoft/sqlserver/jdbc/datatypes/SQLServerSpatialDatatypeTest.java
@@ -812,7 +812,7 @@ public class SQLServerSpatialDatatypeTest extends AbstractTest {
         String geoWKT = "POINT(1 2)";
 
         Geometry geomWKT = Geometry.point(1, 2, 0);
-        Geography geogWKT = Geography.point(1, 2, 4326);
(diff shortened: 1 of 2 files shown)


src/test/java/com/microsoft/sqlserver/jdbc/datatypes/SQLServerSpatialDatatypeTest.java
```diff
@@ -1041,6 +1041,48 @@ public void testNull() throws SQLException {
...
+            try (SQLServerPreparedStatement pstmt = (SQLServerPreparedStatement) con.prepareStatement(
+                    "insert into " + AbstractSQLGenerator.escapeIdentifier(geomTableName) + " values (?)");) {
+                pstmt.setGeometry(1, geomWKT);
+                pstmt.execute();
+
+                try {
+                    SQLServerResultSet rs = (SQLServerResultSet) stmt.executeQuery("select * from " + AbstractSQLGenerator.escapeIdentifier(geomTableName));
+                    rs.next();
+                    rs.getGeography(1); // should fail
+                    throw new SQLException ("getGeography against Geometry column should fail");
```
ulvii: Why SQLException? `fail()` would do.

PR #854, opened 2018-10-26
Fix | Intermittent BulkCopyResultSetCursorTest error

rene-ye:
Resolves an intermittent error where rows retrieved were out of order compared to the expected.

E.G:  org.opentest4j.AssertionFailedError: Expected Value:12345.12345, Actual Value: 45.12345 ==> expected: <true> but was: <false>


src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyResultSetCursorTest.java
```diff
@@ -143,7 +143,7 @@ public void testMultiplePreparedStatementAndResultSet() throws SQLException {
                     verifyDestinationTableData(expectedBigDecimals.length * 3);
                 }
 
-                String sql = "insert into " + AbstractSQLGenerator.escapeIdentifier(desTable) + " values (?,?,?,?)";
+                String sql = "insert into " + AbstractSQLGenerator.escapeIdentifier(desTable) + "(c1,c2,c3,c4) values (?,?,?,?)";
```
cheenamalhotra: Adding columns here is not required if the first column is Identity. Identity column cannot be inserted, hence always excluded.

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyResultSetCursorTest.java
```diff
@@ -192,12 +192,11 @@ private static void verifyDestinationTableData(int expectedNumberOfRows) throws
     }
 
     private static void populateSourceTable() throws SQLException {
-        String sql = "insert into " + AbstractSQLGenerator.escapeIdentifier(srcTable) + " values (?,?,?,?)";
+        String sql = "insert into " + AbstractSQLGenerator.escapeIdentifier(srcTable) + "(c1,c2,c3,c4) values (?,?,?,?)";
```
cheenamalhotra: Same applies here.

PR #859, opened 2018-10-26
Delete unused import

simon04:


rene-ye:
Placeholder API which are mandatory according to JDBC 4.3 specs. Even though the feature isn't implemented in the driver, we still need to retain it. Closing PR.

PR #861, opened 2018-10-26
Fix/improve StringBuilder usages

simon04:
- Replace string concatenation w/ StringBuilder in loops
- Avoid string concatenation in `StringBuilder.append()`


simon04:
Thanks for the pointer. I amended my commits. Hopefully, I didn't forget something.

PR #873, opened 2018-11-01
Address TODOs in tests

lilgreenbird:
removed/addressed/created backlog on TODOs

### DIFF
--- a/src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyColumnMappingTest.java
+++ b/src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyColumnMappingTest.java
@@ -383,13 +383,13 @@ public class BulkCopyColumnMappingTest extends BulkCopyTestSetUp {
                 DBResultSet dstResultSet = dstStmt
                         .executeQuery("SELECT * FROM " + destinationTable.getEscapedTableName() + ";")) {
             ResultSetMetaData sourceMeta = ((ResultSet) srcResultSet.product()).getMetaData();
-            int srcTotalColumns = sourceMeta.getColumnCount();
+            int totalColumns = sourceMeta.getColumnCount();
 
             // verify data from sourceType and resultSet
             int numRows = 0;
             while (srcResultSet.next() && dstResultSet.next()) {
                 numRows++;
-                for (int i = 1; i <= srcTotalColumns; i++) {
+                for (int i = 1; i <= totalColumns; i++) {
                     Object srcValue, dstValue;
                     srcValue = srcResultSet.getObject(i);
                     dstValue = dstResultSet.getObject(i);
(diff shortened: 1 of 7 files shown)


src/test/java/com/microsoft/sqlserver/jdbc/datatypes/TVPWithSqlVariantTest.java
```diff
@@ -353,8 +353,10 @@ public void testDateTime() throws SQLException {
      * 
      * @throws SQLException
      * @throws SQLTimeoutException
+     * https://msdn.microsoft.com/en-ca/library/dd303302.aspx?f=255&MSPPError=-2147217396
+     * Data types cannot be NULL when inside a sql_variant
      */
-    @Test // TODO We need to check this later. Right now sending null with TVP is not supported
```
cheenamalhotra: This TODO must stay, as there's a driver bug that throws exception for NULL value. Instead of validating Exception, we should be ideally validating stored NULL value.

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyColumnMappingTest.java
```diff
@@ -403,6 +404,13 @@ private void validateValuesRepetitiveCM(DBConnection con, DBTable sourceTable,
                     }
                 }
             }
+
+            // verify number of rows and columns
+            assertTrue(((ResultSet) dstResultSet.product()).getMetaData().getColumnCount() == totalColumns + 1);
+            if (numRows > 0) {
```
ulvii: Shouldn't the test fail if `numRows` is 0?

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyTestUtil.java
```diff
@@ -280,17 +281,26 @@ static void validateValues(DBConnection con, DBTable sourceTable, DBTable destin
...
 
                     ComparisonUtil.compareExpectedAndActual(destMeta.getColumnType(i), srcValue, dstValue);
                 }
             }
+
+            // verify number of rows and columns
+            assertTrue(((ResultSet) srcResultSet.product()).getMetaData().getColumnCount() == totalColumns);
+            assertTrue(((ResultSet) dstResultSet.product()).getMetaData().getColumnCount() == totalColumns);
+
+            if (numRows > 0) {
```
ulvii: Shouldn't the test fail if `numRows` is 0?

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyTestUtil.java
```diff
@@ -280,17 +281,26 @@ static void validateValues(DBConnection con, DBTable sourceTable, DBTable destin
...
                     srcValue = srcResultSet.getObject(i);
                     dstValue = dstResultSet.getObject(i);
 
                     ComparisonUtil.compareExpectedAndActual(destMeta.getColumnType(i), srcValue, dstValue);
                 }
             }
+
+            // verify number of rows and columns
+            assertTrue(((ResultSet) srcResultSet.product()).getMetaData().getColumnCount() == totalColumns);
+            assertTrue(((ResultSet) dstResultSet.product()).getMetaData().getColumnCount() == totalColumns);
```
ulvii: We get `totalColumns` from `dstResultSet`, so this line is redundant.

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyTestUtil.java
lilgreenbird: no, because this method is used sometimes to just copy a table with no data

Issue #846, opened 2018-10-23
TVP varchar(max) fails with ISQLServerDataRecord, works with SQLServerDataTable


dangets:
## Driver version
7.1.1.jre10-preview AND 6.2.2.jre8 (assuming all other versions too)

## SQL Server version
Microsoft SQL Server 2014 (SP2-CU11) (KB4077063) - 12.0.5579.0 (X64) 
	Feb 21 2018 12:19:47 
	Copyright (c) Microsoft Corporation
	Developer Edition (64-bit) on Windows NT 6.3 <X64> (Build 9600: )

## Client Operating System
Ubuntu 18.04

## JAVA/JVM version
10.0.2+13-1ubuntu0.18.04.2

## Table schema
```sql
create type data.DgTestType AS table (
  fieldA VARCHAR(MAX) NULL
)
grant exec on type :: data.DgTestType to public

create procedure data.SelFromType @tbl data.DgTestType READONLY
AS
BEGIN
  SELECT * FROM @tbl
end
```

## Problem description
Error thrown when trying to use TVP, with string column > 4000 characters.  Using `SQLServerDataTable` works, but `ISQLServerDataRecord` fails.  The code snippet below shows the usages.  This relates to #250 

Error message is `The incoming tabular data stream (TDS) remote procedure call (RPC) protocol stream is incorrect. Table-valued parameter 0 (""), row 1, column 1: Data type 0xE7 has an invalid data length or metadata length.`

## Reproduction code
Code is in Kotlin, but is nearly identical to java version.

```kotlin
fun main(args: Array<String>) {
    val ds = devCustCustodyData()

    val paramString = "abcd".repeat(1001)

    val table = SQLServerDataTable()
    table.addColumnMetadata("fieldA", Types.LONGVARCHAR)
    table.addRow(paramString)
    table.tvpName = "data.DgTestType"

    ds.connection.use { conn ->
        conn.prepareStatement("exec data.SelFromType ?").use { pstmt ->
            pstmt.unwrap(SQLServerPreparedStatement::class.java).use { msStmt ->
                msStmt.setStructured(1, "data.DgTestType", MyRow(listOf(paramString)))  // <---- ISQLServerDataRecord fails
                //msStmt.setStructured(1, "data.DgTestType", table)  // <---- SQLServerDataTable works
                msStmt.executeQuery().use { rs ->
                    while (rs.next())
                        println(rs.getString(1).length)
                }
            }
        }
    }

}

[...]

PR #876, opened 2018-11-08
Add SQLServerMetadata constructor to set precision for string types

lilgreenbird:
fixes issue #846

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerMetaData.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerMetaData.java
@@ -61,6 +61,22 @@ public class SQLServerMetaData {
     }
 
     /**
+     * Constructs a SQLServerMetaData with the column name, SQL type, length.
+     * 
+     * @param columnName
+     *        the name of the column
+     * @param sqlType
+     *        the SQL type of the column
+     * @param length
+     *        the length of the string type
+     */
+    public SQLServerMetaData(String columnName, int sqlType, int length) {
+        this.columnName = columnName;
+        this.javaSqlType = sqlType;
+        this.precision = length;
+    }
+
+    /**
      * Constructs a SQLServerMetaData.
      * 
      * @param columnName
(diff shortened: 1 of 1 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerMetaData.java
```diff
@@ -60,6 +60,25 @@ public SQLServerMetaData(String columnName, int sqlType, int precision, int scal
         this.scale = scale;
     }
 
+    /**
+     * Constructs a SQLServerMetaData with the column name, SQL type, length.
```
ulvii: I think JavaDoc for this constructor should be more descriptive if we decide to add it to the driver. The description should explain in what cases and why it is necessary to use the new constructor when there is already one for string types. We might also want to update JavaDoc for `public SQLServerMetaData(String columnName, int sqlType)` constructor too, to avoid confusions.

lilgreenbird: there isn't already one for string types, that's why this is added.
I could add the comments into the javadoc. Or do you suggest something else?

rene-ye: ... ,SQL type, and length.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerMetaData.java
```diff
@@ -60,6 +60,23 @@ public SQLServerMetaData(String columnName, int sqlType, int precision, int scal
         this.scale = scale;
     }
 
+    /**
+     * Constructs a SQLServerMetaData with the column name, SQL type, and length.
+     * The precision for string types is length which is used to determine whether to send vmax header to the server.
```
peterbae: why did we mention precision to this constructor? shouldn't it be on the one above?
we should also add javadoc to this method to explain why this constructor was necessary.

Issue #879, opened 2018-11-13
The driver could not establish a secure connection to SQL Server by using Secure Sockets Layer (SSL) encryption. Error: "Unexpected rethrowing".


martinm1000:
## Driver version
mssql-jdbc-7.1.2.jre11
mssql-jdbc-7.0.0.jre10 ( i think I got the problem with this one too)

## Problem / Question

I've been refactoring my database connection code, and I probably have a new non-oblivious problem since but I need some pointers on what is going on in the mssql driver.

I seem to have **random** problems creating SQL connections that otherwise would work ok. I get the following stacktraces while connecting to a **NON-ENCRYPTED** sql server 2017 database.

Can someone explain the TDS Prelogin negociations vs SSL vs any other encryption scheme that might explain why I get this exception sometimes ? Or is this wrong exception message ? Is there some SSL going on to login even if we don't have encrption setup on the sql server side ?

```java
com.microsoft.sqlserver.jdbc.SQLServerException: The driver could not establish a secure connection to SQL Server by using Secure Sockets Layer (SSL) encryption. Error: "Unexpected rethrowing".
        at com.microsoft.sqlserver.jdbc@7.1.2.jre11-preview/com.microsoft.sqlserver.jdbc.SQLServerConnection.terminate(SQLServerConnection.java:2826)
        at com.microsoft.sqlserver.jdbc@7.1.2.jre11-preview/com.microsoft.sqlserver.jdbc.TDSChannel.enableSSL(IOBuffer.java:1829)
        at com.microsoft.sqlserver.jdbc@7.1.2.jre11-preview/com.microsoft.sqlserver.jdbc.SQLServerConnection.connectHelper(SQLServerConnection.java:2391)
        at com.microsoft.sqlserver.jdbc@7.1.2.jre11-preview/com.microsoft.sqlserver.jdbc.SQLServerConnection.login(SQLServerConnection.java:2042)
        at com.microsoft.sqlserver.jdbc@7.1.2.jre11-preview/com.microsoft.sqlserver.jdbc.SQLServerConnection.connectInternal(SQLServerConnection.java:1889)
        at com.microsoft.sqlserver.jdbc@7.1.2.jre11-preview/com.microsoft.sqlserver.jdbc.SQLServerConnection.connect(SQLServerConnection.java:1120)
[...]

PR #882, opened 2018-11-16
Improved retry logic for intermittent TLS1.2 issue

ulvii:
Intermittent TLS1.2 issue details can be found [here](https://blogs.msdn.microsoft.com/dataaccesstechnologies/2016/11/30/intermittent-jdbc-connectivity-issue-the-driver-could-not-establish-a-secure-connection-to-sql-server-by-using-secure-sockets-layer-ssl-encryption-error-sql-server-returned-an-incomplete-respons/). 

- Based on the discussion on #879 , changed the retry logic to also check for Exception cause because actual error message can be wrapped into a different message when re-thrown from underlying `InputStream`.
- Some minor cleanup.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
@@ -1801,9 +1801,9 @@ final class TDSChannel {
              * message when re-thrown from underlying InputStream.
              */
             String causeErrMsg = null;
-            if (e.getCause() != null) {
-                causeErrMsg = (e.getCause().getLocalizedMessage() != null) ? e.getCause().getLocalizedMessage()
-                                                                           : e.getCause().getMessage();
+            Throwable cause = e.getCause();
+            if (cause != null) {
+                causeErrMsg = (cause.getLocalizedMessage() != null) ? cause.getLocalizedMessage() : cause.getMessage();
             }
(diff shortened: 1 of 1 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -1799,31 +1799,38 @@ else if (con.getTrustManagerClass() != null) {
...
+            String errMsg = (e.getLocalizedMessage() != null) ? e.getLocalizedMessage() : e.getMessage();
+            /*
+             * Retrieve the error message of the cause too because actual error message can be wrapped into a different
+             * message when re-thrown from underlying InputStream.
+             */
+            String causeErrMsg = null;
+            if (e.getCause() != null) {
+                causeErrMsg = (e.getCause().getLocalizedMessage() != null) ? e.getCause().getLocalizedMessage()
+                                                                           : e.getCause().getMessage();
+            }
```
lilgreenbird: I like conditionals it looks cleaner but this means getLocalizedMessage() and getCause() are called multiple times

ulvii: These are just getters, it wouldn't really make much difference, but I still updated the code in case if there are changes to these methods in the future.

PR #903, opened 2018-12-05
Add ability to perform JUnit testing against Azure DW

peterbae:
Command to perform only Azure DW Test:

mvn clean install -DtestGroup=AzureDWTest
(and you also need to either provide the mssql_jdbc_test_connection_properties environment variable yourself, or modify that variable in the AbstractTest.java file)

This test can be performed along with the build42/build43 profiles.

The regular JUnit test has about ~450 test cases, so the Azure DW test suite currently tests about 25% of the regular JUnit test plan, and this number will be brought up in the near future.

Added test results screenshots to the PR.

test1
test2


src/test/java/com/microsoft/sqlserver/jdbc/TestUtils.java
```diff
@@ -666,4 +676,85 @@ public static boolean supportJDBC43(Connection con) throws SQLException {
...
+        try {
+            Class.forName("com.microsoft.sqlserver.jdbc.SQLServerDriver");
+        } catch (ClassNotFoundException e) {
+            throw new SQLException(e.toString());
+        }
+
+        boolean ownsCon = false;
+
+        if (null == con) {
+            con = DriverManager.getConnection(AbstractTest.getConnectionString());
```
cheenamalhotra: Why can we not own the connection in the method itself and close it in the end? I don't understand the point of `ownCon` property in this class.. even it is implemented for other reasons, doesn't have to be necessarily used in new methods that we add here.

src/test/java/com/microsoft/sqlserver/jdbc/TestUtils.java
```diff
@@ -666,4 +676,85 @@ public static boolean supportJDBC43(Connection con) throws SQLException {
...
+            _isSqlAzure = (engineEdition == ENGINE_EDITION_FOR_SQL_AZURE
+                    || engineEdition == ENGINE_EDITION_FOR_SQL_AZURE_DW);
+            _isSqlAzureDW = (engineEdition == ENGINE_EDITION_FOR_SQL_AZURE_DW);
+            if (ownsCon) {
+                con.close();
+            }
+            return _isSqlAzureDW;
+        }
+    }
+}
```
cheenamalhotra: Need one empty line in here..

src/test/java/com/microsoft/sqlserver/jdbc/TestUtils.java
```diff
@@ -666,4 +676,85 @@ public static boolean supportJDBC43(Connection con) throws SQLException {
...
+     * @return boolean
+     * @throws SQLException
+     */
+    public static boolean isSqlAzureDW(Connection con) throws SQLException {
+        if (_determinedSqlAzureOrSqlServer) {
+            return _isSqlAzureDW;
+        }
+
+        try {
+            Class.forName("com.microsoft.sqlserver.jdbc.SQLServerDriver");
```
cheenamalhotra: This may be unnecessary block

src/test/java/com/microsoft/sqlserver/jdbc/TestUtils.java
peterbae: This is faster as we don't always have to create a new connection if the method that's calling isSqlAzure already has access to a connection. However, since this is a static method that can be called anywhere, there's places where the caller method doesn't have access to their own connection, in which case this method creates/closes a new connection. It's mainly here for better performance.

src/test/java/com/microsoft/sqlserver/jdbc/TestUtils.java
peterbae: This is necessary because this static block can be called in the JUnit test before the classloader has had a chance to load our driver. The JUnit test fails without this line here (because sometimes we want to identify the server before we even use the driver).

src/test/java/com/microsoft/sqlserver/jdbc/TestUtils.java
peterbae: sure, done.

src/test/java/com/microsoft/sqlserver/jdbc/TestUtils.java
```diff
@@ -666,4 +676,85 @@ public static boolean supportJDBC43(Connection con) throws SQLException {
...
+        }
+
+        try (Statement stmt = con.createStatement();
+                ResultSet rs = stmt.executeQuery("SELECT CAST(SERVERPROPERTY('EngineEdition') as INT)")) {
+            rs.next();
+            int engineEdition = rs.getInt(1);
+            _determinedSqlAzureOrSqlServer = true;
+            _isSqlAzure = (engineEdition == ENGINE_EDITION_FOR_SQL_AZURE
+                    || engineEdition == ENGINE_EDITION_FOR_SQL_AZURE_DW);
+            if (ownsCon) {
```
lilgreenbird: in finally? ownsCon not closed if failed above

src/test/java/com/microsoft/sqlserver/testframework/DBTable.java
```diff
@@ -361,32 +361,66 @@ public DBTable cloneSchema() {
...
-        for (int i = 0; i < totalRows; i++) {
-            if (i != 0) {
-                sb.add(COMMA);
-            }
-            sb.add(OPEN_BRACKET);
-            for (int colNum = 0; colNum < totalColumns; colNum++) {
+                sb.add(OPEN_BRACKET);
+                for (int colNum = 0; colNum < totalColumns; colNum++) {
+
+                    // TODO: consider how to enclose data in case of preparedStatemets
```
lilgreenbird: is there a backlog for this?

src/test/java/com/microsoft/sqlserver/jdbc/preparedStatement/BatchExecutionWithBulkCopyTest.java
```diff
@@ -544,9 +555,18 @@ public void testIllegalNumberOfArgNoColumnList() throws Exception {
...
         try (Connection connection = DriverManager.getConnection(connectionString + ";useBulkCopyForBatchInsert=true;");
                 SQLServerPreparedStatement pstmt = (SQLServerPreparedStatement) connection.prepareStatement(invalid);
                 Statement stmt = (SQLServerStatement) connection.createStatement();) {
+            
+            if (TestUtils.isSqlAzureDW(connection)) {
+                expected = "Column name or number of supplied values does not match table definition.";
+            } else {
+                expected = "There are fewer columns in the INSERT statement than values specified in the VALUES clause. The number of values in the VALUES clause must match the number of columns specified in the INSERT statement.";
+            }
+
```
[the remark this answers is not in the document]
cheenamalhotra: If we do, this String should go to TestResource

src/test/java/com/microsoft/sqlserver/testframework/DBTable.java
ulvii: Please do not add TODOs, create backlog items instead.

peterbae: i didn't add this..look at the original text to the left pls (i wouldn't do this, either)

src/test/java/com/microsoft/sqlserver/jdbc/preparedStatement/BatchExecutionWithBulkCopyTest.java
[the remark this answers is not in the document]
peterbae: We were checking the string before. Moving the String to TestResource can be done in another PR (there's a lot more places where this is applicable)

cheenamalhotra:
Please pull changes from https://github.com/peterbae/mssql-jdbc/pull/6

src/test/java/com/microsoft/sqlserver/jdbc/fips/FipsTest.java
```diff
@@ -27,6 +28,7 @@
  * Test class for testing FIPS property settings.
  */
 @RunWith(JUnitPlatform.class)
+@Tag("AzureDWTest")
```
ulvii: Why do we have to skip this test?

ulvii:
Remove unused imports.

src/test/java/com/microsoft/sqlserver/testframework/DBTable.java
```diff
@@ -361,32 +361,61 @@ public DBTable cloneSchema() {
...
+                    }
+
+                    if (colNum < totalColumns - 1) {
+                        sb.add(COMMA);
+                    }
                 }
+                sb.add(CLOSE_BRACKET);
+                sb.add(SEMI_COLON);
+            }
+        } else {
```
ulvii: No need to duplicate same lines in the else block. The insertion that works for DW should also work for the rest of the servers.

src/test/java/com/microsoft/sqlserver/jdbc/unit/UTF8SupportTest.java
```diff
@@ -107,7 +107,7 @@ public void testVarchar() throws SQLException {
 
     @BeforeAll
     public static void setUp() throws ClassNotFoundException, SQLException {
-        connection = PrepUtil.getConnection(getConfiguredProperty("mssql_jdbc_test_connection_properties"));
+        connection = PrepUtil.getConnection(AbstractTest.getConnectionString());
```
ulvii: The test already extends `AbstractTest`, `connection = PrepUtil.getConnection(connectionString);` is enough.

src/test/java/com/microsoft/sqlserver/jdbc/connection/NativeMSSQLDataSourceTest.java
```diff
@@ -29,17 +30,18 @@
 import com.microsoft.sqlserver.testframework.AbstractTest;
 
 
-@RunWith(JUnitPlatform.class)
```
ulvii: Why is this removed?

PR #908, opened 2018-12-12
Workaround Bulk Copy for batch insert operation against specific datatypes.

peterbae:
For now, these datatypes don't seem to work against Azure DW when using bulk copy:
Money/Smallmoney
Temporal data types (Date, Datetime, Datetime2, SmallDateTime, DateTimeOffset, Time).

This PR puts a workaround that makes tables with those columns present fallback to the original batch insert operation.

Also, Binary/Varbinary data types also used to fail, but I fixed that part in this PR as well.


peterbae:
PR moved to #912.

Issue #909, opened 2018-12-13
Single timeout thread never stops and could block


sehrope:
## Driver version
7.1.3 (preview release)

## Problem description
The shared single threaded timeout handler introduced in #842, once started via running a query on a connection with a non-zero `queryTimeout`, never stops. It keeps running even after all connections are closed. This will lead to a classloader issue in any environment with multiple classloaders (ex: Tomcat) that dynamically add and remove the driver class as the associated classes can never be unloaded.

I haven't recreated that situation myself (I don't use Tomcat) but recognized it as an issue as the PostgreSQL JDBC driver had a similar issue a couple years back. For background, the original pgjdbc issue reported by an end user is is here: https://github.com/pgjdbc/pgjdbc/pull/188, PR that fixes it is here: https://github.com/pgjdbc/pgjdbc/pull/197.

The `TimeoutPoller` also executes the timeout specific `interrupt()` action in the single timer thread. From skimming through the code I don't think that's an issue right now as the interrupts being executed are non-blocking. If that changes down the road, then it'd be possible for a single hanging `interrupt()` execution to cause the timer thread to stall indefinitely. That'd stall all timeouts on all connections using the same classloader as the thread is shared. Having the interrupt execution happen in a separate thread (say from a cached thread pool) should fix that.

Again, I'm not sure if that specific problem could happen as the only IO operations inside the `interrupt()` calls I could see are closing sockets but it could happen down the road with unrelated code changes. Synchronized blocks may lead to to a similar issue.

# Expected Behavior
1. When all connections are closed there should not be any extra threads running and all classes should be unloaded.
2. Timeout `interrupt()` operations from one command should not be able to stall timeouts for other commands.

Regarding actual implementation, since the newer versions of the driver target JDK 8+, I'd suggest using a shared `ScheduledThreadPoolExecutor` with a pool size of one (1) that could shutdown when there are no more connections referencing it. That'd be a slightly more modern version of what pgjdbc implemented.

One other advantage compared vs the current implementation of a single thread in a sleep / check loop is that that the standard implementation of ScheduledThreadPoolExecutor uses a heap to sort the timeout commands rather than looping through the entire list every time and it sleeps until the next timeout period rather than waking up every second.

[...]

PR #914, opened 2018-12-19
Refactor TimeoutPoller into SQLServerTimeoutManager

shayaantx:
As discussed in #909, we need to address the edge case where the timeout thread lives forever even after all the connections are closed and the classloader tries to unload the driver package.

Also addressed is the handling blocking interrupt calls within the timeout functionality


shayaantx:
@sehrope @ulvii please let me know of any questions or concerns

It would be kind of weird to write a test to simulate blocking an interrupt call, especially when there is no mocking framework in the project.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
```diff
@@ -1676,8 +1677,8 @@ private void getDestinationMetadata() throws SQLServerException {
                         ResultSet.CONCUR_READ_ONLY, connection.getHoldability(), stmtColumnEncriptionSetting);
 
                 // Get destination metadata
-                rs = stmt.executeQueryInternal("sp_executesql N'SET FMTONLY ON SELECT * FROM "
-                        + escapedDestinationTableName + " '");
+                rs = stmt.executeQueryInternal(
```
shayaantx: i think this formatting was wrong

src/main/java/com/microsoft/sqlserver/jdbc/TimeoutCommand.java
```diff
@@ -5,20 +5,28 @@
...
- * Abstract implementation of a command that can be timed out using the {@link TimeoutPoller}
+ * Abstract implementation of a command that can be timed out using the {@link SQLServerTimeoutManager}
  */
 abstract class TimeoutCommand<T> {
     private final long startTime;
     private final int timeout;
     private final T command;
     private final SQLServerConnection sqlServerConnection;
+    private ScheduledFuture<?> timeoutTask;
+    static AtomicInteger uniqueId = new AtomicInteger();
```
shayaantx: im open to suggestions regarding the id stuff, since its theoretically possible for someone to break this limit

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -7536,8 +7537,8 @@ final TDSReader startResponse(boolean isAdaptive) throws SQLServerException {
         // If command execution is subject to timeout then start timing until
         // the server returns the first response packet.
         if (queryTimeoutSeconds > 0) {
-            this.timeoutCommand = new TdsTimeoutCommand(queryTimeoutSeconds, this, null);
-            TimeoutPoller.getTimeoutPoller().addTimeoutCommand(this.timeoutCommand);
+            this.timeoutCommand = new TdsTimeoutCommand(queryTimeoutSeconds, this, tdsReader.getConnection());
```
cheenamalhotra: I'm wondering if passing connection is the right approach for calling `SQLServerTimeoutManager.startTimeoutCommand()` and would it be better to simply call `startTimeoutCommand()` with this instance as you're calling `releaseTimeoutCommand()`?

shayaantx: I think that would be more ideal, but I'll have to do something different inside of the timeout manager if we want to still follow the whole recommended approach where the connection on close releases the timeout commands.

Specifically make some sort of map of connection-> timeout commands and call release with the connection everytime close is called.

Thoughts?

shayaantx: @cheenamalhotra I made requested changes, please let me know of any other issues/concerns

src/main/java/com/microsoft/sqlserver/jdbc/TimeoutCommand.java
cheenamalhotra: Let's go for AtomicLong which has a very high range and is safer for applications running for longer time periods.

cheenamalhotra:
@sehrope 

Would request your participation in reviewing the PR as well, as we are looking forward to merge the fix soon.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerTimeoutManager.java
```diff
@@ -0,0 +1,139 @@
...
+            }
+        }
+    }
+
+    private static void addTimeoutCommand(TimeoutCommand<?> timeoutCommand) {
+        synchronized (timeoutCommands) {
+            UUID connectionId = timeoutCommand.getSqlServerConnection().getClientConIdInternal();
+            List<TimeoutCommand<?>> timeouts = timeoutCommands.get(connectionId);
+            if (timeouts == null) {
+                timeouts = new ArrayList<>();
```
cheenamalhotra: Using LinkedList will be a better idea if we're only adding/removing timeout commands to/from the list.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerTimeoutManager.java
```diff
@@ -0,0 +1,139 @@
...
+            logger.log(Level.WARNING, "Could not cancel timeout task", e);
+        }
+
+    }
+
+    static void releaseTimeoutCommands(UUID connectionId) {
+        synchronized (timeoutCommands) {
+            List<TimeoutCommand<?>> timeouts = timeoutCommands.get(connectionId);
+            if (timeouts != null && !timeouts.isEmpty()) {
+                for (TimeoutCommand<?> timeoutCommand : new ArrayList<>(timeouts)) {
```
cheenamalhotra: Also releasing with Iterator of LinkedList will be an O(1) as compared to ArrayList.remove() for each element.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerTimeoutManager.java
```diff
@@ -0,0 +1,139 @@
...
+ */
+final class SQLServerTimeoutManager {
+    private static ScheduledExecutorService scheduledTimeoutTasks = Executors.newScheduledThreadPool(1,
+            createThreadFactory("com.microsoft.sqlserver.jdbc.SQLServerTimeoutManager"));
+    private static ExecutorService timeoutTaskWorker = Executors.newSingleThreadExecutor(
+            createThreadFactory("com.microsoft.sqlserver.jdbc.SQLServerTimeoutManager.TimeoutTaskWorker"));
+    private static Map<UUID, List<TimeoutCommand<?>>> timeoutCommands = new HashMap<>();
+    private final static Logger logger = Logger.getLogger("com.microsoft.sqlserver.jdbc.SQLServerTimeoutManager");
+
+    static void startTimeoutCommand(TimeoutCommand<?> timeoutCommand) {
```
rene-ye: A bit hard to read due to the multiple nested runnables. Also shouldn't just blanket catch any exception and attribute the cause to a timeout. Suggested change pasted below:

```
static void startTimeoutCommand(TimeoutCommand<?> timeoutCommand) {
    if (scheduledTimeoutTasks.isShutdown()) {
        // reset id counter for timeout commands
        TimeoutCommand.uniqueId.set(0);
        scheduledTimeoutTasks = Executors.newScheduledThreadPool(1);
        timeoutTaskWorker = Executors.newSingleThreadExecutor();
    }

    Runnable timeoutTaskRunnable = () -> {
        Future<?> timeoutTask = timeoutTaskWorker.submit(() -> {
            timeoutCommand.interrupt();
        });
        try {
            // if the timeout command takes too long to interrupt, cancel it and release
            timeoutTask.get(10, TimeUnit.SECONDS);
        } catch (TimeoutException te) {
            logger.log(Level.WARNING, "Timeout task too long, aborting", te);
        } catch (InterruptedException | ExecutionException e) {
            logger.log(Level.FINE, "Unexpected Exception occured in timeout thread.", e);
        } finally {
            releaseTimeoutCommand(timeoutCommand);
        }
    };

    timeoutCommand.setTimeoutTask(scheduledTimeoutTasks.schedule(() -> {
        timeoutTaskRunnable.run();
    }, timeoutCommand.getTimeout(), TimeUnit.SECONDS));
    addTimeoutCommand(timeoutCommand);
}
```

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerTimeoutManager.java
```diff
@@ -0,0 +1,139 @@
...
+            public void run() {
+                Future<?> timeoutTask = timeoutTaskWorker.submit(new Runnable() {
+                    @Override
+                    public void run() {
+                        timeoutCommand.interrupt();
+                    }
+                });
+                try {
+                    // if the timeout command takes too long to interrupt, cancel it and release
+                    timeoutTask.get(10, TimeUnit.SECONDS);
```
rene-ye: This shouldn't be an arbitrary number, will think about what should be done here.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerTimeoutManager.java
```diff
@@ -0,0 +1,154 @@
...
+        if (timeouts.isEmpty()) {
+            // if there are no more timeouts, remove the connection from cache
+            timeoutCommands.remove(connectionId);
+        }
+    }
+
+    private static void checkForTimeoutThreadsShutdown() {
+        if (!areTimeoutCommandsAvailable()) {
+            scheduledTimeoutTasks.shutdownNow();
+            timeoutTaskWorker.shutdownNow();
```
rene-ye: This change doesn't really accurately resolve the original problem.
>When all connections are closed there should not be any extra threads running and all classes should be unloaded.

This change instead stops all the threads when there are no active commands, which might have some unintended consequences. It's perfectly normal to do work on a single connection. With the current change, threads would be spun up and stopped in between each new query assuming the user consumes all their resources.

shayaantx: In the latest commit I modified shutdown logic to only get called when connections are cleaned up and add a test case validating the threads are still around during a single connections lifespan

sehrope:
I'm hoping to check this out in detail this week. From initial skim through I think it can be simplified a lot. There shouldn't be a need to include unique ids or track connections in the timeout commands as the `.interrupt()` actions can include them inline as needed. Might even be able to remove the `TimeoutCommand` class entirely as the actions can be submitted as implementations of `Callable`.

Also, the lifecycle of the shared timer can be released (if necessary) in the `.close()` method of the Connection. If nothing ever uses a shared timer then no clean would be done. If there was one in use at some point in the lifecycle of the Connection then it'll be released when it is closed, and, if it's the last usage of the shared timer, then the timer's core ScheduledThreadPoolExecutor would also be shutdown. The idea is that rather than tracking individual tasks that are submitted (ex: per time out enabled query), you only need to track the check outs of the timer itself at the connection level.

cheenamalhotra:
Hi @sehrope

I agree with you, in fact I was working on a parallel implementation to fix the leaking thread issue in a simplified way than introducing SQLServerTimeoutManager. My changes can be found here: https://github.com/cheenamalhotra/mssql-jdbc/pull/20.

@shayaantx
Please take a look and let me know if you see any concerns with my implementation and if you too agree on simplifying the logic. 

Either you can pull those changes or I can come up with a PR on the main repo once we all have an agreement.

Thanks!

shayaantx:
@cheenamalhotra @sehrope I made your recommended changes to simply the timeout implementation.

@cheenamalhotra the only downside I found was that once I started relying on the connection references, the full test case suite is **not** close connections consistently. The TimeoutTest suite runs fine by itself, but when run with rest of the tests, I found 35+ connections that were added to the references set, but never cleaned up (the ones I looked at were stale connections that never got GC-ed or explicitly closed). I documented some of them below

- I disabled the 3 tests that check thread state (I don't know why @Ignore wasn't working, so I just commented out the test annotation)

Let me know how you want to proceed, I can go through all the cases and try to fix them in a separate PR if you'd like.

testTVPResultSet - the connection is never closed since the finally block is overriden with different functionality
	- saw this a few times
	
testTVPStoredProcedureResultSet - same thing above, calls above method
	- saw this a few times
	
testBulkCopyResultSet - same thing above
	- saw this a few times
	
TVPWithSqlVarientTest#testSetup - this one was a bit strange, it should get cleaned up in afterAll, however testSetup has an annontation of beforeEach, so my assumption with these are they just haven't been GC-ed yet
	- saw 2 connections from this one

testConnectionPoolGetTwice - I don't know enough the connection pools here to comment, but I assume its hasn't been GC-ed yet
	- also we are calling close on the connection when we don't need to
	
setShardingKeyTest - the connection is never explicitly closed

cheenamalhotra:
Hi @shayaantx 

I'm working on that, and will fix it in my PR #919 :v:

cheenamalhotra:
Hi @shayaantx 

PR #919 is merged, please update your branch and continue working.
I'll also update my PR with timeout Test I added, also includes DataSource, and other APIs to create connection.

cheenamalhotra:
Hi @shayaantx 

We reviewed all 3 changes proposed for fixing up Timeout Thread leak, and looks like our team has decided to go ahead with PR #920. The changes in all propositions were very similar and I believe we all intend to solve a common problem which all 3 PRs resolve in similar fashion.

We thank and welcome your contributions to the JDBC driver and look forward for more contributions in future. For now, we'll close this PR and proceed with merging PR #920.

Best Regards!

PR #920, opened 2019-01-05
Replace timeout handling to use SharedTimer

### LINKED ISSUE #909: Single timeout thread never stops and could block

sehrope:
PR to fix #909 following the reference counting model outlined in the comments. This is an alternative to #914. Was easier to put together a separate PR as the code changes were completely different.

This PR does not track the actual references or connection IDs, just a total count of how many times the SharedTimer has been requested (increment reference) and released (decrement reference). Connections that make use of the SharedTimer cache a reference to it and clean it up in `Connection.close()`. If the reference count drops to zero then the SharedTimer is shutdown and the core thread used for scheduling interrupt tasks is stopped.

Timeout that are actually triggered, i.e. the timer expired and the action hasn't completed yet, create a new thread on the fly to execute the interrupt action. This is to ensure that they do not block the core timer thread. Timeouts that are cancelled prior to expiring, i.e. query completed prior to timeout, do not create a new handler thread. 

All current  tests, including timeout tests, pass with this PR but there are no new tests yet to check if things are being cleaned up correctly. Manual testing via jconsole shows that it is but it'd be nice to eventually include something in the core test suite to cover that. Something similar to https://github.com/cheenamalhotra/mssql-jdbc/pull/20 should work though we'd have to deal with the issue of the rest of the test suite cleaning up all its resources prior to running (probably a good idea anyway!).

Alternatively we could have the thread leak tests be in their own test suite that is only run when explicitly enabled. Travis / Appveyor could then run them in isolation as a secondary step via: `ENABLE_TIMEOUT_TESTS=true mvn -Dtest=com.microsoft.sqlserver.jdbc.timeouts.TimeoutThreadLeakTest test`

I'll be taking a look at adding the tests when I have some more time but if someone else wants to do it in the meantime I can review.

Couple of notes on minor changes in behavior:

* Timeout operations are scheduled via `ScheduledThreadPoolExecutor` which internally sleeps until the next operation will execute. Previously they were being manually polled every 1-second so it was possible for a timeout to not execute for 2-seconds (*ex: poller is running at t=0, t=1, t=2, ... and timeout is scheduled at t=.01*). They will now execute as soon as they need to run.
* The clock for cancellation of bulk inserts now starts when the bulk command is going to execute as opposed to when it is created. The only blocking operation in between is waiting on an internal synchronization monitor, `schedulerLock`, so in practice it'd be microseconds of difference.


cheenamalhotra:
Hi @sehrope 

The implementation looks definitely cleaner although a few improvements can be done further to reduce code and achieve final design. I created PR https://github.com/sehrope/mssql-jdbc/pull/2 to your branch, which when merged will be final implementation for the timer thread that we would like to go ahead with.

sehrope:
Rebased atop latest dev branch and couple of additional commits for cleanup and tests added.

The first commit is the same as before (rebased on dev). The rest are broken down into small chunks to simplify review. Here's a summary:

* Replace timeout handling to use SharedTimer (main work of this PR)
* Normalize newlines in TimeoutTest - Replaces CRNL with just NL (no actual changes)
* Move TimeoutTest to parent package so that it can access package private members (no actual changes)
* Centralize SharedTimer thread checks in TimeoutTest - Adds `@Before`/`@After` methods to handle ensuring timer threads don't exist before and after tests run (rather than manually doing it in each test).
* Centralize and reduce timeout value used in TimeoutTest - Pulls up the 10-second timeout used by the existing tests and changes it to two-seconds (*test time goes down by 20 seconds, yay!*)
* Centralize SQL for WAIT FOR DELAY in TimeoutTest - Pulls up the WAITFOR SQL and formats it a bit to make it easier to read (`00:01` always looks like 1-second to me whereas `00:01:00` is clearly 1-minute)
* Clean up TimeoutTest to use assertThrows(...) - Changes existing tests to use assert
* Add additional SharedTimer tests - Adds a couple tests to verify creation and destruction of core timer thread (similar to @cheenamalhotra's tests but split out into a couple separate ones).

I've left in the `id` field on `SharedTimer` as one of the tests uses it to verify that the same instance is being used by multiple timer retrievals. The comparison could be done on the actual reference instead but I think the `id` based one is better long term as `getTimer()` could be expanded to provide a proxy that only allows the original caller a single dereference call so it wouldn't be the exact same physical reference to each caller (only same logical reference).

sehrope:
Looks like Travis passed fine but AppVeyor. I'm thinking it's just AppVeyor running a tad slower so it didn't have enough time to complete and clean up the thread.

I've pushed  another commit that changes the wait logic to repeatedly check if the thread is stopped in a loop for up to 10-seconds (previously it was 500ms).

src/main/java/com/microsoft/sqlserver/jdbc/TdsTimeoutTask.java
```diff
@@ -0,0 +1,62 @@
...
+                // If the timer wasn't canceled before it ran out of
+                // time then interrupt the registered command.
+                command.interrupt(SQLServerException.getErrString("R_queryTimedOut"));
+            }
+        } catch (SQLServerException e) {
+            // Unfortunately, there's nothing we can do if we fail to time out the request. There
+            // is no way to report back what happened.
+            assert null != command;
+            command.log(Level.WARNING, "Command could not be timed out. Reason: " + e.getMessage());
+            System.err.println(e.getMessage());
```
cheenamalhotra: This shouldn't have been picked up - I think I made another commit to remove this syserr which didn't get here.

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -6390,9 +6395,8 @@ synchronized final boolean readPacket() throws SQLServerException {
             // terminate the connection.
             if ((command.getCancelQueryTimeoutSeconds() > 0 && command.getQueryTimeoutSeconds() > 0)) {
```
rene-ye: I know this wasn't edited but shouldn't this be a `||` and not `&&`.

sehrope: I was curious about that as well but didn't touch it as wasn't entirely sure where the second half of that is loaded from.

Seems like it should be if either is defined with values less than zero considered to be zero:

```java
int seconds = Math.ceil(command.getCancelQueryTimeoutSeconds(),0)
     + Math.ceil(command.getQueryTimeoutSeconds(),0);
if (seconds > 0) {
    // if a timeout is configured with this object, add it to the timeout poller
    this.timeout = con.getSharedTimer().schedule(new TdsTimeoutTask(command, con), seconds);
}
```

rene-ye: I think I figured it out, the intention might've been
`(command.getCancelQueryTimeoutSeconds() > -1 && command.getQueryTimeoutSeconds() > 0)`

cancelQueryTimeout's "infinite" representation is -1, with 0 meaning literally 0. Whereas queryTimeout has an infinite timeout representation value of both 0 and -1. So this is meant to not schedule a timeout when either of the values are infinite.

sehrope: I just read through that link three times and still have no clue what "cancelQueryTimeout" is supposed to do:

> __cancelQueryTimeout__  Beginning with Microsoft JDBC Driver 6.4 for SQL Server, this property can be used to cancel queryTimeout set on the connection. Query execution hangs and does not throw exception if TCP connection to SQL Server is silently dropped. This property is only applicable if 'queryTimeout' is also set on the connection.
> The driver waits the total amount of cancelQueryTimeout + queryTimeout seconds, to drop the connection and close the channel. 

I can't tell if the space in "cancel queryTimeout" is a typo or if this property is meant to *cancel* a `queryTimeout` setting.

cheenamalhotra: @sehrope Yes this property is used to cancel 'QueryTimeout' and is only functional if queryTimeout is provided (not infinite) - hence the && check. This was introduced to fix #525 .

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
```diff
@@ -646,16 +623,14 @@ private void sendBulkLoadBCP() throws SQLServerException {
...
 
             final boolean doExecute() throws SQLServerException {
-                if (null != timeoutCommand) {
-                    if (logger.isLoggable(Level.FINEST))
-                        logger.finest(this.toString() + ": Starting bulk timer...");
-
-                    TimeoutPoller.getTimeoutPoller().addTimeoutCommand(timeoutCommand);
+                int timeoutSeconds = copyOptions.getBulkCopyTimeout();
+                if (timeoutSeconds > 0) {
+                    connection.checkClosed();
```
cheenamalhotra: Is this check required here? This code path is executed after the check has passed.

sehrope: Yes, the previous code had that check done in a ternary in the constructor.

My last note in the first description of this PR references this:

> The clock for cancellation of bulk inserts now starts when the bulk command is going to execute as opposed to when it is created. The only blocking operation in between is waiting on an internal synchronization monitor, schedulerLock, so in practice it'd be microseconds of difference.

src/main/java/com/microsoft/sqlserver/jdbc/TdsTimeoutTask.java
```diff
@@ -0,0 +1,61 @@
...
+
+import java.util.UUID;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.logging.Level;
+
+
+/**
+ * The TDS default implementation of a timeout command
+ */
+class TdsTimeoutTask implements Runnable {
```
peterbae: I think we want to capitalize TDS, for consistency

sehrope:
:tada:

PR #928, opened 2019-01-09
Fix | Security Issue specified by sonarqube

rene-ye:
Add a flag to XML transformer factory


src/main/java/com/microsoft/sqlserver/jdbc/PLPInputStream.java
```diff
@@ -494,6 +507,6 @@ public void reset() throws IOException {
             SQLServerException.makeFromDriverError(null, null, e.getMessage(), null, true);
         }
 
-        return null;
+        return new byte[0];
```
cheenamalhotra: Do not change implementation here.

PR #930, opened 2019-01-10
Fix SonarQube issues

peterbae:


src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -4880,7 +4880,7 @@ else if (DataTypes.UNKNOWN_STREAM_LENGTH == dataLength)
                 isShortValue = columnPair.getValue().precision <= DataTypes.SHORT_VARTYPE_MAX_BYTES;
                 isNull = (null == currentObject);
                 if (currentObject instanceof String)
```
cheenamalhotra: should we not check isNull here ?

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -4880,7 +4880,7 @@ else if (DataTypes.UNKNOWN_STREAM_LENGTH == dataLength)
...
-                    dataLength = isNull ? 0 : (ParameterUtils.HexToBin(currentObject.toString())).length;
+                    dataLength = ParameterUtils.HexToBin(currentObject.toString()).length;
                 else
                     dataLength = isNull ? 0 : ((byte[]) currentObject).length;
```
cheenamalhotra: If isNull is checked above, this check can go away too

peterbae: i think this one stays?

cheenamalhotra: Also, this method has 2 object instances `currentObject` and `currentObjectStringValue` that are compared to null over and over again, which can be done in the beginning of the method and a boolean instance for each can be used everywhere we need to check for null. SqlVariant has no null value check for `currentObject` that should be handled too..

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
peterbae: no, if currentObject is an instance of String, it cannot be null, so we dont have to check isNull here.

cheenamalhotra: Turns out `null instanceOf <Class>` does not error out, so we can avoid null check

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
cheenamalhotra: As discussed, we'll keep that change for later

PR #933, opened 2019-01-14
SonarQube Fixes to abide rules + Warnings fixed

cheenamalhotra:


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
```diff
@@ -1170,6 +1179,18 @@ private String getDestTypeFromSrcType(int srcColIndx, int destColIndx,
             // Encrypted columns are sent as binary data.
             tdsWriter.setCryptoMetaData(destColumnMetadata.get(destColIndx).cryptoMeta);
 
+            /*
+             * if source is encrypted and destination is unenecrypted, use destination's sql type to send since there is
```
rene-ye: >unenecrypted

PR #934, opened 2019-01-14
Fix | Handle Null resources in TVP to avoid NullPointerException

cheenamalhotra:
+Spell Corrections


src/main/java/com/microsoft/sqlserver/jdbc/TVP.java
```diff
@@ -89,18 +89,18 @@ void initTVP(TVPType type, String tvpPartName) throws SQLServerException {
     TVP(String tvpPartName, ResultSet tvpResultSet) throws SQLServerException {
         initTVP(TVPType.ResultSet, tvpPartName);
         sourceResultSet = tvpResultSet;
-        // Populate TVP metdata from ResultSetMetadta.
+        // Populate TVP metadata from ResultSetMetadta.
```
rene-ye: ResultSetMetadta to ResultSetMetadata

PR #940, opened 2019-01-22
Set up CI with Azure Pipelines

cheenamalhotra:


azure-pipelines.yml
```diff
@@ -0,0 +1,51 @@
...
+        Target_SQL: 'SQL-2k17-03'
+      SQL-2008R2:
+        Target_SQL: 'SQL-2k8R2-SP3-1'
+    maxParallel: 2
+  steps:
+  - powershell: |
+      mkdir AE_Certificates
+      cd AE_Certificates
+      $cert = New-SelfSignedCertificate -dns "AlwaysEncryptedCert" -CertStoreLocation Cert:CurrentUser\My
+      $pwd = ConvertTo-SecureString -String "password" -Force -AsPlainText
```
rene-ye: Will we have issues including the password here and below.

azure-pipelines.yml
```diff
@@ -0,0 +1,51 @@
...
+      checkStyleRunAnalysis: true
+      pmdRunAnalysis: true
+  - task: Maven@3
+    displayName: 'Maven build 42'
+    inputs:
+      mavenPomFile: 'pom.xml'
+      goals: 'clean -Dmssql_jdbc_test_connection_properties=jdbc:sqlserver://$(Target_SQL)$(server_domain);$(database);$(user);$(password); install -Pbuild42'
+      testRunTitle: 'Maven build 42'
+      javaHomeOption: Path
+      jdkDirectory: $(JDK11)
```
peterbae: by looking at the context, shouldnt this be JDK8?

azure-pipelines.yml
```diff
@@ -0,0 +1,51 @@
...
+      mavenPomFile: 'pom.xml'
+      goals: 'clean -Dmssql_jdbc_test_connection_properties=jdbc:sqlserver://$(Target_SQL)$(server_domain);$(database);$(user);$(password); install -Pbuild43'
+      testResultsFiles: '**/TEST-*.xml'
+      testRunTitle: 'Maven build 43'
+      javaHomeOption: Path
+      jdkDirectory: $(JDK11)
+      mavenAuthenticateFeed: true
+      checkStyleRunAnalysis: true
+      pmdRunAnalysis: true
+  - task: Maven@3
```
peterbae: should this be something other than Maven@3 since the task above already is named Maven@3 (does it work that way?)

cheenamalhotra: no.. they are independent tasks - Maven@3 means Maven task of version 3

azure-pipelines.yml
cheenamalhotra: No, this is going to build the jre8 driver profile, hence JDK 11 is required

azure-pipelines.yml
cheenamalhotra: Secured

PR #955, opened 2019-02-08
Fix | Fix possible Statement Leak in SQLServerConnection.isValid() API

### LINKED ISSUE #954: [QUESTION] should validate put close infinally block?
## Question
The statement of the validation query is not closed when an exception happen. The close should be moved to a finally block I guess. 

https://github.com/Microsoft/mssql-jdbc/blob/dev/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java#L5584

The comment below suggest you would have to catch exceptions and to set valid=false. This complicates the code but looks like the right thing to do.

cheenamalhotra:
Fixes #954 
- Create `SQLServerStatement` in try-with-resources block.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
@@ -5563,8 +5563,6 @@ public class SQLServerConnection implements ISQLServerConnection, java.io.Serial
      */
     @Override
     public boolean isValid(int timeout) throws SQLException {
-        boolean isValid = false;
-
         loggerExternal.entering(getClassNameLogging(), "isValid", timeout);
 
         // Throw an exception if the timeout is invalid
@@ -5578,6 +5576,7 @@ public class SQLServerConnection implements ISQLServerConnection, java.io.Serial
         if (isSessionUnAvailable())
             return false;
 
+        boolean isValid = true;
         try (SQLServerStatement stmt = new SQLServerStatement(this, ResultSet.TYPE_FORWARD_ONLY,
                 ResultSet.CONCUR_READ_ONLY, SQLServerStatementColumnEncryptionSetting.UseConnectionSetting)) {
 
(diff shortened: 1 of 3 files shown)


ecki:
I am afraid that PR changes the code semantics. The valid=true is set before the close, the following comment explicitly says that an exception on close is also invalid. Maybe it is easier to add a return false to the catch and remove the flag completely. This also allows for nicer logging

cheenamalhotra:
I added `invalid=false` on exception, to keep the exit gate consistent to end of method - takes care of exit logger too.

lilgreenbird:
> I added `invalid=false` on exception, to keep the exit gate consistent to end of method - takes care of exit logger too.

it's unnecessary as valid was init to false, so unless set to true it would already still be false

ecki:
> it's unnecessary as valid was init to false, so unless set to true it would already still be false

As long as isValid=true is inside the try block it will be executed before the auto close. If the autoclose fails the catch is executed while isValid is already true. So setting it to false is one possible fix to keep the old semantic.

(Personally i don’t like single-exit because of exactly that kind of problem but imguess its a good idea to stay consistent with existing conventions)

ulvii:
Would you re-run the CI? Seem like it fails on ConnectionDriverTest?

ecki:
Instill think the `invalid=false` is neeed

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -5581,13 +5580,13 @@ public boolean isValid(int timeout) throws SQLException {
...
             // if the query fails to execute in that time.
             stmt.executeQueryInternal("SELECT 1");
-            stmt.close();
             isValid = true;
         } catch (SQLException e) {
             // Do not propagate SQLExceptions from query execution or statement closure.
             // The connection is considered to be invalid if the statement fails to close,
             // even though query execution succeeded.
             connectionlogger.fine(toString() + " Exception checking connection validity: " + e.getMessage());
+            return false;
```
peterbae: can you set isValid = false here?

src/test/java/com/microsoft/sqlserver/jdbc/connection/ConnectionDriverTest.java
```diff
@@ -332,7 +332,7 @@ public void testDeadConnection() throws SQLException {
                 assertEquals(e.getMessage(), TestResource.getResource("R_connectionReset"),
                         TestResource.getResource("R_unknownException"));
             }
-            assertEquals(conn.isValid(5), false, TestResource.getResource("R_deadConnection"));
+            assertEquals(conn.isValid(0), false, TestResource.getResource("R_deadConnection"));
```
peterbae: Do we want to test isValid with infinite timeout? The method states that if the connection is dead, this method can hang forever.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -5581,18 +5589,14 @@ public boolean isValid(int timeout) throws SQLException {
...
         } catch (SQLException e) {
             // Do not propagate SQLExceptions from query execution or statement closure.
             // The connection is considered to be invalid if the statement fails to close,
             // even though query execution succeeded.
             connectionlogger.fine(toString() + " Exception checking connection validity: " + e.getMessage());
-        } finally {
-            if (null != stmt && !stmt.isClosed()) {
-                stmt.close();
-            }
+            isValid = false;
```
rene-ye: Nit-picky, but unnecessary line? It's by default false.

src/test/java/com/microsoft/sqlserver/jdbc/connection/ConnectionDriverTest.java
```diff
@@ -332,7 +330,7 @@ public void testDeadConnection() throws SQLException {
                 assertEquals(e.getMessage(), TestResource.getResource("R_connectionReset"),
                         TestResource.getResource("R_unknownException"));
             }
-            assertEquals(conn.isValid(5), false, TestResource.getResource("R_deadConnection"));
+            assertEquals(conn.isValid(10), false, TestResource.getResource("R_deadConnection"));
```
rene-ye: Do we still need this change after fixing the hanging issue.

cheenamalhotra: I'm gonna reset all other changes.. just a sec

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
cheenamalhotra: Here its needed, coz if `stmt.close()` fails.. we still want to invalidate.. or I can reverse the logic as in your PR..

PR #962, opened 2019-02-13
Fix for updating datetime value

peterbae:
Starting from SQL Server 2016, the rounding logic for inserting datetime value has been changed. Instead of uniformly rounding up values normally (<=4 rounds down, >=5 rounds up), the logic present in the getRoundedMilliseconds method shows how the numbers are actually rounded.

The writeRPCDateTime method that gets used during updating a datetimecolumn wasn't correctly reflecting this, and this PR addresses that issue.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
@@ -5337,16 +5337,9 @@ final class TDSWriter {
         int daysSinceSQLBaseDate = DDC.daysSinceBaseDate(cal.get(Calendar.YEAR), cal.get(Calendar.DAY_OF_YEAR),
                 TDS.BASE_YEAR_1900);
 
-        // Next, figure out the number of milliseconds since midnight of the current day.
-        int millisSinceMidnight;
         // Millis into the current second
-        if (con.getServerMajorVersion() >= SQL_SERVER_VERSION_2016) {
-            millisSinceMidnight = getRoundedMilliseconds(subSecondNanos);
-        } else {
-            millisSinceMidnight = (subSecondNanos + Nanos.PER_MILLISECOND / 2) / Nanos.PER_MILLISECOND;
-        }
-
-        millisSinceMidnight = millisSinceMidnight + 1000 * cal.get(Calendar.SECOND) + // Seconds into the current minute
+        int millisSinceMidnight = (subSecondNanos + Nanos.PER_MILLISECOND / 2) / Nanos.PER_MILLISECOND + 
+                1000 * cal.get(Calendar.SECOND) + // Seconds into the current minute
                 60 * 1000 * cal.get(Calendar.MINUTE) + // Minutes into the current hour
                 60 * 60 * 1000 * cal.get(Calendar.HOUR_OF_DAY); // Hours into the current day
 
@@ -5629,37 +5622,6 @@ final class TDSWriter {
         int roundedNanos = ((subSecondNanos + (Nanos.PER_MAX_SCALE_INTERVAL / 2)) / Nanos.PER_MAX_SCALE_INTERVAL)
                 * Nanos.PER_MAX_SCALE_INTERVAL;
         return roundedNanos;
-    }
-    
-    /**
-     * Returns subSecondNanos rounded to a millisecond value for SQL Server compatibility level >= 130 (SQL Server 2016 and after).
-     * The rounding logic below represents the actual rounding logic employed by datetime columns when determining their last millisecond digit.
-     */
-    private int getRoundedMilliseconds(int subSecondNanos) {
-        int fourthDigitOnward = subSecondNanos % 1000000;
-        int millis = subSecondNanos / 1000000;
-        int lastMillisDigit = millis % 10;
-        switch (lastMillisDigit) {
-            case 1:
-                if (fourthDigitOnward >= 666650) {
-                    millis++;
-                }
-                break;
-            case 4:
-                if (fourthDigitOnward >= 999950) {
-                    millis++;
-                }
-                break;
-            case 8:
-                if (fourthDigitOnward >= 333350) {
-                    millis++;
-                }
-                break;
-            default:
-                break;
-        }
-
-        return millis;
     }
 
     /**
--- a/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
@@ -5340,7 +5340,7 @@ final class TDSWriter {
         // Next, figure out the number of milliseconds since midnight of the current day.
         int millisSinceMidnight;
         // Millis into the current second
-        if (con.getServerMajorVersion() >= SQL_SERVER_VERSION_2016 || con.isAzure()) {
+        if (con.getServerMajorVersion() >= SQL_SERVER_VERSION_2016) {
             millisSinceMidnight = getRoundedMilliseconds(subSecondNanos);
         } else {
             millisSinceMidnight = (subSecondNanos + Nanos.PER_MILLISECOND / 2) / Nanos.PER_MILLISECOND;
--- a/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
@@ -3007,6 +3007,7 @@ final class SocketConnector implements Runnable {
 final class TDSWriter {
     private static Logger logger = Logger.getLogger("com.microsoft.sqlserver.jdbc.internals.TDS.Writer");
     private final String traceID;
+    private static final double SQL_SERVER_VERSION_2016 = 13.0;
 
     final public String toString() {
         return traceID;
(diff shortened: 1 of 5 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -5337,10 +5338,15 @@ void writeRPCDateTime(String sName, GregorianCalendar cal, int subSecondNanos,
                 TDS.BASE_YEAR_1900);
 
         // Next, figure out the number of milliseconds since midnight of the current day.
-        int millisSinceMidnight = (subSecondNanos + Nanos.PER_MILLISECOND / 2) / Nanos.PER_MILLISECOND + // Millis into
-                                                                                                         // the current
-                                                                                                         // second
-                1000 * cal.get(Calendar.SECOND) + // Seconds into the current minute
+        int millisSinceMidnight;
+        // Millis into the current second
+        if (con.getServerMajorVersion() >= SQL_SERVER_VERSION_2016) {
```
lilgreenbird: can you please put a little comment explaining why we have to do different things for different sql server versions?

lilgreenbird: and can we do that just once in the beginning so we don't have to check every time?

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -5337,10 +5338,15 @@ void writeRPCDateTime(String sName, GregorianCalendar cal, int subSecondNanos,
...
+            millisSinceMidnight = getRoundedMilliseconds(subSecondNanos);
+        } else {
+            millisSinceMidnight = (subSecondNanos + Nanos.PER_MILLISECOND / 2) / Nanos.PER_MILLISECOND;
+        }
+
+        millisSinceMidnight = millisSinceMidnight + 1000 * cal.get(Calendar.SECOND) + // Seconds into the current minute
```
lilgreenbird: it's nice to list out the calculations so it's easier to understand, but I'm wondering if we should just do all the calculations of the constant values 1st and just put this into the comments so we don't need to do this every time for performance reason?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -6223,35 +6228,57 @@ public void onEviction(CityHash128Key key, PreparedStatementHandle handle) {
...
                 rs.next();
-                isAzureDW = rs.getInt(1) == ENGINE_EDITION_FOR_SQL_AZURE_DW;
+
+                int engineEdition = rs.getInt(1);
+                isAzure = (engineEdition == ENGINE_EDITION_FOR_SQL_AZURE || engineEdition == ENGINE_EDITION_FOR_SQL_AZURE_DW || engineEdition == ENGINE_EDITION_FOR_SQL_AZURE_MI);
+                isAzureDW = (engineEdition == ENGINE_EDITION_FOR_SQL_AZURE_DW);
+                isAzureMI = (engineEdition == ENGINE_EDITION_FOR_SQL_AZURE_MI);
+
+            } catch (SQLException e) {
+                if (loggerExternal.isLoggable(java.util.logging.Level.FINER))
```
lilgreenbird: isn't it pretty bad if we can't get the version? should we die or at least log at a higher level than FINER?

peterbae:
I re-ran the CI build and it passed: https://dev.azure.com/sqlclientdrivers-ci/mssql-jdbc/_build/results?buildId=278

src/main/java/com/microsoft/sqlserver/jdbc/dtv.java
```diff
@@ -741,10 +741,15 @@ private void sendTemporal(DTV dtv, JavaType javaType, Object value) throws SQLSe
                     case DATETIME:
                     case DATETIME2:
                         // Default and max fractional precision is 7 digits (100ns)
-                        // Send DateTime2 to DateTime columns to let the server handle nanosecond rounding.
+                        // Send DateTime2 to DateTime columns to let the server handle nanosecond rounding. Also
+                        // adjust scale accordingly to avoid rounding on driver's end.
+                        int scale = typeInfo.getScale();
+                        if (typeInfo.getSSType() == SSType.DATETIME) {
+                            scale+=4;
+                        }
```
rene-ye:
```java
int scale = (typeInfo.getSSType() == SSType.DATETIME) ? typeInfo.getScale()+=4 : typeInfo.getScale()
```

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
peterbae: This is a very simple statement, if we can't get the version, the entire connection would have had some sort of other problem, but it isn't isAzure's job to handle/report it.

src/main/java/com/microsoft/sqlserver/jdbc/dtv.java
```diff
@@ -738,11 +738,15 @@ private void sendTemporal(DTV dtv, JavaType javaType, Object value) throws SQLSe
             if (null != typeInfo) // updater
             {
                 switch (typeInfo.getSSType()) {
+                    case DATETIME:
                     case DATETIME2:
                         // Default and max fractional precision is 7 digits (100ns)
+                        // Send DateTime2 to DateTime columns to let the server handle nanosecond rounding. Also
+                        // adjust scale accordingly to avoid rounding on driver's end.
```
lilgreenbird: sorry to nitpick, but multi-line comments should be in the /* */ format

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/DataTypesTest.java
```diff
@@ -1670,4 +1672,134 @@ public void testUpdateMisc() throws Exception {
...
     }
+    
+    /*
+     * SQL Server compatibility level >= 130 (SQL Server 2016 and after) doesn't round up or down equally for
+     * datetime. The rounding logic below represents the actual rounding logic employed by datetime
+     * columns when determining their last millisecond digit.
+     */
+    @Test
+    public void testDateTimeInsertUpdate() throws Exception {
+        assumeTrue(!isSqlAzureDW(), TestResource.getResource("R_cursorAzureDW"));
```
cheenamalhotra: Can this test be made compatible for Azure DW too?

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/DataTypesTest.java
```diff
@@ -1670,4 +1672,134 @@ public void testUpdateMisc() throws Exception {
...
+                DateTimeOffset dto6 = DateTimeOffset.valueOf(Timestamp.from(ist), 0);
+                
+                pstmt.setObject(1, dto6);
+                pstmt.addBatch();
+                
+                pstmt.executeBatch();
+                
+                Method method;
+                try {
+                    method = conn.getClass().getSuperclass().getDeclaredMethod("getServerMajorVersion");
```
cheenamalhotra: Please use `conn.getMetaData().getDatabaseMajorVersion()` instead.

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/DataTypesTest.java
peterbae: I tried, but the cursor type needs to be updatable, which means it's not supported for DW.

src/main/java/com/microsoft/sqlserver/jdbc/dtv.java
peterbae: done.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -6223,35 +6228,57 @@ public void onEviction(CityHash128Key key, PreparedStatementHandle handle) {
...
+    boolean isAzure() {
+        if (null == isAzure) {
+            try (Statement stmt = this.createStatement(); ResultSet rs = stmt.executeQuery("SELECT CAST(SERVERPROPERTY('EngineEdition') as INT)")) {
                 rs.next();
-                isAzureDW = rs.getInt(1) == ENGINE_EDITION_FOR_SQL_AZURE_DW;
+
+                int engineEdition = rs.getInt(1);
+                isAzure = (engineEdition == ENGINE_EDITION_FOR_SQL_AZURE || engineEdition == ENGINE_EDITION_FOR_SQL_AZURE_DW || engineEdition == ENGINE_EDITION_FOR_SQL_AZURE_MI);
+                isAzureDW = (engineEdition == ENGINE_EDITION_FOR_SQL_AZURE_DW);
+                isAzureMI = (engineEdition == ENGINE_EDITION_FOR_SQL_AZURE_MI);
```
rene-ye: Isn't this really redundant? Can't we just have an `EngineEdition` int and just set it here. To check if it's Azure/DW/MI we can just compare the engine edition variable to the statics. The 3 booleans are kind of messy and unnecessary.

peterbae: We could, but I wouldn't say it's unnecessary. Keeping them as variables lets us be able to compare the values once and return them as is, instead of having to compare against `EngineEdition` every time.

rene-ye: It's not really an issue with performance, an int compare instead of a boolean lookup has negligible impact. It just looks cluttered and we end up with a bunch of `if-else` scenarios.

peterbae: We have to do the boolean lookup anyways even in your case, so we will end up doing another comparison, every time isAzure() is called.

PR #966, opened 2019-02-20
Fix | Generate primary key constraints randomly

cheenamalhotra:
Fixes issues with parallel executions.


sehrope:
Both variables named `pk_constraint` should be renamed to camel case. Something like: `primaryKeyConstraintName`

Not specific to this, but we should also add consistent `private static final` qualifiers on all those constants. There's a random mix of some private, some static, some final, throughout the tests but I don't think any of them are ever reassigned.

PR #973, opened 2019-02-28
Improvements | SQLServerParameterMetadata improvements with improved code coverage

cheenamalhotra:


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerParameterMetaData.java
```diff
@@ -771,15 +775,19 @@ public int getParameterMode(int param) throws SQLServerException {
...
+                Map<String, Object> info = getParameterInfo(param);
+                if (null != info) {
+                    int n = (int) info.get("COLUMN_TYPE");
+                    switch (n) {
+                        case 1:
+                            return parameterModeIn;
+                        case 2:
+                            return parameterModeOut;
+                        default:
+                            return parameterModeUnknown;
```
ulvii: The `deafult` case can be removed.

cheenamalhotra: There's no harm in keeping it, plus static analysis tools would always need it

rene-ye: Can't the switch just be removed? There's not really enough results to warrant one.
```java
Map<String, Object> info = getParameterInfo(param);
if (null != info) {
    int n = (int) info.get("COLUMN_TYPE");
    if  (n == 1) {
        return parameterModeIn;
    } else if (n == 2) {
        return parameterModeOut;
    } else {
        return parameterModeUnknown;
    }
}

File: src/main/java/com/microsoft/sqlserver/jdbc/SQLServerParameterMetaData.java [hunk 2]
```diff
@@ -735,15 +735,19 @@ private void checkParam(int n) throws SQLServerException {
...
                 checkParam(param);
-                return queryMetaMap.get(param).parameterClassName;
+                parameterClassName = queryMetaMap.get(param).parameterClassName;
             } else {
-                JDBCType jdbcType = JDBCType.of((short) getParameterInfo(param).get("DATA_TYPE"));
-                return jdbcType.className();
+                Map<String, Object> info = getParameterInfo(param);
+                if (null != info) {
+                    parameterClassName = JDBCType.of((short) info.get("DATA_TYPE")).className();
+                }
```
rene-ye: So if getParameterInfo(param) returns null, we won't call any APIs which may cause a nullptr exception. However, the code in its current state will return a String object that's `null` in these scenarios instead of an actual empty String such as `""`. Do we want this behavior? If there are calls to `String.length()` using the String returned by these methods, we'll get nullptr issues again.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerParameterMetaData.java
ulvii: Line 912 has similar code (`if` statements are missing `{`, `}` by the way ), without the last redundant `else`.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerParameterMetaData.java
```diff
@@ -888,11 +909,14 @@ public int isNullable(int param) throws SQLServerException {
...
-                    return parameterNullable;
-                if (nNull == 0)
-                    return parameterNoNulls;
+                Map<String, Object> info = getParameterInfo(param);
+                if (null != info) {
+                    int nNull = (int) info.get("NULLABLE");
+                    if (nNull == 1)
+                        return parameterNullable;
+                    if (nNull == 0)
+                        return parameterNoNulls;
```
rene-ye: Similar code compared to the switch statement at the top. Consider keeping both these consistent? E.g if we decide to use switch statements for the top implementation, also use a switch here.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerParameterMetaData.java
```diff
@@ -929,6 +958,10 @@ public boolean isSigned(int param) throws SQLServerException {
 
     String getTVPSchemaFromStoredProcedure(int param) throws SQLServerException {
         checkClosed();
-        return (String) getParameterInfo(param).get("SS_TYPE_SCHEMA_NAME");
+        Map<String, Object> info = getParameterInfo(param);
+        if (null != info) {
+            return (String) info.get("SS_TYPE_SCHEMA_NAME");
+        }
+        return null;
```
rene-ye: return `null` or `""`?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerParameterMetaData.java
lilgreenbird: while we're at this, change the if above
            if (null == procMetadata) {

lilgreenbird: I dont see a big deal using switch or if/else. If keeping switch I don't think we should remove default. Default case should always be there for completeness and maintainability - if nothing then just break and do nothing

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerParameterMetaData.java
peterbae: i don't think that's the point of this change. the objective of this change is to avoid calling getParameterInfo(param).get("DATA_TYPE"), which assumes that getParameterInfo(param) is not going to be null (which it could). The PR is not concerned with how the String returned from this method is consumed, it's the caller's responsibility to check for nulls from this method (which is could return before, anyways).

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerParameterMetaData.java
peterbae: should return null imo

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerParameterMetaData.java
rene-ye: I don't think this change would be out of scope for the PR. Worst case scenario we could just  backlog it, but I don't think the fix is complicated enough for that. Returning `null` from this method makes no sense. We should either be throwing an error or returning `""`. According to javadocs, the function `Retrieves the fully-qualified name of the Java class whose instances should be passed to the method PreparedStatement.setObject`. `null` is not a valid "fully-qualified name" and the value is expected to be used with `setObject`, which may result in confusion as null is a "valid" value for any Object class and not only String.

peterbae: you're probably overthinking it, the method was already returning null before. I imagine this PR was made to account for the static analysis tool results, and I think the PR is aimed at solving that complaint. Not to mention that changing the method to return an empty string instead of null is a behavioral change, and it could potentially break existing applications.

rene-ye: It's not really about overthinking, it's just bad design. We'll probably have to take a call on the backwards compatibility thing, but returning `null` when there's no reason to isn't good practice. Creating PRs to please static code analysis tools but ignoring bad practice is a little convoluted since that's why we even use these tools.

cheenamalhotra: I think these APIs are returning null since beginning by design.. should rather throw exception. Will fix this behavior overall.

peterbae: Returning an empty string instead of a null when we do not have a corresponding entry for the class name isn't necessarily a better design. If the required information hasn't been provided, we should return it as is (null), instead of returning an instance of something that doesn't exist and giving it an empty String. Again, this is a public API that's being used by customers, and it could break existing applications for not a very strong reason.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerParameterMetaData.java
cheenamalhotra: It's as per existing design.. but I guess we should throw exception if no metadata is available for specified `param`.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerParameterMetaData.java
```diff
@@ -703,58 +703,66 @@ public boolean isWrapperFor(Class<?> iface) throws SQLException {
...
-                return procMetadata.get(param - 1);
-            }
+            return procMetadata.get(param - 1);
         } else {
             // Note row 1 is the 'return value' meta data
-            paramFound = procMetadata.size() > param;
-            if (paramFound) {
-                return procMetadata.get(param);
-            }
+            return procMetadata.get(param);
```
lilgreenbird: do we not want to check and error if procMetadata.size() < param anymore?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerParameterMetaData.java
```diff
@@ -703,58 +703,66 @@ public boolean isWrapperFor(Class<?> iface) throws SQLException {
...
         }
-        if (!paramFound) {
+    }
+
+    private boolean isValidParamProc(int n) {
+        // Note row 1 is the 'return value' meta data
+        return (((stmtParent).bReturnValueSyntax && isTVP && procMetadata.size() >= n) || procMetadata.size() > n);
```
peterbae: (stmtParent).bReturnValueSyntax => stmtParent.bReturnValueSyntax 👌

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerParameterMetaData.java
cheenamalhotra: `checkParam()` takes care of that - it wouldn't reach here in that case.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerParameterMetaData.java
```diff
@@ -765,170 +773,121 @@ public int getParameterCount() throws SQLServerException {
...
+                case microsoft.sql.Types.SMALLDATETIME:
+                    parameterType = SSType.DATETIME2.getJDBCType().asJavaSqlType();
+                    break;
+                case microsoft.sql.Types.MONEY:
+                case microsoft.sql.Types.SMALLMONEY:
+                    parameterType = SSType.DECIMAL.getJDBCType().asJavaSqlType();
+                    break;
+                case microsoft.sql.Types.GUID:
+                    parameterType = SSType.CHAR.getJDBCType().asJavaSqlType();
+                    break;
```
lilgreenbird: Not a big deal but I would've preferred to see an explicit default case f does nothing but break just for completeness and maintainability.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerParameterMetaData.java
```diff
@@ -765,170 +772,123 @@ public int getParameterCount() throws SQLServerException {
     @Override
     public int getParameterMode(int param) throws SQLServerException {
         checkClosed();
-        try {
-            if (procMetadata == null) {
-                checkParam(param);
-                // if it is not a stored proc, the param can only be input.
+        checkParam(param);
```
peterbae: I think this checkParam should be moved to line 777. The logic is different now from before - we didn't require isValidParamProc to be true to be able to call this method before (and successfully return from the else statement), but now we do. If this was intentional, I think we should cover this case specifically as well.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerParameterMetaData.java
```diff
@@ -765,170 +772,123 @@ public int getParameterCount() throws SQLServerException {
...
+                return parameterModeOut;
+            else
+                return parameterModeUnknown;
         }
     }
 
     @Override
     public int getParameterType(int param) throws SQLServerException {
         checkClosed();
+        checkParam(param);
```
peterbae: Similar comment here, to the comment I made above.

PR #974, opened 2019-03-06
Initial implementation of NTLM authentication

lilgreenbird:
- initial implementation of NTLM authentication
- MD4.java was obtained from OpenJDK with modifications to reduce dependencies


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -1581,6 +1586,22 @@ Connection connectInternal(Properties propsIn,
                         SQLServerException.getErrString("R_SetAuthenticationWhenIntegratedSecurityTrue"), null);
             }
 
+            if (authenticationString.equalsIgnoreCase(SqlAuthentication.NTLM.toString())) {
+                // domain and no user or password
+                if (!StringUtils.isEmpty(domainName) && ((activeConnectionProperties
```
cheenamalhotra:
```suggestion
                if (null == domainName || domainName.isEmpty() || (!domainName.isEmpty() && ((activeConnectionProperties...
```

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -4775,10 +4803,16 @@ final boolean complete(LogonCommand logonCommand, TDSReader tdsReader) throws SQ
         if (serverName != null && serverName.length() > 128)
             serverName = serverName.substring(0, 128);
 
+        String domainName = (null != currentConnectPlaceHolder) ? currentConnectPlaceHolder.getDomainName()
+                                                                : activeConnectionProperties.getProperty(
+                                                                        SQLServerDriverStringProperty.DOMAIN_NAME
+                                                                                .toString());
+        ntlmAuthentication = !StringUtils.isEmpty(domainName);
```
cheenamalhotra: Why this??

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -1581,6 +1586,22 @@ Connection connectInternal(Properties propsIn,
                         SQLServerException.getErrString("R_SetAuthenticationWhenIntegratedSecurityTrue"), null);
             }
 
+            if (authenticationString.equalsIgnoreCase(SqlAuthentication.NTLM.toString())) {
+                // domain and no user or password
```
cheenamalhotra: Read DomainName property here, if it’s not NTLM, we don’t need to read it..

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -1587,11 +1583,16 @@ Connection connectInternal(Properties propsIn,
...
+                String sPropValueDomain = activeConnectionProperties.getProperty(sPropKeyDomain);
+                String domainName = sPropValueDomain;
+
                 // domain and no user or password
-                if (!StringUtils.isEmpty(domainName) && ((activeConnectionProperties
-                        .getProperty(SQLServerDriverStringProperty.USER.toString()).isEmpty())
-                        || (activeConnectionProperties.getProperty(SQLServerDriverStringProperty.PASSWORD.toString())
-                                .isEmpty()))) {
+                if (null == domainName || domainName.isEmpty()
+                        || !domainName.isEmpty() && ((activeConnectionProperties
```
cheenamalhotra: You need an opening bracket before !domainName.isEmpty() ending at the end of condition..

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResource.java
```diff
@@ -385,6 +386,8 @@ static String getResource(String key) {
             {"R_UnknownFeatureAck", "Unknown feature acknowledge is received."},
             {"R_SetAuthenticationWhenIntegratedSecurityTrue",
                     "Cannot set \"Authentication\" with \"IntegratedSecurity\" set to \"true\"."},
+            {"R_NoUserPasswordForDomain",
+                    "Both \"User\" (or \"UserName\") and \"Password\" connection string keywords must be specified for \"domain\"."},
```
ulvii: I have a few concerns regarding this error message.
- The driver throws this error message even if both username and password are present, but domain name is missing. (`authentication=NTLM;username=user;password=password;`). I am suggesting to modify the error message to something along `User, Password, domain must be specified for NTLM authentication.`
- `connection string keywords` does not sound correct, because these properties can be specified through SQLServerDataSource too. Perhaps `Datasource/Connection String keywords` is a better option.

ulvii:
SQLServerDataSource APIs are missing.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -1581,6 +1582,27 @@ Connection connectInternal(Properties propsIn,
                         SQLServerException.getErrString("R_SetAuthenticationWhenIntegratedSecurityTrue"), null);
             }
 
+            if (authenticationString.equalsIgnoreCase(SqlAuthentication.NTLM.toString())) {
```
ulvii: SQLServerDataSource APIs are missing.

PR #976, opened 2019-03-07
Temporary PR to fix infinite wait on last executed command

rene-ye:


rene-ye:
Investigations finished.

PR #977, opened 2019-03-08
Fix | Fix SQL Exception Error State length to respect SQLSTATE Standards

### LINKED ISSUE #956: [BUG] Driver returns SQL State longer than 5 characters
## Driver version
7.2.1

## SQL Server version
N/A

## Client Operating System
N/A

## JAVA/JVM version
1.8, 11

## Table schema
N/A

## Problem description
1. Expected behaviour: when an error occurs the driver should throw a SQL exception with a 5 character SQL state
2. Actual behaviour: SQL state is sometimes 6 characters due to this code: https://github.com/Microsoft/mssql-jdbc/blob/9c5cc97f661fe14de11df6be16c8acbc3379d1ff/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerException.java#L364

The `EXECUTE AS USER = 'DOMAIN/USER'` can for example result in SQL state `S00011` (6 chars!) and the message `Could not obtain information about Windows NT group/user '[USER_NAME]', error code 0x4bc.`

This was reported to us by a customer: See https://github.com/flyway/flyway/issues/2275

We believe the driver's behavior is incorrect in this case as it violates [the SQLSTATE standard](https://en.wikipedia.org/wiki/SQLSTATE).

cheenamalhotra:
Fixes #956 

Calling `SQLException.getErrorState()` would result in SQL State of length more than 5 characters allowed as per [SQL State Specifications](https://en.wikipedia.org/wiki/SQLSTATE). This fix restricts length of SQLException Error State and appends zeroes (0) as needed to Error State received from SQL Server.

As per SQL Server specifications, the values of [ERROR_STATE](https://docs.microsoft.com/en-us/sql/t-sql/functions/error-state-transact-sql?view=sql-server-2017#return-value) ranges from 0 - 255 as one can raise an error with custom Error State using [RAISEERROR](https://docs.microsoft.com/en-us/sql/t-sql/language-elements/raiserror-transact-sql?view=sql-server-2017) as well. Accordingly, we must adjust trailing zeroes to the SQLException Error State.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerException.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerException.java
@@ -323,13 +323,13 @@ public final class SQLServerException extends java.sql.SQLException {
      *        the database state
      * @return the state code
      */
-    static String generateStateCode(SQLServerConnection con, int errNum, int databaseState) {
+    static String generateStateCode(SQLServerConnection con, int errNum, Integer databaseState) {
         // Generate a SQL 99 or XOPEN state from a database generated error code
         boolean xopenStates = (con != null && con.xopenStates);
         if (xopenStates) {
             switch (errNum) {
                 case 4060:
-                    return "08001"; // Database name undefined at loging
+                    return "08001"; // Database name undefined at logging
                 case 18456:
                     return "08001"; // username password wrong at login
                 case 2714:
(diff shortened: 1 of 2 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerException.java
```diff
@@ -360,8 +360,19 @@ static String generateStateCode(SQLServerConnection con, int errNum, int databas
...
+                    String dbState = databaseState.toString();
+                    /*
+                     * Length allowed for SQL State is 5 characters as per SQLSTATE specifications. Append trailing
+                     * zeroes as needed based on length of database error State as length of databaseState is either 1
+                     * or 2 characters.
+                     */
+                    StringBuilder trailingZeroes = new StringBuilder("S");
+                    for (int i = 0; i < 4 - dbState.length(); i++) {
+                        trailingZeroes.append("0");
+                    }
```
rene-ye: `String.format("S%4s", s).replace(' ', '0')` pads 0's to the left of the string up to a length of 4 and appends uppercase 'S' to the beginning. The default block can just become:
```java
default:
    return String.format("S%4s", databaseState.toString()).replace(' ', '0');
```

cheenamalhotra: String.format is not good at performance when compared with StringBuilder, hence not implemented.

For negative values test scenario, the negative values if passed to SQL Server with RAISEERROR, are received from SQL Server as positive integer 1, hence no check is needed.

PR #982, opened 2019-03-11
Tests | Junit improvements + Organize constants

cheenamalhotra:
This PR includes changes as under:
- Manage Constants in one common class and use static reference to access them across all tests.
- Removed magic strings and numerics in tests to use static constants instead.
- Created `getConnection()` method in **AbstractTest** class that returns Connection using available _connectionString_ and updated tests to use this instead of creating connection from `DriverManager`.
- Updated other custom connection requests to get connection from **PrepUtil** class.
- Updated `e.toString()` to `e.getMessage()`
- Closed open statement(s) in some files.
- Removed unused methods (we are using alternative methods to do the same stuff)
- Organize imports + fixed warnings + cleanups.


src/test/java/com/microsoft/sqlserver/testframework/Constants.java
```diff
@@ -0,0 +1,98 @@
...
+import java.util.concurrent.ThreadLocalRandom;
+import java.util.logging.Logger;
+
+import com.microsoft.sqlserver.jdbc.RandomUtil;
+
+
+public class Constants {
+    protected static final ThreadLocalRandom RANDOM = ThreadLocalRandom.current();
+    protected static final Logger LOGGER = Logger.getLogger("AbstractTest");
+    protected static final String JKS_SECRET_STRING = "changeit";
```
rene-ye: Not so secret anymore 🤔

cheenamalhotra: It isn't a secret anyway 😃

PR #983, opened 2019-03-12
Fix | Fix SQLServerConnection.abort() API behavior to clear resources consistently

cheenamalhotra:
As per [Java Specifications](https://docs.oracle.com/javase/7/docs/api/java/sql/Connection.html#abort(java.util.concurrent.Executor)), executing `Connection.abort(Executor executor)`, must also release all resources and mark the connection closed as `Connection.close()` would.

The PR fixes the same by managing resource cleanup separately in a common internal method.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
@@ -3191,9 +3191,10 @@ public class SQLServerConnection implements ISQLServerConnection, java.io.Serial
             SQLServerException.makeFromDriverError(null, null, form.format(msgArgs), null, false);
         } else {
 
-            // Always report the connection as closed for any further use, no matter
-            // what happens when we try to clean up the physical resources associated
-            // with the connection using executor.
+            /*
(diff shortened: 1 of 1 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -3191,11 +3185,19 @@ public void abort(Executor executor) throws SQLException {
...
+        if (null == executor) {
+            MessageFormat form = new MessageFormat(SQLServerException.getErrString("R_invalidArgument"));
+            Object[] msgArgs = {"executor"};
+            SQLServerException.makeFromDriverError(null, null, form.format(msgArgs), null, false);
+        } else {
 
-        setState(State.Closed);
+            // Always report the connection as closed for any further use, no matter
+            // what happens when we try to clean up the physical resources associated
+            // with the connection using executor.
```
lilgreenbird: /*
*/
and, might as well change line 3178 to null != executor?

cheenamalhotra: Done

Issue #961, opened 2019-02-12
[BUG] Data type float is not supported in bulk copy.


pkgajulapalli:
## Driver version
7.0.0.jre8

## SQL Server version
Microsoft SQL Server 2016 (SP2-CU3) (KB4458871) - 13.0.5216.0 (X64)
	Sep 13 2018 22:16:01
	Copyright (c) Microsoft Corporation

## Client Operating System
Mac OS

## JAVA/JVM version
java version "1.8.0_101"

## Problem description
I have to transfer data from one database server to a SQL Server. I'm using SQLServerBulkCopy to do that:
`// connection1 is with the source system and`
`// connection2 is with the destination SQL Server`
`Statement statement = connnection1.createStatement();`
``
`ResultSet resultSet = statement.executeQuery("select * from db.table");`
``
`SQLServerBulkCopy bulkCopy = new SQLServerBulkCopy(connection2);`
`bulkCopy.setDestinationTableName("tableName");`
`bulkCopy.writeToServer(resultSet);`

## JDBC trace logs
> com.microsoft.sqlserver.jdbc.SQLServerException: Data type float is not supported in bulk copy.
> 
> at com.microsoft.sqlserver.jdbc.SQLServerException.makeFromDriverError(SQLServerException.java:226)
> at com.microsoft.sqlserver.jdbc.SQLServerBulkCopy.getDestTypeFromSrcType(SQLServerBulkCopy.java:1443) at com.microsoft.sqlserver.jdbc.SQLServerBulkCopy.createInsertBulkCommand(SQLServerBulkCopy.java:1464) at
>  com.microsoft.sqlserver.jdbc.SQLServerBulkCopy.sendBulkCopyCommand(SQLServerBulkCopy.java:1611) at
>  com.microsoft.sqlserver.jdbc.SQLServerBulkCopy.doInsertBulk(SQLServerBulkCopy.java:1553) at
>  com.microsoft.sqlserver.jdbc.SQLServerBulkCopy.access$200(SQLServerBulkCopy.java:63) at
>  com.microsoft.sqlserver.jdbc.SQLServerBulkCopy$1InsertBulk.doExecute(SQLServerBulkCopy.java:705) at
>  com.microsoft.sqlserver.jdbc.TDSCommand.execute(IOBuffer.java:7240) at
>  com.microsoft.sqlserver.jdbc.SQLServerConnection.executeCommand(SQLServerConnection.java:2869) at
[...]

PR #986, opened 2019-03-12
Support FLOAT data type for bulk copy operation using RowSet

peterbae:
Fixes issue #961

### DIFF
--- a/src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyRowSetTest.java
+++ b/src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyRowSetTest.java
@@ -1,0 +1,70 @@
+package com.microsoft.sqlserver.jdbc.bulkCopy;
+
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.sql.SQLException;
+import java.sql.Statement;
+
+import javax.sql.RowSetMetaData;
+import javax.sql.rowset.CachedRowSet;
+import javax.sql.rowset.RowSetFactory;
+import javax.sql.rowset.RowSetMetaDataImpl;
+import javax.sql.rowset.RowSetProvider;
+
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.opentest4j.TestAbortedException;
+
+import com.microsoft.sqlserver.jdbc.RandomUtil;
+import com.microsoft.sqlserver.jdbc.SQLServerBulkCopy;
+import com.microsoft.sqlserver.jdbc.SQLServerStatement;
+import com.microsoft.sqlserver.jdbc.TestUtils;
+import com.microsoft.sqlserver.testframework.AbstractSQLGenerator;
+import com.microsoft.sqlserver.testframework.AbstractTest;
+
+public class BulkCopyRowSetTest extends AbstractTest {
+    
+    static String tableName = AbstractSQLGenerator.escapeIdentifier(RandomUtil.getIdentifier("BulkCopyFloatTest"));
+    
+    @Test
+    public void testBulkCopyFloatRowSet() throws SQLException {
+        try (Connection con = DriverManager.getConnection(connectionString)) {
+            RowSetFactory rsf = RowSetProvider.newFactory();
+            CachedRowSet crs = rsf.createCachedRowSet();
+            RowSetMetaData rsmd = new RowSetMetaDataImpl();
+            rsmd.setColumnCount(1);
+            rsmd.setColumnName(1, "c1");
+            rsmd.setColumnType(1, java.sql.Types.FLOAT);
+            
+            crs.setMetaData(rsmd);
+            crs.setFloat(1, 123.45f);
+            
+            SQLServerBulkCopy bcOperation = new SQLServerBulkCopy(con);
+            bcOperation.setDestinationTableName(tableName);
+            bcOperation.writeToServer(crs);
+            bcOperation.close();
+        }
+    }
+    
+    @BeforeEach
+    public void testSetup() throws TestAbortedException, Exception {
+        try (Connection connection = DriverManager
+                .getConnection(connectionString)) {
+            try (Statement stmt = (SQLServerStatement) connection.createStatement()) {
+                TestUtils.dropTableIfExists(tableName, stmt);
+                String sql1 = "create table " + tableName + " (c1 float)";
+                stmt.execute(sql1);
+            }
+        }
+    }
+
+    @AfterAll
+    public static void terminateVariation() throws SQLException {
+        try (Connection connection = DriverManager.getConnection(connectionString)) {
+            try (Statement stmt = (SQLServerStatement) connection.createStatement()) {
+                TestUtils.dropTableIfExists(tableName, stmt);
+            }
+        }
+    }
+}
--- a/src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyRowSetTest.java
+++ b/src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyRowSetTest.java
@@ -49,22 +49,19 @@ public class BulkCopyRowSetTest extends AbstractTest {
     
     @BeforeEach
     public void testSetup() throws TestAbortedException, Exception {
-        try (Connection connection = DriverManager
-                .getConnection(connectionString)) {
-            try (Statement stmt = (SQLServerStatement) connection.createStatement()) {
-                TestUtils.dropTableIfExists(tableName, stmt);
-                String sql1 = "create table " + tableName + " (c1 float)";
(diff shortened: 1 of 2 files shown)


src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyRowSetTest.java
```diff
@@ -0,0 +1,68 @@
...
+            crs.setFloat(1, RandomData.generateReal(false));
+            
+            try (SQLServerBulkCopy bcOperation = new SQLServerBulkCopy(con)) {
+                bcOperation.setDestinationTableName(tableName);
+                bcOperation.writeToServer(crs);
+            }
+        }
+    }
+    
+    @BeforeEach
```
cheenamalhotra: I think you wanna run this method as `@BeforeAll` instead? As you could extend this table to include other datatypes in future and set this up once for all..

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyRowSetTest.java
```diff
@@ -0,0 +1,68 @@
...
+    public void testSetup() throws TestAbortedException, Exception {
+        try (Connection connection = DriverManager.getConnection(connectionString);
+                Statement stmt = (SQLServerStatement) connection.createStatement()) {
```
cheenamalhotra: Avoid casting to SQLServerStatement/SQLServerConnection unless using driver specific APIs.

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyRowSetTest.java
```diff
@@ -0,0 +1,68 @@
...
+            rsmd.setColumnName(1, "c1");
+            rsmd.setColumnType(1, java.sql.Types.FLOAT);
+            
+            crs.setMetaData(rsmd);
+            crs.setFloat(1, RandomData.generateReal(false));
+            
+            try (SQLServerBulkCopy bcOperation = new SQLServerBulkCopy(con)) {
+                bcOperation.setDestinationTableName(tableName);
+                bcOperation.writeToServer(crs);
+            }
```
cheenamalhotra: Must also do verification to check if the fetched data is same as inserted.
Consider generating multiple Random Value(s) (different range of data) during setup (beforeAll) and during insert also add additional IDENTITY column to table to be able to fetch in the same order.

peterbae: Added check. I didn't feel like running the same test an arbitrary number of times was necessary though.

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyRowSetTest.java
peterbae: It will probably stay as '@BeforeEach' in the future, so that the table is dropped and re-created before each test, separating each test run from other runs.

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyRowSetTest.java
peterbae: done

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyRowSetTest.java
cheenamalhotra: Then how will u manage different datatypes - do u think you will need same table recreated for all tests?

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyRowSetTest.java
```diff
@@ -0,0 +1,80 @@
...
+        try (Connection con = DriverManager.getConnection(connectionString);
+                Statement stmt = connection.createStatement()) {
+            RowSetFactory rsf = RowSetProvider.newFactory();
+            CachedRowSet crs = rsf.createCachedRowSet();
+            RowSetMetaData rsmd = new RowSetMetaDataImpl();
+            rsmd.setColumnCount(1);
+            rsmd.setColumnName(1, "c1");
+            rsmd.setColumnType(1, java.sql.Types.FLOAT);
+            
+            Float floatData = RandomData.generateReal(false);
```
cheenamalhotra: why not calling generateFloat?

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyRowSetTest.java
cheenamalhotra: What I meant to say was there are 2 types of Floats we should test (least of all) with null value
`RandomData.generateFloat(24, false);`
`RandomData.generateFloat(53, false);`
`RandomData.generateFloat(53, true); //null`

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyRowSetTest.java
peterbae: generateFloat creates a double. generateReal creates the java equivalent of a float (4bytes)

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyRowSetTest.java
peterbae: As for the null case, it seems we can't put nulls when using bulk copy using a RowSet because the CachedRowSetImpl doesn't allow null values.
As for the generateFloat(24) and generateFloat(53), they're actually an identical case in this scenario. When the user specifies `FLOAT` in the RowSetMetaData, the driver will send the java equivalent of a float, which truncates the 53 case down to 24, except now we have to deal with rounding errors in the assertEquals line.

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyRowSetTest.java
peterbae: We might create the same table that contains all datatypes, we don't know what will be included yet. Whoever writes the test in the future can modify this line if they want.

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyRowSetTest.java
```diff
@@ -0,0 +1,83 @@
...
+        try (Connection con = DriverManager.getConnection(connectionString);
+                Statement stmt = connection.createStatement()) {
+            RowSetFactory rsf = RowSetProvider.newFactory();
+            CachedRowSet crs = rsf.createCachedRowSet();
+            RowSetMetaData rsmd = new RowSetMetaDataImpl();
+            rsmd.setColumnCount(2);
+            rsmd.setColumnName(1, "c1");
+            rsmd.setColumnName(2, "c2");
+            rsmd.setColumnType(1, java.sql.Types.FLOAT);
+            rsmd.setColumnType(2, java.sql.Types.REAL);
```
cheenamalhotra:
```suggestion
            rsmd.setColumnType(2, java.sql.Types.FLOAT);
```

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyRowSetTest.java
```diff
@@ -0,0 +1,83 @@
...
+            crs.moveToCurrentRow();
+            
+            try (SQLServerBulkCopy bcOperation = new SQLServerBulkCopy(con)) {
+                bcOperation.setDestinationTableName(tableName);
+                bcOperation.writeToServer(crs);
+            }
+            
+            ResultSet rs = stmt.executeQuery("select * from " + tableName);
+            rs.next();
+            assertEquals(floatData, (Float) rs.getFloat(1));
```
cheenamalhotra:
```suggestion
            assertEquals(floatData, (Float) rs.getFloat(1));
            assertEquals(floatData, (Float) rs.getFloat(2));
```

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyRowSetTest.java
```diff
@@ -0,0 +1,83 @@
...
+        }
+    }
+    
+    @BeforeAll
+    public static void testSetup() throws TestAbortedException, Exception {
+        try (Connection connection = DriverManager.getConnection(connectionString);
+                Statement stmt = connection.createStatement()) {
+            TestUtils.dropTableIfExists(tableName, stmt);
```
cheenamalhotra:
```suggestion
           
```

Issue #539, opened 2017-11-02
Setting a BIT field to 1 with TVPs results in a 0


alexnixon:
```
SQLServerDataTable table = new SQLServerDataTable();
table.addColumnMetadata("Wat", Types.BIT);
table.addRow(1);
```
This table will store a 0 into your database, because the `1` gets converted to `"1"` with `toString`, then parsed to `false` with `Boolean.parseBoolean`.

I personally disagree with `SQLServerDataTable::addRow`'s general approach of "call `toString` on the argument then parse it in a lax manner". I think it would be better to:

- Handle known types specifically, e.g. if you attempt to set a BIT field with a `Byte`, cast it to a `Byte` and handle the value
- Handle known values specifically, e.g. allow 0, 1, true, false to set a BIT field
- Reject unknown types and nonsense values with an exception

gordthompson:
Related: #538 (via #531 ).

cheenamalhotra:
Thanks @gordthompson for relating PR to the issue. 

> Handle known types specifically, e.g. if you attempt to set a BIT field with a Byte, cast it to a Byte and handle the value

I guess that's implemented in PR.

> Handle known values specifically, e.g. allow 0, 1, true, false to set a BIT field

We will update PR to parse inputs (1, 0) for Boolean type (True, False) respectively.

> Reject unknown types and nonsense values with an exception

SQL Server does that for us, so additional logic in Java is not needed. We rely on SQLException and throw it directly. Is there any specific case where it will be helpful [_other than integers 0,1 for Boolean that will now be handled_]?

alexnixon:
Thanks for the update. I disagree with your point that we can rely on SQL Server to reject unknown types because the code frequently calls `toString` on the values passed in then parsing them. So all the downstream code sees is whatever type the string is subsequently parsed as.

To clarify I've come up with some examples by looking at the code (though I haven't run these so correct me if I'm wrong). I believe the PR does fix the BIT case that I reported but there are still several cases where I think the code continues to be overly permissive and others where I think the PR has introduced regressions:

```
Object unknownTypeInstance = new Object() { @Override public String toString() { return "1"; } };

{
    SQLServerDataTable table = new SQLServerDataTable();
    table.addColumnMetadata("c1", Types.BIT);

    //     Before PR: insert 0
    //      After PR: insert 1
    // What I expect: REJECT because the type is unknown
    table.addRow(unknownTypeInstance);

    //     Before PR: insert 0
[...]

PR #990, opened 2019-03-14
Performance | Improve performance of SQLServerDataTable.internalAddRow() function

cheenamalhotra:
Fixes issue #531 and #539 

Rework of PR #538


cheenamalhotra:
@alexnixon @cogman 

I recreated the original PR #538 for improving performance in SQLServerDataTable considering regression scenarios and what we discussed before. I have added tests to cover discussed cases.

**Note**: I relooked over allowing custom object types when adding row data, and I think that should be supported as it is. The underlying datatype being passed decides the type of the value added to the column and should work as expected.

Let me know if you have any concerns.

alexnixon:
Hi Cheena.

Thanks for revisiting this. Initial thoughts are that  having some tests around this is great.

On performance though this PR still leaves lots of cases which go via string serialization. For example using a `java.lang.Integer` to set a `BIGINT` field or using a `java.lang.Byte` to set a `BIT` field.

The type-testing approach I suggested [in this comment](https://github.com/Microsoft/mssql-jdbc/pull/538#issuecomment-343945081) on the previous PR handles all these cases and more - do you have an objection to it?

(defaulting to using `toString` for unknown types as you suggested seems fair enough for backwards compatibility's sake)

cheenamalhotra:
Hi @alexnixon 

I considered that before but it increases code complexity and performs conversions even when not needed (which impacts performance for corresponding types).

With this change, we do give maximum performance for corresponding datatypes, but for values whose datatypes do not correspond to JDBC Type, we support them by converting to String generically... which seems a fair deal to me. It is not something that we recommend for users either way and need not overload implementation for all such possibilities.

alexnixon:
Could you give me specific example of an unneeded conversion? I'm struggling to see any case in which my suggestion is slower<sup>[1]</sup> but can see many important cases which are substantially faster (e.g. the aforementioned using an `java.lang.Integer` to set a `BIGINT` field).

I know performance isn't often a big deal but these cases are _abysmally_ slow and given the client is using TVPs it's not unlikely they're shoveling large amount of data through here.

<sup>[1] unless you count the `(BIGINT, Long)` case, where I didn't write the `Long` type-test first, so technically there are a few instructions being wasted. Easy to shuffle that test to the top if you care.</sup>

cogman:
@alexnixon I believe @cheenamalhotra is referring to all the instance of checks that end up happening.

I believe the complexity tradeoff is fair.  In most (all?) cases I'm matching the types up anyways.   Even if I didn't do that, I could make the decision to do it at the point of creating the TVP to get better performance with @cheenamalhotra's implementation (I could handle the up/down casting myself)

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDataTable.java
```diff
@@ -155,33 +155,43 @@ public synchronized void addRow(Object... values) throws SQLServerException {
      */
     private void internalAddrow(JDBCType jdbcType, Object val, Object[] rowValues,
             Map.Entry<Integer, SQLServerDataColumn> pair) throws SQLServerException {
+        if (null != val) {
+            SQLServerDataColumn currentColumnMetadata = pair.getValue();
+            int nValueLen;
+            switch (jdbcType) {
+                case BIGINT:
+                    rowValues[pair.getKey()] = (val instanceof Long) ? (long) val : Long.parseLong(val.toString());
```
cogman: Rather than casting from the Object to the primitive, it would be better to keep this as the object itself since it is going to be assigned directly into an Object[].  That will prevent unnecessary unboxing and reboxing.

That is to say, this line could read

    (val instanceof Long) ? val : Long.parseLong(val.toString())

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDataTable.java
```diff
@@ -207,92 +217,87 @@ private void internalAddrow(JDBCType jdbcType, Object val, Object[] rowValues,
...
+
+                    if (nValueLen > currentColumnMetadata.precision) {
+                        currentColumnMetadata.precision = nValueLen;
+                        columnMetadata.put(pair.getKey(), currentColumnMetadata);
+                    }
+                    rowValues[pair.getKey()] = (byte[]) val;
+                    break;
+
+                case CHAR:
+                    if (val instanceof UUID)
```
cogman: Why is this special conversion for UUID only present on CHAR and not the other string data types?  Out of scope for this review, just seems a little bit odd.

cheenamalhotra:
@alexnixon There are many computations going on in that approach, and I think its fair to keep the implementation less complicated inside driver. e.g. working with all Numeric types as Long and then converting them to Short/Integer class objects, etc. also opens up endless possibilities with other types too, whereas it doesn't give much of an advantage vs its need for end user.

And as far as performance is concerned, I agree the corresponding types must be quickly passed through and hence this change is well deserved.

@cogman Thanks, I made quick modifications as per your suggestions. It looks clean now.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDataTable.java
```diff
@@ -207,92 +217,86 @@ private void internalAddrow(JDBCType jdbcType, Object val, Object[] rowValues,
...
+                    // Sending temporal types as string. Error from database is thrown if parsing fails
+                    // no need to send precision for temporal types, string literal will never exceed
+                    // DataTypes.SHORT_VARTYPE_MAX_BYTES
+
+                    // java.sql.Date, java.sql.Time and java.sql.Timestamp are subclass of java.util.Date
+                    if (val instanceof java.util.Date || val instanceof microsoft.sql.DateTimeOffset
+                            || val instanceof OffsetDateTime || val instanceof OffsetTime)
+                        rowValues[pair.getKey()] = val.toString();
+                    else
+                        rowValues[pair.getKey()] = val;
```
peterbae: Don't we need to cast this to String?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDataTable.java
```diff
@@ -207,92 +217,86 @@ private void internalAddrow(JDBCType jdbcType, Object val, Object[] rowValues,
...
+                case BINARY:
+                case VARBINARY:
+                case LONGVARBINARY:
+                    nValueLen = ((byte[]) val).length;
+
+                    if (nValueLen > currentColumnMetadata.precision) {
+                        currentColumnMetadata.precision = nValueLen;
+                        columnMetadata.put(pair.getKey(), currentColumnMetadata);
+                    }
+                    rowValues[pair.getKey()] = val;
```
peterbae: Same here, cast to byte[]?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDataTable.java
cheenamalhotra: Changes done as per suggestion here: https://github.com/Microsoft/mssql-jdbc/pull/990#discussion_r268369804
Casting is not required as the object type is expected to be of required type. If it is not, it wouldn't get cast eitherway.

peterbae: That comment is addressing a different scenario. In that scenario, we've already confirmed that `val` is a Long instance, so there is no further casting required, and we were unnecessarily unboxing the value. Here, we haven't confirmed the type of `val`, and we need to send the temporal types as String, as mentioned in the comments.

cheenamalhotra: Temporal types are already covered in the `if` block and their String value is determined, `else` case is for handling String type hence it was cast to String which is not necessary.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDataTable.java
```diff
@@ -155,33 +155,43 @@ public synchronized void addRow(Object... values) throws SQLServerException {
...
+                    rowValues[pair.getKey()] = (val instanceof Long) ? val : Long.parseLong(val.toString());
+                    break;
+
+                case BIT:
+                    if (val instanceof Boolean) {
+                        rowValues[pair.getKey()] = val;
+                    } else if (val.toString().equals("0")
+                            || val.toString().equalsIgnoreCase(Boolean.FALSE.toString())) {
+                        rowValues[pair.getKey()] = Boolean.FALSE;
+                    } else if (val.toString().equals("1") || val.toString().equalsIgnoreCase(Boolean.TRUE.toString())) {
```
rene-ye: Create a local String variable for `val.toString()` so we don't end up calling it 4 times.

src/test/java/com/microsoft/sqlserver/jdbc/tvp/TVPTypesTest.java
```diff
@@ -500,6 +503,154 @@ public void testSmallDateTime() throws SQLException {
...
+            }
+        };
+
+        {
+            SQLServerDataTable table = new SQLServerDataTable();
+            table.addColumnMetadata("c1", Types.BIT);
+
+            try {
+                table.addRow(3);
+                fail();
```
ulvii:
```suggestion
                fail(TestResource.getResource("R_expectedFailPassed"));
```

src/test/java/com/microsoft/sqlserver/jdbc/tvp/TVPTypesTest.java
```diff
@@ -500,6 +503,154 @@ public void testSmallDateTime() throws SQLException {
...
+                table.addRow(1.555d);
+                fail();
+            } catch (SQLException e) {
+                assert (null != e);
+                assert (e.getMessage().contains(TestResource.getResource("R_TVPIncorrectColumnValue")));
+            }
+
+            try {
+                table.addRow(Long.parseLong("344563252234"));
+                fail();
```
ulvii:
```suggestion
                fail(TestResource.getResource("R_expectedFailPassed"));
```

src/test/java/com/microsoft/sqlserver/jdbc/tvp/TVPTypesTest.java
```diff
@@ -500,6 +503,154 @@ public void testSmallDateTime() throws SQLException {
...
+                table.addRow(0.66f);
+                fail();
+            } catch (SQLException e) {
+                assert (null != e);
+                assert (e.getMessage().contains(TestResource.getResource("R_TVPIncorrectColumnValue")));
+            }
+
+            try {
+                table.addRow(1.555d);
+                fail();
```
ulvii:
```suggestion
                fail(TestResource.getResource("R_expectedFailPassed"));
```

src/test/java/com/microsoft/sqlserver/jdbc/tvp/TVPTypesTest.java
```diff
@@ -500,6 +503,154 @@ public void testSmallDateTime() throws SQLException {
...
+                table.addRow((short) -1);
+                table.addRow(-1);
+                table.addRow(unknownTypeInstance);
+            } catch (SQLException e) {
+                fail();
+            }
+
+            try {
+                table.addRow(0.66f);
+                fail();
```
ulvii:
```suggestion
                fail(TestResource.getResource("R_expectedFailPassed"));
```

src/test/java/com/microsoft/sqlserver/jdbc/tvp/TVPTypesTest.java
```diff
@@ -500,6 +503,154 @@ public void testSmallDateTime() throws SQLException {
...
+                table.addRow(0.6655f);
+                fail();
+            } catch (SQLException e) {
+                assert (null != e);
+                assert (e.getMessage().contains(TestResource.getResource("R_TVPIncorrectColumnValue")));
+            }
+
+            try {
+                table.addRow(13243.343d);
+                fail();
```
ulvii:
```suggestion
                fail(TestResource.getResource("R_expectedFailPassed"));
```

src/test/java/com/microsoft/sqlserver/jdbc/tvp/TVPTypesTest.java
```diff
@@ -500,6 +503,154 @@ public void testSmallDateTime() throws SQLException {
...
+                table.addRow(-1);
+                fail();
+            } catch (SQLException e) {
+                assert (null != e);
+                assert (e.getMessage().contains(TestResource.getResource("R_TVPIncorrectColumnValue")));
+            }
+
+            try {
+                table.addRow(0.6655f);
+                fail();
```
ulvii:
```suggestion
                fail(TestResource.getResource("R_expectedFailPassed"));
```

src/test/java/com/microsoft/sqlserver/jdbc/tvp/TVPTypesTest.java
```diff
@@ -500,6 +503,154 @@ public void testSmallDateTime() throws SQLException {
...
+                table.addRow(3);
+                fail();
+            } catch (SQLException e) {
+                assert (null != e);
+                assert (e.getMessage().contains(TestResource.getResource("R_TVPIncorrectColumnValue")));
+            }
+
+            try {
+                table.addRow(-1);
+                fail();
```
ulvii:
```suggestion
                fail(TestResource.getResource("R_expectedFailPassed"));
```

src/test/java/com/microsoft/sqlserver/jdbc/tvp/TVPTypesTest.java
```diff
@@ -500,6 +503,154 @@ public void testSmallDateTime() throws SQLException {
...
+
+            try {
+                table.addRow((short) 3);
+                table.addRow((long) 3);
+                table.addRow(3);
+                table.addRow((short) -1);
+                table.addRow(-1);
+                table.addRow(unknownTypeInstance);
+            } catch (SQLException e) {
+                fail();
```
ulvii:
```suggestion
                fail(TestResource.getResource("R_unexpectedException"));
```

src/test/java/com/microsoft/sqlserver/jdbc/tvp/TVPTypesTest.java
```diff
@@ -500,6 +503,154 @@ public void testSmallDateTime() throws SQLException {
...
+        tvp = new SQLServerDataTable();
+
+        tvp.addColumnMetadata("c1", java.sql.Types.BIT);
+        for (int i = 0; i < values.length; i++) {
+            tvp.addRow(values[i]);
+        }
+
+        final String sql = "{call " + AbstractSQLGenerator.escapeIdentifier(procedureName) + "(?)}";
+
+        try (SQLServerCallableStatement P_C_statement = (SQLServerCallableStatement) connection.prepareCall(sql)) {
```
ulvii: Please rename `P_C_statement `

src/test/java/com/microsoft/sqlserver/jdbc/tvp/TVPTypesTest.java
```diff
@@ -500,6 +503,154 @@ public void testSmallDateTime() throws SQLException {
...
+            P_C_statement.setStructured(1, tvpName, tvp);
+            P_C_statement.execute();
+        }
+        try (Statement stmt = connection.createStatement();
+                ResultSet rs = stmt.executeQuery("select * from " + AbstractSQLGenerator.escapeIdentifier(tableName))) {
```
ulvii: The select needs to be ordered by id.

src/test/java/com/microsoft/sqlserver/jdbc/TestResource.java
```diff
@@ -172,6 +172,7 @@ public static String getResource(String key) {
                     "Column name or number of supplied values does not match table definition."},
             {"R_incorrectSyntaxTable", "Incorrect syntax near the keyword 'table'."},
             {"R_incorrectSyntaxTableDW", "Incorrect syntax near 'table'."},
+            {"R_TVPIncorrectColumnValue","Input data is not in correct format"},
```
ulvii: Please use SQLServerResource instead.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDataTable.java
```diff
@@ -207,92 +220,86 @@ private void internalAddrow(JDBCType jdbcType, Object val, Object[] rowValues,
...
+                case CHAR:
+                case VARCHAR:
+                case NCHAR:
+                case NVARCHAR:
+                case LONGVARCHAR:
+                case LONGNVARCHAR:
+                case SQLXML:
+                    if (val instanceof UUID)
+                        val = val.toString();
+                    nValueLen = (2 * ((String) val).length());
```
ulvii: Why are we multiplying length by 2 for `CHAR` type? I know you didn't introduce this, but seems like an issue to me.

cheenamalhotra: I wouldn't modify unless this is reported as issue. This could be by design.

rene-ye: Probably because CHAR and VARCHAR bytes per character is half the size of the others. The length value here is meant to represent the number of characters and not the length of the buffer. I had to do something similar for clob vs nclob.

ulvii: not really, `CHAR`, `VARCHAR` are single-byte types, meaning their length and number of characters are the same.

cheenamalhotra: We'll keep it in backlog, could be investigated later.

Issue #969, opened 2019-02-25
getMoreResults Skips resultsets[BUG]


michaeldenkmal:
## Problem description

1. Expected behaviour:
2. Actual behaviour:
3. Error message/stack trace:
4. Any other details that can be helpful:

michaeldenkmal:
the following test function skips every secound Result set with gradle:testCompile group: 'com.microsoft.sqlserver', name: 'mssql-jdbc', version: '7.2.1.jre8'
but i doest it right with testCompile 'com.microsoft.sqlserver:mssql-jdbc:6.1.0.jre8'

`    public static void testMoreResults(Connection con) {
        try (Statement stmt = con.createStatement();) {
            String SQL = "exec sp_help 'dbo.docs'";

            boolean results = stmt.execute(SQL);
            int rsCount = 0;

            // Loop through the available result sets.
            do {
                if (results) {
                    ResultSet rs = stmt.getResultSet();
                    String firstColumnName = rs.getMetaData().getColumnName(1);
                    rsCount++;

                    // Show data from the result set.
                    System.out.println(String.format("result #%d: firstColumnName:%s", rsCount, firstColumnName));
                    while (rs.next()) {
                        System.out.println(rs.getString(1));
                    }
                }
                System.out.println();
                results = stmt.getMoreResults();
            } while (results);
        }
        // Handle any errors that may have occurred.
        catch (SQLException e) {
            e.printStackTrace();
        }
    }
`

expected result:
` result #1: firstColumnName:Name
DOCS

result #2: firstColumnName:Column_name
ID
DOC_ART
DOKUMENTPFAD
FIRMA1
SUCHBEGRIFF
PNAME
ERSTELLT_AM
LEG_NR
F_NR
P_NR
LU
ERSTELLT_VON
DATEIGROESSE
NOTIZEN
VORLAGE
GZ_ID
EXTENSION
LEGTR_NR
LEGUTR_NR
GR_ID
login_name
BPRIVATE
betreff
ohintid
AD_ID
WrkFlwStat_Id
ABSSB
ABSFirma1
ABS_F_NR
ABS_P_NR
ABS_PNAME
ABS_AD_ID
dFileCreation
dFileLastAccess
dFileLastWrite
wm_id
bm_id
szrevnr
maBm_Id
szDocRecJson
dLastViewed
xmlInfo
ohintRefTabName_C
ft_idxed_b
archived_b
text_inhalt_m

result #3: firstColumnName:Identity
[...]

PR #991, opened 2019-03-15
Fix | Fix for ResultSets being consumed when reading warnings

cheenamalhotra:
Fixes issue #969


cheenamalhotra:
Jars for testing: [mssql-jdbc-PR991.zip](https://github.com/Microsoft/mssql-jdbc/files/2969011/mssql-jdbc-PR991.zip)

biasb:
I have verified that the jre8 version jar works in our test suite.

cheenamalhotra:
Hi @biasb 

I've added the scenario as you mentioned, please let me know if I misunderstood anything.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCallableStatement.java
```diff
@@ -257,7 +257,7 @@ boolean onDone(TDSReader tdsReader) throws SQLServerException {
         }
 
         ExecDoneHandler execDoneHandler = new ExecDoneHandler();
-        TDSParser.parse(resultsReader(), execDoneHandler);
+        TDSParser.parse(resultsReader(), execDoneHandler, false);
```
sehrope: Why explicitly add `false` parameter here? The two arg version of `TDSParser.parse(...)` has been updated to default to false and call the three arg version via `parse(..., ..., false)` so its not necessary. Ditto for the other ten places that have been updated with the new `false` parameter. If it's meant to be explicit everywhere then the old method should be removed and it should be added across the board but there are still other two-arg invocations.

src/main/java/com/microsoft/sqlserver/jdbc/tdsparser.java
```diff
@@ -23,10 +23,11 @@
      * token handler says to stop by returning false from one of the token handling methods.
      */
     static void parse(TDSReader tdsReader, String logContext) throws SQLServerException {
-        parse(tdsReader, new TDSTokenHandler(logContext));
+        parse(tdsReader, new TDSTokenHandler(logContext), false);
     }
 
-    static void parse(TDSReader tdsReader, TDSTokenHandler tdsTokenHandler) throws SQLServerException {
+    static void parse(TDSReader tdsReader, TDSTokenHandler tdsTokenHandler,
+            boolean readOnlyWarningsFlag) throws SQLServerException {
```
sehrope: The javadoc for `parse(...)` should be updated to have the new parameter documented here. Took a bit of skimming through the code to understand what it does.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCallableStatement.java
cheenamalhotra: Hi @sehrope 

The existing 2 param parse call accepts String argument as second param to define `logContext` while the second method accepts `TokenHandler` instance. The new param is added only for specific case where only TDS_MSG tokens are to be read. I overloaded the original `parse` method so the driver calls will not be modified to include third parameter except where its truly desired.

src/main/java/com/microsoft/sqlserver/jdbc/tdsparser.java
cheenamalhotra: Added.

src/test/java/com/microsoft/sqlserver/jdbc/resultset/ResultSetTest.java
```diff
@@ -510,4 +511,89 @@ public void testResultSetMethods() throws SQLException {
             }
         }
     }
+
+    @Test
+    public void testMultipleResultSets() throws SQLException {
+        try (Statement stmt = connection.createStatement()) {
```
ulvii: Please create a new connection for the test.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerStatement.java
```diff
@@ -1385,6 +1385,7 @@ final void clearLastResult() {
      * @return true if another result (ResultSet or update count) was available; false if there were no more results.
      */
     final boolean getNextResult(boolean clearFlag) throws SQLServerException {
+
```
ulvii: Please remove this.

src/test/java/com/microsoft/sqlserver/jdbc/resultset/ResultSetTest.java
```diff
@@ -510,4 +511,89 @@ public void testResultSetMethods() throws SQLException {
...
+
+            boolean results = stmt.execute(SQL);
+            int rsCount = 0;
+
+            // Loop through the available result sets.
+            while (results) {
+                try (ResultSet rs = stmt.getResultSet()) {
+                    rsCount++;
+                    String firstColumnName = rs.getMetaData().getColumnName(1);
+                    switch (rsCount) {
```
ulvii: This `switch` statement seems to be redundant, `firstColumnValue` can be fetched using the column names instead in the second `switch statement`, which will also validate the column names at the same time (if column names do not match, exception will be thrown). 

```
            // Loop through the available result sets.
            while (results) {
                try (ResultSet rs = stmt.getResultSet()) {
                    rsCount++;
                    int i = 1;
                    while (rs.next()) {
                        String firstColumnValue = null;rs.getString(1);
                        switch (rsCount) {
                            case 1:
                                firstColumnValue = rs.getString("Name");
                                assert (firstColumnValue.equals(tableName));
                                break;
                            case 2:
                                firstColumnValue = rs.getString("Column_name");
                                assert (firstColumnValue.equalsIgnoreCase("c" + i++));
                                break;
                            case 3:
                                firstColumnValue = rs.getString("Identity");
                                assert (firstColumnValue.equalsIgnoreCase("c1"));
                                break;
                            case 4:
                                firstColumnValue = rs.getString("RowGuidCol");
                                assert (firstColumnValue.equalsIgnoreCase("No rowguidcol column defined."));
                                break;
                            case 5:
                                firstColumnValue = rs.getString("Data_located_on_filegroup");
                                assert (firstColumnValue.equalsIgnoreCase("PRIMARY"));
                                break;
                        }
                    }
                }
```

src/test/java/com/microsoft/sqlserver/jdbc/resultset/ResultSetTest.java
```diff
@@ -510,4 +511,89 @@ public void testResultSetMethods() throws SQLException {
...
+                                assert (firstColumnValue.equalsIgnoreCase("No rowguidcol column defined."));
+                                break;
+                            case 5:
+                                assert (firstColumnValue.equalsIgnoreCase("PRIMARY"));
+                                break;
+                        }
+                    }
+                }
+
+                SQLWarning warnings = stmt.getWarnings();
```
ulvii: Shouldn't we also validate the warnings?

cheenamalhotra: These warnings are not consistent and I don't think there is a clean way of testing them. Sometimes they're just blank lines. We just care about letting them pass through.

ulvii: Can we at least validate that warnings exist (not `null`)?

cheenamalhotra: Added a validation for their count (hardcoded value is based on what database responds with). We can confirm with this that all the warnings were captured.

PR #996, opened 2019-03-19
Remove Katmai check

peterbae:
Katmai (SQL Server 2008) server check is no longer needed, because our driver no longer supports SQL server versions below 2008.


peterbae:
Let's leave this as is, so that we don't break the driver for people who use older Java versions.

Issue #696, opened 2018-05-09
Cannot perform non-integrated NTLM authentication 


NathanStrong-Tripwire:
## Driver version or jar name
6.4.0.jre8

## SQL Server version
SQL Server 2017

## Client operating system
Windows 10

## Java/JVM version
Oracle Java 8 1.8.0_144

## Table schema
N/A

## Problem description
Our application currently utilizes jTDS to communicate with SQL Server databases. We
would like to switch to the Microsoft JDBC driver, for a few reasons:
- jTDS does not support using an application-provided trust store for TLS connections
- jTDS does not support advanced SQL Server features
- jTDS is not in active development
- jTDS does not support integrated authentication

The problem we're having is that the Microsoft driver doesn't seem to offer one of the
authentication modes that our application relies on heavily: the ability to log into
the database as an arbitrary domain user.

Please see the JavaDoc comments in the reproduction code for details.

And of course, there's always the case that I'm doing something wrong.

## Expected behavior and actual behavior
I've written a really simple test case. See "Repro code" below.

Expected behavior: "PASS"
Actual behavior: "FAIL" with an error that user '' failed to log in.

## Repro code

Please see the Java class for an A/B example of what we currently do with the jTDS
driver that I've been unsuccessful in getting to work with the Microsoft driver.

https://github.com/NathanStrong-Tripwire/mssql-jdbc-usecase

ulvii:
Hi @NathanStrong-Tripwire ,

Currently Microsoft JDBC driver does not support authentication as an arbitrary domain user. We will investigate what needs to be done to implement it.

vaibhavjain3:
Not really sure why Microsoft team is taking it lightly. It is such an important feature and blocked me from upgrading to latest Sql Server driver.

David-Engel:
@vaibhavjain3 It's not taken lightly. The main reason it has not been implemented up to now is that NTLM is known to be an insecure protocol. Old drivers like jTDS implemented it a long time ago when NTLM wasn't considered so insecure. We are looking into implementing it because users are asking for it. But we will probably recommend that it only be used when you are communicating over a known secure network. We are also investigating options to support Windows username and password authentication using Kerberos which I believe may be a viable alternative to NTLM in this use case.

vaibhavjain3:
@David-Engel Sorry If I was bit harsh, But trust me, this is very important feature and hundreds of developers are stuck on it. Including my self.

stolsvik:
[...]

PR #998, opened 2019-03-19
Feature | Introduce NTLM authentication mode

lilgreenbird:
Introduces support for NTLM Authentication in the driver as requested in issue #696.
This feature currently supports authentication with connection string as under:

`jdbc:sqlserver://<hostname>;integratedSecurity=true;authenticationScheme=NTLM;domain=<server_domain>;user=<userName>;password=<password>`

- Supports **Extended Protection enabled SQL Servers**.
- Does not support Extended Protection + Encrypted connection
Implementation of Channel Binding is in progress.


src/main/java/com/microsoft/sqlserver/jdbc/FailOverInfo.java
```diff
@@ -121,13 +121,25 @@ synchronized void failoverAdd(SQLServerConnection connection, boolean actualUseF
...
     private final String serverName;
+    private final String domainName;
     private final int port;
     private final String instanceName;
     private final boolean checkLink;
     private final SQLServerConnectionSecurityManager securityManager;
 
-    ServerPortPlaceHolder(String name, int conPort, String instance, boolean fLink) {
-        serverName = name;
+    ServerPortPlaceHolder(String serverName, int conPort, String instance, boolean fLink) {
```
ulvii: The old constructor can be refactored:

```
    ServerPortPlaceHolder(String serverName, int conPort, String instance, boolean fLink) {
        this(serverName, "", conPort, instance, fLink);
    }
```

src/main/java/com/microsoft/sqlserver/jdbc/NTLMAuthentication.java
```diff
@@ -0,0 +1,644 @@
...
+
+/**
+ * Provides an implementation of NTLM authentication
+ * 
+ * <pre>
+ * See <a href=https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-nlmp/b38c36ed-2804-4868-a9ff-8dd3182128e4">MS-NLMP:NTLM Authentication Protocol</a>
+ * </pre>
+ */
+final class NTLMAuthentication extends SSPIAuthentication {
+    private final static java.util.logging.Logger logger = java.util.logging.Logger
```
ulvii: The logger is not currently used anywhere, I am assuming we will add later?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -1569,6 +1570,23 @@ Connection connectInternal(Properties propsIn,
...
+                if (activeConnectionProperties.getProperty(SQLServerDriverStringProperty.USER.toString()).isEmpty()
+                        || activeConnectionProperties.getProperty(SQLServerDriverStringProperty.PASSWORD.toString())
+                                .isEmpty()) {
+
+                    if (connectionlogger.isLoggable(Level.SEVERE)) {
+                        connectionlogger.severe(
+                                toString() + " " + SQLServerException.getErrString("R_NtlmNoUserPasswordDomain"));
+                    }
+                    throw new SQLServerException(SQLServerException.getErrString("R_NtlmNoUserPasswordDomain"), null);
+
```
ulvii: Please remove the blank line here.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -1569,6 +1570,23 @@ Connection connectInternal(Properties propsIn,
                     ImpersonatedUserCred = (GSSCredential) activeConnectionProperties.get(sPropKey);
                     isUserCreatedCredential = true;
                 }
+            } else if (intAuthScheme == AuthenticationScheme.ntlm) {
```
ulvii: These lines should be removed. 

```
                String sPropKeyDomain = SQLServerDriverStringProperty.DOMAIN_NAME.toString();
                String sPropValueDomain = activeConnectionProperties.getProperty(sPropKeyDomain);
                String domainName = sPropValueDomain;
                // domain and no user or password
```

src/main/java/com/microsoft/sqlserver/jdbc/FailOverInfo.java
lilgreenbird: removed change in this file as discussed

src/main/java/com/microsoft/sqlserver/jdbc/NTLMAuthentication.java
```diff
@@ -0,0 +1,877 @@
...
+
+            this.serverNameBytes = null != serverFqdn ? unicode(serverFqdn.toUpperCase()) : null;
+
+            this.domainName = null != domainName ? domainName.toUpperCase() : "\0";
+            this.domainBytes = unicode(this.domainName);
+
+            this.userName = con.activeConnectionProperties.getProperty(SQLServerDriverStringProperty.USER.toString());
+            this.userNameBytes = unicode(userName);
+
+            this.password = con.activeConnectionProperties
```
ulvii: We should avoid storing the password unsalted like this. I would suggest to calculate MD4 hash of the password as soon as we read it from connection string and discard the actual password.

src/main/java/com/microsoft/sqlserver/jdbc/NTLMAuthentication.java
```diff
@@ -0,0 +1,877 @@
...
+
+            switch (id) {
+                // verify domain name
+                case NTLM_AVID_MSVAVDNSDOMAINNAME:
+                    if (!Arrays.equals(context.domainBytes, new String(value).toUpperCase().getBytes())) {
+                        MessageFormat form = new MessageFormat(SQLServerException.getErrString("R_ntlmBadDomain"));
+                        Object[] msgArgs = {new String(value), new String(context.domainBytes)};
+
+                        // this may fail if we could not get the FQDN so just log a warning
+                        logger.warning(form.format(msgArgs));
```
ulvii: Please add the `if (logger.isLoggable(Level.WARNING))` check before logging a warning.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResource.java
```diff
@@ -30,7 +30,7 @@ static String getResource(String key) {
             {"R_invalidRoutingInfo",
                     "Unexpected routing information received. Please check your connection properties and SQL Server configuration."},
             {"R_multipleRedirections",
-                    "Two or more redirections have occurred. Only one redirection per login attempt is allowed."},
+                    "Two or more redirections have occurred. Only one redirection per logib attempt is allowed."},
```
ulvii: Please remove this change.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResource.java
```diff
@@ -180,6 +180,7 @@ static String getResource(String key) {
             {"R_userPropertyDescription", "The database user."},
             {"R_passwordPropertyDescription", "The database password."},
             {"R_databaseNamePropertyDescription", "The name of the database to connect to."},
+            {"R_domainPropertyDescription", "The name of the domain that contains the user account."},
```
ulvii: `The Windows domain to authenticate in using NTLM.`

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResource.java
```diff
@@ -385,6 +386,8 @@ static String getResource(String key) {
             {"R_UnknownFeatureAck", "Unknown feature acknowledge is received."},
             {"R_SetAuthenticationWhenIntegratedSecurityTrue",
                     "Cannot set \"Authentication\" with \"IntegratedSecurity\" set to \"true\"."},
+            {"R_NtlmNoUserPasswordDomain",
+                    "\"User\" (or \"UserName\") and \"Password\" connection string properties must be specified for \"NTLM Authentication\"."},
```
ulvii:
```suggestion
                    "\"User\" (or \"UserName\") and \"Password\" connection properties must be specified for NTLM Authentication."},
```

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResource.java
```diff
@@ -550,5 +553,16 @@ static String getResource(String key) {
                     "MSI Token failure: Failed to acquire access token from IMDS, unexpected error occurred."},
             {"R_MSITokenFailureEndpoint", "MSI Token failure: Failed to acquire token from MSI Endpoint"},
             {"R_propertyNotSupported",
-                    "Microsoft JDBC Driver for SQL Server currently does not support the property: {0}"}};
+                    "Microsoft JDBC Driver for SQL Server currently does not support the property: {0}"},
```
ulvii: Please make sure all the files are formatted.

lilgreenbird: it is

src/main/java/com/microsoft/sqlserver/jdbc/NTLMAuthentication.java
ulvii: This is still not resolved.

src/main/java/com/microsoft/sqlserver/jdbc/NTLMAuthentication.java
```diff
@@ -0,0 +1,877 @@
...
+                     * Note doc says getCanonicalHostName uses "Best effort method" only, meaning it may not be able to
+                     * return the fqdn
+                     */
+                    serverFqdn = addr.getCanonicalHostName().toUpperCase();
+                }
+            } catch (UnknownHostException e) {
+                MessageFormat form = new MessageFormat(SQLServerException.getErrString("R_ntlmUnknownServer"));
+                Object[] msgArgs = {serverName, e.getMessage()};
+                throw new SQLServerException(form.format(msgArgs), e);
+            }
```
ulvii: There are a few inconsistencies below:

1. Server name, domain name are passed in as parameters, username and password are retrieved from `activeConnectionProperties`.
2. If `serverFqdn` is `null`, we set `serverNameBytes` to `null`, but if `domainName` is null, we set it to `"\0"`.
3. Domain name null check can be done in one line, just like `serverFqdn`.

src/main/java/com/microsoft/sqlserver/jdbc/NTLMAuthentication.java
```diff
@@ -0,0 +1,877 @@
...
+
+            this.serverNameBytes = null != serverFqdn ? unicode(serverFqdn.toUpperCase()) : null;
+
+            this.domainName = null != domainName ? domainName.toUpperCase() : "\0";
```
ulvii: Why are we setting domain name to `"\0"`? Is this meant to be empty string? 
We should handle all these null checks the same way, I prefer what you have done for `serverNameBytes`. `this.serverNameBytes = null != serverFqdn ? unicode(serverFqdn.toUpperCase()) : null;`. I assume you will also need to add a null check to `generateNtlmAuthenticate()`.

src/main/java/com/microsoft/sqlserver/jdbc/NTLMAuthentication.java
```diff
@@ -0,0 +1,877 @@
...
+            int id = targetInfoBuf.getShort();
+            byte[] value = new byte[targetInfoBuf.getShort()];
+            targetInfoBuf.get(value);
+
+            switch (id) {
+                // verify domain name
+                case NTLM_AVID_MSVAVDNSDOMAINNAME:
+                    if (!Arrays.equals(context.domainBytes, new String(value).toUpperCase().getBytes())) {
+                        MessageFormat form = new MessageFormat(SQLServerException.getErrString("R_ntlmBadDomain"));
+                        Object[] msgArgs = {new String(value), new String(context.domainBytes)};
```
ulvii: Both `value` and `context.domainBytes` are UTF16 sequences. You need use `java.nio.charset.StandardCharsets.UTF_16LE` to decode these bytes into a String before logging. 
Example: `new String(value, java.nio.charset.StandardCharsets.UTF_16LE)`

src/main/java/com/microsoft/sqlserver/jdbc/NTLMAuthentication.java
```diff
@@ -0,0 +1,877 @@
...
+     *     Z(4),                  NTLM_CLIENT_CHALLENGE_RESERVED2
+     *     ServerName,            CHALLENGE_MESSAGE.TargetInfo (from 3.1.5.1.2 Client Receives a CHALLENGE_MESSAGE from the Server)
+     *     Z(4))                  EOL
+     * </pre>
+     *
+     * @param clientNonce
+     *        client challenge nonce
+     * @return client challenge blob
+     */
+    private byte[] genClientChallengeBlob(final byte[] clientNonce) {
```
ulvii: Please keep the method names consistent. `generateClientChallengeBlob`

src/main/java/com/microsoft/sqlserver/jdbc/NTLMAuthentication.java
```diff
@@ -0,0 +1,885 @@
...
+     *
+     * Section 2.2.1.2 CHALLENGE_MESSAGE
+     *
+     * @param inToken
+     *        SSPI input blob
+     * @throws SQLServerException
+     */
+    private void parseNtlmChallenge(final byte[] inToken) throws SQLServerException {
+
+        context.token = ByteBuffer.wrap(inToken).order(ByteOrder.LITTLE_ENDIAN);
```
ulvii: Why do we have to copy the `inToken` into a different array?

src/main/java/com/microsoft/sqlserver/jdbc/NTLMAuthentication.java
```diff
@@ -0,0 +1,885 @@
...
+
+        // verify targetInfo - was requested so should always be sent
+        context.targetInfo = new byte[targetInfoLen];
+        context.token.get(context.targetInfo);
+        if (0 == context.targetInfo.length) {
+            throw new SQLServerException(SQLServerException.getErrString("R_ntlmNoTargetInfo"), null);
+        }
+
+        // verify target info av pairs
+        ByteBuffer targetInfoBuf = ByteBuffer.wrap(context.targetInfo).order(ByteOrder.LITTLE_ENDIAN);
```
ulvii: Why do we have to copy the `context.targetInfo` into a different array?

src/main/java/com/microsoft/sqlserver/jdbc/NTLMAuthentication.java
```diff
@@ -0,0 +1,885 @@
...
+            int id = targetInfoBuf.getShort();
+            byte[] value = new byte[targetInfoBuf.getShort()];
+            targetInfoBuf.get(value);
+
+            switch (id) {
+                // verify domain name
+                case NTLM_AVID_MSVAVDNSDOMAINNAME:
+                    if (null != context.domainBytes && logger.isLoggable(Level.WARNING)) {
+                        // verify domain name
+                        if (!Arrays.equals(context.domainBytes, new String(value).toUpperCase().getBytes())) {
```
ulvii: `new String(value)` depends on the platform's default charset and should not be used like this. Looks like we are doing all these conversions to convert the server's response to uppercase. Maybe we should reconsider `NTLMContext` fields?

src/main/java/com/microsoft/sqlserver/jdbc/NTLMAuthentication.java
```diff
@@ -0,0 +1,906 @@
...
+        ByteBuffer targetInfoBuf = ByteBuffer.wrap(context.targetInfo).order(ByteOrder.LITTLE_ENDIAN);
+        boolean done = false;
+        for (int i = 0; i < context.targetInfo.length && !done;) {
+            int id = targetInfoBuf.getShort();
+            byte[] value = new byte[targetInfoBuf.getShort()];
+            targetInfoBuf.get(value);
+
+            switch (id) {
+                case NTLM_AVID_MSVAVDNSDOMAINNAME:
+                    if (null != context.domainName && logger.isLoggable(Level.WARNING)) {
```
ulvii: Please move `logger.isLoggable(Level.WARNING)` to the next `if` statement. 
`if (!Arrays.equals(context.domainBytes, new String(value).toUpperCase().getBytes()) && logger.isLoggable(Level.WARNING))`

src/main/java/com/microsoft/sqlserver/jdbc/NTLMAuthentication.java
```diff
@@ -0,0 +1,906 @@
...
+
+    /**
+     * Generates a key from password to get NTLMv2 hash Section 3.3.2 NTLM v2 Authentication
+     *
+     * @return NT response key
+     * @throws InvalidKeyException
+     */
+    private byte[] ntowfv2() throws InvalidKeyException {
+
+        return hmacMD5(context.passwordHash, (null != context.userName) ? unicode(context.userName + context.domainName)
```
ulvii: Can username be `null` here?

src/main/java/com/microsoft/sqlserver/jdbc/NTLMAuthentication.java
```diff
@@ -0,0 +1,906 @@
...
+     *
+     * Section 2.2.2.1 AV_PAIR
+     *
+     * MsvAvTimestamp is always sent in the CHALLENGE_MESSAGE.
+     *
+     * Section 7 Appendix B: Product Behavior
+     *
+     * MsvAvTimestamp AV_PAIR type is not supported in Windows NT, Windows 2000, Windows XP, and Windows Server 2003.
+     */
+    private static final byte[] NTLM_LMCHALLENAGERESPONSE = {0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
```
ulvii: `NTLM_LMCHALLENAGERESPONSE` is not used anywhere. Only length of it is used in `generateNtlmAuthenticate()`, which can be part of `NTLM_AUTHENTICATE_PAYLOAD_OFFSET`.

src/main/java/com/microsoft/sqlserver/jdbc/NTLMAuthentication.java
```diff
@@ -0,0 +1,906 @@
...
+            // MIC flag
+            newTargetInfo.putShort(NTLM_AVID_MSVAVFLAGS);
+            newTargetInfo.putShort((short) 4);
+            newTargetInfo.putInt((int) NTLM_AVID_VALUE_MIC);
+
+            // EOL
+            newTargetInfo.putShort(NTLM_AVID_MSVAVEOL);
+            newTargetInfo.putShort((short) 0);
+
+            context.token.put(newTargetInfo.array(), 0, newTargetInfo.array().length);
```
ulvii: No need to copy `newTargetInfo` to `context.token` here.

src/main/java/com/microsoft/sqlserver/jdbc/NTLMAuthentication.java
```diff
@@ -0,0 +1,906 @@
...
+            // copy targetInfo up to NTLM_AVID_MSVAVEOL
+            newTargetInfo.put(context.targetInfo, 0, context.targetInfo.length - 4);
+
+            // MIC flag
+            newTargetInfo.putShort(NTLM_AVID_MSVAVFLAGS);
+            newTargetInfo.putShort((short) 4);
+            newTargetInfo.putInt((int) NTLM_AVID_VALUE_MIC);
+
+            // EOL
+            newTargetInfo.putShort(NTLM_AVID_MSVAVEOL);
```
ulvii: I don't think these two lines are required, you are writing zeroes when the array is already initialized to zeroes.

src/main/java/com/microsoft/sqlserver/jdbc/NTLMAuthentication.java
```diff
@@ -0,0 +1,906 @@
...
+            // not used = encrypted random session key fields
+            context.token.putShort((short) 0);
+            context.token.putShort((short) 0);
+            context.token.putInt(offset);
+
+            // same negotiate flags sent in negotiate message
+            context.token.putInt((int) context.negotiateFlags);
+
+            // version - not used but need to send blank to separate from MIC otherwise server confuses this with MIC
+            context.token.put(NTLMSSP_VERSION, 0, NTLMSSP_VERSION.length);
```
ulvii: This line is not required. The array is already initialized to zeroes, you just need to increment the index by 8. 

`context.token.position(context.token.position() + 8);`

src/main/java/com/microsoft/sqlserver/jdbc/NTLMAuthentication.java
```diff
@@ -0,0 +1,906 @@
...
+     */
+    private static final byte[] NTLM_LMCHALLENAGERESPONSE = {0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+            0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00};
+
+    /**
+     * Section 2.2.2.10 VERSION
+     *
+     * used for debugging purposes only and its value does not affect NTLM message processing
+     */
+    private static final byte[] NTLMSSP_VERSION = new byte[] {0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00};
```
ulvii: This can be removed too.

src/main/java/com/microsoft/sqlserver/jdbc/NTLMAuthentication.java
lilgreenbird: you're right it can't because it should be set to default value in SQLServerConnection. But, good style to null check anyway?

src/main/java/com/microsoft/sqlserver/jdbc/NTLMAuthentication.java
lilgreenbird: it's used in authenticate message. It's nicer to separate so it's clear what it is instead of just lumping it into the payload?

src/main/java/com/microsoft/sqlserver/jdbc/NTLMAuthentication.java
lilgreenbird: it makes it more clear, you just happen to know EOL is 0 but it may not be :)

src/main/java/com/microsoft/sqlserver/jdbc/NTLMAuthentication.java
lilgreenbird: you're right, but is that any faster than putting 0's? I thought it makes it more clear..and in case we ever want to send version

src/main/java/com/microsoft/sqlserver/jdbc/NTLMAuthentication.java
```diff
@@ -0,0 +1,906 @@
...
+        // workstation field
+        context.token.putShort((short) workstationLen);
+        context.token.putShort((short) workstationLen);
+        context.token.putInt(offset);
+        offset += workstationLen;
+
+        // version not used - for debug only
+
+        // payload
+        context.token.put(context.domainBytes, 0, domainNameLen);
```
ulvii: This line throws `NullPointerException` if I do `dataSource.setDomain(null);`

src/main/java/com/microsoft/sqlserver/jdbc/NTLMAuthentication.java
lilgreenbird: actually it should be 1st

src/main/java/com/microsoft/sqlserver/jdbc/NTLMAuthentication.java
lilgreenbird: see above

src/test/java/com/microsoft/sqlserver/jdbc/NTLMConnectionTest.java
```diff
@@ -0,0 +1,689 @@
...
+    }
+
+    @Test
+    /**
+     * Test Bad NTLM Initialization
+     * 
+     * @throws SQLException
+     */
+    public void testNTLMBadInit() throws SQLException {
+        try {
```
ulvii:
```suggestion
        try {
        @SuppressWarnings("unused")
```

src/test/java/com/microsoft/sqlserver/jdbc/NTLMConnectionTest.java
```diff
@@ -0,0 +1,689 @@
...
+     */
+    private void verifyDatabase(Connection con, String databaseName) throws SQLException {
+        try (Statement stmt = con.createStatement(); ResultSet rs = stmt.executeQuery("SELECT DB_NAME()")) {
+            while (rs.next()) {
+                assertEquals(databaseName, rs.getString(1));
+            }
+        }
+    }
+
+    // TODO Remove when this method is merged to dev in PR #968
```
ulvii: Please remove this and make use of `AbstractTest.updateDataSource()`.

src/test/java/com/microsoft/sqlserver/jdbc/NTLMConnectionTest.java
```diff
@@ -0,0 +1,689 @@
...
+    }
+
+    /**
+     * Tests basic NTLM authentication.
+     * 
+     * @throws SQLException
+     */
+    @Test
+    public void testNTLMBasicConnection() throws SQLException {
+        if (ntlmPropsDefined) {
```
ulvii: If `ntlmPropsDefined` is not `true`, this test will be skipped, but in Maven results it will be marked as passed. It would be better if we find a way of reporting these tests as skipped (perhaps with a tag?) when NTLM is not supported.

src/test/java/com/microsoft/sqlserver/jdbc/NTLMConnectionTest.java
```diff
@@ -0,0 +1,689 @@
...
+            assertTrue(e.getMessage().matches(TestUtils.formatErrorMsg(resourceKeyword)));
+        }
+    }
+
+    /*
+     * Verify NTLM authentication
+     */
+    private void verifyNTLM(Connection con) throws SQLException {
+        try (Statement stmt = con.createStatement(); ResultSet rs = stmt
+                .executeQuery("select auth_scheme from sys.dm_exec_connections where session_id=@@spid")) {
```
ulvii: Please note that this query requires `VIEW SERVER STATE` permission. We could grant it to NTLM user using the original `connectionString` in `setUp()` **assuming** original user has `GRANT` permission.

src/test/java/com/microsoft/sqlserver/jdbc/NTLMConnectionTest.java
```diff
@@ -0,0 +1,689 @@
...
+        }
+    }
+
+    /**
+     * Tests NTLM authentication against a non-default database.
+     * 
+     * @throws SQLException
+     */
+    @Test
+    public void testNTLMNonDeafultDatabase() throws SQLException {
```
ulvii: I changed the test a bit to not use a system database and NTLM connection seems to fail.

```
    @Test
    public void testNTLMNonDeafultDatabase() throws SQLException {
        if (ntlmPropsDefined) {
            String databaseName = RandomUtil.getIdentifier("NTLM");
                stmt.executeUpdate("CREATE DATABASE " + AbstractSQLGenerator.escapeIdentifier(databaseName));
            }
            try (Connection con = DriverManager
                    .getConnection(addOrOverrideProperty(connectionStringNTLM, "database", databaseName))) {
                verifyNTLM(con);
                verifyDatabase(con, databaseName);
            }
                TestUtils.dropDatabaseIfExists(databaseName, stmt);
            }
        }
    }
```

ulvii: The test is fine and was failing because of a permission issue on my server. Please update `testNTLMNonDeafultDatabase()` test to not use `tempdb`.

src/main/java/com/microsoft/sqlserver/jdbc/NTLMAuthentication.java
```diff
@@ -0,0 +1,905 @@
...
+     *
+     * @param con
+     * @param serverName
+     * @param domainName
+     * @param userName
+     * @param password
+     * @param workstation
+     * @throws SQLServerException
+     */
+    NTLMAuthentication(final SQLServerConnection con, final String serverName, final String domainName,
```
ulvii: Why are we passing in the connection to this method?

src/test/java/com/microsoft/sqlserver/jdbc/NTLMConnectionTest.java
```diff
@@ -0,0 +1,689 @@
...
+     * @throws SQLException
+     */
+    public void testNTLMipAddr() throws SQLException {
+        if (ntlmPropsDefined) {
+            String ipAddr;
+            try {
+                ipAddr = InetAddress.getByName(serverFqdn).getHostAddress();
+                try (Connection con = DriverManager.getConnection(connectionStringNTLM + ";servername=" + ipAddr)) {
+                    verifyNTLM(con);
+
```
ulvii: Blank line.

src/test/java/com/microsoft/sqlserver/jdbc/NTLMConnectionTest.java
```diff
@@ -0,0 +1,689 @@
...
+            assertTrue(loggerContent.toString().matches(TestUtils.formatErrorMsg("R_ntlmNoTimestamp")));
+        }
+    }
+
+    /*
+     * Get Server FQDN from connection
+     */
+    private void getServerFqdn(SQLServerConnection con) throws SQLException {
+        try {
+            serverFqdn = InetAddress.getByName(con.currentConnectPlaceHolder.getServerName()).getCanonicalHostName()
```
ulvii: Why do we have to do this? Can't we extract server name from the connection string and just get the `FQDN` of it?

src/test/java/com/microsoft/sqlserver/jdbc/NTLMConnectionTest.java
```diff
@@ -0,0 +1,689 @@
...
+    // offsets for targetinfo av pairs
+    private final int NTLM_CHALLENGE_MSVAVDNSDOMAINNAME_OFFSET = NTLM_CHALLENGE_TARGETINFO_OFFSET + 54;
+    private final int NTLM_CHALLENGE_MSVAVDNSCOMPUTERNAME_OFFSET = NTLM_CHALLENGE_TARGETINFO_OFFSET + 78;
+    private final int NTLM_CHALLENGE_MSVAVTIMESTAMP_OFFSET = -16;
+
+    @Test
+    public void testNTLMBadSignature() throws SQLException {
+        if (ntlmPropsDefined) {
+            try (SQLServerConnection con = (SQLServerConnection) DriverManager.getConnection(connectionStringNTLM)) {
+                getServerFqdn(con);
```
ulvii: There are lots of duplicated code in `testNTLMBad*`, please move them into a common method.

src/test/java/com/microsoft/sqlserver/jdbc/NTLMConnectionTest.java
```diff
@@ -0,0 +1,689 @@
...
+                try {
+                    // modify token with bad computer name
+                    byte[] badComputer = {0, 0, 0, 0, 0, 0, 0, 0, 0};
+                    byte[] badToken = getChallengeToken(NTLM_CHALLENGE_MSVAVDNSCOMPUTERNAME_OFFSET, badComputer);
+
+                    auth.generateClientContext(badToken, done);
+
+                } catch (Exception e) {
+                    // this should just generate a warning but not fail
+                    fail(e.getMessage());
```
ulvii:
```suggestion
                    fail(TestResource.getResource("R_unexpectedException"));
```

src/test/java/com/microsoft/sqlserver/jdbc/NTLMConnectionTest.java
```diff
@@ -0,0 +1,689 @@
...
+                try {
+
+                    byte[] badTimestamp = {0, 0, 0, 0, 0, 0, 0, 0, 0, 0};
+                    byte[] badToken = getChallengeToken(NTLM_CHALLENGE_MSVAVTIMESTAMP_OFFSET, badTimestamp);
+
+                    auth.generateClientContext(badToken, done);
+
+                } catch (Exception e) {
+                    // this should just generate a warning but not fail
+                    fail(e.getMessage());
```
ulvii:
```suggestion
                    fail(TestResource.getResource("R_unexpectedException"));
```

src/test/java/com/microsoft/sqlserver/jdbc/NTLMConnectionTest.java
```diff
@@ -0,0 +1,689 @@
...
+
+        return token.array();
+    }
+
+    /*
+     * Test invalid properties
+     */
+    private void testInvalidProperties(String connectionString, String resourceKeyword) {
+        try (Connection con = DriverManager.getConnection(connectionString)) {
+            fail();
```
ulvii:
```suggestion
            fail(TestResource.getResource("R_expectedFailPassed"));
```

src/main/java/com/microsoft/sqlserver/jdbc/NTLMAuthentication.java
ulvii: Still not resolved. Applies to all occurrences of `new String(value)`.

src/main/java/com/microsoft/sqlserver/jdbc/NTLMAuthentication.java
```diff
@@ -0,0 +1,905 @@
...
+    }
+
+    /**
+     * Generates a key from password to get NTLMv2 hash Section 3.3.2 NTLM v2 Authentication
+     *
+     * @return NT response key
+     * @throws InvalidKeyException
+     */
+    private byte[] ntowfv2() throws InvalidKeyException {
+
```
ulvii: Please format all the files and remove unnecessary blank lines.

src/test/java/com/microsoft/sqlserver/jdbc/NTLMConnectionTest.java
lilgreenbird: same thing, but ok

src/test/java/com/microsoft/sqlserver/jdbc/NTLMConnectionTest.java
```diff
@@ -165,12 +174,18 @@ public void testNTLMEncryptedConnection() throws SQLException {
     @Test
     public void testNTLMNonDeafultDatabase() throws SQLException {
         if (ntlmPropsDefined) {
```
cheenamalhotra: Need not check this now in all tests. If this is 'false' the test class should fail in `@BeforeAll > setup()` method.

azure-pipelines.yml
```diff
@@ -33,7 +33,7 @@ jobs:
     displayName: 'Maven build 43'
     inputs:
       mavenPomFile: 'pom.xml'
-      goals: 'clean -Dmssql_jdbc_test_connection_properties=jdbc:sqlserver://$(Target_SQL)$(server_domain);$(database);$(user);$(password); install -Pbuild43 -DexcludeGroups=$(Ex_Groups)'
+      goals: 'clean -Dmssql_jdbc_test_connection_properties=jdbc:sqlserver://$(Target_SQL)$(server_domain);$(database);$(user);$(password); install -Pbuild43 -DuserNTLM=$(userNTLM) -DpasswordNTLM=$(passwordNTLM) -Ddomain=$(domainNTLM) -DexcludeGroups=$(Ex_Groups)'
```
cheenamalhotra: should also provide `-DincludeGroups=NTLM`

src/main/java/com/microsoft/sqlserver/jdbc/ISQLServerDataSource.java
```diff
@@ -820,4 +820,18 @@
      * @return msiClientId property value
      */
     public String getMSIClientId();
+
+    /**
+     * Sets the Windows domain to authenticate in using NTLM.
```
cheenamalhotra:
```suggestion
     * Sets the Windows domain name to authenticate with using NTLM.
```

src/main/java/com/microsoft/sqlserver/jdbc/NTLMAuthentication.java
```diff
@@ -0,0 +1,905 @@
...
+     * @param responseKeyNT
+     *        NT response hash key
+     * @return computed response
+     * @throws InvalidKeyException
+     * @throws NoSuchAlgorithmException
+     */
+    private byte[] computeResponse(final byte[] responseKeyNT) throws InvalidKeyException, NoSuchAlgorithmException {
+        // get random client challenge nonce
+        byte[] clientNonce = new byte[NTLM_CLIENT_NONCE_LENGTH];
+        (new Random()).nextBytes(clientNonce);
```
rene-ye: Do we want to change this to `ThreadLocalRandom.current().nextBytes(clientNonce);`?

>Normally to generate Random numbers, we either do Create an instance of java.util.Random OR Math.random() - which internally creates an instance of java.util.Random on first invocation. However in a concurrent applications usage of above leads to contention issues.
Random is thread safe for use by multiple threads. But if multiple threads use the same instance of Random, the same seed is shared by multiple threads. It leads to contention between multiple threads and so to performance degradation.
ThreadLocalRandom is solution to above problem. ThreadLocalRandom has a Random instance per thread and safeguards against contention.

src/main/java/com/microsoft/sqlserver/jdbc/NTLMAuthentication.java
```diff
@@ -0,0 +1,905 @@
...
+import javax.crypto.spec.SecretKeySpec;
+
+import mssql.security.provider.MD4;
+
+
+/**
+ * Provides an implementation of NTLM authentication
+ * 
+ * @see <a
+ *      href=https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-nlmp/b38c36ed-2804-4868-a9ff-8dd3182128e4">MS-NLMP:NTLM
```
cheenamalhotra: Can we add content that we need here instead of MS Docs URL. URLs keep on changing/removing, and we may not find information we need later.

src/main/java/com/microsoft/sqlserver/jdbc/NTLMAuthentication.java
```diff
@@ -0,0 +1,905 @@
...
+    /**
+     * Generates the Type 3 NTLM Authentication message
+     *
+     * Section 2.2.1.3 AUTHENTICATE_MESSAGE
+     *
+     * @return NTLM Authenticate message
+     * @throws SQLServerException
+     */
+    private byte[] generateNtlmAuthenticate() throws SQLServerException {
+        int domainNameLen = null != context.domainBytes ? context.domainBytes.length : 0;
```
rene-ye: Might be worth making a helper function instead? Seems like this comparison logic is used a lot for these byte buffers.

```java
private int getByteArrayLength(byte[] b) {
    return null == b ? 0 : b.length();
}

int domainNameLen = getByteArrayLength(context.DomainBytes);
```

src/main/java/com/microsoft/sqlserver/jdbc/NTLMAuthentication.java
```diff
@@ -0,0 +1,905 @@
...
+        int userNameLen = null != context.userNameBytes ? context.userNameBytes.length : 0;
+        int workstationLen = null != context.workstationBytes ? context.workstationBytes.length : 0;
+        byte[] msg = null;
+
+        try {
+            // get NT challenge response
+            byte[] ntChallengeResp = getNtChallengeResp();
+            int ntChallengeLen = ntChallengeResp.length;
```
rene-ye: Can `getNtChallengeResp()` return null?

src/main/java/mssql/security/provider/MD4.java
```diff
@@ -0,0 +1,279 @@
...
+ */
+
+package mssql.security.provider;
+
+// import static sun.security.provider.ByteArrayAccess.*;
+// code updated to remove dependency
+
+/**
+ * The MD4 class is used to compute an MD4 message digest over a given buffer of bytes. It is an implementation of the
+ * RSA Data Security Inc MD4 algorithim as described in internet RFC 1320.
```
cheenamalhotra: algorithim > algorithm
internet > Internet

src/main/java/mssql/security/provider/MD4.java
```diff
@@ -0,0 +1,279 @@
...
+ * for more details (a copy is included in the LICENSE file that accompanied this code). You should have received a copy
+ * of the GNU General Public License version 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA. Please contact Oracle, 500 Oracle Parkway, Redwood
+ * Shores, CA 94065 USA or visit www.oracle.com if you need additional information or have any questions.
+ */
+
+package mssql.security.provider;
+
+// import static sun.security.provider.ByteArrayAccess.*;
+// code updated to remove dependency
```
cheenamalhotra: Please clean the class

src/main/java/mssql/security/provider/MD4.java
```diff
@@ -0,0 +1,279 @@
...
+     */
+    private void implDigest(byte[] out, int ofs) {
+        long bitsProcessed = bytesProcessed << 3;
+
+        int index = (int) bytesProcessed & 0x3f;
+        int padLen = (index < 56) ? (56 - index) : (120 - index);
+        engineUpdate(padding, 0, padLen);
+
+        // i2bLittle4((int)bitsProcessed, buffer, 56);
+        // i2bLittle4((int)(bitsProcessed >>> 32), buffer, 60);
```
cheenamalhotra: Is this safe to be uncommented?

src/test/java/com/microsoft/sqlserver/jdbc/NTLMConnectionTest.java
```diff
@@ -0,0 +1,536 @@
...
+    @BeforeAll
+    public static void setUp() throws Exception {
+        // remove all other logging and add our own handler
+        LogManager.getLogManager().reset();
+        ntlmLogger.setLevel(Level.FINER);
+        ntlmLogger.addHandler(handler);
+        handler.setLevel(Level.FINER);
+
+        // if these properties are defined then NTLM is desired, modify connection string accordingly
+        String domain = System.getProperty("domain");
```
cheenamalhotra: Please move the code to AbstractTest for creating connection String for NTLM.
All properties must be read at same location. They may/may not be available.

src/test/java/com/microsoft/sqlserver/jdbc/NTLMConnectionTest.java
```diff
@@ -0,0 +1,536 @@
...
+    private void getServerFqdn(SQLServerConnection con) throws SQLException {
+        try {
+            serverFqdn = InetAddress
+                    .getByName(con.activeConnectionProperties
+                            .getProperty(SQLServerDriverStringProperty.SERVER_NAME.toString()))
+                    .getCanonicalHostName().toUpperCase();
+
+        } catch (UnknownHostException e) {
+            fail("Error getting server FQDN: " + e.getMessage());
+            e.printStackTrace();
```
cheenamalhotra: No printing stacktraces

src/test/java/com/microsoft/sqlserver/jdbc/NTLMConnectionTest.java
```diff
@@ -0,0 +1,536 @@
...
+     * 
+     * @param connectionString
+     *        original connection string
+     * @param property
+     *        name of the property
+     * @param value
+     *        value of the property
+     * @return The updated connection string
+     */
+    private static String addOrOverrideProperty(String connectionString, String property, String value) {
```
cheenamalhotra: Method should be moved to TestUtils

src/test/java/com/microsoft/sqlserver/jdbc/NTLMConnectionTest.java
```diff
@@ -0,0 +1,536 @@
...
+            fail("Error getting server FQDN: " + e.getMessage());
+            e.printStackTrace();
+        }
+    }
+
+    /**
+     * Send a bad NTLM Challenge Token
+     */
+    private void sendBadToken(byte[] badField, int offset) throws Exception {
+        try (SQLServerConnection con = (SQLServerConnection) DriverManager.getConnection(connectionStringNTLM)) {
```
cheenamalhotra: Use PrepUtils as other tests

src/test/java/com/microsoft/sqlserver/jdbc/NTLMConnectionTest.java
```diff
@@ -0,0 +1,536 @@
...
+    /**
+     * Test NTLM connection with IP address
+     * 
+     * @throws SQLException
+     */
+    public void testNTLMipAddr() throws SQLException {
+        String ipAddr;
+        try {
+            ipAddr = InetAddress.getByName(serverFqdn).getHostAddress();
+            try (Connection con = DriverManager.getConnection(connectionStringNTLM + ";servername=" + ipAddr)) {
```
cheenamalhotra: Please use PrepUtils, applies everywhere.

src/test/java/com/microsoft/sqlserver/jdbc/NTLMConnectionTest.java
```diff
@@ -0,0 +1,536 @@
...
+        public String format(LogRecord record) {
+            StringBuffer sb = new StringBuffer();
+            sb.append(record.getMessage());
+            return sb.toString();
+        }
+    }
+
+    private static SQLServerDataSource dsNTLMLocal = null;
+
+    // updated connection strong with NTLM properties
```
cheenamalhotra: strong > string

src/test/java/com/microsoft/sqlserver/jdbc/NTLMConnectionTest.java
```diff
@@ -0,0 +1,536 @@
...
+        try {
+            byte[] badSignature = {0, 0, 0, 0, 0, 0, 0, 0};
+            sendBadToken(badSignature, NTLM_CHALLENGE_SIGNATURE_OFFSET);
+        } catch (Exception e) {
+            assertTrue(e.getMessage().matches(TestUtils.formatErrorMsg("R_ntlmSignatureError")));
+        }
+    }
+
+    @Test
+    public void testNTLMBadMessageType() throws SQLException {
```
cheenamalhotra: Tests here catch Exception and throw SQLException (which is not possible), either of the two should be removed.

src/test/java/com/microsoft/sqlserver/jdbc/NTLMConnectionTest.java
```diff
@@ -0,0 +1,536 @@
...
+
+        timerStart = System.currentTimeMillis();
+        try (Connection con = DriverManager
+                .getConnection(connectionStringNTLM + ";servername=bogus;instanceName=;loginTimeout=" + loginTimeout)) {
+            fail(TestResource.getResource("R_expectedFailPassed"));
+        } catch (SQLException e) {
+            timerEnd = System.currentTimeMillis();
+        }
+        long elapsedTime = (timerEnd - timerStart) / 1000;
+        assertTrue(elapsedTime < loginTimeout + 1);
```
cheenamalhotra: loginTimeout may not happen consistently (it would max +7 seconds) but not beyond that (Still you could allow 10 seconds relay). That's why I see the test is disabled. You may pass loginTimeout as 2 seconds and assert:

```suggestion
        assertTrue(elapsedTime < loginTimeout + 10);
```
Default timeout is 15 seconds in which case this would fail if applied timeout doesn't work.

src/main/java/com/microsoft/sqlserver/jdbc/NTLMAuthentication.java
```diff
@@ -0,0 +1,905 @@
...
+         * Section 2.2.2.1 AV_PAIR
+         *
+         * Section 7 Appendix B Product Behavior
+         * 
+         * This structure should always be sent in the CHALLENGE_MESSAGE not supported in Windows NT, Windows 2000,
+         * Windows XP, and Windows Server 2003
+         */
+        if (null == context.timestamp || 0 >= context.timestamp.length) {
+            // this SHOULD always be present but for some reason occasionally this had seen to be missing
+            logger.finer(SQLServerException.getErrString("R_ntlmNoTimestamp"));
```
cheenamalhotra: Check for isLoggable FINER?

src/main/java/com/microsoft/sqlserver/jdbc/NTLMAuthentication.java
lilgreenbird: all the relative content has been added, this is added here in case you want more :)

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -6166,9 +6206,17 @@ public void onEviction(CityHash128Key key, PreparedStatementHandle handle) {
     }
 
     /**
+     * <<<<<<< HEAD
```
lilgreenbird: need to fix comment

cheenamalhotra: Conflicts came in comments..

pom.xml
```diff
@@ -47,10 +47,10 @@
 			xAzureSQLDB - - - - For tests not compatible with Azure SQL Database - - 
 			xAzureSQLDW - - - - For tests not compatible with Azure Data Warehouse - 
 			xAzureSQLMI - - - - For tests not compatible with Azure SQL Managed Instance 
+			NTLM - - - - - - - For not running NTLM tests (excluded by default) - - -
```
ulvii: `xNTLM`?

src/main/java/com/microsoft/sqlserver/jdbc/NTLMAuthentication.java
```diff
@@ -0,0 +1,846 @@
...
+         * @throws SQLServerException
+         */
+        NTLMContext(final String domainName, final String userName, final String password,
+                final String workstation) throws SQLServerException {
+
+            this.domainName = null != domainName ? domainName.toUpperCase()
+                                                 : SQLServerDriverStringProperty.DOMAIN.getDefaultValue();
+            this.domainBytes = unicode(this.domainName);
+
+            this.userNameBytes = null != SQLServerDriverStringProperty.USER.getDefaultValue() ? unicode(userName)
```
ulvii: I don't get this line. Why are we checking against `SQLServerDriverStringProperty.USER.getDefaultValue()`?

src/main/java/com/microsoft/sqlserver/jdbc/NTLMAuthentication.java
```diff
@@ -0,0 +1,846 @@
...
+    private void parseNtlmChallenge(final byte[] inToken) throws SQLServerException {
+        // get token buffer
+        ByteBuffer token = ByteBuffer.wrap(inToken).order(ByteOrder.LITTLE_ENDIAN);
+
+        // verify signature
+        byte[] signature = new byte[NTLM_HEADER_SIGNATURE.length];
+        token.get(signature);
+        if (!Arrays.equals(signature, NTLM_HEADER_SIGNATURE)) {
+            MessageFormat form = new MessageFormat(SQLServerException.getErrString("R_ntlmSignatureError"));
+            Object[] msgArgs = {new String(signature, java.nio.charset.StandardCharsets.UTF_16LE)};
```
ulvii: Does it make sense to convert NTLM signature to String? I think it should rather be a hex representation.

src/main/java/com/microsoft/sqlserver/jdbc/NTLMAuthentication.java
```diff
@@ -0,0 +1,846 @@
...
+                case NTLM_AVID_MSVAVFLAGS:
+                case NTLM_AVID_MSVAVNBCOMPUTERNAME:
+                case NTLM_AVID_MSVAVNBDOMAINNAME:
+                case NTLM_AVID_MSVAVDNSTREENAME:
+                case NTLM_AVID_MSVAVSINGLEHOST:
+                case NTLM_AVID_MSVAVTARGETNAME:
+                    break;
+                default:
+                    MessageFormat form = new MessageFormat(SQLServerException.getErrString("R_ntlmUnknownValue"));
+                    Object[] msgArgs = {new String(value, java.nio.charset.StandardCharsets.UTF_16LE)};
```
ulvii: Again, converting to String doesn't make much sense here. Just print out the token.

src/main/java/com/microsoft/sqlserver/jdbc/NTLMAuthentication.java
```diff
@@ -0,0 +1,846 @@
...
+                    throw new SQLServerException(form.format(msgArgs), null);
+            }
+
+            if (logger.isLoggable(Level.FINER)) {
+                MessageFormat form = new MessageFormat(SQLServerException.getErrString("R_ntlmAvp"));
+                Object[] msgArgs = {id, new String(value, java.nio.charset.StandardCharsets.UTF_16LE)};
```
ulvii: Same as above. Also, I think this should be logged when `FINEST` is enabled.

src/main/java/com/microsoft/sqlserver/jdbc/NTLMAuthentication.java
```diff
@@ -0,0 +1,846 @@
...
+                                                 : SQLServerDriverStringProperty.DOMAIN.getDefaultValue();
+            this.domainBytes = unicode(this.domainName);
+
+            this.userNameBytes = null != SQLServerDriverStringProperty.USER.getDefaultValue() ? unicode(userName)
+                                                                                              : null;
+            this.upperUserName = userName.toUpperCase();
+
+            this.passwordHash = null != password ? MD4(unicode(password)) : null;
+
+            this.workstationBytes = null != workstation ? unicode(workstation.toUpperCase()) : null;
```
ulvii: NTLM specs 2.2.1.1

WorkstationName (variable): If WorkstationLen does not equal 0x0000, WorkstationName MUST be a byte array that contains the name of the client machine that MUST be encoded using the OEM character set. Otherwise, this data is not present.

ulvii: Please also note that when server returns workstation name in Authenticate message, it is in negotiated character set, in our case unicode. 

NTLM specs 2.2.1.3

Workstation (variable): The name of the computer to which the user is logged on. Workstation MUST be encoded in the negotiated character set.

pom.xml
cheenamalhotra: x Notation is for "Not compatible with XXX" tests..
NTLM is fine. Since the tag is for tests that do NTLM Auth.

ulvii: Description should be updated to something like "To enable NTLM tests".

cheenamalhotra: Yes.. Description should be fixed to: "For tests using NTLM Authentication mode"
Enabled/Not Enabled (verb) depends on if one puts this tag in "excludedGroups".

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -3621,16 +3644,24 @@ final boolean doExecute() throws SQLServerException {
...
+                    authentication = new KerbAuthentication(this, currentConnectPlaceHolder.getServerName(),
+                            currentConnectPlaceHolder.getPortNumber(), ImpersonatedUserCred, isUserCreatedCredential);
+                } else
+                    authentication = new KerbAuthentication(this, currentConnectPlaceHolder.getServerName(),
+                            currentConnectPlaceHolder.getPortNumber());
+            } else if (ntlmAuthentication) {
+                authentication = new NTLMAuthentication(this, domainName,
+                        activeConnectionProperties.getProperty(SQLServerDriverStringProperty.USER.toString()),
+                        activeConnectionProperties.getProperty(SQLServerDriverStringProperty.PASSWORD.toString()),
+                        hostName);
```
ulvii: Please add `activeConnectionProperties.remove(SQLServerDriverStringProperty.PASSWORD.toString());` to also remove the password from connection properties.

src/main/java/com/microsoft/sqlserver/jdbc/NTLMAuthentication.java
```diff
@@ -0,0 +1,889 @@
...
+
+    /**
+     * Gets the unicode of a string
+     *
+     * @param str
+     *        string to convert to unicode
+     * @return unicode of string
+     */
+    private byte[] unicode(final String str) {
+        return str.getBytes(java.nio.charset.StandardCharsets.UTF_16LE);
```
peterbae: add null check here, i think it's possible for str to be null

src/test/java/com/microsoft/sqlserver/jdbc/TestUtils.java
```diff
@@ -76,7 +76,7 @@
     /**
      * Checks if connection is established to Azure server.
      * 
-     * @see com.microsoft.sqlserver.jdbc.SQLServerConnection#isAzure()
+     * @see com.microsoft.sqlserver.jdbc.SQLServerConnection#isAzure() >>>>>>> upstream/dev
```
peterbae: remove this piece

src/test/java/com/microsoft/sqlserver/jdbc/NTLMConnectionTest.java
```diff
@@ -0,0 +1,413 @@
...
+    @BeforeAll
+    public static void setUp() throws Exception {
+        // reset logging to avoid servere logs due to negative testing
+        LogManager.getLogManager().reset();
+
+        dsNTLMLocal = new SQLServerDataSource();
+        AbstractTest.updateDataSource(connectionString, dsNTLMLocal);
+
+        // grant view server state permission - needed to verify NTLM
+        try (Connection con = PrepUtil
```
ulvii: At this point, you have already executed `connectionString = connectionStringNTLM`, line 43. To be able to grant a permission, the user must have GRANT option, which NTLM user does not necessarily have. Please make sure you execute this statement with `sa` user (original connection string).

src/test/java/com/microsoft/sqlserver/testframework/AbstractTest.java
```diff
@@ -84,10 +85,34 @@ public static void setup() throws Exception {
         applicationKey = getConfiguredProperty("applicationKey");
         keyIDs = getConfiguredProperty("keyID", "").split(Constants.SEMI_COLON);
         connectionString = getConfiguredProperty(Constants.MSSQL_JDBC_TEST_CONNECTION_PROPERTIES);
+        connectionStringNTLM = connectionString;
 
-        ds = updateDataSource(new SQLServerDataSource());
-        dsXA = updateDataSource(new SQLServerXADataSource());
-        dsPool = updateDataSource(new SQLServerConnectionPoolDataSource());
+        // if these properties are defined then NTLM is desired, modify connection string accordingly
+        String domain = System.getProperty("domain");
```
ulvii: Please rename these variables, to `domainNTLM`, `userNTLM`, `passwordNTLM`.

src/test/java/com/microsoft/sqlserver/testframework/AbstractTest.java
```diff
@@ -84,10 +85,34 @@ public static void setup() throws Exception {
...
+
+        if (null != user) {
+            connectionStringNTLM = TestUtils.addOrOverrideProperty(connectionStringNTLM, "user", user);
+        }
+
+        if (null != password) {
+            connectionStringNTLM = TestUtils.addOrOverrideProperty(connectionStringNTLM, "password", password);
+        }
+
+        if (null != domain && null != user && null != password) {
```
ulvii: `domain` property is optional, please remove `null != domain` from the `if` statement.

src/main/java/com/microsoft/sqlserver/jdbc/NTLMAuthentication.java
```diff
@@ -0,0 +1,910 @@
...
+     *        password
+     * @param workstation
+     *        hostname of the workstation
+     * @throws SQLServerException
+     *         if error occurs
+     */
+    NTLMAuthentication(final SQLServerConnection con, final String domainName, final String userName,
+            final String password, final String workstation) throws SQLServerException {
+        if (null == context) {
+            this.context = new NTLMContext(con, domainName, userName, password, workstation);
```
peterbae: writing it this way (differentiating between context and this.context) makes me think that `context` was passed to this method, but it's not...it's not a big deal though.

src/main/java/com/microsoft/sqlserver/jdbc/NTLMAuthentication.java
```diff
@@ -0,0 +1,910 @@
...
+         *         if error occurs
+         */
+        NTLMContext(final SQLServerConnection con, final String domainName, final String userName,
+                final String password, final String workstation) throws SQLServerException {
+
+            this.domainName = null != domainName ? domainName.toUpperCase()
+                                                 : SQLServerDriverStringProperty.DOMAIN.getDefaultValue();
+            this.domainUbytes = unicode(this.domainName);
+
+            this.userNameUbytes = null != userName ? unicode(userName) : null;
```
ulvii: The `null` case is not properly handled here, if `userName` is `null`, the next `this.upperUserName = userName.toUpperCase();` line will result in `NullPointerException`.

src/main/java/com/microsoft/sqlserver/jdbc/SSPIAuthentication.java
```diff
@@ -5,12 +5,193 @@
...
+
 /**
  * Integrated Authentication master file. Common items for kerb and JNI auth are in this interface.
  */
 
 abstract class SSPIAuthentication {
-    abstract byte[] GenerateClientContext(byte[] pin, boolean[] done) throws SQLServerException;
+    abstract byte[] generateClientContext(byte[] pin, boolean[] done) throws SQLServerException;
+
+    abstract int releaseClientContext() throws SQLServerException;
```
peterbae: I know this code was written like this before, but do we need to keep it return an int? I don't think this method needs to return anything.

src/main/java/com/microsoft/sqlserver/jdbc/NTLMAuthentication.java
```diff
@@ -0,0 +1,910 @@
...
+    // Handle to NTLM context
+    private NTLMContext context = null;
+
+    /**
+     * Creates an instance of the NTLM authentication
+     *
+     * @param con
+     *        connection to SQL server
+     * @param domainName
+     *        domain name to authentication in using NTLM
```
ulvii:
```suggestion
     *        domain name used for NTLM authentication
```

src/main/java/com/microsoft/sqlserver/jdbc/NTLMAuthentication.java
```diff
@@ -0,0 +1,910 @@
...
+     * @param userName
+     *        user name
```
ulvii:
```suggestion
     *        domain user
```

src/main/java/com/microsoft/sqlserver/jdbc/NTLMAuthentication.java
```diff
@@ -0,0 +1,910 @@
...
+     * @param password
+     *        password
```
ulvii:
```suggestion
     *        domain password
```

src/main/java/com/microsoft/sqlserver/jdbc/NTLMAuthentication.java
```diff
@@ -0,0 +1,910 @@
...
+         * Section 2.2.2.1 AV_PAIR
+         *
+         * Section 7 Appendix B Product Behavior
+         * 
+         * This structure should always be sent in the CHALLENGE_MESSAGE not supported in Windows NT, Windows 2000,
+         * Windows XP, and Windows Server 2003
+         */
+        if (null == context.timestamp || 0 >= context.timestamp.length) {
+            // this SHOULD always be present but for some reason occasionally this had seen to be missing
+            if (logger.isLoggable(Level.FINEST)) {
```
ulvii: I wonder if this should be a severe warning.

src/main/java/com/microsoft/sqlserver/jdbc/NTLMAuthentication.java
```diff
@@ -0,0 +1,910 @@
...
+                logger.finer(SQLServerException.getErrString("R_ntlmNoTimestamp"));
```
cheenamalhotra: Should log same level as being checked above.

src/main/java/com/microsoft/sqlserver/jdbc/NTLMAuthentication.java
```diff
@@ -0,0 +1,910 @@
...
+     * Parses the Type 2 NTLM Challenge message from server
+     *
+     * Section 2.2.1.2 CHALLENGE_MESSAGE
+     *
+     * @param inToken
+     *        SSPI input blob
+     * @throws SQLServerException
+     *         if error occurs
+     */
+    private void parseNtlmChallenge(final byte[] inToken) throws SQLServerException {
```
cheenamalhotra: Should have Entering/Exiting FINEST logs in all methods.

src/main/java/com/microsoft/sqlserver/jdbc/NTLMAuthentication.java
```diff
@@ -0,0 +1,910 @@
...
+                case NTLM_AVID_MSVAVTARGETNAME:
+                    break;
+                default:
+                    MessageFormat form = new MessageFormat(SQLServerException.getErrString("R_ntlmUnknownValue"));
+                    Object[] msgArgs = {value};
+                    throw new SQLServerException(form.format(msgArgs), null);
+            }
+
+            if (logger.isLoggable(Level.FINEST)) {
+                MessageFormat form = new MessageFormat(SQLServerException.getErrString("R_ntlmAvp"));
```
cheenamalhotra: Log message need not be localized string

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResource.java
```diff
@@ -43,9 +42,11 @@ static String getResource(String key) {
...
             {"R_unknownJDBCType", "Invalid JDBC data type {0}."},
             {"R_notSQLServer",
-                    "The driver received an unexpected pre-login response. Verify the connection properties and check that an instance of SQL Server is running on the host and accepting TCP/IP connections at the port. This driver can be used only with SQL Server 2005 or later."},
+                    "The driver received an unexpected pre-login response. Verify the connection properties and check that an instance of SQL Server is running on the host and accepting "
+                            + "TCP/IP connections at the port. This driver can be used only with SQL Server 2005 or later."},
             {"R_tcpOpenFailed",
-                    "{0}. Verify the connection properties. Make sure that an instance of SQL Server is running on the host and accepting TCP/IP connections at the port. Make sure that TCP connections to the port are not blocked by a firewall."},
+                    "{0}. Verify the connection properties. Make sure that an instance of SQL Server is running on the host and accepting TCP/IP connections at the port. Make sure that TCP "
```
ulvii: Is this really necessary?

src/test/java/com/microsoft/sqlserver/jdbc/NTLMConnectionTest.java
```diff
@@ -0,0 +1,420 @@
...
+     */
+    private void verifyDatabase(Connection con, String databaseName) throws SQLException {
+        try (Statement stmt = con.createStatement(); ResultSet rs = stmt.executeQuery("SELECT DB_NAME()")) {
+            while (rs.next()) {
+                assertEquals(databaseName, rs.getString(1));
+            }
+        }
+    }
+
+    static String extractPort(String server, SQLServerDataSource ds) {
```
ulvii: Why is this method here?

PR #1000, opened 2019-03-21
Align DateTimeOffset's behavior with OffsetDateTime

peterbae:
There's probably multiple ways of handling this problem, but this is my take on it. Not to be merged until everyone agrees with this approach, and in the future we might want to consider offering the ability to create a DateTimeOffset object with a non-Timestamp object.


cheenamalhotra:
PR #1005 adds test to validate driver's behavior is right. Closing this PR.

PR #1007, opened 2019-03-23
Update default value of connection property 'sendTimeAsDatetime' to false.

cheenamalhotra:


barryw-mssql:
These are changes we want to make but this is a breaking change and our test lab is currently unable to test this scenario

PR #1009, opened 2019-03-24
Fix SQLServerDatabaseMetadata.getMaxConnections() API query

cheenamalhotra:
Currently driver was using `sp_configure` to fetch 'user connections' configuration which is not correct way to do it. Also with SQL Server 2019, this configuration is moved to 'Advanced Options' hence cannot be run directly as with older SQL Servers.

This PR modifies the query and executes SELECT from `sys.configurations` which is the recommended way for gathering metadata information.


src/test/java/com/microsoft/sqlserver/jdbc/databasemetadata/DatabaseMetaDataTest.java
```diff
@@ -546,4 +546,21 @@ public void testPreparedStatementMetadataCaching() throws SQLException {
...
+    public void testGetMaxConnections() throws SQLException {
+        try (Statement stmt = connection.createStatement(); ResultSet rs = stmt
+                .executeQuery("select maximum from sys.configurations where name = 'user connections'")) {
+            assert (null != rs);
+            rs.next();
+
+            DatabaseMetaData databaseMetaData = connection.getMetaData();
+            int maxConn = databaseMetaData.getMaxConnections();
+            
+            assert (0 != maxConn);
```
peterbae: why do we not allow this to be 0?

cheenamalhotra: It is impossible to connect to database and get response as 0 - at least 1 connection will be allowed when making this query on an active connection.

cheenamalhotra: Fixed

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
```diff
@@ -1015,15 +1015,19 @@ public int getMaxColumnsInTable() throws SQLServerException {
     public int getMaxConnections() throws SQLException, SQLTimeoutException {
         checkClosed();
         try {
-            String s = "sp_configure 'user connections'";
-            SQLServerResultSet rs = getResultSetFromInternalQueries(null, s);
-            if (!rs.next())
-                return 0;
+            SQLServerResultSet rs = getResultSetFromInternalQueries(null,
```
ulvii: If the user doesn't have required privileges, won't this query result in an exception?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
```diff
@@ -1014,16 +1015,26 @@ public int getMaxColumnsInTable() throws SQLServerException {
     @Override
     public int getMaxConnections() throws SQLException, SQLTimeoutException {
         checkClosed();
+        SQLServerResultSet rs = null;
         try {
-            String s = "sp_configure 'user connections'";
-            SQLServerResultSet rs = getResultSetFromInternalQueries(null, s);
-            if (!rs.next())
+            rs = getResultSetFromInternalQueries(null,
```
ulvii: `rs` is never closed.

ulvii: I noticed `getUserName()` leaks a `ResultSet` too. Would you also double check this class for other leaks?

src/test/java/com/microsoft/sqlserver/jdbc/TestResource.java
```diff
@@ -82,6 +82,7 @@ public static String getResource(String key) {
             {"R_loginFailed", "Login failed"}, {"R_exitedMoreSeconds", "Exited in more than {0} seconds."},
             {"R_exitedLessSeconds", "Exited in less than {0} seconds."},
             {"R_invalidArgumentExecutor", "The argument executor is not valid"},
+            {"R_invalidArgumentCatalog", "The argument catalog is not valid"},
```
ulvii: There is no need for this anymore. You should be able to use SQLServerResource directly by calling `formatErrorMsg()`. See PR #1034.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
cheenamalhotra: `getUserName()` is fine, its in a try-with-resources block. I fixed this one, good catch!

src/test/java/com/microsoft/sqlserver/jdbc/TestResource.java
cheenamalhotra: This error message doesn't exist in SQLServerResource, and is thrown by SQL Server, so this should be fine.

rene-ye: Are these for string comparisons? There are a few issues where SQL Server would return error messages with an extra period depending on the version (mainly DW). E.g "The connection is closed" vs "The connection is closed.".

cheenamalhotra: Yep - I'm comparing with "contains"

ulvii: https://github.com/Microsoft/mssql-jdbc/blob/51d6c40a1193c879d9b10eca254c510c84fe5a15/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResource.java#L159

PR #1015, opened 2019-03-29
Tests | Support multi-platform testing with exclusion tags and fix Skipping tests

cheenamalhotra:
This PR

- Adds support to run JUnit tests with multiple SQL databases without skipping tests.
Tagging tests for exclusions safely excludes test classes/tests from being tested against targeted SQL Server. Exclusion tags are based on non-compatibility of tests against respective SQL Servers.
- Adds plugin to perform `jacoco:merge` to generate cumulative code coverage for the source code.
- Modifies tests to use Static `connection` instance from `AbstractTest` for operations that do not require special connection like creating/dropping objects. This `connection` is created either way with every test. **Reduces execution time by 45 secs.**
- Removes unwanted Test Dependencies from POM.xml


src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/BatchExecuteWithErrorsTest.java
```diff
@@ -120,20 +119,209 @@ private void Repro47239Internal(String mode) throws Exception {
...
+            // On SQL Azure, raising FATAL error by RAISERROR() is not supported and there is no way to
+            // cut the current connection by a statement inside a SQL batch.
+            // Details: Although one can simulate a fatal error (that cuts the connections) by dropping the
+            // database,
+            // this simulation cannot be written entirely in TSQL (because it needs a new connection),
+            // and thus it cannot be put into a TSQL batch and it is useless here.
+            // So we have to skip the last scenario of this test case, i.e. "Test Severe (connection-closing)
+            // errors"
+            // It is worthwhile to still execute the first 5 test scenarios of this test case, in order to have best
+            // test coverage.
```
lilgreenbird: multi-line comments style (same in Repro47239largeInternal)

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/BatchExecuteWithErrorsTest.java
```diff
@@ -142,62 +330,58 @@ private void Repro47239Internal(String mode) throws Exception {
...
-                    log.fine("UpdateCounts:");
-                }
-                for (int updateCount : actualUpdateCounts) {
+
+                log.fine("UpdateCounts:");
+                for (long updateCount : actualUpdateCounts) {
                     log.fine("" + updateCount + ",");
                 }
-                log.fine("");
+
```
lilgreenbird: check if FINE 1st the for loop isn't necessary if not

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/BatchExecuteWithErrorsTest.java
cheenamalhotra: Its the existing test+comments moved over to a different location hence shows changes like that. I don't think that's critical for this PR. Can be worked upon separately.

lilgreenbird: it's not, I only brought this up cos I have another comment on this file :)

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/BatchExecuteWithErrorsTest.java
cheenamalhotra: I'll remove these logs from Junit tests in next PR - we don't need them

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyRowSetTest.java
```diff
@@ -42,41 +47,40 @@ public void testBulkCopyFloatRowSet() throws SQLException {
...
             String sql1 = "create table " + tableName + " (c1 float, c2 real)";
             stmt.execute(sql1);
         }
     }
 
     @AfterAll
     public static void terminateVariation() throws SQLException {
-        try (Connection connection = DriverManager.getConnection(connectionString);
-                Statement stmt = connection.createStatement()) {
+        try (Statement stmt = connection.createStatement()) {
```
ulvii: Why are we changing the tests to use `connection` from `AbstractTest`? Dropping tables will fail if that connection is broken.

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/SQLServerSpatialDatatypeTest.java
```diff
@@ -187,6 +187,7 @@ public void testFullGlobeWkb() throws SQLException {
     }
 
     @Test
+    @Tag("xAzureSQLDW")
```
ulvii: Why don't we skip the entire class instead?

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyRowSetTest.java
cheenamalhotra: This Static Connection is recreated everytime with every test class. This is an attempt to reduce number of connections specially when running tests with Azure databases.

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/SQLServerSpatialDatatypeTest.java
cheenamalhotra: Not all tests are skipped in this class.

ulvii: Spatial datatypes are not supported on DW, I thought we could skip the entire class for that reason, instead of tagging so many tests. Looks like the tests that are not skipped do not contact database at all. I will leave this up to you.

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyRowSetTest.java
cheenamalhotra: As also mentioned in PR description, this change reduces time of Junit execution by 45 secs in normal run (against SQL 2019) and much more on Azure databases.

Issue #1013, opened 2019-03-27
[BUG] DatabaseMetaData.getColumns(...) IS_GENERATEDCOLUMN value icorrect


george-walker:
## Driver version
7.2.1.jre8

## SQL Server version
Azure SQL

## Client Operating System
Ubuntu 18.04

## JAVA/JVM version
openjdk full version "1.8.0_191-8u191-b12-2ubuntu0.18.04.1-b12"

## Table schema
```
CREATE TABLE product_list (
  id int auto_increment primary key,
  name NVARCHAR(100) NOT NULL,
  category NVARCHAR(50),
  business_unit NVARCHAR(50),
  quantity INTEGER,
  price FLOAT,
  realname NVARCHAR AS concat( business_unit, ':', name )
)
```

## Problem description
1. Expected behaviour:

https://docs.oracle.com/javase/7/docs/api/java/sql/DatabaseMetaData.html#getColumns(java.lang.String,%20java.lang.String,%20java.lang.String,%20java.lang.String)

The column IS_GENERATEDCOLUMN should return "YES" or "NO" as per JDBC

In the DDL above the column *realname* is a computed column

2. Actual behaviour:

The column IS_GENERATEDCOLUMN return "1" or "0"

3. Error message/stack trace:
4. Any other details that can be helpful:

peterbae:
Hi @george-walker, thanks for contacting us about the issue. We've confirmed this issue, and will come up with a PR addressing this problem soon.

rene-ye:
Hi @george-walker, #1016 has been merged. Closing the issue.

PR #1016, opened 2019-04-01
Fix | Align DatabaseMetadata.getColumns with JDBC 4.3 Specifications

rene-ye:
Addresses #1013. Aligns `ResultSet` retrieved from `DatabaseMetaData.getColumns()` with what's described in the JDBC Specifications [here](https://docs.oracle.com/en/java/javase/11/docs/api/java.sql/java/sql/DatabaseMetaData.html#getColumns(java.lang.String,java.lang.String,java.lang.String,java.lang.String)). Requires a change in [our current MSDN docs](https://docs.microsoft.com/en-us/sql/connect/jdbc/reference/getcolumns-method-sqlserverdatabasemetadata?view=sql-server-2017) and possibly a major version bump.

Changes in createSqlKeyWords() due to formatting. Removed Katmai checks and explicitly uses sp_columns_100 as we no longer provide support for SQL Server 2008 and below.

SCOPE_CATALOG, SCOPE_SCHEMA, and SCOPE_TABLE explicitly always returns `NULL` as SQL Server currently doesn't seem to support `REF` column types. The description for these fields is pasted below from the JDBC specification link above:
> SCOPE_CATALOG String => catalog of table that is the scope of a reference attribute (null if DATA_TYPE isn't REF)
> SCOPE_SCHEMA String => schema of table that is the scope of a reference attribute (null if the DATA_TYPE isn't REF)
> SCOPE_TABLE String => table name that this the scope of a reference attribute (null if the DATA_TYPE isn't REF)


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
```diff
@@ -601,72 +601,63 @@ private static String EscapeIDName(String inID) throws SQLServerException {
...
+            pstmt.setInt(6, 3);// odbc version
+
+            rs = (SQLServerResultSet) pstmt.executeQuery();
+            rs.getColumn(5).setFilter(new DataTypeFilter());
             rs.getColumn(7).setFilter(new ZeroFixupFilter());
             rs.getColumn(8).setFilter(new ZeroFixupFilter());
             rs.getColumn(16).setFilter(new ZeroFixupFilter());
+        } catch (SQLException e) {
+            errorOnClose = e;
+            pstmt.close();
```
peterbae: Do we need pstmt.close() here as well as the closeOnCompletion above? I think we can get rid of the catch statement and still have the same flow. Perhaps clean up the logic a bit?

rene-ye: The statement needs closeOnCompletion because the ResultSet gets passed onto the user, and we can't close the statement until the user is done with RS. The close in the catch block is needed in case ResultSet fetching fails, in which case the statement will never decrease its ResultSet count, and closeOnCompletion is useless. In this scenario, the pstmt will just leak and never get closed.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
```diff
@@ -601,72 +601,63 @@ private static String EscapeIDName(String inID) throws SQLServerException {
...
-                    getColumnsColumnNames);
-        // Hook in a filter on the DATA_TYPE column of the result set we're
-        // going to return that converts the ODBC values from sp_columns
-        // into JDBC values.
-        rs.getColumn(5).setFilter(new DataTypeFilter());
-
-        if (connection.isKatmaiOrLater()) {
-            rs.getColumn(22).setFilter(new IntColumnIdentityFilter());
+        String originalCatalog = null;
+        originalCatalog = switchCatalogs(catalog);
```
ulvii: This won't work against Azure.

rene-ye: Original implementation did not take Azure into account either, method will just throw w/e error Azure throws when attempting to switch cats. Switchcatalog returns null immediately if no catalog is provided.

rene-ye:
> add test

Existing API with tests, already being tested.

ulvii:
Please add the new columns to `DatabaseMetaDataTest.testGetDBColumn()` to make sure they are tested.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
```diff
@@ -601,72 +601,63 @@ private static String EscapeIDName(String inID) throws SQLServerException {
...
-            /* 3 */ TABLE_NAME, /* 4 */ COLUMN_NAME, /* 5 */ DATA_TYPE, /* 6 */ TYPE_NAME, /* 7 */ COLUMN_SIZE,
-            /* 8 */ BUFFER_LENGTH, /* 9 */ DECIMAL_DIGITS, /* 10 */ NUM_PREC_RADIX, /* 11 */ NULLABLE, /* 12 */ REMARKS,
-            /* 13 */ COLUMN_DEF, /* 14 */ SQL_DATA_TYPE, /* 15 */ SQL_DATETIME_SUB, /* 16 */ CHAR_OCTET_LENGTH,
-            /* 17 */ ORDINAL_POSITION, /* 18 */ IS_NULLABLE, /* 20 */ SS_IS_SPARSE, /* 20 */ SS_IS_COLUMN_SET,
-            /* 21 */ IS_GENERATEDCOLUMN, /* 22 */ IS_AUTOINCREMENT};
-
     @Override
     public java.sql.ResultSet getColumns(String catalog, String schema, String table,
-            String col) throws SQLServerException, SQLTimeoutException {
+            String col) throws SQLException {
```
ulvii: Needs to be formatted.

rene-ye:
Changes are being tested in our framework. Cannot modify `DatabaseMetaDataTest.testGetDBColumn()` because the purpose of the test is to check and make sure the values returned from each of the columns is not `NULL`. This isn't true as some columns here is always returned `NULL`, and the test itself isn't checking to make sure all columns are present. A test that does this has been modified in the framework tests.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
```diff
@@ -602,77 +605,168 @@ private static String EscapeIDName(String inID) throws SQLServerException {
...
-        else
-            rs = getResultSetWithProvidedColumnNames(catalog, CallableHandles.SP_COLUMNS, arguments,
-                    getColumnsColumnNames);
-        // Hook in a filter on the DATA_TYPE column of the result set we're
-        // going to return that converts the ODBC values from sp_columns
-        // into JDBC values.
-        rs.getColumn(5).setFilter(new DataTypeFilter());
+            return rs;
+        } else {
+            try (PreparedStatement stored_proc_pstmt = this.connection
```
ulvii: camelCase

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
```diff
@@ -602,77 +605,168 @@ private static String EscapeIDName(String inID) throws SQLServerException {
...
+                    .prepareStatement("EXEC sp_columns_100 ?,?,?,?,?,?;")) {
+                stored_proc_pstmt.setString(1, (null != table && !table.isEmpty()) ? table : "%");
+                stored_proc_pstmt.setString(2, (null != schema && !schema.isEmpty()) ? schema : "%");
+                stored_proc_pstmt.setString(3,
+                        (null != catalog && !catalog.isEmpty()) ? catalog : this.connection.getCatalog());
+                stored_proc_pstmt.setString(4, (null != col && !col.isEmpty()) ? col : "%");
+                stored_proc_pstmt.setInt(5, 2);// show sparse columns
+                stored_proc_pstmt.setInt(6, 3);// odbc version
+
+                SQLServerResultSet user_rs = null;
```
ulvii: camelCase

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
```diff
@@ -602,77 +605,168 @@ private static String EscapeIDName(String inID) throws SQLServerException {
...
-                    getColumnsColumnNamesKatmai);
-        else
-            rs = getResultSetWithProvidedColumnNames(catalog, CallableHandles.SP_COLUMNS, arguments,
-                    getColumnsColumnNames);
-        // Hook in a filter on the DATA_TYPE column of the result set we're
-        // going to return that converts the ODBC values from sp_columns
-        // into JDBC values.
-        rs.getColumn(5).setFilter(new DataTypeFilter());
+            return rs;
+        } else {
```
ulvii: Need to add `switchCatalogs(catalog);` call here to make sure the behavior is consistent with Azure DB.

ulvii: Or remove `switchCatalogs(catalog);` for Azure DB too.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
```diff
@@ -602,77 +605,174 @@ private static String EscapeIDName(String inID) throws SQLServerException {
...
+                    userRs.getColumn(16).setFilter(new ZeroFixupFilter());
+                    userRs.getColumn(23).setFilter(new IntColumnIdentityFilter());
+                    userRs.getColumn(24).setFilter(new IntColumnIdentityFilter());
+                } catch (SQLException e) {
+                    errorOnClose = e;
+                    if (null != result_pstmt) {
+                        result_pstmt.close();
+                    }
+                } finally {
+                    if (null != errorOnClose) {
```
cheenamalhotra: Why is it needed in finally block?

rene-ye: Because if the error causes the pstmt to close, then calling pstmt.close() in the catch block will throw something like "statement already closed". That causes us to lose the original error that caused the exception.

cheenamalhotra: if SQLException is captured in catch block, try block resource (Resultset rs) is automatically closed and since you set result_pstmt to close on completion, it will be closed too, so you can throw exception from the catch block safely.

rene-ye: Because the finally block always executes and I don't want to throw null.

cheenamalhotra: Ideally, you should be closing `result_pstmt` in try catch block and catch that exception safely, may just not do anything if it occurs. Coz right now if `result_pstmt.close()` throws exception, you are ignoring thrown exception from this statement and throwing another exception which doesn't seem right.

Alternatively you could do `if (null != result_pstmt && !result_pstmt.isClosed())`

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
```diff
@@ -602,77 +605,174 @@ private static String EscapeIDName(String inID) throws SQLServerException {
...
+                pstmt.setInt(6, 3);// odbc version
+
+                rs = (SQLServerResultSet) pstmt.executeQuery();
+                rs.getColumn(5).setFilter(new DataTypeFilter());
+                rs.getColumn(7).setFilter(new ZeroFixupFilter());
+                rs.getColumn(8).setFilter(new ZeroFixupFilter());
+                rs.getColumn(16).setFilter(new ZeroFixupFilter());
+            } catch (SQLException e) {
+                errorOnClose = e;
+                pstmt.close();
```
cheenamalhotra: I guess similar conditions should come here.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
```diff
@@ -602,77 +605,174 @@ private static String EscapeIDName(String inID) throws SQLServerException {
...
+            }
+        }
+    }
 
-        if (connection.isKatmaiOrLater()) {
-            rs.getColumn(22).setFilter(new IntColumnIdentityFilter());
-            rs.getColumn(7).setFilter(new ZeroFixupFilter());
-            rs.getColumn(8).setFilter(new ZeroFixupFilter());
-            rs.getColumn(16).setFilter(new ZeroFixupFilter());
+    private String generateAzureDWSelect(ResultSet rs, Map<Integer, String> columns) throws SQLException {
```
cheenamalhotra: Does it actually throw SQLException?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
rene-ye: This change is to circumvent the issue you described. The exception the statement throws on statement.close() isn't descriptive or relevant to the actual reason the try block failed. We can't add result_pstmt to the try-catch because closing the pstmt that generated the resultset the user needs to read from closes the resultset as well.

cheenamalhotra: What I meant to say was:
```java             
} catch (SQLException e) {
    if (null != result_pstmt) {
         try{
               result_pstmt.close();
          } catch (SQLException e) { 
               //do nothing but catch it
          }
    }
    throw e;
}
```
OR
```java             
} catch (SQLException e) {
    if (null != result_pstmt && !result_pstmt.isClosed()) {
          result_pstmt.close();
    }
    throw e;
}
```

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
```diff
@@ -602,77 +605,181 @@ private static String EscapeIDName(String inID) throws SQLServerException {
...
+                    + "TYPE_NAME, PRECISION AS COLUMN_SIZE, LENGTH AS BUFFER_LENGTH, SCALE AS DECIMAL_DIGITS, RADIX AS NUM_PREC_RADIX,"
+                    + "NULLABLE, REMARKS, COLUMN_DEF, SQL_DATA_TYPE, SQL_DATETIME_SUB, CHAR_OCTET_LENGTH, ORDINAL_POSITION, IS_NULLABLE,"
+                    + "NULL AS SCOPE_CATALOG, NULL AS SCOPE_SCHEMA, NULL AS SCOPE_TABLE, SS_DATA_TYPE AS SOURCE_DATA_TYPE,"
+                    + "CASE SS_IS_IDENTITY WHEN 0 THEN 'NO' WHEN 1 THEN 'YES' WHEN '' THEN '' END AS IS_AUTOINCREMENT,"
+                    + "CASE SS_IS_COMPUTED WHEN 0 THEN 'NO' WHEN 1 THEN 'YES' WHEN '' THEN '' END AS IS_GENERATEDCOLUMN, "
+                    + "SS_IS_SPARSE, SS_IS_COLUMN_SET, SS_UDT_CATALOG_NAME, SS_UDT_SCHEMA_NAME, SS_UDT_ASSEMBLY_TYPE_NAME,"
+                    + "SS_XML_SCHEMACOLLECTION_CATALOG_NAME, SS_XML_SCHEMACOLLECTION_SCHEMA_NAME, SS_XML_SCHEMACOLLECTION_NAME "
+                    + "FROM @mssqljdbc_temp_sp_columns_result;";
+            SQLServerResultSet rs = null;
+            SQLException errorOnClose = null;
```
cheenamalhotra: Don't need this now.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
```diff
@@ -602,77 +605,181 @@ private static String EscapeIDName(String inID) throws SQLServerException {
...
-            String col) throws SQLServerException, SQLTimeoutException {
+    @Override
+    public java.sql.ResultSet getColumns(String catalog, String schema, String table, String col) throws SQLException {
         if (loggerExternal.isLoggable(Level.FINER) && Util.isActivityTraceOn()) {
             loggerExternal.finer(toString() + " ActivityId: " + ActivityCorrelator.getNext().toString());
         }
         checkClosed();
+        if (!this.connection.isAzureDW()) {
+            String originalCatalog = null;
+            originalCatalog = switchCatalogs(catalog);
```
cheenamalhotra: I know its minor, but u could avoid an extra assignment.. 
`String originalCatalog = switchCatalogs(catalog);`

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
```diff
@@ -602,77 +605,181 @@ private static String EscapeIDName(String inID) throws SQLServerException {
...
-        // into JDBC values.
-        rs.getColumn(5).setFilter(new DataTypeFilter());
+            return rs;
+        } else {
+            /**
+             * Can't actually switchCatalogs on Azure DW. This is here to keep consistency in behavior with SQL Azure DB
+             * when user provides a different catalog than the one they're currently connected to. Will throw exception
+             * when it's different and do nothing if it's the same/null.
+             */
+            switchCatalogs(catalog);
```
cheenamalhotra: This is else block and only for DW, you can remove `switchCatalogs` call here. Also the comment above.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
```diff
@@ -602,77 +605,181 @@ private static String EscapeIDName(String inID) throws SQLServerException {
...
+                    userRs.getColumn(8).setFilter(new ZeroFixupFilter());
+                    userRs.getColumn(16).setFilter(new ZeroFixupFilter());
+                    userRs.getColumn(23).setFilter(new IntColumnIdentityFilter());
+                    userRs.getColumn(24).setFilter(new IntColumnIdentityFilter());
+                } catch (SQLException e) {
+                    if (null != resultPstmt) {
+                        try {
+                            resultPstmt.close();
+                        } catch (SQLServerException ignore) {
+                            if (loggerExternal.isLoggable(Level.FINER) && Util.isActivityTraceOn()) {
```
cheenamalhotra: Either append ActivityId in log, or remove the check `Util.isActivityTraceOn()`

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
```diff
@@ -602,77 +605,181 @@ private static String EscapeIDName(String inID) throws SQLServerException {
...
+                storedProcPstmt.setString(1, (null != table && !table.isEmpty()) ? table : "%");
+                storedProcPstmt.setString(2, (null != schema && !schema.isEmpty()) ? schema : "%");
+                storedProcPstmt.setString(3,
+                        (null != catalog && !catalog.isEmpty()) ? catalog : this.connection.getCatalog());
+                storedProcPstmt.setString(4, (null != col && !col.isEmpty()) ? col : "%");
+                storedProcPstmt.setInt(5, 2);// show sparse columns
+                storedProcPstmt.setInt(6, 3);// odbc version
+
+                SQLServerResultSet userRs = null;
+                SQLException errorOnClose = null;
```
cheenamalhotra: Remove errorOnClose

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
rene-ye: Was asked to add that to maintain consistency with Azure SQL.

ulvii: How about calling `switchCatalogs(catalog);` only if the server `!isAzure()`?

cheenamalhotra: You may optionally check for !isAzureDB() in the if block to avoid extra call for Azure DB, but here in else block that we know applies to only DW it makes no sense.

cheenamalhotra: I see one reason why this call is useful is to validate if the catalog user specified is correct or not, as this is public API. How about if we verify Catalog name to match database name and throw exception otherwise without the need to contact database for Azure DB/DW/Managed Instance and otherwise let switch catalog? Don't think that would cause much of performance hit since we are already making call to get database type.

ulvii: I see your point. The same table can exist in both databases, so we definitely need to verify the specified catalog name.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
```diff
@@ -602,77 +605,178 @@ private static String EscapeIDName(String inID) throws SQLServerException {
...
+                rs.getColumn(7).setFilter(new ZeroFixupFilter());
+                rs.getColumn(8).setFilter(new ZeroFixupFilter());
+                rs.getColumn(16).setFilter(new ZeroFixupFilter());
+            } catch (SQLException e) {
+                if (null != pstmt) {
+                    try {
+                        pstmt.close();
+                    } catch (SQLServerException ignore) {
+                        if (loggerExternal.isLoggable(Level.FINER)) {
+                            loggerExternal.finer("findColumn() threw an exception when attempting to close PreparedStatement");
```
ulvii: API name is `getColumns()`

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
```diff
@@ -602,77 +605,178 @@ private static String EscapeIDName(String inID) throws SQLServerException {
...
+                    userRs.getColumn(16).setFilter(new ZeroFixupFilter());
+                    userRs.getColumn(23).setFilter(new IntColumnIdentityFilter());
+                    userRs.getColumn(24).setFilter(new IntColumnIdentityFilter());
+                } catch (SQLException e) {
+                    if (null != resultPstmt) {
+                        try {
+                            resultPstmt.close();
+                        } catch (SQLServerException ignore) {
+                            if (loggerExternal.isLoggable(Level.FINER)) {
+                                loggerExternal.finer("findColumn() threw an exception when attempting to close PreparedStatement");
```
ulvii: API name is `getColumns()`

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
rene-ye: This change isn't the purpose of this PR and if we do decide to do something like this, it should be applied to every method in this class. As it stands calling `switchCatalogs` just throws an error indicating that catalog switching isn't allowed on Azure, which is basically what we want. We can't check this property without contacting the server because isAzure/DW/MI all contact the server, and this API is meant to work on other databases when on a regular server.

cheenamalhotra: Maybe we can leave this change for now then, and handle it in a separate PR. For now, please keep this statement to 'switchCatalog' out of 'if' block as it's called in both cases eitherway.

We may modify the method itself to work differently only for Azure Databases.. to avoid DB call and validate catalog name explicitly, allowing acceptable values only.

PR #1020, opened 2019-04-03
Fix ActivityID map getting not cleaned up properly

peterbae:
Replaces PR #1018.

Every time `ActivityCorrelator::cleanupActivityId` is called, the driver now iterates through the Activity ID map and removes entries that are no longer needed.

To alleviate this change from affecting the driver's performance, I've made changes during the pre-login process to not create an ActivityID if Activity Trace is not on (before, the driver was creating an ActivityID during pre-login regardless of the Activity Trace being on or not), because Activity ID is an optional part of the pre-login header. This means that the Activity ID map will be empty if Activity Trace is turned off, as opposed to before where the Activity ID map was getting populated regardless.

I've also added tests to show that the Activity ID map is no longer leaking memory.


src/main/java/com/microsoft/sqlserver/jdbc/ActivityCorrelator.java
```diff
@@ -18,25 +18,20 @@
...
 
     static void cleanupActivityId() {
-        // remove the ActivityId that belongs to this thread.
-        long uniqueThreadId = Thread.currentThread().getId();
-
-        if (activityIdTlsMap.containsKey(uniqueThreadId)) {
-            activityIdTlsMap.remove(uniqueThreadId);
-        }
+        // remove ActivityIds that belongs to this thread or no longer have an associated thread.
+        activityIdTlsMap.entrySet().removeIf(e -> !e.getValue().getThread().isAlive()
```
David-Engel: I'd include some defensive coding:
        activityIdTlsMap.entrySet().removeIf(e ->
            e.getValue() == null ||
            e.getValue().getThread() == null ||
            e.getValue().getThread() == Thread.currentThread() ||
            !e.getValue().getThread().isAlive());

David-Engel:
I like the additional change to not populate the map at all if tracing is off. Other than one optional check in my comment, looks good to me.

peterbae:
Attaching the latest jars (zip file contains both JRE8/11 versions) for testing:
[mssql-jdbc-7.3.0-preview-PR1020.zip](https://github.com/Microsoft/mssql-jdbc/files/3045724/mssql-jdbc-7.3.0-preview-PR1020.zip)

src/test/java/com/microsoft/sqlserver/jdbc/ActivityIDTest.java
```diff
@@ -0,0 +1,179 @@
...
+            es.shutdown();
+            pooledCon.close();
+        }
+        assertEquals(0, ActivityCorrelator.getActivityIdTlsMap().size());
+    }
+    
+    @BeforeEach
+    @AfterEach
+    public void clearActivityId() {
+        ActivityCorrelator.clear();
```
rene-ye: All of these tests assume ActivityCorrelator size should be 0 at the end of the tests. If this is necessary, doesn't that imply there's something inherently wrong with the tests? Also I don't think it needs to be both Before and After.

cheenamalhotra: The problem is only with HikariCP test - but I think this part can be removed since we figured out an alternative to clear activity ID for Hikari's thread - @peterbae can we run test without this method?

peterbae: I agree it's redundant after the changes we made. I've removed this part.

src/test/java/com/microsoft/sqlserver/jdbc/ActivityIDTest.java
```diff
@@ -0,0 +1,171 @@
...
+    @Test
+    public void testActivityIDPooledConnection() throws Exception {
+        int poolsize = 10;
+        int numPooledExecution = 200;
+
+        PooledConnection pooledCon = ((SQLServerConnectionPoolDataSource) dsPool).getPooledConnection();
+        ExecutorService es = Executors.newFixedThreadPool(poolsize);
+        try {
+            CountDownLatch latchPool = new CountDownLatch(numPooledExecution);
+            es.execute(new Runnable() {
```
rene-ye:
```java
es.execute(() -> {
    IntStream.range(0, numPooledExecution).forEach(i -> {
        try (Connection c = pooledCon.getConnection(); Statement s = c.createStatement();
                ResultSet rs = s.executeQuery("SELECT @@VERSION AS 'SQL Server Version'")) {
            while (rs.next()) {
                rs.getString(1);
            }
        } catch (SQLException e) {
            fail("Pooled Thread #" + i + " failed with: " + e.toString());
        }
        latchPool.countDown();
    });
});
```

rene-ye: Please make similar changes to code below, fully declaring the class/constructor for `Runnable` is unnecessary if there are no additional helper methods needed.

peterbae: Good suggestion, done.

src/main/java/com/microsoft/sqlserver/jdbc/ActivityCorrelator.java
```diff
@@ -17,26 +17,30 @@
 
     private static Map<Long, ActivityId> activityIdTlsMap = new ConcurrentHashMap<>();
 
-    static void cleanupActivityId() {
-        // remove the ActivityId that belongs to this thread.
-        long uniqueThreadId = Thread.currentThread().getId();
-
-        if (activityIdTlsMap.containsKey(uniqueThreadId)) {
-            activityIdTlsMap.remove(uniqueThreadId);
+    static void clear() {
```
cheenamalhotra: Method can be removed as we deleted its usage from test.

src/main/java/com/microsoft/sqlserver/jdbc/ActivityCorrelator.java
```diff
@@ -17,26 +17,30 @@
...
-        // Since the Id for each thread is unique, this assures that the below if statement is run only once per thread.
-        if (!activityIdTlsMap.containsKey(uniqueThreadId)) {
-            activityIdTlsMap.put(uniqueThreadId, new ActivityId());
+        Thread thread = Thread.currentThread();
+        if (!activityIdTlsMap.containsKey(thread.getId())) {
+            activityIdTlsMap.put(thread.getId(), new ActivityId(thread));
         }
-
-        return activityIdTlsMap.get(uniqueThreadId);
+
```
cheenamalhotra: Make sure the file is formatted correctly.

src/test/java/com/microsoft/sqlserver/jdbc/ActivityIDTest.java
```diff
@@ -0,0 +1,171 @@
...
+        String ActivityIDTraceOff = Util.ActivityIdTraceProperty + "=off";
+        try (InputStream is = new ByteArrayInputStream(ActivityIDTraceOff.getBytes());) {
+            LogManager lm = LogManager.getLogManager();
+            lm.readConfiguration(is);
+        }
+    }
+    
+    @BeforeAll
+    public static void testSetup() throws Exception {
+        String ActivityIDTraceOn = Util.ActivityIdTraceProperty + "=on";
```
cheenamalhotra: Please use proper way of turning ON and OFF Activity ID tracing, as an end user would do.
Also use proper naming conventions for String objects.

src/test/java/com/microsoft/sqlserver/jdbc/ActivityIDTest.java
```diff
@@ -0,0 +1,171 @@
...
+            assertEquals(0, ActivityCorrelator.getActivityIdTlsMap().size());
+        } catch (ComparisonFailure e) {
+            Set<Thread> threadSet = Thread.getAllStackTraces().keySet();
+            logger.fine("List of threads alive:");
+            for (Thread thread: threadSet) {
+                logger.fine(thread.toString());
+            }
+            logger.fine("List of entries in the ActivityID map:");
+            logger.fine(ActivityCorrelator.getActivityIdTlsMap().toString());
+            throw new Exception(e);
```
cheenamalhotra: Fail the test instead of throwing exception

src/test/java/com/microsoft/sqlserver/jdbc/ActivityIDTest.java
```diff
@@ -0,0 +1,171 @@
...
+            }
+        } catch (SQLException e) {
+            fail(e.toString());
+        }
+        
+        try {
+            assertEquals(0, ActivityCorrelator.getActivityIdTlsMap().size());
+        } catch (ComparisonFailure e) {
+            Set<Thread> threadSet = Thread.getAllStackTraces().keySet();
+            logger.fine("List of threads alive:");
```
cheenamalhotra: Logger is disabled when JUnit tests are running, writing log messages doesn't help if a failure occurs randomly. Use System.out.println() to print this information that needs to be captured immediately.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -2558,18 +2556,23 @@ void Prelogin(String serverName, int portNumber) throws SQLServerException {
...
         offset += conIdByteArray.length;
 
-        // copy ActivityId
-        System.arraycopy(actIdByteArray, 0, preloginRequest, offset, actIdByteArray.length);
-        offset += actIdByteArray.length;
-
-        long seqNum = activityId.getSequence();
-        Util.writeInt((int) seqNum, preloginRequest, offset);
-        offset += 4;
+        if (Util.IsActivityTraceOn()) {
```
cheenamalhotra: Rename the Util function to follow Java Naming conventions as `isActivityTraceOn`

src/test/java/com/microsoft/sqlserver/jdbc/ActivityIDTest.java
peterbae: This is the proper way of setting Activity Id tracing, where the user would read this value from a .properties file. I'm doing the same without creating an actual file for it.

src/test/java/com/microsoft/sqlserver/jdbc/ActivityIDTest.java
```diff
@@ -0,0 +1,167 @@
...
+        String activityIDTraceOff = Util.ActivityIdTraceProperty + "=off";
+        try (InputStream is = new ByteArrayInputStream(activityIDTraceOff.getBytes());) {
+            LogManager lm = LogManager.getLogManager();
+            lm.readConfiguration(is);
+        }
+    }
+
+    @BeforeAll
+    public static void testSetup() throws Exception {
+        String activityIDTraceOn = Util.ActivityIdTraceProperty + "=on";
```
cheenamalhotra: Could you please also rename Util > `ActivityIdTraceProperty` to `activityIdTraceProperty`

peterbae: done

src/main/java/com/microsoft/sqlserver/jdbc/ActivityCorrelator.java
```diff
@@ -18,25 +18,20 @@
...
 
     static void cleanupActivityId() {
-        // remove the ActivityId that belongs to this thread.
-        long uniqueThreadId = Thread.currentThread().getId();
-
-        if (activityIdTlsMap.containsKey(uniqueThreadId)) {
-            activityIdTlsMap.remove(uniqueThreadId);
-        }
+        // remove ActivityIds that belongs to this thread or no longer have an associated thread.
+        activityIdTlsMap.entrySet().removeIf(e -> e.getValue() == null || e.getValue().getThread() == null
```
lilgreenbird: just a minor nitpick to use null == blah instead

PR #1024, opened 2019-04-08
SQLServerDataTable.clear() now reset columnNames as well

lavinj87:

PR #1025, opened 2019-04-08
Add | Implement hashCode() and equals() APIs for SQLServerDataTable and SQLServerDataColumn

### LINKED ISSUE #1023: [FEATURE REQUEST] SQLServerDataTable and SQLServerDataColumn implementing equals
Given that `SQLServerDataTable` and `SQLServerDataColumn` are final classes (and this is perfectly fine) they can't be mocked with `Mockito` 1.19.

When writing unit tests that interact with those classes, you can build the expected `SQLServerDatatable` and `SQLServerDataColumn` instances and then compare them with the actual instances via `equals()`

Those classes do not implement the equals method, therefore their comparision is via identifical reference, not equality. 

An `equals()` (and therefore a `hashCode()` method for both classes, in order to be able to compare not same instances to be equal

None

## Additional context
The version of the driver is `7.0.0.jre8`

cheenamalhotra:
Resolves #1023 

Merged PR #1024 to move Test class to `tvp` package.


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDataColumn.java
```diff
@@ -45,4 +45,27 @@ public String getColumnName() {
...
+        return hash;
+    }
+
+    public boolean equals(Object object) {
+        if (this == object) {
+            return true;
+        }
+
+        if (null != object && object instanceof SQLServerDataColumn) {
+            if (hashCode() == ((SQLServerDataColumn) object).hashCode())
```
peterbae: For both of these equal methods, even if the hashCodes are equal, the actual objects may be different. After validating that hashCodes are equal, we should individually check that the fields are identical between the two objects to return true for equals.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDataColumn.java
```diff
@@ -45,4 +45,34 @@ public String getColumnName() {
...
+        int hash = 7;
+        hash = 31 * hash + javaSqlType;
+        hash = 31 * hash + precision;
+        hash = 31 * hash + scale;
+        hash = 31 * hash + numberOfDigitsIntegerPart;
+        hash = 31 * hash + (null != columnName ? columnName.hashCode() : 0);
+        return hash;
+    }
+
+    public boolean equals(Object object) {
```
peterbae: add override tag

PR #1027, opened 2019-04-09
Tests | Add logging stream handler for tests

lilgreenbird:
added stream handler logging option so tests can enable logging without having to wait for slow console or file logs


src/test/java/com/microsoft/sqlserver/testframework/AbstractTest.java
```diff
@@ -270,23 +272,32 @@ public static void invokeLogging() {
...
+
+            /*
+             * By default, Loggers also send their output to their parent logger. Typically the root Logger is
+             * configured with a set of Handlers that essentially act as default handlers for all loggers.
+             */
             Logger logger = Logger.getLogger(Constants.MSSQL_JDBC_PACKAGE);
             logger.setLevel(Level.FINEST);
+
+            // enable activity trace
+            TestUtils.setActivityTraceOn();
```
cheenamalhotra: Enabling Activity Trace for all tests could impact ActivityID test added in #1020 . Please update PR with all latest dev changes to verify.

src/main/java/com/microsoft/sqlserver/jdbc/Util.java
```diff
@@ -25,13 +25,14 @@
...
  *
  */
-
 final class Util {
     final static String SYSTEM_SPEC_VERSION = System.getProperty("java.specification.version");
     final static char[] hexChars = {'0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'A', 'B', 'C', 'D', 'E', 'F'};
     final static String WSIDNotAvailable = ""; // default string when WSID is not available
 
     final static String ActivityIdTraceProperty = "com.microsoft.sqlserver.jdbc.traceactivity";
+    static boolean activityIdTraceOn = "on"
```
peterbae: Static initialization of this variable will lock this to false, because the user hasn't provided the properties file yet. Even if they do, this variable won't be updated - we should leave isActivityTraceOn method as it is.

src/test/java/com/microsoft/sqlserver/jdbc/TestUtils.java
```diff
@@ -63,22 +63,23 @@
...
+     * SERVERPROPERTY('EngineEdition') can be used to determine whether the db server is SQL Azure. It should return 6
+     * for SQL Azure DW. This is more reliable than @@version or serverproperty('edition'). Reference:
+     * http://msdn.microsoft.com/en-us/library/ee336261.aspx
+     * 
+     * SERVERPROPERTY('EngineEdition') means Database Engine edition of the instance of SQL Server installed on the
+     * server.
      * 
-     * SERVERPROPERTY('EngineEdition') means
-     * Database Engine edition of the instance of SQL Server installed on the server.
+     * <pre>
```
peterbae: why do we have pre tags here?

lilgreenbird: as discussed, this is necessary to prefer formatter from making this unreadable

src/test/java/com/microsoft/sqlserver/testframework/AbstractTest.java
lilgreenbird: merged and tested, thx

src/test/java/com/microsoft/sqlserver/testframework/AbstractTest.java
```diff
@@ -241,35 +243,42 @@ public static String getConfiguredProperty(String key, String defaultValue) {
...
         try {
             if (Constants.LOGGING_HANDLER_CONSOLE.equalsIgnoreCase(loggingHandler)) {
                 handler = new ConsoleHandler();
+                handler.setFormatter(new SimpleFormatter());
             } else if (Constants.LOGGING_HANDLER_FILE.equalsIgnoreCase(loggingHandler)) {
                 handler = new FileHandler(Constants.DEFAULT_DRIVER_LOG);
+                handler.setFormatter(new SimpleFormatter());
                 System.out.println("Look for Driver.log file in your classpath for detail logs");
+            } else if (Constants.LOGGING_HANDLER_STREAM.equalsIgnoreCase(loggingHandler)) {
+                handler = new StreamHandler(new PrintStream(Constants.LOGGING_STREAM), new SimpleFormatter());
```
cheenamalhotra: You should store this stream in class as a class object, initialize it here and then add to StreamHandler.
Also close it safely after all tests are done in `@AfterAll` method.

Issue #1031, opened 2019-04-11
[QUESTION] geography not valid


vforchi:
## Question
I believe I am doing something wrong, but when I use the Geography class to create WKB I get an invalid geography.

```
byte[] wkb = Geography.point(10, 20, 104001).serialize()
assert wkb[5] == 0x0c
```

This returns false, because that value is 0x08, meaning that the geometry is not valid, if I am not mistaken. But according to the CLR Types specification the valid bit should always be set for version 1.

vforchi:
I have a similar problem with a polygon:
```
String wkt = "POLYGON((142.448916 -51.037471,142.41632 -52.52176,144.406309 -52.52273,144.37449 -51.038436,142.448916 -51.037471))"
byte[] wkb = Geometry.STGeomFromText(wkt, srid).serialize();
assert wkb[5] == 0x04;
```
Also, in this case wkb is 
```
41 96 01 00 01 00 05 00 00 00 d6 54 16 85 5d ce 61 40 32 c6 87 d9 cb 84 49 c0 77 15 52 7e 52 cd 61 40 34 2e 1c 8 c9 42 4a c0 41 62 bb 7b 0 d 62 40 ad 51 f d1 e8 42 4a c0 bb d5 73 d2 fb b 62 40 99 7e 89 78 eb 84 49 c0 d6 54 16 85 5d ce 61 40 32 c6 87 d9 cb 84 49 c0 01 00 00 00 02 00 00 00 00 01 00 00 00 ff ff ff ff 00 00 00 00 03
```
but I got the WKT from an existing geography in the DB, that has this representation:
```
41 96 01 00 01 04 05 00 00 00 32 c6 87 d9 cb 84 49 c0 d6 54 16 85 5d ce 61 40 34 2e 1c 08 c9 42 4a c0 77 15 52 7e 52 cd 61 40 ad 51 0f d1 e8 42 4a c0 41 62 bb 7b 00 0d 62 40 99 7e 89 78 eb 84 49 c0 bb d5 73 d2 fb 0b 62 40 32 c6 87 d9 cb 84 49 c0 d6 54 16 85 5d ce 61 40 01 00 00 00 02 00 00 00 00 01 00 00 00 ff ff ff ff 00 00 00 00 03
```
The length is in both cases 112, the difference is in the valid bit and in the data representation.

peterbae:
[...]

PR #1035, opened 2019-04-13
Fix | Mark spatial datatypes as valid

peterbae:
Addresses issue #1031 and #1038.

For #1031, I changed the isValid bit to be set to `true` by default, as it is stated in the CLR specs. Leaving this isValid bit to false makes some of the Spatial Datatype operations on the server's end fail.

For #1038, I've declared the default constructors as protected, to allow extension of Geometry and Geography classes.

I've also noticed that setObject() using Geography/Geometry weren't working as intended, so I'm fixing that as well.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/Geography.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/Geography.java
@@ -27,7 +27,7 @@ public class Geography extends SQLServerSpatialDatatype {
      * @throws SQLServerException
      *         if an exception occurs
      */
-    private Geography(String wkt, int srid) throws SQLServerException {
+    protected Geography(String wkt, int srid) throws SQLServerException {
         if (null == wkt || wkt.length() <= 0) {
             throwIllegalWKT();
         }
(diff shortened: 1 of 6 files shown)


peterbae:
[mssql-jdbc-7.3.1-SNAPSHOT-preview-PR1035.zip](https://github.com/Microsoft/mssql-jdbc/files/3075537/mssql-jdbc-7.3.1-SNAPSHOT-preview-PR1035.zip)

Attaching the driver from this PR for testing.

src/main/java/com/microsoft/sqlserver/jdbc/Geography.java
```diff
@@ -15,6 +15,8 @@
 
 public class Geography extends SQLServerSpatialDatatype {
 
+    protected Geography() {}
+
     /**
      * Private constructor used for creating a Geography object from WKT and Spatial Reference Identifier.
```
lilgreenbird: protected

cheenamalhotra: Protected is fine - came from other issue, but is this particular constructor needed now?

peterbae: Yes, if we don't specify protected here, the default constructor will have package private visibility, and it won't let other classes outside the package inherit this class. Therefore, we need this line.

PR #1042, opened 2019-04-17
Modify formatter for project

peterbae:
I checked "Never join lines" and unchecked "Remove blank lines" in comments section.


peterbae:
We have an alternative. Closing as not needed.

PR #1043, opened 2019-04-17
Maven | Add Shade plugin configuration to package the driver in an uber-jar

rene-ye:
Packages all the dependencies used by the driver.


ulvii:
PR needs to be updated to reflect [JRE12 changes](https://github.com/microsoft/mssql-jdbc/pull/1050).

PR #1045, opened 2019-04-23
Strictly name the MicroSoft SQL Driver for JLink and Distribution

GedMarc:
Strictly names the Microsoft SQL Driver for JLink and Distribution 

This allows the driver to be deployed with JLink, JPackage (JDK13) and enables production usage of the driver.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -87,7 +87,13 @@
 			<groupId>com.microsoft.rest</groupId>
 			<artifactId>client-runtime</artifactId>
 			<version>${rest.client.version}</version>
-			<optional>true</optional>
+            <exclusions>
+                <exclusion>
+                    <artifactId>okhttp-urlconnection</artifactId>
+                    <groupId>com.squareup.okhttp3</groupId>
+                </exclusion>
+            </exclusions>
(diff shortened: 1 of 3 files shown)


cheenamalhotra:
Hi @GedMarc 

I see the build has some issues, can you please fix them too. Also looks like we probably have to wait for Stable `moditect-maven-plugin` release to use here, have you considered alternatives to `moditect-maven-plugin`?

GedMarc:
Hi @cheenamalhotra 

Thanks for taking a look, I'll run through the build again and see if I've missed anything painfully obvious - 

Class building module-info files is still pretty new going around, there are a couple (literally 2) other options, however with JacksonJSON, Google Guice, JBoss Rest, Apache Commons, Classgraph and JWebMP all going Moditect (That I've found), it seems the most stable, and trustworthy; as well as proofed (ClassGraph is the Duke Choice Winner for last year and uses Moditect. I believe the developer is also on the Moditect team) solution.

GedMarc:
Actually because the profile 43 is compiling in JRE11, there is no point in using any module-info builder for the JRE8 build, I'll remove it!

PR #1046, opened 2019-04-24
Fix | Fix SharedTimer implementation to use class level lock for thread safety

### LINKED ISSUE #1040: Hikari + MSSQL 7.2.1 having issues
Hello All,

We have an spring boot application that is using HikariCP and mssql 6.x libraries and everything seems working fine.

We started moving to the latest version of mssql, which is 7.2.1 and started observing issues connecting to the database, once in a while under load test.

On the initial research, we found that the SharedTimer thingy that got introduced is killing us.

`2019-04-17_00:27:40.315 [42122f1e7f861f55:42122f1e7f861f55:0] [Root=1-5cb672fb-8a9936f01f1e83407ebf0024] [qtp1673605040-61] ERROR com.zaxxer.hikari.pool.HikariPool - HikariPool-278 - Exception during pool initialization.
java.util.concurrent.RejectedExecutionException: Task java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask@746dc536 rejected from java.util.concurrent.ScheduledThreadPoolExecutor@7dd4eba[Terminated, pool size = 0, active threads = 0, queued tasks = 0, completed tasks = 0]
	at java.util.concurrent.ThreadPoolExecutor$AbortPolicy.rejectedExecution(ThreadPoolExecutor.java:2063)
	at java.util.concurrent.ThreadPoolExecutor.reject(ThreadPoolExecutor.java:830)
	at java.util.concurrent.ScheduledThreadPoolExecutor.delayedExecute(ScheduledThreadPoolExecutor.java:326)
	at java.util.concurrent.ScheduledThreadPoolExecutor.schedule(ScheduledThreadPoolExecutor.java:533)
	at com.microsoft.sqlserver.jdbc.SharedTimer.schedule(SharedTimer.java:102)`

Any help on how to fix is highly appreciated.

cheenamalhotra:
Fixes #1040

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SharedTimer.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SharedTimer.java
@@ -63,8 +63,7 @@ class SharedTimer implements Serializable {
             if (refCount.get() <= 0) {
                 throw new IllegalStateException("removeRef() called more than actual references");
             }
-            refCount.getAndDecrement();
-            if (refCount.get() == 0) {
+            if (refCount.decrementAndGet() == 0) {
                 // Removed last reference so perform cleanup
                 executor.shutdownNow();
                 executor = null;
@@ -80,15 +79,15 @@ class SharedTimer implements Serializable {
      *
      * When the caller is finished with the SharedTimer it must be released via {@link#removeRef}
      */
-    public static synchronized SharedTimer getTimer() {
+    public static SharedTimer getTimer() {
         synchronized (lock) {
             if (instance == null) {
                 // No shared object exists so create a new one
                 instance = new SharedTimer();
             }
             instance.refCount.getAndIncrement();
+            return instance;
         }
-        return instance;
     }
 
     /**
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SharedTimer.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SharedTimer.java
@@ -19,10 +19,9 @@ class SharedTimer implements Serializable {
     private static final long serialVersionUID = -4069361613863955760L;
 
     static final String CORE_THREAD_PREFIX = "mssql-jdbc-shared-timer-core-";
-    private static final AtomicLong CORE_THREAD_COUNTER = new AtomicLong();
-    private static volatile SharedTimer instance;
-    private static Object lock = new Object();
 
+    private static final AtomicLong CORE_THREAD_COUNTER = new AtomicLong();
+    private static final Object lock = new Object();
     /**
      * Unique ID of this SharedTimer
      */
@@ -31,6 +30,8 @@ class SharedTimer implements Serializable {
      * Number of outstanding references to this SharedTimer
      */
     private final AtomicInteger refCount = new AtomicInteger();
+
+    private static volatile SharedTimer instance;
     private ScheduledThreadPoolExecutor executor;
 
     private SharedTimer() {
@@ -49,7 +50,7 @@ class SharedTimer implements Serializable {
     /**
      * @return Whether there is an instance of the SharedTimer currently allocated.
      */
-    static synchronized boolean isRunning() {
+    static boolean isRunning() {
         return instance != null;
     }
 
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SharedTimer.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SharedTimer.java
@@ -8,6 +8,7 @@ import java.io.Serializable;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.ScheduledThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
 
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/SharedTimer.java
```diff
@@ -56,16 +58,18 @@ static synchronized boolean isRunning() {
...
-            executor.shutdownNow();
-            executor = null;
-            instance = null;
+    public void removeRef() {
+        synchronized (lock) {
+            if (refCount.get() <= 0) {
+                throw new IllegalStateException("removeRef() called more than actual references");
+            }
+            refCount.getAndDecrement();
+            if (refCount.get() == 0) {
```
eugeniyk: you'd better to use result that was returned by `refCount.getAndDecrement()`

src/main/java/com/microsoft/sqlserver/jdbc/SharedTimer.java
```diff
@@ -77,11 +81,13 @@ public synchronized void removeRef() {
      * When the caller is finished with the SharedTimer it must be released via {@link#removeRef}
      */
     public static synchronized SharedTimer getTimer() {
```
sehrope: The `synchronized` on the static method is no longer required here as its being manually managed via the `lock` object.

src/main/java/com/microsoft/sqlserver/jdbc/SharedTimer.java
```diff
@@ -77,11 +81,13 @@ public synchronized void removeRef() {
...
-            instance = new SharedTimer();
+        synchronized (lock) {
+            if (instance == null) {
+                // No shared object exists so create a new one
+                instance = new SharedTimer();
+            }
+            instance.refCount.getAndIncrement();
         }
-        instance.refCount += 1;
         return instance;
```
sehrope: The `return instance;` should be moved up into the `synchronized (lock) {...}` block. Otherwise there's a potential race condition if the instance is cleared by a concurrent call to `removeRef()` that completes after this synchronized block completes but prior to this method returning the value of `instance`.

src/main/java/com/microsoft/sqlserver/jdbc/SharedTimer.java
```diff
@@ -16,10 +17,11 @@
      * Always update serialVersionUID when prompted
      */
     private static final long serialVersionUID = -4069361613863955760L;
-    
+
     static final String CORE_THREAD_PREFIX = "mssql-jdbc-shared-timer-core-";
     private static final AtomicLong CORE_THREAD_COUNTER = new AtomicLong();
-    private static SharedTimer instance;
+    private static volatile SharedTimer instance;
+    private static Object lock = new Object();
```
sehrope: This should be final as it must never change.

Could also reorder it above the SharedTimer so that the final fields come before all non-final fields.

sehrope:
The static `isRunning()` method should be updated to use the explicit lock object as well.

sehrope:
Actually maybe it doesn't need the explicit synchronized lock at all as it's accessing the volatile field.

The method level synchronized can be removed though as the only thing it would be synchronizing is concurrent callers to that one method.

src/main/java/com/microsoft/sqlserver/jdbc/SharedTimer.java
```diff
@@ -56,16 +59,17 @@ static synchronized boolean isRunning() {
...
-        refCount -= 1;
-        if (refCount == 0) {
-            // Removed last reference so perform cleanup
-            executor.shutdownNow();
-            executor = null;
-            instance = null;
+    public void removeRef() {
+        synchronized (lock) {
+            if (refCount.get() <= 0) {
+                throw new IllegalStateException("removeRef() called more than actual references");
```
ulvii: @cheenamalhotra , shouldn't this be a `SQLServerException` instead?

src/main/java/com/microsoft/sqlserver/jdbc/SharedTimer.java
```diff
@@ -56,16 +59,17 @@ static synchronized boolean isRunning() {
...
-        if (refCount <= 0) {
-            throw new IllegalStateException("removeRef() called more than actual references");
-        }
-        refCount -= 1;
-        if (refCount == 0) {
-            // Removed last reference so perform cleanup
-            executor.shutdownNow();
-            executor = null;
-            instance = null;
+    public void removeRef() {
```
ulvii: `removeRef()` doesn't have to be `public`.

src/main/java/com/microsoft/sqlserver/jdbc/SharedTimer.java
```diff
@@ -76,13 +80,15 @@ public synchronized void removeRef() {
      *
      * When the caller is finished with the SharedTimer it must be released via {@link#removeRef}
      */
-    public static synchronized SharedTimer getTimer() {
-        if (instance == null) {
-            // No shared object exists so create a new one
-            instance = new SharedTimer();
+    public static SharedTimer getTimer() {
```
ulvii: Please remove `public`.

src/main/java/com/microsoft/sqlserver/jdbc/SharedTimer.java
cheenamalhotra: The class 'SharedTimer' is a package level class, these methods are not accessible to public eitherway. I didn't modify any implementation so this stayed.

src/main/java/com/microsoft/sqlserver/jdbc/SharedTimer.java
cheenamalhotra: It's not SQL Server Exception or driver generated exception if this code is reached. This is driver's internal class and the APIs are called internally. So technically if this exception is thrown, it is an illegal state of the driver/application just like a "NullPointerException" that can occur anytime a null value is referenced.

ulvii: Is the error message meant to be localized?

cheenamalhotra: This is "non-expected" error as it's driver internal implementation. It's not meant to occur.
So no, we wouldn't need to localize it. If this was a driver thrown exception, we would have localized it when this was added.

PR #1049, opened 2019-04-25
Fix | Fix issues with Login Timeout not getting applied appropriately.

### LINKED ISSUE #1048: [BUG]loginTimeout in URL not valid if DriverManager.setLoginTimeout
## Driver version
6.2.2 

## SQL Server version
Any

## Client Operating System
Linux or Mac or Windows

## JAVA/JVM version
1.8

## Table schema
No need

## Problem description
When I set logintimeout in URL, I expect that the connection would timeout by the setting in URL. But it would timeout by the global loginTimeout setting. 

And the problem is caused by the loginTimeout property is overwritten in SQLServerDriver.parseAndMergeProperties
image

## Reproduction code
With the following code, the connection would timeout in 60 seconds not 5 seconds.
DriverManager.setLoginTimeout(60)
println(new Date().toString())
try {
    url = "jdbc:sqlserver://10.0.2.15;userName=sa;password=PASSW0RD;database=master;socketTimeout=10000;logintimeout=5"
    def sql = Sql.newInstance(url, "system", "123456", "com.microsoft.sqlserver.jdbc.SQLServerDriver")
}
catch (Exception e) {
    println(e.getMessage())
    println(new Date().toString())
}

cheenamalhotra:
Fixes #1048

### DIFF
--- a/src/test/java/com/microsoft/sqlserver/jdbc/connection/TimeoutTest.java
+++ b/src/test/java/com/microsoft/sqlserver/jdbc/connection/TimeoutTest.java
@@ -93,21 +93,21 @@ public class TimeoutTest extends AbstractTest {
     public void testDMLoginTimeoutNotApplied() {
         long timerEnd = 0;
         int timeout = 10;
+        try {
+            DriverManager.setLoginTimeout(timeout * 3); // 30 seconds
+            long timerStart = System.currentTimeMillis();
 
-        DriverManager.setLoginTimeout(timeout * 3); // 30 seconds
-        long timerStart = System.currentTimeMillis();
-
-        try (Connection con = PrepUtil
-                .getConnection("jdbc:sqlserver://" + randomServer + ";user=sa;password=pwd;loginTimeout=" + timeout)) {
-            fail(TestResource.getResource("R_shouldNotConnect"));
-        } catch (Exception e) {
-            assertTrue(e.getMessage().contains(TestResource.getResource("R_tcpipConnectionToHost")));
-            timerEnd = System.currentTimeMillis();
+            try (Connection con = PrepUtil.getConnection(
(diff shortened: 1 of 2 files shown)


src/test/java/com/microsoft/sqlserver/jdbc/connection/TimeoutTest.java
```diff
@@ -29,66 +30,124 @@
...
             timerEnd = System.currentTimeMillis();
         }
-        assertTrue(0 != timerEnd, TestResource.getResource("R_shouldNotConnect"));
 
-        long timeDiff = timerEnd - timerStart;
-        assertTrue(timeDiff > 14000);
+        verifyTimeout(timerEnd - timerStart, defaultTimeout);
+    }
+
+    private void verifyTimeout(long timeDiff, int timeout) {
```
rene-ye: Questionable boundaries for a "pass", should probably revisit this in the future.

src/test/java/com/microsoft/sqlserver/jdbc/connection/TimeoutTest.java
```diff
@@ -29,66 +30,124 @@
...
+                .getConnection("jdbc:sqlserver://" + randomServer + ";user=sa;password=pwd;loginTimeout=" + timeout)) {
+            fail(TestResource.getResource("R_shouldNotConnect"));
+        } catch (Exception e) {
+            assertTrue(e.getMessage().contains(TestResource.getResource("R_tcpipConnectionToHost")));
+            timerEnd = System.currentTimeMillis();
+        }
+
+        verifyTimeout(timerEnd - timerStart, timeout);
+
+        DriverManager.setLoginTimeout(0); // Default to 0 again
```
ulvii: Please add these 2 lines to a `finally` block to make sure DM timeout is set back to its default value even if the test fails.

PR #1050, opened 2019-04-26
Feature | Introduce Java 12 Support to driver and build additional "jre12" JAR

cheenamalhotra:
This PR adds support to build `jre12` Jar for the driver along with `jre11` and `jre8` with both Maven and Gradle build scripts.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -41,9 +41,11 @@
 
 	<properties>
 		<!-- Allowed values for excluded Groups here - - - - - - - - - - - - - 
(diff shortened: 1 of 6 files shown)


pom.xml
```diff
@@ -285,39 +256,40 @@
...
+				</plugins>
+			</build>
+		</profile>
+		<profile>
+			<id>jre12</id>
+			<activation>
+				<activeByDefault>true</activeByDefault>
+			</activation>
+			<build>
+				<finalName>${project.artifactId}-${project.version}.jre11-preview</finalName>
```
peterbae: jre12-preview

PR #1062, opened 2019-05-22
Fix | Unexpected Token 0x0 fix + additional check for columnName 

cheenamalhotra:
* Unexpected Token 0x0
An unexpected token 0x0 is received intermittently from SQL Server which causes intermittent exceptions. There is no documentation for when/why this token is sent/thrown by SQL Server in TDS documentation. The only possible workaround looks like we should stop streaming when this token is received, similar to `-1` token.

Putting up the PR for further extensive testing, triggering discussions for any known issues with this approach. If not, we should look forward to fix the behavior to avoid intermittent driver exceptions.

* Additional check for ColumnName
Change for Static Code Analysis complains, tests already cover this scenario.


cheenamalhotra:
Closing this for now.

Issue #1055, opened 2019-05-13
SQLServerResultSet.findColumn(String columnName) should cache result


stulscott:
The performance of retrieving a value from SQLServerResultSet by column name could be vastly improved by caching the result of the findColumn(columnName) method. Since this is such a simple to implement but hugely beneficial change I am curious if there is a reason why this has not yet been implemented. I am happy to create a patch if it is agreed by the community here that it is desirable to add this feature to the driver.

rene-ye:
Hi @stulscott, please feel free to create a PR against the repo and the team will review it in the near future.

rene-ye:
Hi @stulscott, #1066 attempts to address this issue, please take a look to see if it's within expectations or even feel free to review it.

stulscott:
> Hi @stulscott, #1066 attempts to address this issue, please take a look to see if it's within expectations or even feel free to review it.

#1066 appears to only affect SQLServerCallableStatement. SQLServerResultSet.findColumn() still iterates through all columns on every call.

rene-ye:
@stulscott, I took a look at the implementation for `findColumn()` in `SQLServerResultSet`, and came to the conclusion that implementing caching would not always improve performance, and will degrade performance in some cases. The implementation as of now just iterates through the list of columns and does a string compare via a getter. If I understand what you want, you'd like the driver to cache column names (either as it iterates or all at once), and perform some kind of quick lookup (HashMap) instead of iterating through the column list every time. This approach might sometimes be faster but it leads to some issues.

1. Requires HashMap as well as a TreeMap. HashMap is incapable of doing both a case sensitive AND a case insensitive lookup.
2. Requires more memory. If we were to implement both the maps, in worst case scenarios the driver would require 3x the memory it's currently using for column name storage.
3. The time it takes to hash/construct these maps may outweigh the current approach is findColumn() is not called many times.

There are also other open ended implementation decisions that impact the performance such as:

- When do we populate the maps? Do we do it immediately after we receive all columns or do we do it on the first findColumns() call?
- Do we populate the maps all at once or only the columns we've already iterated through? If not all at once, are cache misses going to significantly impact performance?

I'd like your opinion on the points above and on what solution you envisioned when creating the issue.

stulscott:
[...]

PR #1066, opened 2019-05-22
Performance | Improved performance of column name string lookups

rene-ye:
Addresses #1055.


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResultSet.java
```diff
@@ -675,14 +685,15 @@ public int findColumn(String columnName) throws SQLServerException {
         // Per JDBC spec, 27.3 "The driver will do a case-insensitive search for
         // columnName in it's attempt to map it to the column's index".
         // Use VM supplied String.equalsIgnoreCase to do the "case-insensitive search".
-        for (i = 0; i < columns.length; i++) {
-            if (columns[i].getColumnName().equalsIgnoreCase(columnName)) {
+        for (int i = 0; i < columns.length; i++) {
+            if (columns[i].getColumnName().equalsIgnoreCase(userProvidedColumnName)) {
```
cheenamalhotra: Do we use `equalsIgnoreCase` or `TreeMap` 🤔 
Use one approach in both classes.

rene-ye: This is separate from the treemap, we have to do it differently in Resultset/CallableStatement because one's an array of columns and the other just has some names.

cheenamalhotra: The String in `columnNames` HashMap contains names of column in both implementations, its just that here its fetched from column Array. The purpose of Map is same.

rene-ye: You can't do a String insensitive look up on a hashmap Key.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCallableStatement.java
```diff
@@ -1301,47 +1306,19 @@ private int findColumn(String columnName) throws SQLServerException {
...
-            for (i = 0; i < l; i++) {
-                String sParam = parameterNames.get(i);
-                sParam = sParam.substring(1, sParam.length());
-                if (sParam.equalsIgnoreCase(columnNameWithoutAtSign)) {
-                    matchPos = i;
-                    break;
-                }
-            }
+        // 2. Search using case-insensitive, non-locale specific (binary) compare last.        
+        Integer matchPos = parameterNames.get(columnNameWithSign);
```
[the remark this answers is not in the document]
rene-ye: Needs to be Integer. Maps all work with Integer so making it an int would just cause a lot of null checks and conditional statements to change it to -1 if null.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCallableStatement.java
```diff
@@ -1247,7 +1249,7 @@ public Ref getRef(String parameterName) throws SQLException {
      * @return the index
      */
     private int findColumn(String columnName) throws SQLServerException {
```
peterbae: columnName -> userProvidedColumnName to keep consistency with the other findColumn method

peterbae: nvm, its not a big deal

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCallableStatement.java
```diff
@@ -1281,11 +1283,15 @@ private int findColumn(String columnName) throws SQLServerException {
...
-                while (rs.next()) {
-                    String parameterName = rs.getString(4);
-                    parameterNames.add(parameterName.trim());
+                try (ResultSet rs = s.executeQueryInternal(metaQuery.toString())) {
+                    parameterNames = new HashMap<>();
+                    insensitiveParameterNames = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
+                    int columnIndex = 0;
+                    while (rs.next()) {
+                        String p = rs.getString(4);
+                        parameterNames.put(p, columnIndex);
```
peterbae: We used to trim `parameterName` before, why do we not call p.trim() here?

PR #1075, opened 2019-06-04
Performance | Code cleanup to improve performance 

ulvii:


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
```diff
@@ -614,7 +626,7 @@ public void writeToServer(ISQLServerBulkRecord sourceData) throws SQLServerExcep
      * Initializes the defaults for member variables that require it.
      */
     private void initializeDefaults() {
-        columnMappings = new LinkedList<>();
+        columnMappings = new ArrayList<>();
```
rene-ye: Removed the unused LinkedList import

dmitri-blinov:
Replacing `String.startsWith(x)` with `String.charAt(0) == x` may lead to an exception being thrown for the empty string "".

ulvii:
Hi @dmitri-blinov ,
Thank you for bringing this to our attention. I reverted the changes to not introduce a regression.

PR #1095, opened 2019-06-24
Update SQLServerConnection to store password hash for NTLM

lilgreenbird:


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -3691,11 +3693,15 @@ private void logon(LogonCommand command) throws SQLServerException {
                             currentConnectPlaceHolder.getPortNumber());
                 }
             } else if (ntlmAuthentication) {
+                if (null == ntlmPasswordHash) {
+                    ntlmPasswordHash = NTLMAuthentication.getNtlmPasswordHash(
+                            activeConnectionProperties.getProperty(SQLServerDriverStringProperty.PASSWORD.toString()));
```
ulvii: Remove the password from `activeConnectionProperties` here.

src/main/java/com/microsoft/sqlserver/jdbc/NTLMAuthentication.java
```diff
@@ -890,4 +890,8 @@ private int getByteArrayLength(byte[] arr) {
 
         return msg;
     }
+
+    public static byte[] getNtlmPasswordHash(String password) {
+        return null != password ? md4(unicode(password)) : null;
```
rene-ye: Is having a `null` password even allowed. Should just throw here if it isn't.

PR #1100, opened 2019-06-30
Tests | Improve assertions in ComparisonUtil

harawata:
Failure message should always include the expected/actual values.
Currently, they are added manually in some cases, but it seems to make more sense to use `assertEquals`.

BTW, I noticed that the assertion comparing `java.sql.Date`s only compares `DAY_OF_MONTH`.
Is it intentional?
It was committed in https://github.com/microsoft/mssql-jdbc/commit/115336296d70eb49b6e792aabe621ab8b93dee51#diff-d4cfc7fdd584aa2ec6cec3e2fcaef964R146 .


ulvii:
Hi @harawata,

Thanks for the contribution, the team will review the PR.

rene-ye:
The comparison isn't intentional, we will address this in a later PR.

PR #1102, opened 2019-07-02
Fix | SonarQube issues

rene-ye:
Addresses issues identified by static analysis


rene-ye:
Closing PR as CI's seem to be bugged

PR #1120, opened 2019-08-01
Revert | Code changes for 7.4.1 Hotfix

rene-ye:


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDataTable.java
```diff
@@ -54,7 +53,6 @@ public synchronized void clear() {
         rowCount = 0;
         columnCount = 0;
         columnMetadata.clear();
-        columnNames.clear();
```
ulvii: This can stay I think

ulvii:
Duplicate of #1121

PR #1122, opened 2019-08-02
Add | Default to null if val is null

rene-ye:
Compensating for a change meant to be pushed into a later PR but was reverted with the hotfix.


cogman:
I'm not sure this change is necessary.

Java specifies that all values in an array of Object are set to null by default.  This value couldn't be set to anything else.  Because rowValues isn't shared/copied/etc.  The value would be null if nothing sets it.

rene-ye:
Was kept to keep code aligned. This line was added in a later PR, but was removed after changing the buffer to an arraylist.

cheenamalhotra:
This is needed change as there is no "default" value on that index. The change comes from the PR #990 where null value assignment was separated from the original block.

PR #1133, opened 2019-08-29
Fix issue with truststore password being removed too early for XA transaction

peterbae:
The creation of an SQLServerXAConnection requires two connections to be made, one for the transaction and one for the control connection. However, after the creation of the transaction connection, truststore password is removed (for security purposes), and this prevents the control connection from being established successfully. The problem occurs if the truststore is password protected.

My solution temporarily re-adds the truststore password to the driver's list of properties between the time when a transaction connection is made and the control connection is made (the password will still be removed after the control connection is established). I've tested against both cases where the user provides the truststore password through the DataSource or from the connection property.


peterbae:
Adding the driver jar with the solution (Java 8/Java 12) for testing.

[github-PR-1133-fix.zip](https://github.com/microsoft/mssql-jdbc/files/3557106/github-PR-1133-fix.zip)

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerXAConnection.java
```diff
@@ -52,6 +52,20 @@
...
+        String trustStorePassword = ds.getTrustStorePassword();
+        if (null == trustStorePassword) {
+                // trustStorePassword can either come from the connection string or added via SQLServerXADataSource::setTrustStorePassword.
+                // if trustStorePassword is null at this point, then check the connection string.
+                Properties urlProps = Util.parseUrl(ds.getURL(), xaLogger);
+                trustStorePassword = urlProps.getProperty(SQLServerDriverStringProperty.TRUST_STORE_PASSWORD.toString());
+        }
+
+        // if trustStorePassword is still null, it wasn't provided. Do not set the property as null to avoid NPE.
+        if (null != trustStorePassword) {
```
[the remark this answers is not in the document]
lilgreenbird: general comment (not specific to this line - add test?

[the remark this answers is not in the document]
peterbae: Adding meaningful tests here would mean we need the test to generate a password protected truststore beforehand / remove it afterwards, and I think that'll be more trouble than what it's worth.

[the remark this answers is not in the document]
lilgreenbird: but this bug occurred because we didn't have a test case for it, so I think it's worth it ;)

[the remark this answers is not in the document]
peterbae: This is a really straightforward case that either works or shouldn't work, I don't think adding a test case here would be of much value to the functionality of the driver.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerXAConnection.java
```diff
@@ -52,6 +52,20 @@
             controlConnectionProperties.setProperty(SQLServerDriverStringProperty.PASSWORD.toString(), pwd);
         }
 
+        // Add truststore password property for creating the control connection. This will be removed again
```
ulvii: Please format the files

peterbae: Thanks, done.

src/test/java/com/microsoft/sqlserver/jdbc/connection/XADataSourceTest.java
```diff
@@ -0,0 +1,94 @@
...
+     * @throws Exception
+     */
+    @Test
+    public void testPKCS12() throws Exception {
+        SQLServerXADataSource ds = new SQLServerXADataSource();
+
+        // populate certificates arraylist with certificates
+        // Only re-populate the truststore if need arises in the future.
+
+        // populateCertificates();
```
ulvii: Needs to be cleaned up

src/test/java/com/microsoft/sqlserver/jdbc/connection/XADataSourceTest.java
```diff
@@ -0,0 +1,44 @@
...
+
+import com.microsoft.sqlserver.jdbc.SQLServerXADataSource;
+import com.microsoft.sqlserver.testframework.AbstractTest;
+import com.microsoft.sqlserver.testframework.Constants;
+
+
+@RunWith(JUnitPlatform.class)
+@Tag(Constants.reqExternalSetup)
+public class XADataSourceTest extends AbstractTest {
+    private static String connectionUrlSSL = connectionString + "encrypt=true;trustServerCertificate=false;";
```
ulvii:
```suggestion
    private static String connectionUrlSSL = connectionString + ";encrypt=true;trustServerCertificate=false;";
```

src/test/java/com/microsoft/sqlserver/testframework/Constants.java
```diff
@@ -25,6 +25,7 @@ private Constants() {}
      * xAzureSQLDW - - - - For tests not compatible with Azure Data Warehouse
      * xAzureSQLMI - - - - For tests not compatible with Azure SQL Managed Instance
      * NTLM  - - - - - - - For NTLM tests
+     * XA  - - - - - - - - For XA tests
```
ulvii: Tag needs to be renamed here too.

Issue #1136, opened 2019-08-30
[BUG] getColumns not escaping Wildcards


Springrbua:
## Driver version
Version 7.3.1 or later

## SQL Server version
SQL Server 2017

## Client Operating System
Windows 10 x64

## JAVA/JVM version
OpenJDK 1.8.0_202

## Problem description
1. Expected behaviour:
I have a table with a column `ID_Article`.
When calling `getColumns` with `ID\_Article` as `col` Parameter, it should return the description of table column `ID_Article` .
2. Actual behaviour:
Since version 7.3.1 it does not seem to find the column anymore, as it does not escape the `_` correctly.

3. Error message/stack trace:
No error is thrown, it just does not find the column.

4. Any other details that can be helpful:
The method [`EscapeIDName`](https://github.com/microsoft/mssql-jdbc/blob/dev/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java#L562) states, that `\_n` will be mapped to `'[_]n'` as the SQL-Server expects this syntax. However, this mapping does not seem to occur anymore.  
The bug was probably introduced by #1016.
Also #1134 might be related to this.

peterbae:
Hi @Springrbua, thanks for contacting us. I've identified the issue and I'm testing my fix. As soon as the internal testings are done, I'll create a PR and provide you with a custom jar that contains my fix.

peterbae:
Hi @Springrbua, @mprins, I've made a PR that fixes this issue. I've attached the driver jars with the fix in them (for Java 8/12), so if you have time, please try these and confirm if the fix is good. Thanks!

[github-1136-fix.zip](https://github.com/microsoft/mssql-jdbc/files/3570786/github-1136-fix.zip)

Springrbua:
@peterbae Thank you very much, the patched driver fixed the issue for me.

mprins:
I'm out-of-office, so won't be able to test until (maybe) friday

dqmdev:
When will MS release the build to customers?

In other words, when will the successor to 7.4.1 be announced at 

https://techcommunity.microsoft.com/t5/SQL-Server/bg-p/SQLServer/label-name/SQLServerDrivers

peterbae:
We're aiming to release the next preview version of the driver by the end of this month, and it will include this fix.

mprins:
[...]

PR #1138, opened 2019-09-03
Fix SQLServerDatabaseMetada.getColumns not escaping wildcard

peterbae:
Fixes issue #1136. Added extra test that tests this case as well.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
@@ -677,11 +677,11 @@ public final class SQLServerDatabaseMetaData implements java.sql.DatabaseMetaDat
              */
             try (PreparedStatement storedProcPstmt = this.connection
                     .prepareStatement("EXEC sp_columns_100 ?,?,?,?,?,?;")) {
-                storedProcPstmt.setString(1, (null != table && !table.isEmpty()) ? table : "%");
-                storedProcPstmt.setString(2, (null != schema && !schema.isEmpty()) ? schema : "%");
+                storedProcPstmt.setString(1, (null != table && !table.isEmpty()) ? EscapeIDName(table) : "%");
+                storedProcPstmt.setString(2, (null != schema && !schema.isEmpty()) ? EscapeIDName(schema) : "%");
                 storedProcPstmt.setString(3,
                         (null != catalog && !catalog.isEmpty()) ? catalog : this.connection.getCatalog());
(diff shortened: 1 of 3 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
```diff
@@ -638,10 +638,10 @@ private static String EscapeIDName(String inID) throws SQLServerException {
             PreparedStatement pstmt = (SQLServerPreparedStatement) this.connection.prepareStatement(spColumnsSql);
             pstmt.closeOnCompletion();
             try {
-                pstmt.setString(1, (null != table && !table.isEmpty()) ? table : "%");
-                pstmt.setString(2, (null != schema && !schema.isEmpty()) ? schema : "%");
+                pstmt.setString(1, (null != table && !table.isEmpty()) ? EscapeIDName(table) : "%");
+                pstmt.setString(2, (null != schema && !schema.isEmpty()) ? EscapeIDName(schema) : "%");
                 pstmt.setString(3, (null != catalog && !catalog.isEmpty()) ? catalog : this.connection.getCatalog());
-                pstmt.setString(4, (null != col && !col.isEmpty()) ? col : "%");
+                pstmt.setString(4, (null != col && !col.isEmpty()) ? EscapeIDName(col) : "%");
```
rene-ye: There is an `else` block below that handles this logic with Azure DW. That section needs similar changes.

peterbae: Thanks. Done

src/test/java/com/microsoft/sqlserver/jdbc/databasemetadata/DatabaseMetaDataTest.java
```diff
@@ -592,7 +604,7 @@ public void testGetMaxConnections() throws SQLException {
     public static void setupTable() throws SQLException {
         try (Statement stmt = connection.createStatement()) {
             stmt.execute("CREATE TABLE " + AbstractSQLGenerator.escapeIdentifier(tableName)
-                    + " (col1 int NOT NULL, col2 varchar(200), col3 decimal(15,2))");
+                    + " (col_1 int NOT NULL, col_2 varchar(200), col_3 decimal(15,2))");
```
lilgreenbird: should we add more test cases other than just "_"?

peterbae: I added more test cases.

PR #1146, opened 2019-09-12
Add | Implement hashCode() and equals() APIs for SQLServerDataTable and SQLServerDataColumn

ulvii:
Re-introduces `hashCode()/equals()` APIs for SQLServerTable and SQLServerDataColumn.
Created on top of #1025, instead of changing return type of `ISQLServerBulkRecord.getRowData()` to `List<Object>`, the PR implements custom equals()/hashCode() methods for `Map<Integer,Object[]>` to keep the driver backward compatible. Fix for #1023


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDataTable.java
```diff
@@ -334,4 +335,75 @@ public String getTvpName() {
...
+        }
+        int h = 0;
+        for (Entry<Integer, Object[]> entry : rows.entrySet()) {
+            h += entry.getKey() ^ Arrays.hashCode(entry.getValue());
+        }
+        return h;
+    }
+
+    private boolean compareRows(Map<Integer, Object[]> otherRows) {
+        if (rows == otherRows)
```
peterbae: If statements should have {}s even if it has one line

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDataTable.java
```diff
@@ -334,4 +335,75 @@ public String getTvpName() {
...
+                Object[] value = e.getValue();
+                if (value == null) {
+                    if (!(otherRows.get(key) == null && otherRows.containsKey(key)))
+                        return false;
+                } else {
+                    if (!Arrays.equals(value, otherRows.get(key)))
+                        return false;
+                }
+            }
+        } catch (ClassCastException unused) {
```
peterbae: I don't think this try block is necessary, as we're only catching runtime exceptions here.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDataTable.java
```diff
@@ -334,4 +335,75 @@ public String getTvpName() {
...
+        if (rows == otherRows)
+            return true;
+        if (rows.size() != otherRows.size())
+            return false;
+        try {
+            for (Entry<Integer, Object[]> e : rows.entrySet()) {
+                Integer key = e.getKey();
+                Object[] value = e.getValue();
+                if (value == null) {
+                    if (!(otherRows.get(key) == null && otherRows.containsKey(key)))
```
peterbae: It would be more performant if `otherRows.containsKey(key)` preceded the check for `otherRows.get(key)`

src/test/java/com/microsoft/sqlserver/jdbc/tvp/SQLServerDataTableTest.java
```diff
@@ -0,0 +1,98 @@
+package com.microsoft.sqlserver.jdbc.tvp;
```
peterbae: We need our MIT licensing header here

peterbae:
Other than the comments I put, the PR looks good to me.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDataTable.java
ulvii: `get()` and `containsKey()` have the same performance, so this should be fine.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDataTable.java
```diff
@@ -334,4 +335,79 @@ public String getTvpName() {
...
+        if (rows.size() != otherRows.size()) {
+            return false;
+        }
+        try {
+            for (Entry<Integer, Object[]> e : rows.entrySet()) {
+                Integer key = e.getKey();
+                Object[] value = e.getValue();
+                if (value == null) {
+                    if (!(otherRows.get(key) == null && otherRows.containsKey(key))) {
+                        return false;
```
lilgreenbird: null == blah

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDataTable.java
```diff
@@ -334,4 +335,79 @@ public String getTvpName() {
...
+            return true;
+        }
+        if (rows.size() != otherRows.size()) {
+            return false;
+        }
+        try {
+            for (Entry<Integer, Object[]> e : rows.entrySet()) {
+                Integer key = e.getKey();
+                Object[] value = e.getValue();
+                if (value == null) {
```
lilgreenbird: null == value

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDataColumn.java
```diff
@@ -45,4 +45,35 @@ public String getColumnName() {
...
+    public boolean equals(Object object) {
+        if (this == object) {
+            return true;
+        }
+
+        if (null != object && object.getClass() == SQLServerDataColumn.class) {
+            SQLServerDataColumn aSQLServerDataColumn = (SQLServerDataColumn) object;
+            if (hashCode() == aSQLServerDataColumn.hashCode()) {
+                // Compare objects to avoid collision
+                return ((null == columnName && null == aSQLServerDataColumn.columnName
```
TheConstructor: This will probably throw a NPE when `this.columnName` is null, but `aSQLServerDataColumn.columnName` is not null. Better would be `java.util.Objects.equals(columnName, aSQLServerDataColumn.columnName)` or `(columnName == null ? aSQLServerDataColumn.columnName == null : columnName.equals(aSQLServerDataColumn.columnName))`

ulvii: Hi @TheConstructor,
Good catch 👍 I created #1168  with your suggestion.

Issue #1112, opened 2019-07-23
[FEATURE REQUEST] support @fUsePattern paramter for sp_sproc_columns


hyparki:
## Is your feature request related to a problem? If so, please give a short summary of the problem and how the feature would resolve it
Yes, we use SQLServerCallableStatement.java class on JPA to SQL Server and we got performance issue related with sp_sproc_columns.
Our most of user defined procedure name includes underscore wildcard(ex, up_procname_purpose..), so sp_sproc_columns use pattern search. If we can use @fUsePattern=0, the performance will be improve.
The default value of @fUsePattern is 1.
In general, there are 100,000 batch requests/secs on the customer's SQL Server, which have a significant performance impact.

## Describe the preferred solution

I hope that it can be to use @fUsePattern parameter in this section on the SQLServerCallableStatement.java.

                ThreePartName threePartName = ThreePartName.parse(procedureName);
                StringBuilder metaQuery = new StringBuilder("exec sp_sproc_columns ");
                if (null != threePartName.getDatabasePart()) {
                    metaQuery.append("@procedure_qualifier=");
                    metaQuery.append(threePartName.getDatabasePart());
                    metaQuery.append(", ");
                }
                if (null != threePartName.getOwnerPart()) {
                    metaQuery.append("@procedure_owner=");
                    metaQuery.append(threePartName.getOwnerPart());
                    metaQuery.append(", ");
                }
                if (null != threePartName.getProcedurePart()) {
                    // we should always have a procedure name part
                    metaQuery.append("@procedure_name=");
                    metaQuery.append(threePartName.getProcedurePart());
                    metaQuery.append(" , @ODBCVer=3");

## Additional context

This is comparison result of io & time statistics according to use @fUsePattern paramter.
The result 234ms vs 141ms may not seems important by a difference of viewpoint.
However I hope you consider that the performance impact can be growing up if this type of requests are repeated.

 EXEC SYS.SP_SPROC_COLUMNS @procedure_name ='testproc123456789123456789123456789_10143', 
           @procedure_owner ='dbo', @procedure_qualifier= 'testdb11', **@fUsePattern=1**

                Table 'Worktable'. Scan count 0, logical reads 0, physical reads 0, read-ahead reads 0, lob logical reads 0, lob physical reads 0, lob read-ahead reads 0.
[...]

PR #1149, opened 2019-09-18
Performance | Disabled pattern matching when using CallableStatements and SQLServerParameterMetaData

ulvii:
Fix for #1112.
The driver executes [`sp_sproc_columns`](https://docs.microsoft.com/en-us/sql/relational-databases/system-stored-procedures/sp-sproc-columns-transact-sql?view=sql-server-2017) when using  CallableStatements with named parameters. `sp_sproc_columns` has pattern matching enabled by default, which makes a big difference when an application uses large amount of procedures with names that contain [wildcard characters](https://docs.microsoft.com/en-us/sql/t-sql/language-elements/like-transact-sql?view=sql-server-2017#arguments). See #1112 for more details. The PR disables pattern matching by default.


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCallableStatement.java
```diff
@@ -1279,7 +1279,7 @@ private int findColumn(String columnName) throws SQLServerException {
                     // we should always have a procedure name part
                     metaQuery.append("@procedure_name=");
                     metaQuery.append(threePartName.getProcedurePart());
-                    metaQuery.append(" , @ODBCVer=3");
+                    metaQuery.append(" , @ODBCVer=3, @fUsePattern=0");
```
peterbae: Do you also want to consider applying the same change to SQLServerParameterMetadata's constructor? It also has sp_sproc_columns and won't have to use pattern matching.

ulvii: Done.

Issue #1156, opened 2019-10-09
[QUESTION] Make internal model public for SQLServerSpatialDatatype class


MrAolen:
## Question

I am working on a project where I need to translate a `Geometry` object into another spatial class that implement a standard other than WKT/WKB.

I created a class that inherits from `Geometry` in the same package as the `Geometry` object. and I can access to `figureList` field. But the mssql-jdbc jar is signed, I got an error. Is that possible to retrieve figures information with an another way ?

peterbae:
Hi @MrAolen, I've attached my custom preview driver jars to this comment (Java 8/13 builds) that contain the following public getter methods:

SQLServerSpatialDatatype.getPointList()
SQLServerSpatialDatatype.getFigureList()
SQLServerSpatialDatatype.getShapeList()
SQLServerSpatialDatatype.getSegmentList()

Please let me know if this is what you needed, in which case, I'll go ahead and open a PR.

[github-1156.zip](https://github.com/microsoft/mssql-jdbc/files/3708532/github-1156.zip)

MrAolen:
Hello @peterbae, thanks for the fast answer

Yeah it can be useful but the problem is `Point`, `Figure`, `Shape`and `Segment`class are package private. So I think we can put them in public visibility, right?

peterbae:
Hi @MrAolen, those fields are set once the Geometry object is created, and making them public will make them vulnerable to illegal modification (as is the point with any non-public visibility). I've provided the getters instead of changing the fields' visibility because I don't think the Geometry object should be modified once it's been created.

MrAolen:
Yeah I understand, but when I try to use the getter like this:`Geometry.STGeomFromWKB(bytes).getPointList().get(0).getX()`, I got the following error : `: error: Point.getX() is defined in an inaccessible class or interface`.

peterbae:
Hi @MrAolen, the public getters I've provided will let you see the internal values of the spatial datatypes, but it won't let you retrieve the inner classes (point, segment, etc) programmatically since they're internal classes. I don't think we intend to make these internal parts of spatial datatypes as part of our public API, so if you want to be able to manipulate them yourself, you can fork your own branch of the driver and make changes yourself (make the ArrayLists public, extract the internal classes into another file and make them public, etc).

MrAolen:
[...]

PR #1169, opened 2019-10-23
Make internal model public for SQLServerSpatialDatatype class

peterbae:
Fixes issue #1156. Added getters and made internal classes public.


src/main/java/com/microsoft/sqlserver/jdbc/Figure.java
```diff
@@ -0,0 +1,48 @@
...
+ * available under the terms of the MIT License. See the LICENSE file in the project root for more information.
+ */
+
+package com.microsoft.sqlserver.jdbc;
+
+/**
+ * Represents the internal makings of a Figure.
+ *
+ */
+public class Figure {
```
rene-ye: Do we want to rename these classes to something like `SQLServerFigure` to maintain consistency.

ulvii: These classes don't seem to belong to microsoft.sql package. I would suggest to create a `com.microsoft.sqlserver.jdbc.spatialdatatypes` package and move **all** related classes under that package. See `com.microsoft.sqlserver.jdbc.dataclassification` for reference.

Issue #1135, opened 2019-08-30
[BUG] MSI authentication fails due to expiry date format mismatch


yjo:
## Driver version
7.4.1

## Client details
Running on an Azure App Server Java 11 / Linux instance in the UK west region
OS: Alpine 3.8 Docker container
JVM: OpenJDK Runtime Environment 19.1-(Zulu-11.29+3-linux-musl-x64)-Microsoft-Azure-restricted (build 11.0.2+7-LTS)

## Problem description
1. Configure an App Services resource (Java 11 in "UK West" region) and Azure SQL DB for authentication using MSI, following instructions [here](https://docs.microsoft.com/en-us/sql/connect/jdbc/connecting-using-azure-active-directory-authentication?view=sql-server-2017#connecting-using-activedirectorymsi-authentication-mode)
2. Deploy code that attempts MSI authentication, e.g.:
```java
//...
public static void main(String[] args) throws Exception {
    SQLServerDataSource ds = new SQLServerDataSource();
    ds.setServerName("my-db-instance.database.windows.net");
    ds.setDatabaseName("my-db-name");
    ds.setAuthentication("ActiveDirectoryMSI");
    Connection connection = ds.getConnection(); 
}
//...
```
3. The connection fails, with an opaque "MSI Token failure: Failed to acquire token from MSI Endpoint" error.

Unfortunately there's no stack trace or log output that shows the underlying problem (even at `java.util.logging.Level.ALL`), but by retrying with a patched version of the driver with extra logging I was able to narrow the issue down to [SQLServerConnection.java:4376](https://github.com/microsoft/mssql-jdbc/blob/v7.4.1/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java#L4376) which throws a `java.text.ParseException` when attempting to parse the token expiry timestamp.  The driver implementation is looking for a 12hr time format - "M/d/yyyy h:mm:ss a X", but the MSI API server (for my App Service instance in UK West at least) is returning a 24hr format, e.g. 
`"08/30/2019 01:08:02 +00:00"`.

It's not clear to me whether the issue is with the MSI server (possibly just in some regions?) or the driver.  Even if the issue is with the MSI server (in which case I'd appreciate advice on how to raise such an issue), I'd suggest it's probably a bug that the relevant exception details are swallowed by the driver without any logging.  I've attached the full log traces from the unmodified 7.4.1 driver from the start of the MSI request so you can see what I mean.
[...]

PR #1177, opened 2019-11-01
Add fallback to 24h time format when MSI authentication fails due to date format mismatch

walk0r:
Addressing this bug - https://github.com/microsoft/mssql-jdbc/issues/1135
Proper solution for the problem would be to use unix timestamps instead of some locale-dependent string representation of date for expires_on field, but since it would be breaking change for existing users, I propose to add fallback logic to allow MSI work on Linux machines. Tests include actual responses from MSI endpoints.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
@@ -34,6 +34,7 @@ import java.sql.Savepoint;
 import java.sql.Statement;
 import java.text.DateFormat;
 import java.text.MessageFormat;
+import java.text.ParseException;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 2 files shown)


ulvii:
Hi @walk0r ,
Thank you for the contribution.
We are currently waiting for Azure side changes that might affect the behavior of the driver, we will revisit your PR once the changes are official.

peterbae:
Hi @walk0r, thanks for your contribution. I've contacted the Azure team and confirmed that the String format for MSI expiry time being different across region/Azure Web App platform is a bug on Azure's side, and no changes should be necessary from the driver's end once the fix is merged in around Feb/Mar timeline. In the meantime, I've provided our driver with a temporary fix for this solution in the original issue #1135. I'll keep this PR open for now, and I will look to close it once the Azure fix is deployed and confirms that the driver doesn't need any code changes.

walk0r:
@peterbae sure, I'm ok with it.

Issue #1170, opened 2019-10-24
[QUESTION] AuthenticationJNI auth DLL loading


Pryanic:
## Relevant Issues and Pull Requests

I found in code of driver interesting class AuthenticationJNI and this class contains 

```
 String libName = "sqljdbc_auth";
 System.loadLibrary(libName);
```
In my application I should load drivers dynamically and I use few version of ms sql jdbc drivers, when I try to connect to MS Sql via Windows Authentication   all work fine, but when I change my driver I get the following exception 
 Native Library D:\sqljdbc_auth.dll already loaded in another classloader. 
How can I fix it?

peterbae:
Hi @Pryanic, if you're using Tomcat, I recommend taking a look at this page:

[I'm encountering classloader problems when using JNI under Tomcat](https://cwiki.apache.org/confluence/display/tomcat/HowTo#HowTo-I'mencounteringclassloaderproblemswhenusingJNIunderTomcat)

If this is not applicable to you, then I suggest sticking to one version of MSSQL JDBC driver. Let me know if this solves your problem.

Pryanic:
@peterbae  Thank you for your response. Unfortunately I don't use tomcat and sticking to one version doesn't help me. Each user of my application has own driver which loads via class loader for communication with Ms Sql server, so can we try to find other solution. Is it sqljdbc_auth thread safe?

Pryanic:
I think  we can add small check and it helps me:
`

`

peterbae:
Could you provide some reproduction code as to how your users are loading different versions of our driver within the same JVM context?

Pryanic:
I think I can make small prototype for you. 

```
public Driver getDriver(DriverInformation info){
     URL[] urls = getUrlsByDriverVersion(info.getDriverVersion());
     URLClassLoader loader =new URLClassLoader(urls, 
     Thread.currentThread().getContextClassLoader()); 
     return (Driver) loader.loadClass(info.getMainDriverClass()).newInstance();
}
```

peterbae:
I'm not sure about the context of DriverInformation, but I don't think it's necessary for users to load the driver versions themselves. Instead, I suggest the below code:

```
Class.forName("com.microsoft.sqlserver.jdbc.SQLServerDriver");
Connection con = DriverManager.getConnection(connectionUrl);
```

[...]

Issue #907, opened 2018-12-08
Please publish Maven artifacts for the x86 and x64 auth DLLs


bturner:
Thanks for publishing Maven artifacts for the SQL Server JDBC driver!

In Atlassian Bitbucket Server, we bundle the JDBC driver jar, and we also bundle the x86 and x64 versions of `sqljdbc_auth.dll` to facilitate using integrated authentication on Windows. Prior to Microsoft publishing their own jars for the JDBC driver we would manually publish them to our internal Maven repository, along with the DLLs. Now we can use the Microsoft-published jar for the JDBC driver, but we still need to manually publish the auth DLLs.

We use classifiers to differentiate the x86 and x64 DLLs, so our dependencies look like this:
```xml
<dependency>
    <groupId>com.microsoft.sqlserver</groupId>
    <artifactId>mssql-jdbc_auth</artifactId>
    <classifier>x64</classifier>
    <type>dll</type>
</dependency>
<dependency>
    <groupId>com.microsoft.sqlserver</groupId>
    <artifactId>mssql-jdbc_auth</artifactId>
    <classifier>x86</classifier>
    <type>dll</type>
</dependency>
```

Is there any chance Microsoft could start publishing the auth DLLs (and maybe the `sqljdbc_xa` DLLs for completeness, though we don't need those)? I assume the DLLs change from time to time, and that would imply it's important to use the DLL version that matches the JDBC driver version. (That's why we package the DLLs; that way we're certain the DLL and driver match.)

Thanks for your consideration!

David-Engel:
Thanks for the suggestion. That's actually a good idea. They do not change frequently, although there are occasional bug fixes. I'll have to see if there are any restrictions on the legal side to publishing them on Maven since they are not included under the MIT license that this repo is under. We will discuss it with the team.

Fuerte1:
Hello, we would also like to see sqljdbc_auth.dll in Maven, both x86 and x64.

ulvii:
Hi everyone,
We are working on publishing the DLLs on Maven, which will automatically rename the DLLs to also contain version and bitness information. For example, the definitions below would generate mssql-jdbc_auth-8.1.1.x64-preview.dll and mssql-jdbc_auth-8.1.1.x86-preview.dll respectively.
```
<dependency>
  <groupId>com.microsoft.sqlserver</groupId>
  <artifactId>mssql-jdbc_auth</artifactId>
  <version>8.1.1.x64-preview</version>
[...]

PR #1178, opened 2019-11-06
Release | Changes to publish mssql-jdbc_auth DLLs on Maven

ulvii:
- DLL name changed from `sqljdbc_auth.dll` to `mssql-jdbc_auth-${version}.${bitness}-${releaseExt}.dll`
- Every time there is release, `releaseExt` field in `SQLJdbcVersion.java`  needs to be updated to reflect the release type.
- The newer versions of the driver will only be able to load the DLLs that match the driver version. For example, `mssql-jdbc-8.1.1.jre13-preview.jar` will require either `mssql-jdbc_auth-8.1.1.x64-preview.dll` or `mssql-jdbc_auth-8.1.1.x86-preview.dll` depending on the bitness of JVM.


ulvii:
#907 #1170

src/main/java/com/microsoft/sqlserver/jdbc/Util.java
```diff
@@ -40,6 +40,10 @@ static boolean isIBM() {
         return SYSTEM_JRE.startsWith("IBM");
     }
 
+    static String getJVMArch() {
+        return System.getProperty("os.arch").contains("64") ? "x64" : "x86";
```
rene-ye: Is there no possibility that an OS is x64 and won't contain that information in its name? It wouldn't be too weird for certain Unix systems that've never supported 32-bit to not include it's bitness in the OS name. The following link seems to confirm that although unusual, it's possible: https://stackoverflow.com/questions/10846105/all-possible-values-os-arch-in-32bit-jre-and-in-64bit-jre/36926327. I'm personally in favor of the "load x64 bit dll first and load x86 if it errors" approach.

ulvii: - The property name is `os.arch`, but it returns JVM bitness. So if you install 32-bit JVM on 64-bit OS, the property will return `x86`.
- DLL will only be loaded on Windows, perhaps I can rename the method to `getJVMArchWindows()` and my tests show that the results are consistent(x86 for 32-bit, amd64 for 64-bit) for all Windows versions. See #907 for more details.
- We can't do "load x64 bit dll first and load x86 if it errors", because of https://github.com/microsoft/mssql-jdbc/blob/8d2c52ee049ca126ca9bcbdec4b8a0beb3096626/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java#L6530.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResource.java
```diff
@@ -571,7 +571,7 @@ static String getResource(String key) {
                     "The client key used to access the Key Vault where the column encryption master key is stored."},
             {"R_ADALMissing", "Failed to load ADAL4J Java library for performing {0} authentication."},
             {"R_DLLandADALMissing",
-                    "Failed to load both sqljdbc_auth.dll and ADAL4J Java library for performing {0} authentication. Please install one of them to proceed."},
+                    "Failed to load both mssql-jdbc_auth and ADAL4J Java library for performing {0} authentication. Please install one of them to proceed."},
```
David-Engel: Should we spell out the full filename for mssql-jdbc_auth so it is clear to users what file they need when they encounter this error?

src/samples/azureactivedirectoryauthentication/src/main/java/AzureActiveDirectoryAuthentication.java
```diff
@@ -46,7 +46,7 @@ public static void main(String[] args) {
 
             System.out.println("For testing 'ActiveDirectoryIntegrated' Authentication, do one of the following:");
             System.out.println("   1. Generate Kerberos Ticket and validate its availability with klist tool, or");
-            System.out.println("   2. Place sqljdbc_auth.dll in the same directory as the pom.xml file.");
+            System.out.println("   2. Place mssql-jdbc_auth in the same directory as the pom.xml file.");
```
David-Engel: Same as my previous comment. We should probably spell out the full mssql-jdbc_auth filename to make it obvious to users what we have done.

Issue #1181, opened 2019-11-08
sendStringParametersAsUnicode not considered for SQLServerBulkCopy?


pbcornelius:
## Driver version
7.4.1

## Problem description
1. Expected behaviour:
`SQLServerBulkCopy` should treat `VARCHAR` and `LONGVARCHAR` like `LONGNVARCHAR` and `NVARCHAR` if `sendStringParametersAsUnicode = True`; or, alternatively, provide a seperate configuration for this (if there is one, I apologise, but I could not find it [here](https://docs.microsoft.com/en-us/sql/connect/jdbc/using-bulk-copy-with-the-jdbc-driver?view=sql-server-ver15) or [here](https://javadoc.io/doc/com.microsoft.sqlserver/mssql-jdbc/8.1.0.jre13-preview/com/microsoft/sqlserver/jdbc/SQLServerBulkCopyOptions.html)).

2. Actual behaviour:
It does not. As a result, unicode characters in `VARCHAR` columns are lost during transfer. This means one cannot transfer unicode from databases that do not distinguish between `VARCHAR` and `NVARCHAR` (such as H2).

3. Error message/stack trace:
None

4. Any other details that can be helpful:
A workaround is to wrap the ResultSetMetaData and manually swap `VARCHAR` for `NVARCHAR`, e.g.:

```
public class UnicodeResultSetMetaData implements ResultSetMetaData {

...

private static List<Integer> UNICODE_EQUIVALENT = List.of(Types.VARCHAR, Types.LONGVARCHAR);

private static List<String> UNICODE_EQUIVALENT_NAMES = List.of("VARCHAR",
		"CHARACTER VARYING",
		"LONGVARCHAR",
		"VARCHAR2",
		"VARCHAR_CASESENSITIVE");

public int getColumnType(int column) throws SQLException {
	int type = meta.getColumnType(column);
	if (UNICODE_EQUIVALENT.contains(type)) {
		return Types.NVARCHAR;
	} else {
		return type;
	}
}

public String getColumnTypeName(int column) throws SQLException {
	String type = meta.getColumnTypeName(column);
	if (UNICODE_EQUIVALENT_NAMES.contains(type)) {
		return "NVARCHAR";
	} else {
		return type;
	}
}
```

peterbae:
@pbcornelius thanks for letting us know, the team will look into this.

ulvii:
Hi @pbcornelius ,
[...]

PR #1193, opened 2019-11-27
Fix | Send bulk data as unicode when destination column is nchar/nvarchar

ulvii:
#1181

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
@@ -944,12 +944,8 @@ public class SQLServerBulkCopy implements java.lang.AutoCloseable, java.io.Seria
                     tdsWriter.writeByte((byte) 0x10);
                 } else {
                     // For the case when source database stores unicode data in CHAR and destination column is NCHAR.
-                    if (!isStreaming && unicodeConversionRequired(srcJdbcType, destSSType)) {
-                        if (SSType.NCHAR == destSSType) {
-                            tdsWriter.writeByte(TDSType.NCHAR.byteValue());
-                        } else {
-                            tdsWriter.writeByte(TDSType.NVARCHAR.byteValue());
-                        }
+                    if (unicodeConversionRequired(srcJdbcType, destSSType)) {
+                        tdsWriter.writeByte(TDSType.NCHAR.byteValue());
                         tdsWriter.writeShort(isBaseType ? (short) (srcPrecision) : (short) (2 * srcPrecision));
                     } else {
                         tdsWriter.writeByte(TDSType.BIGCHAR.byteValue());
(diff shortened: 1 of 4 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
```diff
@@ -3528,4 +3561,16 @@ protected void setStmtColumnEncriptionSetting(
...
+    /**
+     * For the case when source database stores unicode data in CHAR/VARCHAR and destination column is NCHAR/NVARCHAR.
+     *
+     * @param jdbcType
+     * @param ssType
+     * @return whether conversion to unicode is required.
+     */
+    private boolean unicodeConversionRequired(int jdbcType, SSType ssType) {
+        return ((java.sql.Types.CHAR == jdbcType || java.sql.Types.VARCHAR == jdbcType
+                || java.sql.Types.LONGNVARCHAR == jdbcType) && (SSType.NCHAR == ssType || SSType.NVARCHAR == ssType));
```
peterbae: Shouldn't we also check for SSType.NVARCHARMAX?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
```diff
@@ -2168,33 +2193,41 @@ else if (null != sourceCryptoMeta) {
...
+                                    tdsWriter.writeBytes(bytes);
+                                } else {
+                                    tdsWriter.writeShort((short) (colValueStr.length()));
+                                    // converting string into destination collation using Charset
 
-                                SQLCollation destCollation = destColumnMetadata.get(destColOrdinal).collation;
-                                if (null != destCollation) {
-                                    tdsWriter.writeBytes(colValueStr
-                                            .getBytes(destColumnMetadata.get(destColOrdinal).collation.getCharset()));
+                                    SQLCollation destCollation = destColumnMetadata.get(destColOrdinal).collation;
```
lilgreenbird: ? is that a tab?

src/test/java/com/microsoft/sqlserver/jdbc/SQLServerDriverTest.java
```diff
@@ -36,12 +36,12 @@
...
+     * @throws SQLException
      */
     @Test
-    public void testDriverDM() throws ClassNotFoundException {
-        Stream<Driver> drivers = DriverManager.drivers();
-        Object[] driversArray = drivers.toArray();
-        assertEquals(driversArray[0].getClass(), Class.forName(Constants.MSSQL_JDBC_PACKAGE + ".SQLServerDriver"),
+    public void testDriverDM() throws SQLException, ClassNotFoundException {
+        Driver driver = DriverManager.getDriver(connectionString);
+        assertEquals(driver.getClass(), Class.forName(Constants.MSSQL_JDBC_PACKAGE + ".SQLServerDriver"),
```
lilgreenbird: expected is 1st param

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
ulvii: No it's not, the file is formatted.

Issue #1192, opened 2019-11-27
getColumns on Azure returns only one column


ilkerc:
## Driver version
Tried on both `8.1.0.jre8.preview` and `7.4.1.jre8`. Works on `7.0.0.jre8` as expected.

## Problem description
`connection.metaData.getColumns("myDatabaseName", "mySchemaName", "myTableName", "%")`
returns only the first row on Azure. Expected to fetch all the columns in "myTableName". The same thing works perfectly on SQL Server.

ulvii:
Hi @ilkerc ,
Are you using Azure SQL DB or Azure SQL DW?

ilkerc:
> Hi @ilkerc ,
> Are you using Azure SQL DB or Azure SQL DW?

Hi @ulvii I'm using Azure SQL DW

ulvii:
I confirmed the issue, thanks for reporting. We will work on a fix.

ulvii:
#1197 merged.

PR #1197, opened 2019-11-28
Fix | Fixed the issue with getColumns() only returning the first column against Azure DW

rene-ye:
Fix for #1192

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
@@ -688,48 +688,57 @@ public final class SQLServerDatabaseMetaData implements java.sql.DatabaseMetaDat
                 SQLServerResultSet userRs = null;
                 PreparedStatement resultPstmt = null;
                 try (ResultSet rs = storedProcPstmt.executeQuery()) {
-                    rs.next();
-                    // Use LinkedHashMap to force retrieve elements in order they were inserted
-                    Map<Integer, String> columns = new LinkedHashMap<>();
-                    columns.put(1, "TABLE_CAT");
-                    columns.put(2, "TABLE_SCHEM");
-                    columns.put(3, "TABLE_NAME");
-                    columns.put(4, "COLUMN_NAME");
-                    columns.put(5, "DATA_TYPE");
-                    columns.put(6, "TYPE_NAME");
-                    columns.put(7, "COLUMN_SIZE");
-                    columns.put(8, "BUFFER_LENGTH");
-                    columns.put(9, "DECIMAL_DIGITS");
-                    columns.put(10, "NUM_PREC_RADIX");
-                    columns.put(11, "NULLABLE");
-                    columns.put(12, "REMARKS");
-                    columns.put(13, "COLUMN_DEF");
-                    columns.put(14, "SQL_DATA_TYPE");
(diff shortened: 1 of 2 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
```diff
@@ -756,6 +772,13 @@ private static String EscapeIDName(String inID) throws SQLServerException {
         }
     }
 
+    /*
+     * Helper method to create static maps used by the functions.
+     */
+    private static final SimpleEntry<Integer, String> createEntry(Integer i, String s) {
```
peterbae: where is this method being used?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
```diff
@@ -605,6 +607,47 @@ private static String EscapeIDName(String inID) throws SQLServerException {
         return outID.toString();
     }
 
+    // Use LinkedHashMap to force retrieve elements in order they were inserted
+    private static final LinkedHashMap<Integer, String> getColumnsDWColumns = new LinkedHashMap<>();
+    static {
```
ulvii: I can see why you would want this block static, but I think a better option is to populate this map only if the connection is against Azure DW and it is empty.

ulvii:
Needs a test.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
```diff
@@ -675,6 +680,46 @@ private static String EscapeIDName(String inID) throws SQLServerException {
              * when user provides a different catalog than the one they're currently connected to. Will throw exception
              * when it's different and do nothing if it's the same/null.
              */
+            if (null == getColumnsDWColumns) {
```
peterbae: This should be in a synchronized block

ulvii: Please rename this map, there will be other APIs that require different column maps, so better the map name reflects the API name.

rene-ye: it's meant to reflect the method where it's used. This is for `getColumns`, if `getImportedKeys` needs one it'd be `getImportedKeysDWColumns`.

src/test/java/com/microsoft/sqlserver/jdbc/databasemetadata/DatabaseMetaDataTest.java
```diff
@@ -608,6 +610,59 @@ public void testGetMaxConnections() throws SQLException {
...
+            try (ResultSet resultSet = databaseMetaData.getColumns(null, null, tableName, "%");) {
+                ResultSetMetaData rsmd = resultSet.getMetaData();
+                int rowCount = 0;
+                while (resultSet.next()) {
+                    for (int i = 1; i < 33; i++) {
+                        assertEquals(rsmd.getColumnName(i), getColumnsDWColumns.values().toArray()[i - 1]);
+                    }
+                    rowCount++;
+                }
+                assertEquals(rowCount, 3);
```
ulvii: assertEquals(3, rowCount);

Issue #1070, opened 2019-05-30
[BUG]  Performance Issues reading LocalDate,LocalDateTime, and LocalTime


lavabear:
## Driver version
7.2.1.jre8

## SQL Server version
Microsoft SQL Server 2014 (SP3) (KB4022619) - 12.0.6024.0 (X64) 
	Sep  7 2018 01:37:51 
	Copyright (c) Microsoft Corporation
	Developer Edition (64-bit) on Windows NT 6.3 <X64> (Build 9600: )

## Client Operating System
Ubuntu 16.04

## JAVA/JVM version
openjdk version "12" 2019-03-19
OpenJDK Runtime Environment (build 12+33)
OpenJDK 64-Bit Server VM (build 12+33, mixed mode, sharing)

## Table schema
```sql
CREATE TABLE test.DatePerformance (
	someDate DATE
)

-- Populate Table with Random Dates
INSERT INTO test.DatePerformance
SELECT DATEADD(DAY, -RAND(12313456) * 1000, GETDATE())

WHILE((SELECT COUNT(*) FROM test.DatePerformance) < 2500)
INSERT INTO test.DatePerformance
SELECT DATEADD(DAY, -RAND() * 1000, GETDATE())
```

## Problem description
1. Expected behaviour:
    - Reading Date objects from the database should have low overhead.

2. Actual behaviour:
    - There is substantial overhead creating GregorianCalendars and converting them to the LocalDate, LocalDateTime, LocalTime classes
    - Calling Calendar.getInstance(TimeZone.getTimeZone("UTC")) is 30% of the getObject call for LocalDate, LocalDateTime, LocalTime classes

3. Error message/stack trace:

4. Any other details that can be helpful:
    - Result Set - Calendar.getInstance https://github.com/microsoft/mssql-jdbc/blob/v7.2.1/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResultSet.java#L2384-L2398
    - DDC.convertTemporalToObject - https://github.com/microsoft/mssql-jdbc/blob/v7.2.1/src/main/java/com/microsoft/sqlserver/jdbc/DDC.java#L755-L1061
```
Sample Profiling Snapshots

	com.microsoft.sqlserver.jdbc.SQLServerResultSet.getObject () 15.094s
>	com.microsoft.sqlserver.jdbc.DDC.convertTemporalToObject () 9.206s
>>	java.util.GregorianCalendar.<init> () 6.01s
[...]

Issue #1143, opened 2019-09-08
[BUG] java.time support goes through java.sql.Timestamp


marschall:
## Driver version
7.4.1.jre8

## SQL Server version
Microsoft SQL Server 2017 (RTM-CU13) (KB4466404) - 14.0.3048.4 (X64) 
	Nov 30 2018 12:57:58 
	Copyright (C) 2017 Microsoft Corporation
	Developer Edition (64-bit) on Linux (Ubuntu 16.04.5 LTS)

## Client Operating System
Linux hostname 5.0.0-27-generic #28-Ubuntu SMP Tue Aug 20 19:53:07 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux

## JAVA/JVM version
openjdk version "1.8.0_222"
OpenJDK Runtime Environment (Zulu 8.40.0.25-CA-linux64) (build 1.8.0_222-b10)
OpenJDK 64-Bit Server VM (Zulu 8.40.0.25-CA-linux64) (build 25.222-b10, mixed mode)

## Table schema
empty

## Problem description

1. To create a `java.time` type first a `java.sql.Timestamp` is created see
https://github.com/microsoft/mssql-jdbc/blob/8a5fca2207dd9a5f4f22bfbfb45652992a53cb3a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResultSet.java#L2384

This is an issue because `java.sql.Timestamp` is bound to the JVM time zone. If the value does not exist in a JVM time zone because the DST offset changes a wrong value is returned.

## Reproduction code

```java

  @Test
  public void timestampConversion() throws SQLException {
    LocalDateTime unstorableValue = this.getUnstorableValue();
    try (Connection connection = this.dataSource.getConnection();
         PreparedStatement selectStatement = connection.prepareStatement("SELECT '" + unstorableValue.toLocalDate() + " " + unstorableValue.toLocalTime() + "'");
         ResultSet resultSet = selectStatement.executeQuery()) {
      while (resultSet.next()) {
        LocalDateTime actual = resultSet.getObject(1, LocalDateTime.class);
        if (!unstorableValue.equals(actual)) {
          System.out.println("expected: " + unstorableValue + " actual: " + actual);
        }
      }
    }
  }


  LocalDateTime getUnstorableValue() {
    ZoneId systemTimezone = ZoneId.systemDefault();
    Instant now = Instant.now();

    ZoneRules rules = systemTimezone.getRules();
[...]

Issue #1088, opened 2019-06-15
[BUG] rs.getObject(x, LocalDateTime.class) returns incorrect values prior to 1582-10-15


gordthompson:
## Driver version
7.3.1.0 (7.3.1.jre8-preview)

## SQL Server version
2014

## Client Operating System
Windows 7

## JAVA/JVM version
Oracle 1.8.0_201

## Table schema
CREATE TABLE #tmp (id int PRIMARY KEY, dt datetime2)

## Problem description
`ResultSet#getObject(x, LocalDateTime.class)` returns incorrect values for dates prior to 1582-10-15 (the adoption of the Gregorian Calendar).

## Reproduction code

```java
try (Connection conn = DriverManager.getConnection(connUrl, myUid, myPwd)) {
    System.out.printf("%s version %s%n", conn.getMetaData().getDriverName(),
            conn.getMetaData().getDriverVersion());
    // Microsoft JDBC Driver 7.3 for SQL Server version 7.3.1.0
    
    try (Statement st = conn.createStatement()) {
        // test data (The Battle of Hastings)
        st.execute("CREATE TABLE #tmp (id int PRIMARY KEY, dt datetime2)");
        st.execute("INSERT INTO #tmp (id, dt) VALUES (1, '1066-10-14 00:00:00')");
        
        // retrieve as string to verify that the date is stored correctly
        ResultSet rs = st.executeQuery("SELECT CAST(dt AS VARCHAR) FROM #tmp WHERE id = 1");
        rs.next();
        String strActual = rs.getString(1);
        String strExpected = "1066-10-14 00:00:00.0000000";
        System.out.printf("%s - actual: \"%s\", expected: \"%s\"%n", 
                strActual.equals(strExpected) ? "PASS" : "FAIL", strActual, strExpected);
        // PASS - actual: "1066-10-14 00:00:00.0000000", expected: "1066-10-14 00:00:00.0000000"
        
        // retrieve as LocalDateTime
        rs = st.executeQuery("SELECT dt FROM #tmp WHERE id = 1");
        rs.next();
        LocalDateTime ldtActual = rs.getObject(1, LocalDateTime.class);
        LocalDateTime ldtExpected = LocalDateTime.of(1066, 10,14, 0, 0);
        System.out.printf("%s - actual: \"%s\", expected: \"%s\"%n", 
                ldtActual.equals(ldtExpected) ? "PASS" : "FAIL", ldtActual, ldtExpected);
[...]

PR #1200, opened 2019-12-03
Performance | Optimized temporal datatype getter methods

peterbae:
Fixes issues #1070, #1088, and #1143.

[JMH test suite](https://github.com/peterbae/JDBC_JMH_Benchmark) was used for performance measurement.


src/main/java/com/microsoft/sqlserver/jdbc/DDC.java
```diff
@@ -893,16 +901,9 @@ static final Object convertTemporalToObject(JDBCType jdbcType, SSType ssType, Ca
             default:
                 throw new AssertionError("Unexpected SSType: " + ssType);
         }
```
lilgreenbird: why do this instead of our usual SQLServerException.makeFromDriverError and use an error string?

src/main/java/com/microsoft/sqlserver/jdbc/DDC.java
```diff
@@ -1060,6 +1064,136 @@ static final Object convertTemporalToObject(JDBCType jdbcType, SSType ssType, Ca
...
+                ldt = ldt.plusDays(daysSinceBaseDate);
+                if (jdbcType.category != JDBCType.Category.DATE && ssType != SSType.DATE) {
+                    ldt = ldt.plusNanos(ticksSinceMidnight);
+                }
+                subSecondNanos = (int) (ticksSinceMidnight % Nanos.PER_SECOND);
+                break;
+            }
+
+            case DATETIME: // and SMALLDATETIME
+            {
```
lilgreenbird: why don't we need to add "case SMALLDATETIME:" here?

src/main/java/com/microsoft/sqlserver/jdbc/DDC.java
```diff
@@ -1060,6 +1064,136 @@ static final Object convertTemporalToObject(JDBCType jdbcType, SSType ssType, Ca
...
+                    ldt = ldt.plusNanos(ticksSinceMidnight * Nanos.PER_MILLISECOND);
+                }
+
+                subSecondNanos = (int) ((ticksSinceMidnight * Nanos.PER_MILLISECOND) % Nanos.PER_SECOND);
+                break;
+            }
+
+            default:
+                throw new AssertionError("Unexpected SSType: " + ssType);
+        }
```
lilgreenbird: same comment as above, use makeFromDriverError? (multiple occurrences)

src/main/java/com/microsoft/sqlserver/jdbc/DDC.java
```diff
@@ -1060,6 +1064,136 @@ static final Object convertTemporalToObject(JDBCType jdbcType, SSType ssType, Ca
         }
     }
 
+    private static Object convertTemporalToObject(JDBCType jdbcType, SSType ssType, int daysSinceBaseDate,
+            long ticksSinceMidnight, int fractionalSecondsScale) {
+        int subSecondNanos;
```
ulvii: Is subSecondNanos required anymore? I tried removing a few setNanos() calls and didn't seem to break anything.

src/main/java/com/microsoft/sqlserver/jdbc/DDC.java
```diff
@@ -1060,6 +1064,136 @@ static final Object convertTemporalToObject(JDBCType jdbcType, SSType ssType, Ca
...
+                subSecondNanos = (int) (ticksSinceMidnight % Nanos.PER_SECOND);
+                break;
+            }
+
+            case DATE:
+            case DATETIME2:
+            case DATETIMEOFFSET: {
+                ldt = LocalDateTime.of(1, 1, 1, 0, 0, 0);
+                ldt = ldt.plusDays(daysSinceBaseDate);
+                if (jdbcType.category != JDBCType.Category.DATE && ssType != SSType.DATE) {
```
ulvii: Can you explain this check?

src/main/java/com/microsoft/sqlserver/jdbc/DDC.java
peterbae: That variable is the millisecond value after DD:MM:SS. JUnit tests should fail without this variable.

ulvii: Wouldn't plusNanos() already take care of that?

src/main/java/com/microsoft/sqlserver/jdbc/DDC.java
peterbae: It's just the way code was before, which I didn't touch. I think we can leave this as is.

src/main/java/com/microsoft/sqlserver/jdbc/DDC.java
peterbae: plusNanos() is a way to offload the calculation of converting the amount of nanoseconds from midnight to LocalDateTime, whereas subSecondNanos is the actual variable that holds the nanosecond values after DD:MM:SS. plusNanos already gives the LocalDateTime values with the correct nanosecond values, but depending on the target datatype, the nanosecond value inside LocalDateTime might be lost during conversion from LocalDateTime (for example, cases where the target datatype is java.sql.Timestamp), so we need this variable to hold onto the nanosecond value.

src/main/java/com/microsoft/sqlserver/jdbc/DDC.java
peterbae: This comment is misleading, but basically the cases are based off of SSType.Category. DATETIME and SMALLDATETIME both go under Category.DATETIME, so we don't need to add a case for SMALLDATETIME.

src/main/java/com/microsoft/sqlserver/jdbc/DDC.java
peterbae: I've added comment for it, but I put it there for optimization since java.sql.Date doesn't need its time component.

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/DataTypesTest.java
```diff
@@ -1779,4 +1785,85 @@ public void testDateTimeInsertUpdate() throws Exception {
...
+     */
+    @Test
+    public void testGetLocalDateTimePriorGregorian() throws Exception {
+        try (Connection conn = getConnection(); Statement st = conn.createStatement();) {
+            String ldtTable = AbstractSQLGenerator.escapeIdentifier(RandomUtil.getIdentifier("ldtTable"));
+            TestUtils.dropTableIfExists(ldtTable, st);
+            // test data (The Battle of Hastings)
+            LocalDateTime ldtExpected = LocalDateTime.of(1066, 10, 14, 0, 0);
+            st.execute("CREATE TABLE " + ldtTable + " (id int PRIMARY KEY, dt datetime2)");
+            PreparedStatement ps = conn.prepareStatement("INSERT INTO " + ldtTable + " (id, dt) VALUES (1, ?)");
```
ulvii: Move the PreparedStatement  to try block

peterbae: great catch

Issue #1145, opened 2019-09-12
DatabaseMetaData#getImportedKeys() fails on Azure SQL Data Warehouse


FejfarKamil:
## Driver version
7.4.1

## SQL Server version
Microsoft Azure SQL Data Warehouse - 10.0.9999.0 Aug 16 2019 21:18:18 Copyright (c) Microsoft Corporation

## Client Operating System
Windows 10 Pro

## JAVA/JVM version
java version "1.8.0_221

## Problem description
We want to obtain imported keys on Azure SQL Data Warehouse.

`DatabaseMetaData#getImportedKeys()` calls `DECLARE @jdbc_temp_fkeys_result table (...`, that fails, because variable of a table type is not supported on Azure SQL Data Warehouse (https://docs.microsoft.com/en-us/sql/t-sql/language-elements/declare-local-variable-transact-sql?view=sql-server-2017).

Caused by: com.microsoft.sqlserver.jdbc.SQLServerException: Parse error at line: 1, column: 33: Incorrect syntax near 'table'.
	at com.microsoft.sqlserver.jdbc.SQLServerException.makeFromDatabaseError(SQLServerException.java:262)
	at com.microsoft.sqlserver.jdbc.SQLServerStatement.getNextResult(SQLServerStatement.java:1624)
	at com.microsoft.sqlserver.jdbc.SQLServerPreparedStatement.doExecutePreparedStatement(SQLServerPreparedStatement.java:594)
	at com.microsoft.sqlserver.jdbc.SQLServerPreparedStatement$PrepStmtExecCmd.doExecute(SQLServerPreparedStatement.java:524)
	at com.microsoft.sqlserver.jdbc.TDSCommand.execute(IOBuffer.java:7194)
	at com.microsoft.sqlserver.jdbc.SQLServerConnection.executeCommand(SQLServerConnection.java:2979)
	at com.microsoft.sqlserver.jdbc.SQLServerStatement.executeCommand(SQLServerStatement.java:248)
	at com.microsoft.sqlserver.jdbc.SQLServerStatement.executeStatement(SQLServerStatement.java:223)
	at com.microsoft.sqlserver.jdbc.SQLServerPreparedStatement.executeQuery(SQLServerPreparedStatement.java:446)
	at com.microsoft.sqlserver.jdbc.SQLServerDatabaseMetaData.executeSPFkeys(SQLServerDatabaseMetaData.java:1026)
[...]

PR #1205, opened 2019-12-07
Fix | Fixed getImportedKeys API against Azure Data Warehouse

ulvii:
#1145

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
@@ -264,6 +264,10 @@ public final class SQLServerDatabaseMetaData implements java.sql.DatabaseMetaDat
     private static final String IS_AUTOINCREMENT = "IS_AUTOINCREMENT";
     private static final String SQL_KEYWORDS = createSqlKeyWords();
 
+    // Use LinkedHashMap to force retrieve elements in order they were inserted
+    private static LinkedHashMap<Integer, String> getColumnsDWColumns = null;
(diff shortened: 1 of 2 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
```diff
@@ -678,7 +679,7 @@ private static String EscapeIDName(String inID) throws SQLServerException {
              * when user provides a different catalog than the one they're currently connected to. Will throw exception
              * when it's different and do nothing if it's the same/null.
              */
-            synchronized (SQLServerDatabaseMetaData.class) {
+            synchronized (this) {
```
peterbae: I think this synchronized block should be on a class level, to ensure that no race condition occurs when trying to initialize getColumnsDWColumns

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
```diff
@@ -1021,41 +1021,86 @@ public String getIdentifierQuoteString() throws SQLServerException {
...
+            if (null != currentDB) {
+                switchCatalogs(currentDB);
+            }
+            return rs;
+        } else {
+            // Azure DW does not support foreign keys, return an empty result set with correct metadata.
+            ResultSet userRs = null;
+            PreparedStatement pstmt = null;
+            StringBuilder azureDwSelectBuilder = new StringBuilder();
+            synchronized (this) {
```
peterbae: if the above comment applies, then also apply the same changes here

PR #1166, opened 2019-10-22
Refactored AE tests to add new Enclave tests for testing AEv2

lilgreenbird:

PR #1208, opened 2019-12-10
Fix creating lots of Sessions

rene-ye:
Calling sp_describe_column_encryption with 3 parameters causes the server to create an enclave session. Only call it with 2 parameters if enclave is already established. Deletes a swap file accidently checked in from #1166.


src/main/java/com/microsoft/sqlserver/jdbc/ISQLServerEnclaveProvider.java
```diff
@@ -148,7 +161,7 @@ default void processAev1SPDE(String userSql, String preparedTypeDefinitions, Par
 
         // Process the second resultset.
         if (!stmt.getMoreResults()) {
-            throw new SQLServerException(this, SQLServerException.getErrString("R_UnexpectedDescribeParamFormat"), null,
+            throw new SQLServerException(null, SQLServerException.getErrString("R_UnexpectedDescribeParamFormat"), null,
```
peterbae: why is this null

rene-ye: It's an AEv1 thing msg that we don't really even handle in AEv2. It needed to be null because I considered putting it in Util file. Still not sure where the best place is for these methods, looking to see if I can refractor AEv1 too.

src/main/java/com/microsoft/sqlserver/jdbc/Util.java
```diff
@@ -991,6 +999,123 @@ static boolean use43Wrapper() {
     static String escapeSingleQuotes(String name) {
         return name.replace("'", "''");
     }
+
+    static ResultSet executeSDPEv2(PreparedStatement stmt, String userSql, String preparedTypeDefinitions,
```
lilgreenbird: I don't think Util is a good place for this. utils contain very generic types utils that are used everywhere, this is very specific for AE

rene-ye: We already have this weird "shoudHonorAEFor..." method, it doesn't seem like we actually have a specific methodology for what goes into Utils. I was looking for something like AEUtils but that doesn't exist.

Issue #1201, opened 2019-12-05
Calling Blob#length() on an image column causes an InputStream obtained from Blob#getBinaryStream() to close


jborgland:
## Driver version
7.4.1 (appears to be valid for other versions as well)

## SQL Server version
Microsoft SQL Server 2017 (RTM) - 14.0.1000.169 (X64) 
	Aug 22 2017 17:04:49 
	Copyright (C) 2017 Microsoft Corporation
	Developer Edition (64-bit) on Windows Server 2016 Standard 10.0 <X64> (Build 14393: ) (Hypervisor)

(but appears to be valid for other versions as well)

## Client Operating System
Windows 10

## JAVA/JVM version
java version "1.8.0_171"
Java(TM) SE Runtime Environment (build 1.8.0_171-b11)
Java HotSpot(TM) 64-Bit Server VM (build 25.171-b11, mixed mode)

## Table schema
create table mytable (
  mydata image
);

## Problem description
With the fix for https://github.com/Microsoft/mssql-jdbc/issues/611 it's now _usually_ possible to:
1. Obtain an InputStream from a BLOB with `myBlob.getBinaryStream()`.
2. Call `myBlob.length()` to get the size of the BLOB.
3. Consume the InputStream.

However, if the data type of the column is `image` then the following exception is thrown:

```
Caused by: java.io.IOException: The stream is closed.
	at com.microsoft.sqlserver.jdbc.BaseInputStream.checkClosed(SimpleInputStream.java:95) ~[mssql-jdbc.jar:?]
	at com.microsoft.sqlserver.jdbc.SimpleInputStream.read(SimpleInputStream.java:263) ~[mssql-jdbc.jar:?]
	at com.example.copyStreams(JdbcBlobStreamTester .java:41) 
```

This happens when the `value` byte array in `com.microsoft.sqlserver.jdbc.SQLServerBlob` is `null` and the first active input stream isn't an instance of `PLPInputStream` (the call to `stream.getBytes()` in `getBytesFromStream()` causes the stream to be closed).

## Reproduction code
Use the same code as in https://github.com/Microsoft/mssql-jdbc/issues/611.

rene-ye:
Hi @jborgland, the team will take a look.

rene-ye:
I added the fix for this issue into a PR for another semi-related issue. Please feel free to test it out.

rene-ye:
PR merged, closing issue.

PR #1214, opened 2019-12-17
Added backwards compatibility for calling Clob#length() on NVarchar columns

rene-ye:
Also includes a fix for image columns #1201.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerClob.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerClob.java
@@ -68,17 +68,6 @@ public class SQLServerClob extends SQLServerClobBase implements Clob {
     final JDBCType getJdbcType() {
         return JDBCType.CLOB;
     }
-
-    @Override
-    public long length() throws SQLException {
-        long length = super.length();
-        if (null != typeInfo) {
-            String columnTypeName = typeInfo.getSSTypeName();
-            return ("nvarchar".equalsIgnoreCase(columnTypeName) || "ntext".equalsIgnoreCase(columnTypeName)) ? length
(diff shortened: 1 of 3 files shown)


rene-ye:
Attaching compiled jre8 jar

[mssql-jdbc-8.2.0-SNAPSHOT.jre8.zip](https://github.com/microsoft/mssql-jdbc/files/3996058/mssql-jdbc-8.2.0-SNAPSHOT.jre8.zip)

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerClob.java
```diff
@@ -328,7 +328,13 @@ public String getSubString(long pos, int length) throws SQLException {
     public long length() throws SQLException {
         checkClosed();
         if (null == value && activeStreams.get(0) instanceof BaseInputStream) {
-            return (long) ((BaseInputStream) activeStreams.get(0)).payloadLength;
+            int length = ((BaseInputStream) activeStreams.get(0)).payloadLength;
+            if (null != typeInfo) {
+                String columnTypeName = typeInfo.getSSTypeName();
+                return ("nvarchar".equalsIgnoreCase(columnTypeName)
+                        || "ntext".equalsIgnoreCase(columnTypeName)) ? length / 2 : length;
```
peterbae: SSType's name can return nchar as well as well as nvarchar and ntext. Should we also cover for nchar here?

rene-ye: We don't claim to support Char/NChar for clob/nclob: https://docs.microsoft.com/en-us/sql/connect/jdbc/using-advanced-data-types?view=sql-server-ver15

PR #1216, opened 2019-12-20
Fix negative update counts

rene-ye:
ExecuteUpdate() and related APIs are supposed to return "either (1) the row count for SQL Data Manipulation Language (DML) statements or (2) 0 for SQL statements that return nothing". Currently in the driver, if an 'if' operation evaluates false, the driver will return -1.

```sql
IF (1 = 0) UPDATE FOO SET BAR = 'foobar'
```


rene-ye:
Basically will never happen due to test errors. Closing PR.

PR #1217, opened 2019-12-23
Feat add custom socket factory

sehrope:
Adds a new connection property, `socketFactoryClass`, to allow overriding the SocketFactory used to create Sockets. Similar to the custom trust manager, there's also an optional string property, `socketFactorConstructorArg`, to pass as a constructor arg to further customize the SocketFactory. If not specified then the system default via `SocketFactory.getDefault()` is used.

First commit adds an internal helper for instantiating custom class instances as the logic is the same as for the custom trust manager.

Second commit has the new feature and adds a test that configures a basic SocketFactory that keeps logs its usage (to test that it's actually being used).

The SocketFactory is created once for each connection and cached in private variable. It may be used multiple times if the driver attempts to concurrently open sockets (ex: via multiple address resolution for the DB host). My reading of that code suggests that no thread synchronization is required as the sockets are created by the parent prior to executing in their own threads. Would be good for another set of eyes to confirm: https://github.com/microsoft/mssql-jdbc/blob/dev/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java#L2638-L2654


rene-ye:
Is this PR for a particular issue or is it just an overall improvement suggestion?

sehrope:
I haven't opened a separate issue for it but I can if it helps track things. I'd like to be able to customize the `Socket` created internally by the driver.

Without this PR there is no way to do so as the driver issues a `new Socket()` to directly create its sockets (which end up using the default socket factory). This PR allows customizing the socket creation by specifying a custom class to act as the socket factory for that one connection without having to override the default JDK-wide socket factory.

FYI, it's a common option on most Java drivers (both JDBC and non-JDBC):

* PostgreSQL JDBC Driver - https://jdbc.postgresql.org/documentation/head/connect.html
* MariaDB JDBC Driver - https://mariadb.com/kb/en/about-mariadb-connector-j/#infrequently-used-parameters
- MySQL JDBC Driver -https://docs.oracle.com/cd/E17952_01/connector-j-5.1-en/connector-j-reference-configuration-properties.html
* MongoDB Java Driver - https://mongodb.github.io/mongo-java-driver/3.6/javadoc/com/mongodb/MongoClientOptions.Builder.html#socketFactory-javax.net.SocketFactory-

rene-ye:
Just a quick update to this PR. The team will need some time to review the contents of this PR, but it will probably start no earlier than early February as our bandwidth is currently limited.

sehrope:
@rene-ye Thanks for the update.

sehrope:
Rebased to resolve some merge conflicts on dev, fixed a typo, and added one more test for invalid class values (similar to the one for the trust manager property).

ulvii:
Hi @sehrope ,
We made a few changes to your branch around error messages and file formatting. Merging the PR, appreciate the contribution. 👍

sehrope:
Nice. Thanks for the review and merging.

zxcvboy:
Hi. @sehrope 

    Is this property(socketFactoryClass) ready to use? My jdbc driver`s version is "9.1.0.jre8-preview". 
    I tried to use custom SocketFactory, however it seemed not be adopted by DriverManager. 
    Below is my connection url.
    "jdbc:sqlserver://testdb:1433;databaseName=LIFERAY_TEST;user=MEHOLiferaySearch;password=1qaz@WSX;socketFactoryClass=customer.capjdbc.DummySocketFactory;socketFactoryConstructorArg=TEST-CUSTOM-ARG"
  
Thanks a lot.

PR #1220, opened 2019-12-24
Synchronized native AE calls to DLL

lilgreenbird:
fixes issue when multiple threads calls AE methods

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/AuthenticationJNI.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/AuthenticationJNI.java
@@ -157,9 +157,9 @@ final class AuthenticationJNI extends SSPIAuthentication {
             String servicePrincipalName, String clientConnectionId, String clientId, long expirationFileTime,
             java.util.logging.Logger log);
 
-    static native byte[] DecryptColumnEncryptionKey(String masterKeyPath, String encryptionAlgorithm,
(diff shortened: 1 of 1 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/AuthenticationJNI.java
```diff
@@ -157,9 +157,9 @@ private static native FedAuthDllInfo ADALGetAccessTokenForWindowsIntegrated(Stri
             String servicePrincipalName, String clientConnectionId, String clientId, long expirationFileTime,
             java.util.logging.Logger log);
 
-    static native byte[] DecryptColumnEncryptionKey(String masterKeyPath, String encryptionAlgorithm,
+    static synchronized native byte[] DecryptColumnEncryptionKey(String masterKeyPath, String encryptionAlgorithm,
             byte[] encryptedColumnEncryptionKey) throws DLLException;
 
-    static native boolean VerifyColumnMasterKeyMetadata(String keyPath, boolean allowEnclaveComputations,
+    static synchronized native boolean VerifyColumnMasterKeyMetadata(String keyPath, boolean allowEnclaveComputations,
```
rene-ye: Do both these functions call the same problematic handle getter? If so, don't we need to prevent both of them from calling the dll function at the same time? Might need an Object lock here instead of synchronized functions.

PR #1226, opened 2020-01-03
Fixed errors identified by Semmle

lilgreenbird:


src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -3776,7 +3776,14 @@ void writeString(String value) throws SQLServerException {
...
                 valueBytes[bytesCopied++] = (byte) ((ch >> 0) & 0xFF);
-                valueBytes[bytesCopied++] = (byte) ((ch >> 8) & 0xFF);
+
+                if (bytesCopied > bytesToCopy) {
+                    MessageFormat form = new MessageFormat(SQLServerException.getErrString("R_indexOutOfRange"));
+                    Object[] msgArgs = {bytesCopied};
+                    error(form.format(msgArgs), SQLState.DATA_EXCEPTION_NOT_SPECIFIC, DriverError.NOT_SET);
+                }
+                valueBytes[bytesCopied] = (byte) ((ch >> 8) & 0xFF);
+                bytesCopied++;
```
rene-ye: This is not an issue, Java char/string are always 2 bytes per character. It's not necessary to do this check.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerXAResource.java
```diff
@@ -845,8 +845,19 @@ public void forget(Xid xid) throws XAException {
...
+                MessageFormat form = new MessageFormat(SQLServerException.getErrString("R_indexOutOfRange"));
+                Object[] msgArgs = {offset};
+                XAException xex = new XAException(form.format(msgArgs));
+                xex.errorCode = XAException.XAER_RMERR;
+                if (xaLogger.isLoggable(Level.FINER))
+                    xaLogger.finer(toString() + " exception:" + xex);
+                throw xex;
+            }
+            int bid_len = (r.bData[offset] & 0x00FF);
+            offset++;
```
rene-ye: Same here, if we're really worried this may cause an out of index access then we can catch and rethrow it.

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
lilgreenbird: I did this cos adding a try doesn't *fix* the error for Semmle. It's stupid. Ok I changed it

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerXAResource.java
lilgreenbird: I did this cos adding a try doesn't *fix* the error for Semmle. It's stupid. Ok I changed it

PR #1240, opened 2020-01-29
Fix | Hide BouncyCastle from JDK 11/13 runtime

rene-ye:


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerVSMEnclaveProvider.java
```diff
@@ -333,7 +333,7 @@ void validateStatementSignature() throws SQLServerException, GeneralSecurityExce
              * RSASSA-PSS was added in JDK 11, the user might be using an older version of Java. Use BC as backup.
              * Remove this logic if JDK 8 stops being supported or backports RSASSA-PSS
              */
-            Security.addProvider(new org.bouncycastle.jce.provider.BouncyCastleProvider());
+            SQLServerBouncyCastleLoader.loadBouncyCastle();
```
lilgreenbird: update comment

rene-ye: The comment needs to remain because this whole section needs to be deleted, not just the SQLServerBouncyCastleProvider file.

Issue #1249, opened 2020-02-03
NPE in SQLServerResultSet.getObject()


albertsputa:
## Driver version

8.2.0-jre11

## SQL Server version

Microsoft SQL Azure (RTM) - 12.0.2000.8

NPE occurs on:
Caused by: java.lang.NullPointerException
	at com.microsoft.sqlserver.jdbc.SQLServerResultSet.getObject(SQLServerResultSet.java:2400)
	at com.microsoft.sqlserver.jdbc.SQLServerResultSet.getObject(SQLServerResultSet.java:2465)

and

Caused by: java.lang.NullPointerException
	at com.microsoft.sqlserver.jdbc.SQLServerResultSet.getObject(SQLServerResultSet.java:2402)
	at com.microsoft.sqlserver.jdbc.SQLServerResultSet.getObject(SQLServerResultSet.java:2465)

ulvii:
Hi @albertsputa ,
Thanks a lot for reporting the issue. We created a #1250  with the fix.

peterbae:
[github-1249.zip](https://github.com/microsoft/mssql-jdbc/files/4155324/github-1249.zip)

Hi @albertsputa, please try the attached zip that contains the 8.2.0 driver jars with the fix, and let me know if you encounter any other problems. Thanks!

ulvii:
Hi @albertsputa ,

The issue is now fixed in 8.2.1 release, which can be downloaded from [Download Center](https://www.microsoft.com/en-us/download/details.aspx?id=100855), [Maven](https://search.maven.org/artifact/com.microsoft.sqlserver/mssql-jdbc), [GitHub](https://github.com/microsoft/mssql-jdbc/releases/tag/v8.2.1).

PR #1250, opened 2020-02-03
Add null check for getObject() with LocalTime and LocalDate

peterbae:
Added a null check to prevent NPE.

Fixes issue #1249.


leefine:
yes,now i can upgrade the latest one.

PR #1251, opened 2020-02-04
Fix bracket escaping in Util.parseUrl()

### LINKED ISSUE #1188: SqlServerDriver.acceptUrl works wrong
## Driver version

7.2.0.jre8
8.1.0.jre11-preview
## SQL Server version

MsSql16
## Client Operating System

MacOs
## JAVA/JVM version

Java12
## Table schema

My connection string looks like 

> jdbc:sqlserver://myIp:1433;database=master;password=pass**;**sdfsdf%;user=my_user;

As you see my password contains ;. I use Hikari CP and this pool uses acceptUrl method. I don't want change my password and I want to use Hikari CP. Can you help me with it?

peterbae:
Fixes #1188

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/Util.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/Util.java
@@ -416,11 +416,12 @@ final class Util {
                     break;
                 }
                 case inEscapedValueStart: {
-                    // check for escaped }.
-                    // when we see a }, first check to see if this is before the end of the string to avoid index out of range exception
-                    // then check if the character immediately after is also a }.
-                    // if it is, then we have a }}, which is not the closing of the escaped state.
-                    if (ch == '}' && i + 1 <= tmpUrl.trim().length() && tmpUrl.charAt(i + 1) == '}') {
+                    /* check for escaped }.
+                    * when we see a }, first check to see if this is before the end of the string to avoid index out of range exception
+                    * then check if the character immediately after is also a }.
+                    * if it is, then we have a }}, which is not the closing of the escaped state.
+                    */
+                    if (ch == '}' && i + 1 <= tmpUrl.length() && tmpUrl.charAt(i + 1) == '}') {
                         builder = new StringBuilder();
                         builder.append(value);
                         builder.append(ch);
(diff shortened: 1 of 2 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/Util.java
```diff
@@ -416,29 +416,42 @@ else if (ch == ':')
...
                         value = builder.toString();
+                        i++; // escaped }} into a }, so increment the counter once more
                         // same state
+                    } else {
+                        if (ch == '}') {
+                            // no trimming use the value as it is.
+                            name = SQLServerDriver.getNormalizedPropertyName(name, logger);
+                            if (null != name) {
+                                if (logger.isLoggable(Level.FINE)) {
+                                    if (!name.equals(SQLServerDriverStringProperty.USER.toString())
```
lilgreenbird: just nit-picking...instead of so many ident levels why not just &&?

peterbae: I think this code looks fine. Each layer of if statement has a purpose to it, and it's easy to read.

src/main/java/com/microsoft/sqlserver/jdbc/Util.java
```diff
@@ -416,29 +416,42 @@ else if (ch == ':')
...
-                            p.put(name, value);
-                        }
-
-                        name = "";
-                        value = "";
-                        // to eat the spaces until the ; potentially we could do without the state but
-                        // it would not be clean
-                        state = inEscapedValueEnd;
-                    } else {
+                    // check for escaped }.
```
ulvii: Use multi line comments here.

src/main/java/com/microsoft/sqlserver/jdbc/Util.java
```diff
@@ -416,29 +416,42 @@ else if (ch == ':')
...
+                    // when we see a }, first check to see if this is before the end of the string to avoid index out of range exception
+                    // then check if the character immediately after is also a }.
+                    // if it is, then we have a }}, which is not the closing of the escaped state.
+                    if (ch == '}' && i + 1 <= tmpUrl.trim().length() && tmpUrl.charAt(i + 1) == '}') {
```
ulvii: `trim()` should be removed here, preceding white spaces in connection string would make `i` smaller than `tmpUrl.trim().length()`.

Issue #1278, opened 2020-03-05
Driver can't retrieve server informations in SQLServerXAResource On docker mssql


baptistemesta:
Hello,

## Driver version
7.2.1.jre8

## SQL Server version
@@VERSION:
```
Microsoft SQL Server 2017 (RTM-CU17) (KB4515579) - 14.0.3238.1 (X64) 
	Sep 13 2019 15:49:57 
	Copyright (C) 2017 Microsoft Corporation
	Express Edition (64-bit) on Linux (Ubuntu 16.04.6 LTS)
```

SERVERPROPERTY('Edition')
```
Express Edition (64-bit)
```

SERVERPROPERTY('ProductVersion')
```
14.0.3238.1
```

## Client Operating System
Docker in MacOs

## JAVA/JVM version
java version "1.8.0_211"
Java(TM) SE Runtime Environment (build 1.8.0_211-b12)
Java HotSpot(TM) 64-Bit Server VM (build 25.211-b12, mixed mode)

## Problem description

There is a warning when using XA transaction with Microsoft SQL Server 2017 running in docker:

When the driver try to retrieve server informations:
[sql call made by the driver](https://github.com/microsoft/mssql-jdbc/blob/f53b3bfd344dc4c30ea5d8cc6aa9c493ecb4dd8f/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerXAResource.java#L461)
and then parse the result to retrieve `ArchitectureOS`
[parsing of the sql result 'ArchitectureOS'](https://github.com/microsoft/mssql-jdbc/blob/f53b3bfd344dc4c30ea5d8cc6aa9c493ecb4dd8f/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerXAResource.java#L485)

```
select @@Version
```
returns
```
Microsoft SQL Server 2017 (RTM-CU17) (KB4515579) - 14.0.3238.1 (X64) 
	Sep 13 2019 15:49:57 
	Copyright (C) 2017 Microsoft Corporation
	Express Edition (64-bit) on Linux (Ubuntu 16.04.6 LTS)

```
and 
```
select convert(varchar(100), SERVERPROPERTY('Edition'))as edition, 
                                    convert(varchar(100), SERVERPROPERTY('InstanceName'))as instance,
                                    convert(varchar(100), SERVERPROPERTY('ProductVersion')) as version,
[...]

PR #1279, opened 2020-03-05
Fix | Fixed a warning when retrieving Operating System information from SQL Server Linux

ulvii:
SQL Server Linux is x64-compatible only.


educhastenier:
Closes #1278

Issue #1281, opened 2020-03-11
[FEATURE REQUEST] Pull TVP name from SQLServerDataTable to avoid execution of sp_sproc_columns_100


Logic-32:
## Is your feature request related to a problem? If so, please give a short summary of the problem and how the feature would resolve it
We use a lot of TVPs via `PreparedStatement.setObject()`.  Often times, we're able to do this by passing in a `SQLServerDataTable` instance but rarely are we ever able to use `SQLServerPreparedStatement.setStructured()`.  This causes SQLServerPreparedStatement to [forcibly report the tvpName as null](https://github.com/microsoft/mssql-jdbc/blob/v8.2.1/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java#L1469) and then [execute a query to load the tvpName](https://github.com/microsoft/mssql-jdbc/blob/v8.2.1/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerParameterMetaData.java#L291) (sp_sproc_columns_100 for modern SQL Servers).  The execution of that query happens ends up executing about 1.4 million times per hour and causes a high CPU as a result.

## Describe the preferred solution
In situations were we can still infer the `tvpName` without using `setStructured()`, such as when a `SQLServerDataTable` is passed in, use the provided `tvpName`.

## Describe alternatives you've considered
We could alternatively use `SQLServerPreparedStatement.setStructured()`.  That requires calling `unwrap()` on the PreparedStatement we have which breaks a metrics utility we use.  Additionally, the problem is further complicated by frameworks like JdbcTemplate and Apache QueryRunner which may not provide access to the underlying `SQLServerPreparedStatement` but still accept `SQLServerDataTable` instances.

## Additional context
This information was gained by querying sys.dm_exec_query_stats in SQL Server 2014.

ulvii:
Hi @Logic-32 ,
We will investigate and get back to you soon.

ulvii:
Hi @Logic-32 ,
To understand the scenario better, please provide a reproduction code/table schema with `setObject()`.

Logic-32:
Type:
```sql
CREATE TYPE dbo.Integers AS TABLE
(
	Token INT NOT NULL
)
```

Proc:
```sql
CREATE PROCEDURE echoTVP
(
	@accountIds dbo.Integers READONLY
)
AS
BEGIN
	SELECT * FROM @accountIds
END
```

Java:
```java
[...]

PR #1282, opened 2020-03-14
Get TVP name from SQLServerDataTable when using PreparedStatement.setObject()

ulvii:
PR For #1281.

When `SQLServerDataTable` is passed in as an argument to `PreparedStatement.setObject()`, it makes sense to first attempt to get the TVP name from that object, before querying database for metadata.

### DIFF
--- a/src/test/java/com/microsoft/sqlserver/jdbc/TestUtils.java
+++ b/src/test/java/com/microsoft/sqlserver/jdbc/TestUtils.java
@@ -280,7 +280,7 @@ public final class TestUtils {
      * @param tableName
      * @throws SQLException
      */
-    public static void clearTable(java.sql.Connection con, String tableName) throws SQLException {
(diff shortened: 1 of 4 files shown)


src/test/java/com/microsoft/sqlserver/jdbc/TestUtils.java
```diff
@@ -274,6 +274,18 @@ public static void dropTableIfExists(String tableName, java.sql.Statement stmt)
         dropObjectIfExists(tableName, "U", stmt);
     }
 
+    /**
+     * Deletes the contents of a table.
+     * @param con
+     * @param tableName
+     * @throws SQLException
+     */
+    public static void clearTable(java.sql.Connection con, String tableName) throws SQLException {
```
lilgreenbird: not a big deal...but just "Connection" would do we already import java.sql.Connection

PR #1284, opened 2020-03-16
Feature | client certificate authentication

peterbae:
The JDBC driver will add three connection properties for this feature:

clientCertificate – specifies the certificate to be used for authentication. The JDBC driver will support PFX, PEM, DER and CER file extensions. Format:
•	clientCertificate=<file_location> 
The driver uses a certificate file. For certificates in PEM, DER and CER formats clientKey attribute is required.
 
clientKey – specifies a file location of the private key for PEM, DER and CER certificates specified by the clientCertificate attribute. Format:
•	clientKey=<file_location>
Specifies location of the private key file. In case if private key file is password protected then password keyword is required.
clientKeyPassword – optional password string provided to access the clientKey file’s private key.


pom.xml
```diff
@@ -122,6 +122,13 @@
 			<version>1.64</version>
 			<optional>true</optional>
 		</dependency>
+		
+		<!-- dependencies for Client Certificate Authentication -->
+		<dependency>
```
saurabh500: Does this need any sort of open source approval?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCertificateUtils.java
```diff
@@ -0,0 +1,276 @@
...
+        if (keyPath != null && keyPath.length() > 0) {
+            return readPKCS8Certificate(certPath, keyPath, keyPassword);
+        } else {
+            return readPKCS12Certificate(certPath, keyPassword);
+        }
+    }
+
+    private static KeyManager[] readPKCS12Certificate(String certPath,
+            String keyPassword) throws NoSuchAlgorithmException, CertificateException, FileNotFoundException, IOException, UnrecoverableKeyException, KeyStoreException {
+        KeyStore keystore = KeyStore.getInstance("PKCS12");
```
saurabh500: Can "PKCS12" be turned into a constant string? Or is there a constant in JDK that can be used instead?

pom.xml
peterbae: Yes, we will have to talk to crypto team for approval.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCertificateUtils.java
```diff
@@ -0,0 +1,276 @@
...
+        } else if (privateKeyPem.contains(PEM_RSA_PRIVATE_START)) { // PKCS#1 format
+            return loadPrivateKeyFromPKCS1(privateKeyPem, privateKeyPassword);
+        } else {
+            return loadPrivateKeyFromPVK(privateKeyPemPath, privateKeyPassword);
+        }
+    }
+
+    private static PrivateKey loadPrivateKeyFromPKCS8(
+            String key) throws IOException, NoSuchAlgorithmException, InvalidKeySpecException {
+        key = key.replace(PEM_PRIVATE_START, "").replace(PEM_PRIVATE_END, "");
```
saurabh500: Is it possible to use the StringBuilder to do these manipulations on the key? 
The current code will create new instances of String since String is immutable. Can we optimize a bit for memory?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCertificateUtils.java
```diff
@@ -0,0 +1,276 @@
...
+        KeyFactory factory = KeyFactory.getInstance(RSA_ALG);
+        return factory.generatePrivate(new PKCS8EncodedKeySpec(pkcs8EncodedKey));
+    }
+
+    private static PrivateKey loadPrivateKeyFromPKCS1(String key,
+            String keyPass) throws IOException, NoSuchAlgorithmException, InvalidKeySpecException {
+        try {
+            Security.addProvider(new BouncyCastleProvider());
+        } catch (SecurityException se) {
+            // fall through, provider already loaded
```
saurabh500: Instead of relying on empty exception handling, is there a way to check if the provider is loaded, and use that ?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCertificateUtils.java
```diff
@@ -0,0 +1,276 @@
...
+
+            RSAPrivateCrtKeySpec spec = new RSAPrivateCrtKeySpec(modulus, publicExponent, privateExponent, prime1,
+                    prime2, primeExponent1, primeExponent2, crtCoefficient);
+            KeyFactory factory = KeyFactory.getInstance(RSA_ALG);
+            return factory.generatePrivate(spec);
+        }
+    }
+
+    private static boolean startsWithMagic(byte[] b) {
+        for (int i = 0; i < 4; i++) {
```
saurabh500: Magic number 4. can you create a constant out of it?

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -1774,13 +1782,31 @@ else if (con.getTrustManagerClass() != null) {
...
+                    KeyManager[] km = SQLServerCertificateUtils.getKeyManagerFromFile(clientCertificate, clientKey,
+                            clientKeyPassword);
+                    
+                    sslContext = SSLContext.getInstance(sslProtocol);
+                    sslContextProvider = sslContext.getProvider();
 
-            if (logger.isLoggable(Level.FINEST))
-                logger.finest(toString() + " Initializing SSL context");
+                    if (logger.isLoggable(Level.FINEST))
+                        logger.finest(toString() + " Initializing SSL context");
```
lilgreenbird: are these tabs?

src/main/java/com/microsoft/sqlserver/jdbc/ISQLServerDataSource.java
```diff
@@ -903,5 +903,50 @@
...
+     *        Private key file path.
+     */
+    void setClientKey(String keyPath);
+    
+    /**
+     * Returns the password to be used for Private key provided by the user for client certificate authentication.
+     * 
+     * @return Private key password.
+     */
+    String getClientKeyPassword();
```
lilgreenbird: should we have this? we don't usually have getters for passwords

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCertificateUtils.java
```diff
@@ -0,0 +1,274 @@
...
+
+    private static KeyManager[] readPKCS8Certificate(String certPath, String keyPath,
+            String keyPassword) throws IOException, GeneralSecurityException, SQLServerException {
+        Certificate clientCertificate = loadCertificate(certPath);
+        PrivateKey privateKey = loadPrivateKey(keyPath, keyPassword);
+
+        KeyStore keyStore = KeyStore.getInstance("JKS");
+        keyStore.load(null, null);
+        keyStore.setCertificateEntry("client-cert", clientCertificate);
+        keyStore.setKeyEntry("client-key", privateKey, keyPassword.toCharArray(),
```
lilgreenbird: are these hardcoded aliases defined in the specs? can we define them somewhere else?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCertificateUtils.java
```diff
@@ -0,0 +1,274 @@
...
+
+            buffer.position(buffer.position() + 8); // skip btype(1b), version(1b), reserved(2b), and keyalg(4b)
+
+            byte[] key = new byte[keyLength - 8];
+            buffer.get(key);
+
+            if (encrypted) {
+                MessageDigest digest = MessageDigest.getInstance("SHA1");
+                digest.update(salt);
+                if (keyPass != null) {
```
lilgreenbird: null !=

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -4960,7 +4985,7 @@ final boolean complete(LogonCommand logonCommand, TDSReader tdsReader) throws SQ
                 + 4; // AE is always on;
 
         // only add lengths of password and username if not using SSPI or requesting federated authentication info
-        if (!integratedSecurity && !(federatedAuthenticationInfoRequested || federatedAuthenticationRequested)) {
+        if (!integratedSecurity && !(federatedAuthenticationInfoRequested || federatedAuthenticationRequested) && clientCertificate == null) {
```
lilgreenbird: null ==

src/test/java/com/microsoft/sqlserver/clientcertauth/ClientCertificateAuthenticationTest.java
```diff
@@ -0,0 +1,133 @@
...
+    @Test
+    public void invalidCertPassword() throws Exception {
+        String conStr = connectionString + ";clientCertificate=" + clientCertificate + ".pfx;"
+                + "clientKeyPassword=invalid_password;";
+        try (Connection conn = DriverManager.getConnection(conStr)) {
+        } catch (SQLServerException e) {
+            assertTrue(e.getMessage().contains(TestResource.getResource("R_keystorePassword")));
+        }
+    }
+}
```
lilgreenbird: no test for datasource getters and setters
I think you also need to update AbstractTest.updateDataSource()

src/test/java/com/microsoft/sqlserver/clientcertauth/ClientCertificateAuthenticationTest.java
```diff
@@ -0,0 +1,133 @@
...
+public class ClientCertificateAuthenticationTest extends AbstractTest {
+
+    @Test
+    public void pkcs1Test() throws Exception {
+        String conStr = connectionString + ";clientCertificate=" + clientCertificate + ".pem;" + "clientKey="
+                + clientKey + "-pkcs1.key;";
+        try (Connection conn = DriverManager.getConnection(conStr); Statement stmt = conn.createStatement()) {
+            ResultSet rs = stmt.executeQuery("SELECT @@VERSION AS 'SQL Server Version'");
+            rs.next();
+            assertTrue(rs.getString(1).contains(TestResource.getResource("R_microsoft")));
```
lilgreenbird: ? why are we only checking if it contains the string Microsoft?

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
peterbae: Will fix.

src/main/java/com/microsoft/sqlserver/jdbc/ISQLServerDataSource.java
peterbae: Good point, I'll remove this one.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCertificateUtils.java
peterbae: Will do.

src/test/java/com/microsoft/sqlserver/clientcertauth/ClientCertificateAuthenticationTest.java
peterbae: Well...it can check any part of the string that @@VERSION will return. I just decided to use the keyword microsoft.

src/test/java/com/microsoft/sqlserver/clientcertauth/ClientCertificateAuthenticationTest.java
peterbae: Thanks. I added a test for data source.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCertificateUtils.java
rene-ye: Hi @saurabh500, in my opinion this is the best way to handle loading this dependency. The empty catch is there in case the user has loaded BouncyCastle themselves, and this is more efficient and reliable then trying to determine whether they've loaded BC or not.

pom.xml
```diff
@@ -122,6 +123,14 @@
...
+			<groupId>org.bouncycastle</groupId>
+			<artifactId>bcpkix-jdk15on</artifactId>
+			<version>1.64</version>
```
ulvii: create a variable for the version and use it for both bcpkix-jdk15on and bcprov-jdk15on

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -1774,13 +1782,31 @@ else if (con.getTrustManagerClass() != null) {
...
-            if (logger.isLoggable(Level.FINEST))
-                logger.finest(toString() + " Initializing SSL context");
+                    sslContext = SSLContext.getInstance(sslProtocol);
+                    sslContextProvider = sslContext.getProvider();
+
+                    if (logger.isLoggable(Level.FINEST))
+                        logger.finest(toString() + " Initializing SSL context");
+
+                    sslContext.init(km, tm, null);
+                } catch (NullPointerException | FileNotFoundException e) {
```
ulvii: Catching NullPointerException ???

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -1774,13 +1782,31 @@ else if (con.getTrustManagerClass() != null) {
             if (logger.isLoggable(Level.FINEST))
                 logger.finest(toString() + " Getting TLS or better SSL context");
 
-            sslContext = SSLContext.getInstance(sslProtocol);
-            sslContextProvider = sslContext.getProvider();
+            if (null != clientCertificate) {
```
ulvii: The code in if/else are almost the same. You just need to initialize `km` to `null` and have `km = SQLServerCertificateUtils.getKeyManagerFromFile(clientCertificate, clientKey,clientKeyPassword);` inside if block.

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
peterbae: We forgot to remove this exception, I've added null checks in the finally blocks inside the SQLServerCertificateUtils.java. We shouldn't need to catch NPE.

pom.xml
peterbae: done.

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
peterbae: thanks. done

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -1880,6 +1891,8 @@ else if (con.getTrustManagerClass() != null) {
                     && (SQLServerException.getErrString("R_truncatedServerResponse").equals(errMsg)
                             || SQLServerException.getErrString("R_truncatedServerResponse").equals(causeErrMsg))) {
                 con.terminate(SQLServerException.DRIVER_ERROR_INTERMITTENT_TLS_FAILED, form.format(msgArgs), e);
+            } else if (e instanceof FileNotFoundException) {
```
ulvii: This exception to be handled from the method that reads from the file, not here.

peterbae: ok, I've reverted it to the way it was being handled before.

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -1597,9 +1598,16 @@ private void validateServerNameInCertificate(X509Certificate cert) throws Certif
      *        Server Host Name for SSL Handshake
      * @param port
      *        Server Port for SSL Handshake
+     * @param clientCertificate
+     *        Client certificate path
```
ulvii: Is this absolute path? Can applications also provide relative path?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCertificateUtils.java
```diff
@@ -0,0 +1,300 @@
...
+        FileInputStream fis = null;
+        DataInputStream dis = null;
+        try {
+            fis = new FileInputStream(fname);
+            dis = new DataInputStream(fis);
+            byte[] bytes = new byte[dis.available()];
+            dis.readFully(bytes);
+            ByteArrayInputStream bais = new ByteArrayInputStream(bytes);
+            return bais;
+        } catch (FileNotFoundException e) {
```
ulvii: I think it makes sense to catch IOException here and remove IOException from throws.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCertificateUtils.java
ulvii: - SecurityException is thrown if a security manager exists and its method denies access to add a new provider. So there can be multiple reasons why this exception is thrown, not only when the provider exists.
- Why is this approach better than just checking if provider is loaded using `Security.getProvider(name)`?

src/test/java/com/microsoft/sqlserver/clientcertauth/ClientCertificateAuthenticationTest.java
ulvii: I would suggest to use the existing `conn.isValid(timeout);` API.

src/test/java/com/microsoft/sqlserver/clientcertauth/ClientCertificateAuthenticationTest.java
```diff
@@ -0,0 +1,172 @@
...
+        }
+    }
+
+    @Test
+    public void invalidCert() throws Exception {
+        String conStr = connectionString + ";clientCertificate=invalid_path;" + "clientKeyPassword=" + clientKeyPassword
+                + ";";
+        try (Connection conn = DriverManager.getConnection(conStr)) {
+        } catch (SQLServerException e) {
+            assertTrue(e.getMessage().contains(TestResource.getResource("R_invalidPath")));
```
ulvii: You can use `R_clientCertError` from SQLServerResource.

src/test/java/com/microsoft/sqlserver/clientcertauth/ClientCertificateAuthenticationTest.java
```diff
@@ -0,0 +1,172 @@
...
+            ResultSet rs = stmt.executeQuery("SELECT @@VERSION AS 'SQL Server Version'");
+            rs.next();
+            assertTrue(rs.getString(1).contains(TestResource.getResource("R_microsoft")));
+        }
+    }
+    
+    @Test
+    public void pkcs1EncryptedTest() throws Exception {
+        String conStr = connectionString + ";clientCertificate=" + clientCertificate + ".pem;" + "clientKey="
+                + clientKey + "-encrypted-pkcs1.key;" + "clientKeyPassword=" + clientKeyPassword + ";";
```
ulvii: We should not be hardcoding the file names like this. Please use config properties file and specify full file names in it.

src/test/java/com/microsoft/sqlserver/clientcertauth/ClientCertificateAuthenticationTest.java
```diff
@@ -0,0 +1,172 @@
...
+                + "clientKeyPassword=invalid_password;";
+        try (Connection conn = DriverManager.getConnection(conStr)) {
+        } catch (SQLServerException e) {
+            assertTrue(e.getMessage().contains(TestResource.getResource("R_keystorePassword")));
+        }
+    }
+    
+    @Test
+    public void testDataSource() throws Exception {
+        String conStr = connectionString + ";clientCertificate=" + clientCertificate + ".pem;" + "clientKey="
```
ulvii: Why aren't you using new DataSource APIs? setClientCertificate(), setClientKey(), setClientKeyPassword

src/test/java/com/microsoft/sqlserver/clientcertauth/ClientCertificateAuthenticationTest.java
```diff
@@ -0,0 +1,172 @@
...
+        }
+    }
+
+    @Test
+    public void invalidCertPassword() throws Exception {
+        String conStr = connectionString + ";clientCertificate=" + clientCertificate + ".pfx;"
+                + "clientKeyPassword=invalid_password;";
+        try (Connection conn = DriverManager.getConnection(conStr)) {
+        } catch (SQLServerException e) {
+            assertTrue(e.getMessage().contains(TestResource.getResource("R_keystorePassword")));
```
ulvii: Use `R_pvkParseError` from SQLServerResource instead.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -2629,7 +2654,7 @@ void Prelogin(String serverName, int portNumber) throws SQLServerException {
                 0, 0, 0, 0, 0, 0,
 
                 // - Encryption -
-                requestedEncryptionLevel,
+                (null == clientCertificate) ? requestedEncryptionLevel : (byte) (requestedEncryptionLevel | (byte) 0x80),
```
lilgreenbird: define this constant

src/test/java/com/microsoft/sqlserver/clientcertauth/ClientCertificateAuthenticationTest.java
```diff
@@ -0,0 +1,172 @@
...
+ */
+@RunWith(JUnitPlatform.class)
+@Tag(Constants.xSQLv12)
+@Tag(Constants.xSQLv14)
+@Tag(Constants.xAzureSQLDW)
+@Tag(Constants.xAzureSQLDB)
+@Tag(Constants.clientCertAuth)
+public class ClientCertificateAuthenticationTest extends AbstractTest {
+
+    @Test
```
lilgreenbird: add description of what the tests are testing

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
peterbae: It accepts both relative and absolute path.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
peterbae: done

src/test/java/com/microsoft/sqlserver/clientcertauth/ClientCertificateAuthenticationTest.java
peterbae: done

src/test/java/com/microsoft/sqlserver/clientcertauth/ClientCertificateAuthenticationTest.java
peterbae: Why do we need to use a config file? The user can only provide one connection property for `clientCertificate`, and that one property needs to be re-used multiple times for different private key formats. The hardcoded strings here are the formats for the names that the test has defined. I can turn these hardcoded strings into constants, but I wouldn't use a config file here.

src/test/java/com/microsoft/sqlserver/clientcertauth/ClientCertificateAuthenticationTest.java
peterbae: SQLServerResource is not visible in this package, we can just use the TestResource for test related texts.

ulvii: See SQLServerConnectionTest.executeInvalidFmt() for example use.

`assertTrue(e.getMessage().matches(TestUtils.formatErrorMsg("R_invalidArgument")));`

peterbae: Thanks. done.

src/test/java/com/microsoft/sqlserver/clientcertauth/ClientCertificateAuthenticationTest.java
```diff
@@ -0,0 +1,227 @@
...
+    /**
+     * Tests client certificate authentication feature with PKCS1 private key that has been encrypted with a password.
+     * 
+     * @throws Exception
+     */
+    @Test
+    public void pkcs1EncryptedTest() throws Exception {
+        String conStr = connectionString + ";clientCertificate=" + clientCertificate + PEM_SUFFIX + "clientKey="
+                + clientKey + ENCRYPTED_PKCS1_KEY_SUFFIX + "clientKeyPassword=" + clientKeyPassword + ";";
+        try (Connection conn = DriverManager.getConnection(conStr); Statement stmt = conn.createStatement()) {
```
ulvii: No need to create a statement

peterbae: Thanks, done.

src/test/java/com/microsoft/sqlserver/clientcertauth/ClientCertificateAuthenticationTest.java
peterbae: done

src/test/java/com/microsoft/sqlserver/clientcertauth/ClientCertificateAuthenticationTest.java
ulvii: Why is this resolved?

peterbae: For this error message, the error message isn't coming from the driver, it's coming from Java. So there's no SQLServerResource text to refer here.

PR #1285, opened 2020-03-18
Provide an option to configure trusted Azure Key Vault endpoints

ulvii:
- The driver will read from `mssql-jdbc.properties` file from the current working directory.
- Set `AKVTrustedEndpoints` property to a semicolon-delimeted list. If the value begins with a semicolon, it extends the default list; otherwise, it replaces it.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerColumnEncryptionAzureKeyVaultProvider.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerColumnEncryptionAzureKeyVaultProvider.java
@@ -7,6 +7,8 @@ package com.microsoft.sqlserver.jdbc;
 
 import static java.nio.charset.StandardCharsets.UTF_16LE;
 
+import java.io.FileInputStream;
+import java.io.IOException;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.nio.ByteBuffer;
@@ -14,8 +16,12 @@ import java.nio.ByteOrder;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.text.MessageFormat;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Locale;
+import java.util.Properties;
 import java.util.concurrent.ExecutorService;
+import java.util.logging.Level;
 
 import com.microsoft.azure.AzureResponseBuilder;
 import com.microsoft.azure.keyvault.KeyVaultClient;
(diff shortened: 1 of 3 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerColumnEncryptionAzureKeyVaultProvider.java
```diff
@@ -628,4 +633,58 @@ public boolean verifyColumnMasterKeyMetadata(String masterKeyPath, boolean allow
...
+         * List of Azure trusted endpoints
+         * https://docs.microsoft.com/en-us/azure/key-vault/key-vault-secure-your-key-vault
+         */
+        if (append) {
+            trustedEndpoints.add("vault.azure.net");
+            trustedEndpoints.add("vault.azure.cn");
+            trustedEndpoints.add("vault.usgovcloudapi.net");
+            trustedEndpoints.add("vault.microsoftazure.de");
+        }
+        System.out.println(trustedEndpoints);
```
peterbae: remove this

peterbae: other than this line, the PR looks good to me.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerColumnEncryptionAzureKeyVaultProvider.java
```diff
@@ -628,4 +633,57 @@ public boolean verifyColumnMasterKeyMetadata(String masterKeyPath, boolean allow
...
+    }
+
+    /**
+     * Attempt to read MSSQL_JDBC_PROPERTIES.
+     *
+     * @return corresponding Properties object or null if failed to read the file.
+     */
+    private static Properties getMssqlJdbcProperties() {
+        Properties props = null;
+        try (FileInputStream in = new FileInputStream(MSSQL_JDBC_PROPERTIES)) {
```
David-Engel: What path are we loading the file from here? Is this a potential attack vector?

ulvii: File needs to be in application working directory, where JVM is started from.

src/test/java/com/microsoft/sqlserver/testframework/AbstractTest.java
```diff
@@ -145,8 +148,22 @@ public static void setup() throws Exception {
         }
 
         if (null == akvProvider) {
-            akvProvider = new SQLServerColumnEncryptionAzureKeyVaultProvider(applicationClientID, applicationKey);
-            map.put(Constants.AZURE_KEY_VAULT_NAME, akvProvider);
+            File file = null;
+            try {
+                file = new File("mssql-jdbc.properties");
```
peterbae: Why did we define MSSQL_JDBC_PROPERTIES and AKV_TRUSTED_ENDPOINTS_KEYWORD as constants and not use them?

ulvii: weird.

Issue #1069, opened 2019-05-27
SQL Server Always Encrypted feature with KeyVault and MSI


venkatareddym:
## Question

We are trying to use SQL Server Always Encrypted feature with KeyVault and MSI. Azure App Service MSI feature is used on the database on KeyVault. Please provide an example without keyvault ClientId and ClientSecret.

We have a working solution based on keyvault ClientId and ClientSecret, But we are checking a solution without using  ClientId and ClientSecret. 

```
SQLServerColumnEncryptionAzureKeyVaultProvider akvProvider = new SQLServerColumnEncryptionAzureKeyVaultProvider(alwaysOnEncyrptionClientId, alwaysOnEncyrptionClientSecret);
	    Map<String, SQLServerColumnEncryptionKeyStoreProvider> keyStoreMap = new HashMap<>();
	    keyStoreMap.put(akvProvider.getName(), akvProvider);
	    SQLServerConnection.registerColumnEncryptionKeyStoreProviders(keyStoreMap);
```

cheenamalhotra:
Hi @venkatareddym 

In order to work with Always Encrypted feature, you can choose one of three methods:


Documentation Reference:

venkatareddym:
We already using an Always Encrypted feature with Key Vault with ClientId and ClientSecret. We want to eliminate the ClientId and ClientSecret with the Azure app services MSI feature.  For SQL server we already using Azure App services MSI feature. 

Our main goal is to eliminate the secrets and passwords on config files.  We are checking examples for SQLServerKeyVaultAuthenticationCallback with Keyvalut MSI ( Managed Service Identity )features. 

https://github.com/microsoft/mssql-jdbc/blob/master/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerColumnEncryptionAzureKeyVaultProvider.java

cheenamalhotra:
Hi @venkatareddym 

Yes, that is not supported by driver yet. As of now MSI is only supported for acquiring access token for Azure database connection. I will keep the issue open until we support Azure Key Vault access with MSI in the driver.

ronneyramon:
Hi,

@venkatareddym, you can get the Azure Key Vault access token using the AzureServiceTokenProvider (from Microsoft.Azure.Services.AppAuthentication 1.2.0 preview).

`SqlColumnEncryptionAzureKeyVaultProvider azureKeyVaultProvider =
[...]

PR #1286, opened 2020-03-18
Add support to use Managed Identity for AKV authentication

lilgreenbird:
addresses issue [1069](https://github.com/microsoft/mssql-jdbc/issues/1069).

New connection properties added to support to authenticate to AKV using Managed Identities:
```
KeyStoreAuthentication=KeyVaultClientSecret|KeyVaultManagedIdentity
keyStorePrincipalId=<valid Azure Directory Application Client ID>
```
and use existing connection property:
```
keyStoreSecret= <key store secret>
```


pom.xml
```diff
@@ -94,6 +95,11 @@
 			<version>${azure.adal4j.version}</version>
 			<optional>true</optional>
 		</dependency>
+		<dependency>
+			<groupId>com.microsoft.azure</groupId>
+			<artifactId>azure-client-authentication</artifactId>
+			<version>1.7.2</version>
```
ulvii: Please check how version is specified for other dependencies.

src/main/java/com/microsoft/sqlserver/jdbc/KeyVaultCredential.java
```diff
@@ -37,11 +55,19 @@
...
+        String accessToken = null;
         if (null == authenticationCallback) {
-            AuthenticationResult token = getAccessTokenFromClientCredentials(authorization, resource, clientId,
-                    clientKey);
-            accessToken = token.getAccessToken();
+            if (null != msiCred) {
+                try {
+                    accessToken = msiCred.getToken(resource);
+                } catch (IOException e) {
+                    throw new RuntimeException(e);
```
peterbae: Why are we wrapping the exception into a RuntimException here? Shouldn't we wrap the exception into a SQLServerException instead?

lilgreenbird: should we? I used RuntimeException to be consistent with what we're currently doing already.  getAccessTokenFromClientCredentials throws RuntimeException which just gets passed up to the caller here

peterbae: ok, I see that it's being done in the other method as well...but I think we usually wrap any exception into SQLServerException. But I'm ok with following the previous code.

src/test/java/com/microsoft/sqlserver/jdbc/AlwaysEncrypted/MSITest.java
```diff
@@ -0,0 +1,166 @@
+package com.microsoft.sqlserver.jdbc.AlwaysEncrypted;
```
peterbae: Add license header

src/test/java/com/microsoft/sqlserver/jdbc/AlwaysEncrypted/AECommon.java
```diff
@@ -0,0 +1,302 @@
+package com.microsoft.sqlserver.jdbc.AlwaysEncrypted;
```
peterbae: put license header here

src/test/java/com/microsoft/sqlserver/jdbc/AlwaysEncrypted/MSITest.java
```diff
@@ -0,0 +1,166 @@
...
+                keyStorePrincipalId);
+        AETestConnectionString = TestUtils.addOrOverrideProperty(AETestConnectionString, Constants.KEYSTORE_SECRET,
+                keyStoreSecret);
+        testCharAkv();
+    }
+
+    /*
+     * Test AKV with MSI
+     */
+    public void testCharAkvWithMSI() throws SQLException {
```
peterbae: needs @Test tag

src/test/java/com/microsoft/sqlserver/jdbc/AlwaysEncrypted/MSITest.java
```diff
@@ -0,0 +1,171 @@
...
+    /*
+     * Test AKV with with credentials
+     */
+    @Test
+    public void testCharAkvWithCred() throws SQLException {
+        // unregister the custom providers registered in AESetup
+        SQLServerConnection.unregisterColumnEncryptionKeyStoreProviders();
+
+        // add credentials to connection string
+        AETestConnectionString = TestUtils.addOrOverrideProperty(AETestConnectionString,
```
peterbae: I think currently your testing might not be testing the correct inputs from the user. This AETestConnectionString is a static variable that keeps getting added more parameters as individual tests get executed, resulting in latter tests having the same connection property defined multiple times. You should reset this variable to the base connection string at the beginning of each test run.

src/main/java/com/microsoft/sqlserver/jdbc/ISQLServerDataSource.java
```diff
@@ -821,6 +821,21 @@
      */
     String getMSIClientId();
 
+    /**
```
ulvii: Aren't you adding 2 new connection properties?

lilgreenbird: not anymore, latest discussion with Jakub we're aligning with ODBC so using keyStorePrincipalId (new) and keyStoreSecret (existing)

src/main/java/com/microsoft/sqlserver/jdbc/ISQLServerDataSource.java
```diff
@@ -821,6 +821,21 @@
...
+     * Sets the Azure Key Vault (AKV) Provider user principal id.
```
ulvii: This is not necessarily AKV user principal ID. Please align the docs with [ODBC connection property documentation.](https://docs.microsoft.com/en-us/sql/connect/odbc/using-always-encrypted-with-the-odbc-driver?view=sql-server-ver15#using-the-azure-key-vault-provider)

src/main/java/com/microsoft/sqlserver/jdbc/ISQLServerDataSource.java
```diff
@@ -821,6 +821,21 @@
...
+     * 
+     * @param keyVaultPrincipalId
+     *        principal Id of Azure Key Vault (AKV) Provider to be used for column encryption.
```
ulvii: Same here.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -720,6 +722,23 @@ public static synchronized void registerColumnEncryptionKeyStoreProviders(
...
+     */
+    public static synchronized void unregisterColumnEncryptionKeyStoreProviders() {
+        loggerExternal.entering(SQLServerConnection.class.getName(), "unregisterColumnEncryptionKeyStoreProviders",
+                "Removing Column Encryption Key Store Provider");
+
+        if (null != globalCustomColumnEncryptionKeyStoreProviders) {
+            globalCustomColumnEncryptionKeyStoreProviders.clear();
+        }
+
+        loggerExternal.exiting(SQLServerConnection.class.getName(), "unregisterColumnEncryptionKeyStoreProviders",
```
rene-ye: Please do not do this class.getName() in logger, it traverses the stack and is extremely costly performance-wise. I also recommend checking the logging level before calling these entering/exiting methods, although I'm not sure if the intent here is to always log. As for the class.forName(), just use a static string, so SQLServerConnection.class.getName() would be "SQLServerConnection".

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -1291,7 +1316,30 @@ private void registerKeyStoreProviderOnConnection(String keyStoreAuth, String ke
...
+                    // need a secret use use the secret method
+                    if (null == keyStoreSecret) {
+                        throw new SQLServerException(
+                                SQLServerException.getErrString("R_keyStoreSecretOrLocationNotSet"), null);
+                    } else {
+                        SQLServerColumnEncryptionAzureKeyVaultProvider provider = new SQLServerColumnEncryptionAzureKeyVaultProvider(
+                                keyStorePrincipalId, keyStoreSecret);
+                        Map<String, SQLServerColumnEncryptionKeyStoreProvider> keyStoreMap = new HashMap<String, SQLServerColumnEncryptionKeyStoreProvider>();
+                        keyStoreMap.put(provider.getName(), provider);
+                        registerColumnEncryptionKeyStoreProviders(keyStoreMap);
```
rene-ye: Feels like a waste to create a hash-map w/ 1 entry just to conform to the existing constructor. Maybe this can be done better?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerSecurityUtility.java
```diff
@@ -215,4 +234,164 @@ static void verifyColumnMasterKeyMetadata(SQLServerConnection connection, String
...
+                // Below code applicable only when !isAzureFunctcion (VM)
+                if (retry > maxRetry) {
+                    // Do not retry if maxRetry limit has been reached.
+                    break;
+                } else {
+                    try {
+                        int responseCode = connection.getResponseCode();
+                        // Check Error Response Code from Connection
+                        if (410 == responseCode || 429 == responseCode || 404 == responseCode
+                                || (500 <= responseCode && 599 >= responseCode)) {
```
rene-ye: Response codes shouldn't be magic numbers, better to create constants and document a little of their meaning in comments. I think this part of the code is just moved though so please document/create a work item.

src/main/java/com/microsoft/sqlserver/jdbc/ISQLServerDataSource.java
```diff
@@ -821,6 +821,21 @@
      */
     String getMSIClientId();
 
+    /**
+     * Sets the Azure Active Directory Application Client ID.
```
ulvii: From ODBC docs:

When KeyStoreAuthentication = KeyVaultPassword, set this value to a valid Azure Active Directory User Principal Name.
When KeyStoreAuthetication = KeyVaultClientSecret set this value to a valid Azure Active Directory Application Client ID

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -720,6 +722,23 @@ public static synchronized void registerColumnEncryptionKeyStoreProviders(
                         + globalCustomColumnEncryptionKeyStoreProviders.size());
     }
 
+    /**
+     * Unregisters the custom key store providers from the globalCustomColumnEncryptionKeyStoreProviders.
```
ulvii: Please add proper javadoc, this is a public API.

src/main/java/com/microsoft/sqlserver/jdbc/ISQLServerDataSource.java
ulvii: Please set Javadoc to this

```
    /**
     * Sets the value for the connection property 'keyStorePrincipalId'.
     * 
     * @param keyStorePrincipalId
     * 
     *        <pre>
     *        When keyStoreAuthentication = keyVaultPassword, set this value to a valid Azure Active Directory User Principal Name.
     *        When keyStoreAuthentication = keyVaultClientSecret, set this value to a valid Azure Active Directory Application Client ID.
     *        When keyStoreAuthentication = keyVaultManagedIdentity, set this value to a valid Azure Active Directory Application Object ID (optional, for user-assigned only).
     *        </pre>
     */
```

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -1291,7 +1301,30 @@ private void registerKeyStoreProviderOnConnection(String keyStoreAuth, String ke
                         systemColumnEncryptionKeyStoreProvider.put(provider.getName(), provider);
                     }
                     break;
-
+                case KeyVaultClientSecret:
+                    // need a secret use use the secret method
+                    if (null == keyStoreSecret) {
+                        throw new SQLServerException(
+                                SQLServerException.getErrString("R_keyStoreSecretOrLocationNotSet"), null);
```
ulvii: `R_keyStoreSecretOrLocationNotSet` message does not make sense for this future.

src/test/java/com/microsoft/sqlserver/jdbc/AlwaysEncrypted/MSITest.java
```diff
@@ -0,0 +1,189 @@
...
+ */
+@RunWith(JUnitPlatform.class)
+@Tag(Constants.MSI)
+public class MSITest extends AESetup {
+    /*
+     * Test basic MSI auth
+     */
+    @Test
+    public void testAuth() throws SQLException {
+        try (SQLServerConnection con = PrepUtil.getConnection(connectionString)) {} catch (Exception e) {
```
ulvii: How is this testing MSI auth? You are using the global connection stirng.

src/test/java/com/microsoft/sqlserver/jdbc/AlwaysEncrypted/MSITest.java
```diff
@@ -0,0 +1,189 @@
...
+        }
+    }
+
+    /*
+     * Test MSI auth using datasource
+     */
+    @Test
+    public void testDSAuth() throws SQLException {
+        SQLServerDataSource ds = new SQLServerDataSource();
+        AbstractTest.updateDataSource(connectionString, ds);
```
ulvii: Same here, how is this testing MSI auth?

src/test/java/com/microsoft/sqlserver/jdbc/AlwaysEncrypted/MSITest.java
```diff
@@ -0,0 +1,189 @@
...
+                                                        : (SQLServerResultSet) stmt.executeQuery(sql)) {
+                int numberOfColumns = rs.getMetaData().getColumnCount();
+                while (rs.next()) {
+                    AECommon.testGetString(rs, numberOfColumns, values);
+                    AECommon.testGetObject(rs, numberOfColumns, values);
+                    AECommon.testGetBigDecimal(rs, numberOfColumns, values);
+                    AECommon.testWithSpecifiedtype(rs, numberOfColumns, values);
+                }
+            } catch (Exception e) {
+                fail(TestResource.getResource("R_loginFailed") + e.getMessage());
```
ulvii: Login failed? This catch block implies that, the only exception that can happen inside try block is due to a login failure. Just remove the catch block and let the test throw the original exception.

src/test/java/com/microsoft/sqlserver/jdbc/AlwaysEncrypted/MSITest.java
```diff
@@ -0,0 +1,189 @@
...
+     * Test AKV with with bad credentials
+     */
+    @Test
+    public void testNumericAkvWithBadCred() throws SQLException {
+        // unregister the custom providers registered in AESetup
+        SQLServerConnection.unregisterColumnEncryptionKeyStoreProviders();
+
+        // add credentials to connection string
+        String connStr = AETestConnectionString;
+        connStr = TestUtils.addOrOverrideProperty(connStr, Constants.KEYSTORE_AUTHENTICATION, "KeyVaultClientSecret");
```
ulvii: I would prefer to use `KeyVaultManagedIdentity` here, that is what the feature is about.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerSecurityUtility.java
```diff
@@ -5,21 +5,46 @@
...
 /**
  * Various SQLServer security utilities.
  *
  */
 class SQLServerSecurityUtility {
+    static final private java.util.logging.Logger connectionlogger = java.util.logging.Logger
+            .getLogger("com.microsoft.sqlserver.jdbc.internals.SQLServerConnection");
+
+    static final int GONE = 410;
+    static final int TOO_MANY_RESQUESTS = 429;
```
rene-ye: Requests

src/test/java/com/microsoft/sqlserver/jdbc/AlwaysEncrypted/MSITest.java
```diff
@@ -0,0 +1,257 @@
...
+public class MSITest extends AESetup {
+
+    /*
+     * Test MSI auth
+     */
+    @Tag(Constants.xSQLv12)
+    @Tag(Constants.xSQLv14)
+    @Tag(Constants.xSQLv15)
+    @Test
+    public void testMSIAuth() throws SQLException {
```
ulvii, line 47: Please also add similar tests that test ActiveDirectoryMSI along with msiClientId

src/test/java/com/microsoft/sqlserver/jdbc/AlwaysEncrypted/MSITest.java
```diff
@@ -0,0 +1,307 @@
...
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Statement;
+
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+import org.junit.platform.runner.JUnitPlatform;
+import org.junit.runner.RunWith;
+
+import com.microsoft.aad.adal4j.AuthenticationException;
```
ulvii, line 19: There are unused imports here

src/test/java/com/microsoft/sqlserver/jdbc/AlwaysEncrypted/MSITest.java
```diff
@@ -0,0 +1,307 @@
...
+    public void testMSIAuth() throws SQLException {
+        // unregister the custom providers registered in AESetup
+        SQLServerConnection.unregisterColumnEncryptionKeyStoreProviders();
+
+        String connStr = connectionString;
+        connStr = TestUtils.addOrOverrideProperty(connStr, Constants.USER, "");
+        connStr = TestUtils.addOrOverrideProperty(connStr, Constants.PASSWORD, "");
+        connStr = TestUtils.addOrOverrideProperty(connStr, Constants.AUTHENTICATION, "ActiveDirectoryMSI");
+
+        try (SQLServerConnection con = PrepUtil.getConnection(connectionString)) {} catch (Exception e) {
```
ulvii, line 56: aren't you supposed to use connStr ?

lilgreenbird:
hi all, as of the latest [8.3.0 preview release](https://github.com/microsoft/mssql-jdbc/releases/tag/v8.3.0) the driver added support for authentication to AKV using Managed Identity.  Here is a [wiki](https://github.com/microsoft/mssql-jdbc/wiki/Key-Vault-authentication-with-Managed-Identities) on how to use this feature. Please give this a try and let us know if you have any questions.

PR #1289, opened 2020-03-25
Release | Update version to 8.3.0-SNAPSHOT

ulvii:


pom.xml
```diff
@@ -6,7 +6,7 @@
```
peterbae: Do we not add -preview on line 232, 278 and 314?

ulvii: releaseExt takes care of that

PR #1306, opened 2020-04-06
Socket Factory Test Fix

peterbae:
Connection attempt can be made multiple times depending on the environment (such as URL redirection is involved in the case of the client being inside Azure network), so remove this part of the test.

### DIFF
--- a/src/test/java/com/microsoft/sqlserver/jdbc/socketfactory/CustomSocketFactoryTest.java
+++ b/src/test/java/com/microsoft/sqlserver/jdbc/socketfactory/CustomSocketFactoryTest.java
@@ -91,7 +91,6 @@ public class CustomSocketFactoryTest extends AbstractTest {
         try (Connection con = PrepUtil.getConnection(url)) {
             Assert.assertTrue(con != null);
         }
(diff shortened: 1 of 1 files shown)


src/test/java/com/microsoft/sqlserver/jdbc/socketfactory/CustomSocketFactoryTest.java
```diff
@@ -91,7 +91,6 @@ public void testCustomSocketFactory() throws Exception {
         try (Connection con = PrepUtil.getConnection(url)) {
             Assert.assertTrue(con != null);
         }
-        Assert.assertEquals("The custom socket factory should have been used once", 1, dummyLog.size());
```
lilgreenbird, line 94: is this a valid test when it's not AzureDB? If so, add xAzureSQLDB tag instead?

peterbae: It's not about whether the test server is against AzureDB, it has to do with the environment in which the test is performed. I don't think we should be adding a tag for this, and the number of times the custom socket factory being called is a trivial test.

PR #1308, opened 2020-04-09
MSI expiry date format fix for Azure Functions

peterbae:
Fixes issue #1135.

The format in which the expires_on property is sent to the driver from Azure server used to be in the form of a string, but starting from the latest API version (2019-08-01), this value is now an integer value, which is consistent across all regions and operating systems (Windows/Linux).

Note that this fix only applies to retrieving access tokens for MSI authentication from Azure Functions (and Azure Web Apps), and it doesn't affect the usual Azure IMDS routine for MSI authentication.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerSecurityUtility.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerSecurityUtility.java
@@ -309,7 +309,7 @@ class SQLServerSecurityUtility {
                 try (InputStream stream = connection.getInputStream()) {
 
                     BufferedReader reader = new BufferedReader(new InputStreamReader(stream, UTF_8), 100);
-                    String result = reader.readLine();
+                    StringBuilder result = new StringBuilder(reader.readLine());
 
                     int startIndex_AT = result.indexOf(ActiveDirectoryAuthentication.ACCESS_TOKEN_IDENTIFIER)
                             + ActiveDirectoryAuthentication.ACCESS_TOKEN_IDENTIFIER.length();
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerSecurityUtility.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerSecurityUtility.java
@@ -15,9 +15,7 @@ import java.net.HttpURLConnection;
 import java.net.URL;
 import java.security.InvalidKeyException;
 import java.security.NoSuchAlgorithmException;
-import java.text.DateFormat;
 import java.text.MessageFormat;
-import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Calendar;
 import java.util.Date;
@@ -270,7 +268,7 @@ class SQLServerSecurityUtility {
                 && !msiSecret.isEmpty();
 
         if (isAzureFunction) {
-            urlString.append(msiEndpoint).append("?api-version=2017-09-01&resource=").append(resource);
+            urlString.append(msiEndpoint).append("?api-version=2019-08-01&resource=").append(resource);
         } else {
             urlString.append(ActiveDirectoryAuthentication.AZURE_REST_MSI_URL).append("&resource=").append(resource);
             // Retry acquiring access token upto 20 times due to possible IMDS upgrade (Applies to VM only)
@@ -283,11 +281,7 @@ class SQLServerSecurityUtility {
 
         // Append Client Id if available
         if (null != msiClientId && !msiClientId.isEmpty()) {
-            if (isAzureFunction) {
-                urlString.append("&clientid=").append(msiClientId);
-            } else {
-                urlString.append("&client_id=").append(msiClientId);
-            }
+            urlString.append("&client_id=").append(msiClientId);
         }
 
         // Loop while maxRetry reaches its limit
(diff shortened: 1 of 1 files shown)


peterbae:
[github-1135.zip](https://github.com/microsoft/mssql-jdbc/files/4457675/github-1135.zip)

Attaching the latest driver jars with this fix included.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerSecurityUtility.java
```diff
@@ -324,30 +318,23 @@ static SqlFedAuthToken getMSIAuthToken(String resource, String msiClientId) thro
...
                                 .indexOf(ActiveDirectoryAuthentication.ACCESS_TOKEN_EXPIRES_IN_IDENTIFIER)
                                 + ActiveDirectoryAuthentication.ACCESS_TOKEN_EXPIRES_IN_IDENTIFIER.length();
-                        String accessTokenExpiry = result.substring(startIndex_ATX,
-                                result.indexOf("\"", startIndex_ATX + 1));
-                        cal.add(Calendar.SECOND, Integer.parseInt(accessTokenExpiry));
                     }
 
+                    String accessTokenExpiry = result.substring(startIndex_ATX,
+                            result.indexOf("\"", startIndex_ATX + 1));
+                    cal.add(Calendar.SECOND, Integer.parseInt(accessTokenExpiry));
```
rene-ye, line 336:
```java
int startIndex_ATX = isAzureFunction ? result.indexOf(ActiveDirectoryAuthentication.ACCESS_TOKEN_EXPIRES_ON_IDENTIFIER) + ActiveDirectoryAuthentication.ACCESS_TOKEN_EXPIRES_ON_IDENTIFIER.length() : result.indexOf(ActiveDirectoryAuthentication.ACCESS_TOKEN_EXPIRES_IN_IDENTIFIER) + ActiveDirectoryAuthentication.ACCESS_TOKEN_EXPIRES_IN_IDENTIFIER.length();
                    
StringBuilder sb = new StringBuilder(result);
String accessTokenExpiry = sb.subString(startIndex_ATX, sb.indexOf("\"", startIndex_ATX + 1));
cal.add(Calendar.SECOND, Integer.parseInt(accessTokenExpiry));
```

See if we can just use StringBuilder from the beginning instead of a string.

peterbae: Thanks. I used a stringbuilder for this now.

sgarcia2101:
Hi! I found that you introduce a new bug. When library use Azure function or App Service, the parameter should be clientid, without underline, but in the fix it is missed as you can see in the image below:
image

peterbae:
Hi @sgarcia2101, thank you for your input. The API version for Azure function / App Service that the driver uses has been updated to `2019-08-01` version, and in this version Azure accepts `clientid` instead of `client_id`. Please let me know if you encounter any issues using the driver related to this change.

sgarcia2101:
I dont know why, but if I use "clientid" paramete, it works, and if I use "client_id" parameter, I receive this error: {"statusCode":400,"message":"No UserAssigned Managed Identity found for specified ClientId/ResourceId/PrincipalId.","correlationId":"xxxxxxxx"}

peterbae:
I think the older versions of the API versions still accept "clientId", but the latest API version accepts "client_id". Could you show me how you're sending your HTTP GET request to the Azure endpoint? (but this would be unrelated to the driver)

Also, you could refer to this page for details: https://docs.microsoft.com/en-us/azure/app-service/overview-managed-identity?tabs=dotnet#using-the-rest-protocol

sgarcia2101:
Yes, sure.

I have this code in Java deployed in an App Service with System Assigned Managed Identity enabled. The parameters are the clientId generated by System Assigned MI and the resource is "https://database.windows.net/".

```
private String getMSIToken(String clientId, String resource) throws UnsupportedEncodingException {

	Configuration configuration = Configuration.getGlobalConfiguration().clone();

	String msiEndpoint = configuration.get(Configuration.PROPERTY_MSI_ENDPOINT);
	String msiSecret = configuration.get(Configuration.PROPERTY_MSI_SECRET);
	

	StringBuilder uriString = new StringBuilder();
	uriString.append(String.format("%s?resource=%s&api-version=%s", msiEndpoint,
			URLEncoder.encode(resource, "UTF-8"), URLEncoder.encode("2019-08-01", "UTF-8")));
	
	if(clientId != null) {
		uriString.append(String.format("&clientid=%s", URLEncoder.encode(clientId, "UTF-8")));
	}

	URI uri = URI.create(uriString.toString());

	HttpClient client = HttpClient.newHttpClient();
	HttpRequest request = HttpRequest.newBuilder()
			.uri(uri)
			.header("X-IDENTITY-HEADER", msiSecret)
			.GET().build();

	HttpResponse<String> respuesta;
	try {
		respuesta = client.send(request, BodyHandlers.ofString());
		return respuesta.body();
	} catch (Exception e) {
		logger.error("Error getting a token", e);
	}
	return null;
}
```

Then the final request is: 
```
GET /msi/token?resource=https://database.windows.net/&api-version=2019-08-01&client_id=xxxxxxx HTTP/1.1
X-IDENTITY-HEADER: xxxxxxxxx
```

peterbae:
This is an example of my GET request that works for me:

`http://127.0.0.1:41539/MSI/token/?api-version=2019-08-01&resource=https://database.windows.net/&client_id=xxxxxxx`

I think your GET request looks correct, I'm not sure why it's failing for you. Perhaps it's because in your code you showed me, you're still sending `clientid` instead of `client_id` along with `X-IDENTITY-HEADER` for your secret header? If you intend to use the latest api version of `2019-08-01`, you have to use `clientid` along with `X-IDENTITY-HEADER`, but if you use the older version, you use `client_id` along with `Secret`. Hope this helps.

sgarcia2101:
Sorry, I pasted the code that is working, because I use clientid with X-IDENTITY-HEADER. 

As you said, the latest api version 2019-08-01 works with client_id and X-IDENTITY-HEADER, but if I test with this, it is not working.

peterbae:
I see. I'm not sure if I can help you further on this front, because I think your example looks like it should work to me. You can submit a support ticket to Azure regarding this issue, if you want more help on this.

PR #1310, opened 2020-04-10
Adding Delayed Durability

dbrownems:
Adding an overload of commit() to support delayed durability.  Fixed compilation error.  Didn't add method to ISQLServerConnection interface.  Not sure if that is desirable.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
@@ -3333,17 +3333,19 @@ public class SQLServerConnection implements ISQLServerConnection, java.io.Serial
 
     @Override
     public void commit() throws SQLServerException {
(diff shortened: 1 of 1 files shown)


peterbae:
Hi @dbrownems, thank you for your contribution. I think adding an option to enable delayed durability could be useful to some users. Could you modify the commit() method so that it calls the commit(false) method to remove some code duplication here?

dbrownems:
Do you think it needs to be added to ISqlServerConnection and to the proxy type too?  Also I can't get the project to build.  Are there any better instructions?  I tried from the commandline and various IDEs, but no luck.

peterbae:
We have instructions for building the JAR files using maven in our README file (let us know if you're stuck on one of the steps). This won't need to be added to ISqlServerConnection or the proxy type.

ulvii:
Thanks for contribution @dbrownems

Issue #1173, opened 2019-10-29
SQLServerBulkCSVFileRecord should CSV library to split data


gauravjhs:
## Is your feature request related to a problem? If so, please give a short summary of the problem and how the feature would resolve it
Currently SQLServerBulkCSVFileRecord uses String split method to get data to insert in columns. CSV are generally not as simple as just splitting them using the separator. It gets frustrating when there is some text which contains separator as part of the value.

## Describe the preferred solution
SQLServerBulkCSVFileRecord should use a proper CSV library to extract data from the csv file.

## Describe alternatives you've considered
Generate CSV's using a unique separator which is not present in data(which at times becomes very difficult when data comes from user)

cogman:
Rather than pulling in a library, I'd prefer it if there was an interface to extract that data.  That way the user can provide whatever library they like to handle the parsing logic.

I say this because I'd rather not bring in a CSV parsing lib in my apps for the DB driver.  It isn't a feature I use and I'd rather not pay for it.

Could be as simple as a new constructor that takes in something that looks like this

 `interface DataExtractor extends Iterable<Object[]>{}`

From there it would be simple to make a new DataExtractor that can open any sort of file (Or io stream, or whatever) CSV or otherwise.

Perhaps, at that point, it makes sense to make a new class (rather than attaching this onto the BulkCSVFileRecord)

ulvii:
Hi @gauravjhs, @cogman, 

Thanks for suggestions. I don't think adding a new library as a dependency will magically solve the mentioned limitation without enforcing a CSV format standard. I am proposing to use something similar to [RFC4180 ](https://tools.ietf.org/html/rfc4180) for input CSV files. According to RFC4180, if a field contains the delimiter or line breaks, the field needs to be enclosed in quotes. This can be implemented in the driver without introducing a dependency. Please let me know what you think.

gauravjhs:
@ulvii 
This is a good idea. It will help in removing some shortcomings in existing version without adding any new library. The only reason for me to suggest a CSV framework was to get close to CSV format standard and driver user will not have to worry about CSV if it follows standard specifications.

cogman:
@ulvii Works for me.

I don't know what the transport protocol is for BulkCSV (is it the same as TVPs?).  I know it comes up in docs occasionally as the fastest way to get data into SQL Server.  If there was a way to do this without having to go through a file first that would be really nice.

[...]

PR #1312, opened 2020-04-14
Added an option to escape delimiters in CSV files when using bulk copy

ulvii:
Fix for #1173 

Adding `SQLServerBulkCSVFileRecord.setEscapeColumnDelimitersCSV`. When set to `true`, the following rules will apply.

- Each field may or may not be enclosed in double quotes.
- If fields are not enclosed with double quotes, then double quotes may not appear inside the fields.
- Fields containing double quotes, and delimiters should be enclosed in double quotes. 
- If double-quotes are used to enclose fields, then a double-quote appearing inside a field must be escaped by preceding it with another double quote.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCSVFileRecord.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCSVFileRecord.java
@@ -51,6 +51,7 @@ public class SQLServerBulkCSVFileRecord extends SQLServerBulkRecord implements j
 
     private boolean escapeDelimiters;
 
+    // Regex to ignore delimiter when the field is enclosed in quotes.
     private static final String escapeSplitPattern = "(?=(?:[^\"]*\"[^\"]*\")*[^\"]*$)";
 
     /*
@@ -244,16 +245,14 @@ public class SQLServerBulkCSVFileRecord extends SQLServerBulkRecord implements j
                                                                              ? escapeQuotesRFC4180(currentLine.split(
                                                                                      delimiter + escapeSplitPattern))
                                                                              : currentLine.split(delimiter, -1);
-            /*
-             * Cannot go directly from String[] to Object[] and expect it to act as an array.
-             */
+
+            // Cannot go directly from String[] to Object[] and expect it to act as an array.
+
             Object[] dataRow = new Object[data.length];
 
             for (Entry<Integer, ColumnMetadata> pair : columnMetadata.entrySet()) {
                 ColumnMetadata cm = pair.getValue();
-                /*
-                 * Reading a column not available in csv positionInFile > number of columns retrieved after split
-                 */
+                // Reading a column not available in csv positionInFile > number of columns retrieved after split
                 if (data.length < pair.getKey() - 1) {
                     MessageFormat form = new MessageFormat(SQLServerException.getErrString("R_invalidColumn"));
                     Object[] msgArgs = {pair.getKey()};
@@ -538,7 +537,7 @@ public class SQLServerBulkCSVFileRecord extends SQLServerBulkRecord implements j
      * double quotes. If fields are not enclosed with double quotes, then double quotes may not appear inside the
      * fields. Fields containing double quotes, and delimiters should be enclosed in double quotes. If double-quotes are
      * used to enclose fields, then a double-quote appearing inside a field must be escaped by preceding it with another
-     * double quote.
+     * double quote. Spaces are considered part of a field. Spaces before and after enclosing double quotes are ignored.
      *
      * @param escapeDelimiters
      *        true if the rules above to be used.
@@ -547,7 +546,7 @@ public class SQLServerBulkCSVFileRecord extends SQLServerBulkRecord implements j
         this.escapeDelimiters = escapeDelimiters;
     }
 
-    private static String[] escapeQuotesRFC4180(String[] tokens) {
+    private static String[] escapeQuotesRFC4180(String[] tokens) throws SQLServerException {
         if (null == tokens) {
             return tokens;
         }
@@ -555,7 +554,11 @@ public class SQLServerBulkCSVFileRecord extends SQLServerBulkRecord implements j
             boolean escaped = false;
             int j = 0;
             StringBuilder sb = new StringBuilder();
-            if (tokens[i].contains("\"")) {
+            long quoteCount = tokens[i].chars().filter(ch -> ch == '"').count();
+            if (0 != quoteCount % 2) {
+                throw new SQLServerException(SQLServerException.getErrString("R_InvalidCSVOddQuotes"), null, 0, null);
+            }
+            if (quoteCount > 0) {
                 tokens[i] = tokens[i].trim();
             }
             while (j < tokens[i].length()) {
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCSVFileRecord.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCSVFileRecord.java
@@ -536,9 +536,9 @@ public class SQLServerBulkCSVFileRecord extends SQLServerBulkRecord implements j
     /**
      * When set to true, the following rules will be used to parse CSV files: Each field may or may not be enclosed in
      * double quotes. If fields are not enclosed with double quotes, then double quotes may not appear inside the
-     * fields. Fields containing double quotes, and delimiters. If double-quotes are used to enclose fields, then a
-     * double-quote appearing inside a field must be escaped by preceding it with another double quote. should be
-     * enclosed in double-quotes.
+     * fields. Fields containing double quotes, and delimiters should be enclosed in double quotes. If double-quotes are
+     * used to enclose fields, then a double-quote appearing inside a field must be escaped by preceding it with another
+     * double quote.
      *
      * @param escapeDelimiters
      *        true if the rules above to be used.
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCSVFileRecord.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCSVFileRecord.java
@@ -555,11 +555,12 @@ public class SQLServerBulkCSVFileRecord extends SQLServerBulkRecord implements j
             int j = 0;
             StringBuilder sb = new StringBuilder();
             long quoteCount = tokens[i].chars().filter(ch -> ch == '"').count();
-            if (0 != quoteCount % 2) {
-                throw new SQLServerException(SQLServerException.getErrString("R_InvalidCSVOddQuotes"), null, 0, null);
-            }
             if (quoteCount > 0) {
                 tokens[i] = tokens[i].trim();
+            }
+            if (0 != quoteCount % 2 || (quoteCount > 0
+                    && ('"' != tokens[i].charAt(0) || '"' != tokens[i].charAt(tokens[i].length() - 1)))) {
+                throw new SQLServerException(SQLServerException.getErrString("R_InvalidCSVQuotes"), null, 0, null);
             }
             while (j < tokens[i].length()) {
                 if ('"' == tokens[i].charAt(j)) {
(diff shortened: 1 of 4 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCSVFileRecord.java
```diff
@@ -516,4 +523,58 @@ public boolean next() throws SQLServerException {
         }
         return (null != currentLine);
     }
+
+    /**
+     * Returns whether the rules to escape delimiters are used.
+     *
+     * @return true if the rules are used, false otherwise.
+     */
+    public boolean isEscapeColumnDelimitersCSV() {
```
peterbae, line 532: I think we should rename this to getEscapeColumnDelimitersCSV, for the setter/getter convention.

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyCSVTest.java
```diff
@@ -118,6 +134,51 @@ public void testCSVFromURL() throws SQLException {
...
+                SQLServerBulkCSVFileRecord fileRecord = new SQLServerBulkCSVFileRecord(fileName, encoding, delimiter,
+                        false)) {
+            bulkCopy.setDestinationTableName(tableName);
+            fileRecord.setEscapeColumnDelimitersCSV(true);
+            fileRecord.addColumnMetadata(1, null, java.sql.Types.INTEGER, 0, 0);
+            fileRecord.addColumnMetadata(2, null, java.sql.Types.VARCHAR, 50, 0);
+            fileRecord.addColumnMetadata(3, null, java.sql.Types.VARCHAR, 50, 0);
+            fileRecord.addColumnMetadata(4, null, java.sql.Types.VARCHAR, 50, 0);
+            fileRecord.addColumnMetadata(5, null, java.sql.Types.VARCHAR, 50, 0);
+            stmt.executeUpdate(
```
peterbae, line 164: My formatter shows that this line needs to be formatted. Can you try re-formatting this file?

src/test/resources/BulkCopyCSVTestInputDelimiterEscape.csv
```diff
@@ -0,0 +1,5 @@
+1,"test", " test""", "no,split" , testNoQuote
```
peterbae, line 1: Cases such as ` " test"""` and ` "test""test"` are actually invalid cases. RFC 4180 says that spaces need to be insides double quotes if double quotes are provided, so spaces cannot come before or after the double quotes (refer to https://stackoverflow.com/questions/4863852/space-before-quote-in-csv-field for some explanation on this too). 

Currently the code allows this case and simply ignores the spaces, but I think we should not allow these cases (and throw an exception saying that there is an invalid row), then change this CSV file to not test for these invalid cases.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCSVFileRecord.java
```diff
@@ -49,6 +49,10 @@
      */
     private final String delimiter;
 
+    private boolean escapeDelimiters;
+
+    private static final String escapeSplitPattern = "(?=(?:[^\"]*\"[^\"]*\")*[^\"]*$)";
```
peterbae, line 54: Can we add a comment here that explains this regex briefly?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCSVFileRecord.java
```diff
@@ -229,21 +236,24 @@ public void close() throws SQLServerException {
...
+            /*
+             * Binary data may be corrupted The limit in split() function should be a negative value, otherwise trailing
+             * empty strings are discarded. Empty string is returned if there is no value.
+             */
+            String[] data = (escapeDelimiters && currentLine.contains("\""))
+                                                                             ? escapeQuotesRFC4180(currentLine.split(
+                                                                                     delimiter + escapeSplitPattern))
+                                                                             : currentLine.split(delimiter, -1);
+            /*
+             * Cannot go directly from String[] to Object[] and expect it to act as an array.
```
lilgreenbird, line 248: [not a big deal, and only if you're changing this file anyway]
this and line 254-256 is only a single line comment

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCSVFileRecord.java
```diff
@@ -189,15 +193,18 @@ private void initFileReader(InputStreamReader sr, String encoding, String demlim
         if (firstLineIsColumnNames) {
             currentLine = fileReader.readLine();
             if (null != currentLine) {
-                columnNames = currentLine.split(delimiter, -1);
+                columnNames = (escapeDelimiters && currentLine.contains("\""))
```
peterbae, line 196: We should keep the limit argument (-1) to the currentLine.split method to preserve the spacing at the end.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCSVFileRecord.java
peterbae, line 248: This line and line 255 should be single line comments

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyCSVTest.java
```diff
@@ -118,6 +134,51 @@ public void testCSVFromURL() throws SQLException {
         }
     }
 
+    @Test
```
lilgreenbird, line 137: test needs description block

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCSVFileRecord.java
```diff
@@ -516,4 +523,58 @@ public boolean next() throws SQLServerException {
...
+            StringBuilder sb = new StringBuilder();
+            if (tokens[i].contains("\"")) {
+                tokens[i] = tokens[i].trim();
+            }
+            while (j < tokens[i].length()) {
+                if ('"' == tokens[i].charAt(j)) {
+                    if (!escaped) {
+                        escaped = true;
+                    } else {
+                        if ((j < tokens[i].length() - 1) && '"' == tokens[i].charAt(j + 1)) {
```
peterbae, line 566: I think we should add an else statement to this if, to handle the case where the CSV data is corrupted and has a double quote that has not been escaped.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCSVFileRecord.java
ulvii: I named it this way mainly to keep it consistent with SQLServerBulkCopyOptions and SQLServerBulkRecord.

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyCSVTest.java
ulvii: You are right.

src/test/resources/BulkCopyCSVTestInputDelimiterEscape.csv
ulvii: You are right that strictly based on the following statement, there should be no spaces before and after enclosing quotes. 

> Spaces are considered part of a field and should not be ignored.

But I found this [WIKI page](https://en.wikipedia.org/wiki/Comma-separated_values#RFC_4180_standard):

> According to RFC 4180, spaces outside quotes in a field are not allowed; however, the RFC also says that "Spaces are considered part of a field and should not be ignored." and "Implementors should 'be conservative in what you do, be liberal in what you accept from others' (RFC 793, section 2.10) when processing CSV files."

> 1997, "Ford" ,E350

I also did some research online and found multiple issues where users run into difficulties because of this limitation in the specs. And I don't see what benefit we would get from throwing an exception instead of ignoring it, so I chose to give flexibility to applications that won't impact the functionality of the feature.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCSVFileRecord.java
ulvii: Sure.

src/test/resources/BulkCopyCSVTestInputDelimiterEscape.csv
peterbae: Ok, I think that's a valid option too. We can leave the code as it is, but we might want to add a short comment saying that the code will ignore spaces that come before and after double quotes, since it might be relevant to some users. I'll resolve my other comment related to this as well.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCSVFileRecord.java
peterbae: Since we're going to ignore spaces outside double quotes, this code can stay as it is.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCSVFileRecord.java
peterbae: Let me know if you'd like to leave the code as it is here as well, we can make the call as to how we're going to handle invalid cases. I think being lenient and letting un-closed double quotes in (the code is removing the un-closed double quote) works for me.

src/test/resources/BulkCopyCSVTestInputDelimiterEscape.csv
ulvii: I added a comment to the Javadoc of `setEscapeColumnDelimitersCSV`.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCSVFileRecord.java
ulvii: That's a good catch. I will modify to throw an exception when there are odd number of quotes.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCSVFileRecord.java
```diff
@@ -516,4 +522,62 @@ public boolean next() throws SQLServerException {
...
+
+    private static String[] escapeQuotesRFC4180(String[] tokens) throws SQLServerException {
+        if (null == tokens) {
+            return tokens;
+        }
+        for (int i = 0; i < tokens.length; i++) {
+            boolean escaped = false;
+            int j = 0;
+            StringBuilder sb = new StringBuilder();
+            long quoteCount = tokens[i].chars().filter(ch -> ch == '"').count();
```
peterbae, line 557: This lambda expression would slow down the parsing of data by scanning through every character. The if statement on line 569 already implicitly does this check, where adding an else statement to

```if ((j < tokens[i].length() - 1) && '"' == tokens[i].charAt(j + 1)) {```

this if statement would guarantee that the corrupt String would be caught.

peterbae: Actually, adding an else statement wouldn't catch all the cases; if the String starts with a double quote but doesn't end in one, or if the String starts with a double quote but ends with an escaped double quote would not fail in the else statement. If we want to make it easy, we can just increment a counter to count the number of double quotes as we go along in the while loop, instead of counting it here.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCSVFileRecord.java
lilgreenbird: not a bit deal...but typically these kind of errors give an error like "Unclosed quotation mark after the character string 'blah'"

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCSVFileRecord.java
ulvii: I improved the error handling a bit, but kept the quoteCount part as it is. `filter()` has the same time complexity as `contains()` and I need to use one of them to tell if the driver is parsing a double quote enclosed field. I looked into counting double quotes in the while loop, but it becomes significantly challenging to ignore the whitespaces after the last quote. I think it is okay to keep `filter()` because the time complexity still remains O(n).

peterbae: Thanks, that sounds good to me.

PR #1313, opened 2020-04-15
Edit build file to include new dependencies and exclude new tests

rene-ye:


src/test/java/com/microsoft/sqlserver/jdbc/clientcertauth/ClientCertificateAuthenticationTest.java
```diff
@@ -2,7 +2,7 @@
  * Microsoft JDBC Driver for SQL Server Copyright(c) Microsoft Corporation All rights reserved. This program is made
  * available under the terms of the MIT License. See the LICENSE file in the project root for more information.
  */
-package com.microsoft.sqlserver.clientcertauth;
+package com.microsoft.sqlserver.jdbc.clientcertauth;
```
lilgreenbird, line 5: this change should be in the clientcertauth PR

PR #1315, opened 2020-04-16
Fix | Close resources and skip AE tests by default

rene-ye:
Moved Client Certificate Authentication tests to proper folder.
Explicitly close closable resources.
Skip AE tests by default so users don't require the AE folder to run tests.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCertificateUtils.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCertificateUtils.java
@@ -206,11 +206,8 @@ final class SQLServerCertificateUtils {
     private static Certificate loadCertificate(
             String certificatePem) throws IOException, GeneralSecurityException, SQLServerException {
         CertificateFactory certificateFactory = CertificateFactory.getInstance("X509");
-        InputStream certStream = fileToStream(certificatePem);
-        try {
+        try (InputStream certStream = fileToStream(certificatePem)) {
             return certificateFactory.generateCertificate(certStream);
-        } finally {
-            certStream.close();
         }
     }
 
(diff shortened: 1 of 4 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCertificateUtils.java
```diff
@@ -81,10 +81,15 @@
...
             String keyPassword) throws NoSuchAlgorithmException, CertificateException, FileNotFoundException, IOException, UnrecoverableKeyException, KeyStoreException, SQLServerException {
         KeyStore keystore = KeyStore.getInstance(PKCS12_ALG);
+        FileInputStream certStream = null;
         try {
-            keystore.load(new FileInputStream(certPath), keyPassword.toCharArray());
+            certStream = new FileInputStream(certPath);
+            keystore.load(certStream, keyPassword.toCharArray());
         } catch (FileNotFoundException e) {
             throw new SQLServerException(SQLServerException.getErrString("R_clientCertError"), null, 0, null);
+        } finally {
```
ulvii, line 90: Can't you put this into try with resources?

src/test/java/com/microsoft/sqlserver/jdbc/clientcertauth/ClientCertificateAuthenticationTest.java
```diff
@@ -2,7 +2,7 @@
  * Microsoft JDBC Driver for SQL Server Copyright(c) Microsoft Corporation All rights reserved. This program is made
  * available under the terms of the MIT License. See the LICENSE file in the project root for more information.
  */
-package com.microsoft.sqlserver.clientcertauth;
+package com.microsoft.sqlserver.jdbc.clientcertauth;
```
peterbae, line 5: This line is already covered in PR 1313, you should remove this in one of them (probably in the other one)

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCertificateUtils.java
```diff
@@ -277,7 +287,7 @@ private static InputStream fileToStream(String fname) throws IOException, SQLSer
             dis.readFully(bytes);
             ByteArrayInputStream bais = new ByteArrayInputStream(bytes);
             return bais;
-        } catch (FileNotFoundException e) { 
+        } catch (FileNotFoundException e) {
             throw new SQLServerException(SQLServerException.getErrString("R_clientCertError"), null, 0, null);
         } finally {
             if (null != dis) {
```
lilgreenbird, line 293: can use try with resource here too

rene-ye:
I'm not moving it to this PR, it doesn't matter which PR it goes in.

Issue #1304, opened 2020-04-03
Unneeded temporary objects while URL parsing


ecki:
The URL parser state engine generates for each character read a StringBuilder and a new String object. This is not needed. Instead you can keep the StringBuilder, append it until the result step where you use a single toString(). Also in the initial state you dont need the append(result).

https://github.com/microsoft/mssql-jdbc/blob/c8a08c8509b1033f0500ed8545ae60162d3c2645/src/main/java/com/microsoft/sqlserver/jdbc/Util.java#L281

BTW: it looks like there is no state für ipv6 [1:2:3:4] literals.

ecki:
Even less allocations you would get by remembering only the Index position and use substring() at the end of each state, not sure if this is faster, however.

ulvii:
#1317 merged.

PR #1317, opened 2020-04-17
Fix | Use a StringBuilder for 'result'

rene-ye:
Fixes issue https://github.com/microsoft/mssql-jdbc/issues/1304.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/Util.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/Util.java
@@ -245,7 +245,6 @@ final class Util {
         StringBuilder result = new StringBuilder();
         String name = "";
         String value = "";
-        StringBuilder builder;
 
         if (!tmpUrl.startsWith(sPrefix))
             return null;
@@ -284,11 +283,11 @@ final class Util {
                 case inServerName: {
                     if (ch == ';' || ch == ':' || ch == '\\') {
                         // non escaped trim the string
-                        String propertyValue = result.toString().trim();
-                        if (propertyValue.length() > 0) {
-                            p.put(SQLServerDriverStringProperty.SERVER_NAME.toString(), propertyValue);
+                        String property = result.toString().trim();
+                        if (property.length() > 0) {
+                            p.put(SQLServerDriverStringProperty.SERVER_NAME.toString(), property);
                             if (logger.isLoggable(Level.FINE)) {
-                                logger.fine("Property:serverName " + "Value:" + propertyValue);
+                                logger.fine("Property:serverName " + "Value:" + property);
                             }
                         }
                         result.setLength(0);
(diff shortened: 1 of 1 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/Util.java
```diff
@@ -370,7 +357,7 @@ else if (ch == ':')
                         }
                         // same state
                     } else {
-                        builder = new StringBuilder();
+                        StringBuilder builder = new StringBuilder();
```
peterbae, line 360: Why are these else cases handled differently? I think these should also be replaced by `result.append(ch);`?

src/main/java/com/microsoft/sqlserver/jdbc/Util.java
```diff
@@ -478,27 +465,27 @@ else if (ch == ':')
...
-                        logger.fine("Property:serverName " + "Value:" + result);
+                        logger.fine("Property:serverName " + "Value:" + property);
                     }
-                    p.put(SQLServerDriverStringProperty.SERVER_NAME.toString(), result);
+                    p.put(SQLServerDriverStringProperty.SERVER_NAME.toString(), property);
                 }
                 break;
             case inPort:
-                result = result.trim();
+                property = result.toString().trim();
```
peterbae, line 477: Lines 477 and 484 will probably error out, because `property` variable is actually not defined here. You should declare property as a String every time, or move up the declaration of line 468 to somewhere above.

src/main/java/com/microsoft/sqlserver/jdbc/Util.java
```diff
@@ -242,10 +242,9 @@ static void writeLong(long value, byte valueBytes[], int offset) {
         Properties p = new Properties();
         String tmpUrl = url;
         String sPrefix = "jdbc:sqlserver://";
-        String result = "";
```
ulvii, line 245: Please add tests for the changes you are making.

src/main/java/com/microsoft/sqlserver/jdbc/Util.java
lilgreenbird: odd, I'm surprised it's not complaining about that

rene-ye: It doesn't error out, it's declared in the first switch statement.

src/main/java/com/microsoft/sqlserver/jdbc/Util.java
rene-ye: Because this appends a name and not result. This method has fall through logic that forces us to retain some original logic.

src/main/java/com/microsoft/sqlserver/jdbc/Util.java
rene-ye: The PR is to change a String variable to a StringBuilder, I don't believe testing will be beneficial.

src/main/java/com/microsoft/sqlserver/jdbc/Util.java
peterbae: But there is a `break;` after the inServerName case. I think the tests are not failing because every test case had `inServerName` get called before other cases. Can you test this method by providing it a connection string that has maybe the port number come before the server name?

Issue #1302, opened 2020-04-02
Cast exception when getting value of a sql_variant


zoladkow:
## Driver version
- 7.4.1
- 8.2.1

## SQL Server version
12.0.2000.8

## Client Operating System
Win10

## JAVA/JVM version
Oracle's 1.8.0

## Table schema
INFORMATION_SCHEMA

## Problem description
1. Expected behaviour: `resultset.getLong(colName)` to return numeric value as long
2. Actual behaviour: `java.lang.ClassCastException`: `java.lang.Integer` cannot be cast to `java.lang.Long`
3. Error message/stack trace:
```
java.lang.ClassCastException: java.lang.Integer cannot be cast to java.lang.Long
	at com.microsoft.sqlserver.jdbc.SQLServerResultSet.getLong(SQLServerResultSet.java:2340)
```
4. Any other details that can be helpful:
 - Detected while investigating problem why `StoredProcedureQuery` parameters were not bound by name, but instead indexes. 
 - Currently Hibernate is using `select * from INFORMATION_SCHEMA.SEQUENCES` for obtaining sequence information - using `getLong(...)` for `start_value` etc. 
 - **BUT** since `start_value` is a `sql_variant` and if particular sequence is of `int` data type class cast exception is thrown, when reading that value and Hibernate falls back to using safe defaults instead of actual capabilities (including `supportsNamedParameters = false`)
 - Quick workaround in the project was to correct the datatype of two sequences. But clearly there is a problem in the driver with `sql_variant` support.
 - normally a `BIGINT` parameter is propagated down the call stack, but if the the actual column type is `sql_variant` another logic overrides the expected `BIGINT` type with the type detected from `sql_variant`. So an `Integer` is being returned up the stack where explicit casting is performed to `Long`. Which fails.
 - `sql_variant` reading logic should consider the data type provided by the caller to avoid class cast exception. Normally it is expected that a wider data type should contain the narrow data type implicitly.

## JDBC trace logs
[jdbcTrace.log](https://github.com/microsoft/mssql-jdbc/files/4423110/jdbcTrace.log)

## Reproduction code
```java
ResultSet rs = conn
                .prepareStatement("select cast(1 as sql_variant) as [sad_face]")
                .executeQuery();
if (rs.next()) {
     println ("Long value: " + rs.getLong("sad_face"));
}
```

peterbae:
[...]

PR #1320, opened 2020-04-28
Fix issue with cast exception from sql_variant

peterbae:
Fixes issue #1302.

Even though the code has remained this way for a long time, I think the current behavior of sql_variant is incorrect. The `jdbcType` variable in dtv.java is the JDBC type that the user specifies when they try to retrieve the data. This variable is telling the driver what the user wants to retrieve the data as, but this variable was getting modified to become the base type of the server's column type when retrieving sql_variant data, causing this issue (this effectively ignores the data type that the user was asking to retrieve sql_variant as). Leaving `jdbcType` as it is (which we should, since that's what the user wants) solves the issue.

However, when the driver performs bulk copy on a sql_variant column, the driver does need to designate `jdbcType` as the underlying data type to perform datatype-specific conversions. This made it difficult for applying my changes to all datatypes (for the scope of this PR), and for now only the datatypes that do not require any specific conversions from going from one another (for example, numeric values and String values) have the fixes applied.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/DDC.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/DDC.java
@@ -73,10 +73,12 @@ final class DDC {
             case BINARY:
                 return convertIntToBytes(intValue, valueLength);
             case SQL_VARIANT:
-                // specifically return short if the underlying datatype of sql_variant is tinyint or smallint
+                // return short or bit if the underlying datatype of sql_variant is tinyint, smallint or bit
                 // otherwise, return integer
                 // Longer datatypes such as double and float are handled by convertLongToObject instead.
-                if (valueLength == 3 || valueLength == 4) {
+                if (valueLength == 1) {
+                    return 0 != intValue;
+                } else if (valueLength == 3 || valueLength == 4) {
                     return (short) intValue;
                 } else {
                     return intValue;
@@ -353,6 +355,7 @@ final class DDC {
             case NUMERIC:
             case MONEY:
             case SMALLMONEY:
+            case SQL_VARIANT:
                 return bigDecimalVal;
             case FLOAT:
             case DOUBLE:
--- a/src/main/java/com/microsoft/sqlserver/jdbc/DDC.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/DDC.java
@@ -72,6 +72,15 @@ final class DDC {
                 return (float) intValue;
             case BINARY:
                 return convertIntToBytes(intValue, valueLength);
+            case SQL_VARIANT:
+                // specifically return short if the underlying datatype of sql_variant is tinyint or smallint
+                // otherwise, return integer
+                // Longer datatypes such as double and float are handled by convertLongToObject instead.
+                if (valueLength == 3 || valueLength == 4) {
+                    return (short) intValue;
+                } else {
+                    return intValue;
+                }
             default:
                 return Integer.toString(intValue);
         }
@@ -93,6 +102,7 @@ final class DDC {
     static final Object convertLongToObject(long longVal, JDBCType jdbcType, SSType baseSSType, StreamType streamType) {
         switch (jdbcType) {
             case BIGINT:
+            case SQL_VARIANT:
                 return longVal;
             case INTEGER:
                 return (int) longVal;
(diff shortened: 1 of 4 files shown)


src/test/java/com/microsoft/sqlserver/jdbc/datatypes/SQLVariantResultSetTest.java
```diff
@@ -1061,6 +1061,16 @@ public void testTimeClassAsSqlVariant() throws SQLException {
             assertEquals(object.getClass(), java.sql.Time.class);;
         }
     }
+    
+    @Test
+    public void testCastThenGetNumeric() throws SQLException {
+        try (Connection con = getConnection(); Statement stmt = con.createStatement();) {
+            SQLServerResultSet rs = (SQLServerResultSet) stmt.executeQuery("select cast(123 as sql_variant) as c1");
```
ulvii, line 1068: ResultSet in try block.

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/SQLVariantResultSetTest.java
```diff
@@ -1061,6 +1061,16 @@ public void testTimeClassAsSqlVariant() throws SQLException {
...
+            rs.next();
+            long longValue = rs.getLong("c1");
+            assertEquals(longValue, 123L);
```
ulvii, line 1071: Expected value first.

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/SQLVariantResultSetTest.java
```diff
@@ -1061,6 +1061,16 @@ public void testTimeClassAsSqlVariant() throws SQLException {
             assertEquals(object.getClass(), java.sql.Time.class);;
         }
     }
+    
+    @Test
+    public void testCastThenGetNumeric() throws SQLException {
```
ulvii, line 1066: Please add test cases for rest of the datatypes this PR applies to.

rene-ye:
Can we document the "supported" sql variant datatype for now and the ones that aren't? Just so users have a reference in case they notice inconsistencies.

src/main/java/com/microsoft/sqlserver/jdbc/DDC.java
```diff
@@ -72,6 +72,17 @@ static final Object convertIntegerToObject(int intValue, int valueLength, JDBCTy
                 return (float) intValue;
             case BINARY:
                 return convertIntToBytes(intValue, valueLength);
+            case SQL_VARIANT:
+                // return short or bit if the underlying datatype of sql_variant is tinyint, smallint or bit
+                // otherwise, return integer
+                // Longer datatypes such as double and float are handled by convertLongToObject instead.
+                if (valueLength == 1) {
+                    return 0 != intValue;
+                } else if (valueLength == 3 || valueLength == 4) {
```
lilgreenbird, line 81: can you add a comment here on why these 2 values?

peterbae: I do have a comment, on line 76

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/SQLVariantResultSetTest.java
peterbae: Done (applies to other comments too)

src/main/java/com/microsoft/sqlserver/jdbc/DDC.java
```diff
@@ -72,6 +72,17 @@ static final Object convertIntegerToObject(int intValue, int valueLength, JDBCTy
                 return (float) intValue;
             case BINARY:
                 return convertIntToBytes(intValue, valueLength);
+            case SQL_VARIANT:
+                // return short or bit if the underlying datatype of sql_variant is tinyint, smallint or bit
+                // otherwise, return integer
+                // Longer datatypes such as double and float are handled by convertLongToObject instead.
+                if (valueLength == 1) {
```
lilgreenbird, line 79: 1 == valueLength

peterbae: I don't think this is necessary

PR #1321, opened 2020-04-28
Fix | Add certificate expire validation when using Always Encrypted with secure enclaves

rene-ye:
Explicitly check certificate expiry.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerVSMEnclaveProvider.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerVSMEnclaveProvider.java
@@ -16,6 +16,7 @@ import java.security.Security;
 import java.security.Signature;
 import java.security.SignatureException;
 import java.security.cert.CertificateException;
+import java.security.cert.CertificateExpiredException;
 import java.security.cert.CertificateFactory;
 import java.security.cert.X509Certificate;
 import java.security.spec.MGF1ParameterSpec;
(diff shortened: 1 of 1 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerVSMEnclaveProvider.java
```diff
@@ -275,9 +276,10 @@ void validateCert(byte[] b) throws SQLServerException {
                         .generateCertificates(new ByteArrayInputStream(b));
                 for (X509Certificate cert : certs) {
                     try {
+                        cert.checkValidity();
                         healthCert.verify(cert.getPublicKey());
                         return;
-                    } catch (SignatureException e) {
+                    } catch (SignatureException | CertificateExpiredException e) {
                         // Doesn't match, but continue looping through the rest of the certificates
```
lilgreenbird, line 283: should we log a warning?

rene-ye: I don't think it's necessary, up to the user to make sure their certificates haven't expired.

lilgreenbird: I think would be helpful if someone complains that they do have the certificate why isn't it working. But, I don't feel strongly about it so ok.

PR #1322, opened 2020-04-28
Check tls protocol version and throw warning

rene-ye:


src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -1593,6 +1594,9 @@ private void validateServerNameInCertificate(X509Certificate cert) throws Certif
         SSL_HANDHSAKE_COMPLETE
     }
 
+    static final String TLS_1 = "TLSv1";
```
ulvii, line 1596: There is already SSLProtocol enum in the driver

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -1826,6 +1826,16 @@ else if (con.getTrustManagerClass() != null) {
             // SSL is now enabled; switch over the channel socket
             channelSocket = sslSocket;
 
+            // Check the TLS version
+            String tlsProtocol = sslSocket.getSession().getProtocol();
+            if (SSLProtocol.TLS_V10.toString().equalsIgnoreCase(tlsProtocol)
+                    || SSLProtocol.TLS_V11.toString().equalsIgnoreCase(tlsProtocol)) {
+                String warningMsg = tlsProtocol
+                        + " was negotiated. Please update server and client to use TLSv1.2 at minimum.";
```
lilgreenbird, line 1834: add to SQLServerResource?

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -1826,6 +1826,16 @@ else if (con.getTrustManagerClass() != null) {
             // SSL is now enabled; switch over the channel socket
             channelSocket = sslSocket;
 
+            // Check the TLS version
+            String tlsProtocol = sslSocket.getSession().getProtocol();
+            if (SSLProtocol.TLS_V10.toString().equalsIgnoreCase(tlsProtocol)
+                    || SSLProtocol.TLS_V11.toString().equalsIgnoreCase(tlsProtocol)) {
```
lilgreenbird, line 1832: I think it might be better to put add a isMinTLS() (or something) method in SQLServerDriver? That way when a new version comes out it can be changed in 1 place only.

rene-ye: this is the only place we check tls version.

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
rene-ye: Not an error message.

PR #1328, opened 2020-05-05
Updates for java 8 compliance

lilgreenbird:


src/main/java/com/microsoft/sqlserver/jdbc/DDC.java
```diff
@@ -614,9 +614,9 @@ private static LocalDateTime parseStringIntoLDT(String s) {
         if (firstDash > 0 && secondDash > 0 && secondDash < dividingSpace - 1) {
             if (firstDash == YEAR_LENGTH && (secondDash - firstDash > 1 && secondDash - firstDash <= MONTH_LENGTH + 1)
                     && (dividingSpace - secondDash > 1 && dividingSpace - secondDash <= DAY_LENGTH + 1)) {
-                year = Integer.parseInt(s, 0, firstDash, 10);
-                month = Integer.parseInt(s, firstDash + 1, secondDash, 10);
-                day = Integer.parseInt(s, secondDash + 1, dividingSpace, 10);
+                year = Util.parseInt(s, 0, firstDash, 10);
```
ulvii, line 617:
```suggestion
                year = Integer.parseInt(s.subSequence(0, firstDash).toString());
```

src/main/java/com/microsoft/sqlserver/jdbc/DDC.java
```diff
@@ -614,9 +614,9 @@ private static LocalDateTime parseStringIntoLDT(String s) {
...
+                month = Util.parseInt(s, firstDash + 1, secondDash, 10);
```
ulvii, line 618:
```suggestion
                month = Integer.parseInt(s.subSequence(firstDash + 1, secondDash).toString());
```

src/main/java/com/microsoft/sqlserver/jdbc/DDC.java
```diff
@@ -614,9 +614,9 @@ private static LocalDateTime parseStringIntoLDT(String s) {
...
+                day = Util.parseInt(s, secondDash + 1, dividingSpace, 10);
```
ulvii, line 619:
```suggestion
                day = Integer.parseInt(s.subSequence(secondDash + 1, dividingSpace).toString());
```

src/main/java/com/microsoft/sqlserver/jdbc/DDC.java
```diff
@@ -630,16 +630,16 @@ private static LocalDateTime parseStringIntoLDT(String s) {
         // Convert the time; default missing nanos
         int len = s.length();
         if (firstColon > 0 && secondColon > 0 && secondColon < len - 1) {
-            hour = Integer.parseInt(s, dividingSpace + 1, firstColon, 10);
-            minute = Integer.parseInt(s, firstColon + 1, secondColon, 10);
+            hour = Util.parseInt(s, dividingSpace + 1, firstColon, 10);
```
ulvii, line 633:
```suggestion
            hour = Integer.parseInt(s.subSequence(dividingSpace + 1, firstColon).toString());
```

src/main/java/com/microsoft/sqlserver/jdbc/DDC.java
```diff
@@ -630,16 +630,16 @@ private static LocalDateTime parseStringIntoLDT(String s) {
...
+            minute = Util.parseInt(s, firstColon + 1, secondColon, 10);
```
ulvii, line 634:
```suggestion
            minute = Integer.parseInt(s.subSequence(firstColon + 1, secondColon).toString());
```

src/main/java/com/microsoft/sqlserver/jdbc/DDC.java
```diff
@@ -630,16 +630,16 @@ private static LocalDateTime parseStringIntoLDT(String s) {
...
             if (period > 0 && period < len - 1) {
-                second = Integer.parseInt(s, secondColon + 1, period, 10);
+                second = Util.parseInt(s, secondColon + 1, period, 10);
```
ulvii, line 636:
```suggestion
                second = Integer.parseInt(s.subSequence(secondColon + 1, period).toString());
```

src/main/java/com/microsoft/sqlserver/jdbc/DDC.java
```diff
@@ -630,16 +630,16 @@ private static LocalDateTime parseStringIntoLDT(String s) {
...
                 int nanoPrecision = len - (period + 1);
                 if (nanoPrecision > 9)
                     throw new java.lang.IllegalArgumentException(formatError);
                 if (!Character.isDigit(s.charAt(period + 1)))
                     throw new java.lang.IllegalArgumentException(formatError);
-                int tmpNanos = Integer.parseInt(s, period + 1, len, 10);
+                int tmpNanos = Util.parseInt(s, period + 1, len, 10);
```
ulvii, line 642:
```suggestion
                int tmpNanos = Integer.parseInt(s.subSequence(period + 1, len).toString());
```

src/main/java/com/microsoft/sqlserver/jdbc/DDC.java
```diff
@@ -648,7 +648,7 @@ private static LocalDateTime parseStringIntoLDT(String s) {
             } else if (period > 0) {
                 throw new java.lang.IllegalArgumentException(formatError);
             } else {
-                second = Integer.parseInt(s, secondColon + 1, len, 10);
+                second = Util.parseInt(s, secondColon + 1, len, 10);
```
ulvii, line 651:
```suggestion
                second = Integer.parseInt(s.subSequence(secondColon + 1, len).toString());
```

src/main/java/com/microsoft/sqlserver/jdbc/Util.java
```diff
@@ -1032,6 +1033,90 @@ static String convertInputStreamToString(java.io.InputStream is) throws IOExcept
         }
         return result.toString();
     }
+
```
ulvii, line 1036: Once you commit the suggestions above, you can delete this method.

src/main/java/com/microsoft/sqlserver/jdbc/DDC.java
```diff
@@ -614,9 +614,9 @@ private static LocalDateTime parseStringIntoLDT(String s) {
         if (firstDash > 0 && secondDash > 0 && secondDash < dividingSpace - 1) {
             if (firstDash == YEAR_LENGTH && (secondDash - firstDash > 1 && secondDash - firstDash <= MONTH_LENGTH + 1)
                     && (dividingSpace - secondDash > 1 && dividingSpace - secondDash <= DAY_LENGTH + 1)) {
-                year = Integer.parseInt(s, 0, firstDash, 10);
-                month = Integer.parseInt(s, firstDash + 1, secondDash, 10);
-                day = Integer.parseInt(s, secondDash + 1, dividingSpace, 10);
+                year = Integer.parseInt(s.subSequence(0, firstDash).toString());
```
rene-ye, line 617: Is there a reason why we're using `subSequence(int,int)` here instead of `subString(int,int)`. It seems like `subSequence` actually uses `subString` internally and the only major difference is that it returns a `CharSequence` instead of a `String`. Considering it seems we want a `String` anyways, lets just use `subString(int,int)`.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerColumnEncryptionAzureKeyVaultProvider.java
```diff
@@ -667,7 +667,7 @@ public boolean verifyColumnMasterKeyMetadata(String masterKeyPath, boolean allow
         boolean append = true;
         if (null != mssqlJdbcProperties) {
             String endpoints = mssqlJdbcProperties.getProperty(AKV_TRUSTED_ENDPOINTS_KEYWORD);
-            if (null != endpoints && !endpoints.isBlank()) {
+            if (null != endpoints && !endpoints.trim().isEmpty()) {
```
rene-ye, line 670: `endpoints.trim().length() > 0` would be more performant than an `isEmpty()` check.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCertificateUtils.java
```diff
@@ -153,22 +154,23 @@ private static PrivateKey loadPrivateKeyFromPVK(String keyPath,
         ByteBuffer buffer = ByteBuffer.allocate((int) f.length());
         try (FileInputStream in = new FileInputStream(f)) {
             in.getChannel().read(buffer);
-            buffer.order(ByteOrder.LITTLE_ENDIAN).rewind();
+            ((Buffer) buffer.order(ByteOrder.LITTLE_ENDIAN)).rewind();
```
rene-ye, line 157: Is this really necessary, the [Oracle Java docs](https://docs.oracle.com/javase/7/docs/api/java/nio/ByteBuffer.html) says rewind is inherited from Buffer since Java SE7.

src/test/java/com/microsoft/sqlserver/jdbc/unit/lobs/LobsStreamingTest.java
```diff
@@ -145,8 +146,7 @@ public void testClobsVarcharASCII() throws SQLException, IOException {
                         Clob c = rs.getClob(2);
                         assertEquals(c.length(), lob_data.get(index).length());
                         try (InputStream is = c.getAsciiStream();
-                                Scanner s = new Scanner(is, java.nio.charset.StandardCharsets.US_ASCII)
-                                        .useDelimiter("\\A")) {
+                                Scanner s = new Scanner(is, "US-ASCII").useDelimiter("\\A")) {
```
rene-ye, line 149: There should be an unclosed resource warning here, but since I wrote this test I can address it in the LOBs PR if necessary.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCertificateUtils.java
peterbae: I'm curious about this one as well. Does Java try to use the Java 9 version of rewind from ByteBuffer without this cast?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerColumnEncryptionAzureKeyVaultProvider.java
lilgreenbird: more readable and minimal difference

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCertificateUtils.java
lilgreenbird: yes that is the reason, ByteBuffer doesn't have this method so it will throw a NoSuchMethodError

PR #1331, opened 2020-05-07
Extended bulk copy support for Azure DW

peterbae:
Adds new connection property `sendTemporalDataTypesAsStringForBulkCopy`.

This connection property, when set to FALSE, will send DATE, DATETIME, DATIMETIME2 DATETIMEOFFSET, SMALLDATETIME, and TIME datatypes as their respective types instead of sending them as String.

Additionally, even without setting this connection string to FALSE, MONEY and SMALLMONEY datatypes will be sent as MONEY / SMALLMONEY datatypes instead of DECIMAL after these changes.

Note that with this connection property set to FALSE, the driver will only accept the default string literal formatof each temporal datatype, for example:
 
DATE: YYYY-MM-DD
DATETIME: YYYY-MM-DD hh:mm:ss[.nnn]
DATETIME2: YYYY-MM-DD hh:mm:ss[.nnnnnnn]
DATETIMEOFFSET: YYYY-MM-DD hh:mm:ss[.nnnnnnn] [{+|-}hh:mm]
SMALLDATETIME:YYYY-MM-DD hh:mm:ss
TIME: hh:mm:ss[.nnnnnnn]

I've created two separate test classes for this, one for normal cases and one for AE cases. The AE cases are not supported against Azure DW, but I wanted both normal and AE cases to be run against their respective supported servers without one of them not being run when they should be, therefore two separate test classes.

I also made a small change to AESetup.java so that it doesn't throw a NPE when I'm trying to debug/run tests locally through eclipse.


src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopySendTemporalDataTypesAsString.java
```diff
@@ -0,0 +1,227 @@
+/*
```
ulvii, line 1: Make sure to follow JUnit test name standards.

ulvii:
Please apply the formatter to your changes.

src/main/java/com/microsoft/sqlserver/jdbc/DDC.java
```diff
@@ -323,6 +323,28 @@ static final Object convertDoubleToObject(double doubleVal, JDBCType jdbcType, S
...
+        byte[] valueBytes = new byte[bLength];
+        
+        BigInteger bi = bigDecimalVal.unscaledValue();
+        
+        if (bLength == 8) {
+            // money
+            byte[] longbArray = new byte[bLength];
+            Util.writeLong(bi.longValue(), longbArray, 0);
+            // TDS 2.2.5.5.1.4 Fixed-Point Numbers
+            // Money is represented as a 8 byte signed integer, with one 4-byte integer that represents
```
ulvii, line 337: Please use multi-line comments. Apply this to all the changes you made.

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopySendTemporalDataTypesAsString.java
```diff
@@ -0,0 +1,227 @@
...
+            validateValuesFromCSV(stmt, destTableName2, inputFile);
+        } catch (Exception e) {
+            fail(e.getMessage());
+        }
+    }
+
+    static void validateValuesFromCSV(Statement stmt, String destinationTable, String inputFile) {
+        try (BufferedReader br = new BufferedReader(
+                new InputStreamReader(new FileInputStream(filePath + inputFile), encoding));
+                ResultSet rs = stmt.executeQuery("SELECT * FROM " + destinationTable)) {
```
ulvii, line 176: This is assuming insertion order is the same as select order, which is not true. Make sure to have an ID column to be able to order the select result by ID.

src/test/java/com/microsoft/sqlserver/jdbc/AlwaysEncrypted/BulkCopySendTemporalDataTypesAsStringAE.java
```diff
@@ -0,0 +1,173 @@
...
+            validateValuesFromCSV(stmt, tableName, inputFile);
+        } catch (Exception e) {
+            fail(e.getMessage());
+        }
+    }
+
+    static void validateValuesFromCSV(Statement stmt, String destinationTable, String inputFile) {
+        try (BufferedReader br = new BufferedReader(
+                new InputStreamReader(new FileInputStream(filePath + inputFile), encoding));
+                ResultSet rs = stmt.executeQuery("SELECT * FROM " + destinationTable)) {
```
ulvii, line 111: This is assuming insertion order is the same as select order, which is not true. Make sure to have an ID column to be able to order the select result by ID.

src/main/java/com/microsoft/sqlserver/jdbc/ISQLServerDataSource.java
```diff
@@ -991,4 +991,19 @@
      */
     void setClientKeyPassword(String password);
 
+    /**
```
ulvii, line 994: Please make javadoc consistent with other methods.

src/test/java/com/microsoft/sqlserver/jdbc/AlwaysEncrypted/BulkCopySendTemporalDataTypesAsStringAE.java
ulvii: This is still not fixed.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
```diff
@@ -2122,6 +2163,25 @@ else if (null != sourceCryptoMeta) {
...
+                        /*
+                         * if the precision that user provides is smaller than the precision of the actual value, the
+                         * driver assumes the precision that user provides is the correct precision, and throws
+                         * exception
+                         */
+                        if (bulkPrecision < Util.getValueLengthBaseOnJavaType(colValue, JavaType.of(colValue), null,
+                                null, JDBCType.of(bulkJdbcType))) {
+                            MessageFormat form = new MessageFormat(
+                                    SQLServerException.getErrString("R_valueOutOfRange"));
+                            Object[] msgArgs = {SSType.DECIMAL};
```
ulvii, line 2178: Shouldn't this be `destSSType` instead?

src/test/java/com/microsoft/sqlserver/jdbc/AlwaysEncrypted/BulkCopySendTemporalDataTypesAsStringAE.java
peterbae: Why is this not fixed?

peterbae: oh never mind, you meant the row order. I'll fix that.

ulvii: I have a similar tests, you can refer to this PR: https://github.com/microsoft/mssql-jdbc/pull/1312

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
```diff
@@ -919,23 +919,37 @@ private void writeTypeInfo(TDSWriter tdsWriter, int srcJdbcType, int srcScale, i
...
-                    tdsWriter.writeByte((byte) TDSWriter.BIGDECIMAL_MAX_LENGTH); // maximum length
-                    tdsWriter.writeByte((byte) srcPrecision); // unsigned byte
-                    tdsWriter.writeByte((byte) srcScale); // unsigned byte
+                /*
+                 * SQL Server allows the insertion of decimal and numeric into a money (and smallmoney) column,
+                 * but Azure DW only accepts money types for money column.
+                 * To make the code compatible against both SQL Server and Azure DW, always send decimal and
+                 * numeric as money/smallmoney if the destination column is money/smallmoney
+                 * and the source is decimal/numeric.
+                 */
```
ulvii, line 936: I don't get this part of the changes.  So if `destSSType` is money/smallmoney and `srcJdbcType` is decimal/numeric, the code will write both `TDSType.MONEYN.byteValue()`/`DSType.MONEYN.byteValue()` and `tdsWriter.writeByte(TDSType.DECIMALN.byteValue())`/`tdsWriter.writeByte(TDSType.NUMERICN.byteValue())`. Is this expected?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResource.java
```diff
@@ -264,6 +264,8 @@ static String getResource(String key) {
             {"R_clientKeyPropertyDescription", "Private key file path for client certificate authentication feature."},
             {"R_clientKeyPasswordPropertyDescription",
                     "Password for private key if the private key is password protected."},
+            {"R_sendTemporalDataTypesAsStringForBulkCopyPropertyDescription",
+                    "Flag to indicate whether the driver will send temporal datatypes as String value to the server for bulk copy."},
```
lilgreenbird, line 268: suggest that we be consistent in how we describe our other sendBlahAsBlah properties, ie "Determines whether the driver will send..."

peterbae: I don't think my new addition here is inconsistent because other descriptive properties don't follow the "Determines whether..." starters either. I think the suggestion is good though.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
peterbae: Sorry, which part of this code should be destSSTyle?

src/main/java/com/microsoft/sqlserver/jdbc/ISQLServerDataSource.java
peterbae: done

src/test/java/com/microsoft/sqlserver/jdbc/AlwaysEncrypted/BulkCopySendTemporalDataTypesAsStringAE.java
peterbae: Thanks, done

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
peterbae: No, if `destSSType` is money and `srcJdbcType` is decimal, the code will only write `TDSType.MONEYN.byteValue()` and break out of the switch statement.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
peterbae: Actually, I understood. Yes, this should be destSSType instead. I've made similar changes to the decimal/numeric case as well because I think it applies there as well, too.

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -3574,27 +3596,81 @@ void writeTime(java.sql.Timestamp value, int scale) throws SQLServerException {
...
+            calendar = new GregorianCalendar(timeZone, Locale.US);
+
+            String timestampString = stringValue.substring(0, lastColon - 4);
+            int year = Integer.valueOf(timestampString.substring(0, 4));
+            int month = Integer.valueOf(timestampString.substring(5, 7));
+            int day = Integer.valueOf(timestampString.substring(8, 10));
+            int hour = Integer.valueOf(timestampString.substring(11, 13));
+            int minute = Integer.valueOf(timestampString.substring(14, 16));
+            int second = Integer.valueOf(timestampString.substring(17, 19));
+
```
lilgreenbird, line 3639: why define all these variables they're only used once below

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopySendTemporalDataTypesAsStringTest.java
```diff
@@ -0,0 +1,225 @@
...
+     * Test basic case with sendTemporalDataTypesAsStringForBulkCopy connection property, using a data source.
+     * 
+     * @throws SQLException
+     */
+    @Test
+    public void testSendTemporalDataTypesAsStringForBulkCopyDS() throws SQLException {
+        beforeEachSetup();
+        SQLServerDataSource dsLocal = new SQLServerDataSource();
+        AbstractTest.updateDataSource(connectionString, dsLocal);
+        dsLocal.setSendTemporalDataTypesAsStringForBulkCopy(false);
```
lilgreenbird, line 111: this test doesn't actually test the datasource settings as updateDataSource hasn't been updated to include the new properties

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopySendTemporalDataTypesAsStringTest.java
```diff
@@ -0,0 +1,225 @@
...
+    static String filePath = null;
+
+    /**
+     * Test basic case with sendTemporalDataTypesAsStringForBulkCopy connection property.
+     * 
+     * @throws SQLException
+     */
+    @Test
+    public void testSendTemporalDataTypesAsStringForBulkCopy() throws SQLException {
+        beforeEachSetup();
```
lilgreenbird, line 74: why not just use @BeforeEach?

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -3574,27 +3596,81 @@ void writeTime(java.sql.Timestamp value, int scale) throws SQLServerException {
...
+         * introduction of extended bulk copy support for Azure DW, we now need to support this scenario.
+         * Parse the DTO as string if it's coming from a CSV.
+         */
+        if (value instanceof String) {
+            String stringValue = (String) value;
+            int lastColon = stringValue.lastIndexOf(':');
+
+            String offsetString = stringValue.substring(lastColon - 3);
+            minutesOffset = 60 * Integer.valueOf(offsetString.substring(1, 3))
+                    + Integer.valueOf(offsetString.substring(4, 6));
```
rene-ye, line 3615: Need to handle indexoutofbounds here in case the input isn't in the format expected.

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -3574,27 +3596,81 @@ void writeTime(java.sql.Timestamp value, int scale) throws SQLServerException {
...
+             * If the target data type is DATETIMEOFFSET, then use UTC for the calendar that
+             * will hold the value, since writeRPCDateTimeOffset expects a UTC calendar.
+             * Otherwise, when converting from DATETIMEOFFSET to other temporal data types,
+             * use a local time zone determined by the minutes offset of the value, since
+             * the writers for those types expect local calendars.
+             */
+            timeZone = (SSType.DATETIMEOFFSET == destSSType) ? UTC.timeZone
+                                                             : new SimpleTimeZone(minutesOffset * 60 * 1000, "");
+
+            calendar = new GregorianCalendar(timeZone, Locale.US);
```
rene-ye, line 3630: Doesn't this break if the user's locale isn't US?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
```diff
@@ -919,23 +919,37 @@ private void writeTypeInfo(TDSWriter tdsWriter, int srcJdbcType, int srcScale, i
...
+                    break;
+                } else if (destSSType == SSType.SMALLMONEY) {
+                    tdsWriter.writeByte(TDSType.MONEYN.byteValue());
+                    tdsWriter.writeByte((byte) 4);
+                    break;
                 }
+                if (java.sql.Types.DECIMAL == srcJdbcType)
+                    tdsWriter.writeByte(TDSType.DECIMALN.byteValue()); // 0x6A
+                else
+                    tdsWriter.writeByte(TDSType.NUMERICN.byteValue()); // 0x6C
```
rene-ye, line 949:
```byte byteType = (java.sql.Types.DECIMAL == srcJdbcType) ? TDSType.DECIMALN.byteValue() : TDSType.NUMERICN.byteValue();
tdsWriter.writeByte(byteType);```

Change if block to above to shorten the code and make it more clear.

File: src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java (hunk 12 above)
Comment (peterbae) [reply src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java:3615]: Thanks, done.

File: src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java (hunk 13 above)
Comment (peterbae) [reply src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java:3630]: The locale doesn't matter. It's only used to determine what is the first day of the week (either saturday, sunday or monday) but that function isn't being used here, so I just put US locale (but anything works)

File: src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java (hunk 9 above)
Comment (peterbae) [reply src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java:3639]: I think it's easier to see this way.

File: src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java (hunk 14 above)
Comment (peterbae) [reply src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java:949]: done

File: src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java [hunk 15]
```diff
@@ -3574,27 +3596,87 @@ void writeTime(java.sql.Timestamp value, int scale) throws SQLServerException {
...
+        /*
+         * Out of all the supported temporal datatypes, DateTimeOffset is the only datatype that doesn't
+         * allow direct casting from java.sql.timestamp (which was created from a String).
+         * DateTimeOffset was never required to be constructed from a String, but with the
+         * introduction of extended bulk copy support for Azure DW, we now need to support this scenario.
+         * Parse the DTO as string if it's coming from a CSV.
+         */
+        if (value instanceof String) {
+            try {
+                String stringValue = (String) value;
```
ulvii, line 3611: Please add a comment which contains the format you are parsing.

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopySendTemporalDataTypesAsStringTest.java
peterbae: I think it is testing the datasource, because I'm calling setSendTemporalDataTypesAsStringForBulkCopy instead of using updateDataSource. I just checked as well to make sure it is using the datasource with the connection property set to false.

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopySendTemporalDataTypesAsStringTest.java
peterbae: Sometimes beforeEach didn't work for me properly, so I just call beforeEachSetup. There shouldn't be any functional differences.

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopySendTemporalDataTypesAsStringTest.java
```diff
@@ -0,0 +1,225 @@
...
+    private void testBulkCopyResultSet(Connection conn, SQLServerBulkCSVFileRecord fileRecord) {
+        try (SQLServerBulkCopy bulkCopy = new SQLServerBulkCopy(conn); Statement stmt = conn.createStatement()) {
+
+            fileRecord.addColumnMetadata(1, "id", java.sql.Types.INTEGER, 0, 0);
+            fileRecord.addColumnMetadata(2, "c1", java.sql.Types.DATE, 0, 0); // with Date
+            fileRecord.addColumnMetadata(3, "c2", java.sql.Types.TIMESTAMP, 0, 0); // with Datetime
+            fileRecord.addColumnMetadata(4, "c3", java.sql.Types.TIMESTAMP, 0, 7); // with Datetime2
+            fileRecord.addColumnMetadata(5, "c4", java.sql.Types.TIME, 0, 7); // with time
+            fileRecord.addColumnMetadata(6, "c5", microsoft.sql.Types.DATETIMEOFFSET, 0, 7); // with datetimeoffset
+            fileRecord.addColumnMetadata(7, "c6", java.sql.Types.TIMESTAMP, 0, 0); // with SmallDatetime
```
ulvii, line 152: The test fails when I switch to:
`fileRecord.addColumnMetadata(7, "c6", microsoft.sql.Types.SMALLDATETIME, 0, 0); // with SmallDatetime`

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
peterbae: done.

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -3574,27 +3596,100 @@ void writeTime(java.sql.Timestamp value, int scale) throws SQLServerException {
...
+                 * If the target data type is DATETIMEOFFSET, then use UTC for the calendar that
+                 * will hold the value, since writeRPCDateTimeOffset expects a UTC calendar.
+                 * Otherwise, when converting from DATETIMEOFFSET to other temporal data types,
+                 * use a local time zone determined by the minutes offset of the value, since
+                 * the writers for those types expect local calendars.
+                 */
+                timeZone = (SSType.DATETIMEOFFSET == destSSType) ? UTC.timeZone
+                                                                 : new SimpleTimeZone(minutesOffset * 60 * 1000, "");
+
+                calendar = new GregorianCalendar(timeZone, Locale.CHINA);
```
ulvii, line 3645:
```suggestion
                calendar = new GregorianCalendar(timeZone);
```

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopySendTemporalDataTypesAsStringTest.java
peterbae: It's supposed to fail. When performing bulk copy, the users are expected to provide the correct column type. The corresponding column type for smalldatetime is TIMESTAMP, and microsoft.sql.Types.SMALLDATETIME is not a supported column type. If you were to test that scenario without my changes, it will fail as well. The table that outlines which datatypes need to be specified for each datatype for bulk copy is outlined here: https://docs.microsoft.com/en-us/sql/connect/jdbc/using-basic-data-types?view=sql-server-2017

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
```diff
@@ -2397,8 +2468,20 @@ else if (4 >= bulkScale)
                             tdsWriter.writeByte((byte) 0x04);
                         else
                             tdsWriter.writeByte((byte) 0x05);
-
-                        tdsWriter.writeTime((java.sql.Timestamp) colValue, bulkScale);
+                        if (colValue instanceof String) {
```
ulvii, line 2471: Can you explain this with code comments? Also why do we have to parse for TIME, not for other types?

peterbae: I added comments. Other datatypes are already handled as if the data was string (this is the case for date, datetime and smalldatetime), so there was no need for extra steps. Time was the only datatype (other than DTO) that the code was assuming was coming on the form of a timestamp, so I needed to adjust accordingly.

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopySendTemporalDataTypesAsStringTest.java
lilgreenbird: sorry, you're right you are, as you're calling that explicitly after. But that's not how the other connection properties are handled. And updateDataSource needs to be updated to parse all the connection properties properly

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -3574,27 +3596,100 @@ void writeTime(java.sql.Timestamp value, int scale) throws SQLServerException {
...
-        subSecondNanos = dtoValue.getTimestamp().getNanos();
-        minutesOffset = dtoValue.getMinutesOffset();
+        /*
+         * Out of all the supported temporal datatypes, DateTimeOffset is the only datatype that doesn't
+         * allow direct casting from java.sql.timestamp (which was created from a String).
+         * DateTimeOffset was never required to be constructed from a String, but with the
+         * introduction of extended bulk copy support for Azure DW, we now need to support this scenario.
+         * Parse the DTO as string if it's coming from a CSV.
+         */
+        if (value instanceof String) {
```
ulvii, line 3609: Need to check for `connection.getSendTemporalDataTypesAsStringForBulkCopy()` here too.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
```diff
@@ -2397,8 +2468,30 @@ else if (4 >= bulkScale)
...
```
ulvii, line 2471: Need to check for `connection.getSendTemporalDataTypesAsStringForBulkCopy()` here too.

src/test/resources/BulkCopyCSVSendTemporalDataTypesAsStringForBulkCopy2.csv
```diff
@@ -0,0 +1,2 @@
+int,date,datetime,datetime2,time,datetimeoffset,smalldatetime,money,smallmoney
+1,0001-01-01,1753-01-01 00:00:00,0001-01-01 00:00:00,00:00:00,2025-12-10 12:32:10,1900-01-01 06:56:00,100.2523,100.2523
```
ulvii, line 2: New line here.

src/test/resources/BulkCopyCSVSendTemporalDataTypesAsStringForBulkCopy2.csv
```diff
@@ -0,0 +1,2 @@
+int,date,datetime,datetime2,time,datetimeoffset,smalldatetime,money,smallmoney
```
ulvii, line 1: Why do you need separate file for this test?

src/test/java/com/microsoft/sqlserver/jdbc/AlwaysEncrypted/BulkCopySendTemporalDataTypesAsStringAETest.java
```diff
@@ -0,0 +1,172 @@
+/*
```
ulvii, line 1: Please format all the files.

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
peterbae: The check is done before this method is called. Checking for String guarantees that that connection property is set to false, and also that the data is coming from a CSV. Near the beginning of writeColumnToTdsWriter method in SQLServerBulkCopy, this line:

`        } else if (null != serverBulkData && connection.getSendTemporalDataTypesAsStringForBulkCopy()) {
`

ensures that if getSendTemporalDataTypesAsStringForBulkCopy is true then bulkJdbcType will be varchar, which means that if getSendTemporalDataTypesAsStringForBulkCopy is false, we couldn't have come to this line.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
peterbae: Same logic applies here, but these are really good observations / questions.

src/test/resources/BulkCopyCSVSendTemporalDataTypesAsStringForBulkCopy2.csv
peterbae: will add.

src/test/resources/BulkCopyCSVSendTemporalDataTypesAsStringForBulkCopy2.csv
peterbae: I wanted to test the cases where the data in CSV doesn't "exactly" match the returned data. For example, it's legal to omit the nanosecond part of datetime2 value in the CSV, but the result from the server will have `0000000` (7 digits of nanoseconds) attached to the data. I wanted a separate CSV file to test those cases easily. It's only used in the `testSendTemporalDataTypesAsStringForBulkCopyOptional` method in my test class.

src/test/java/com/microsoft/sqlserver/jdbc/AlwaysEncrypted/BulkCopySendTemporalDataTypesAsStringAETest.java
peterbae: This file is formatted. Do you see something that should've been formatted? My formatter doesn't change anything in this file.

src/test/resources/BulkCopyCSVSendTemporalDataTypesAsStringForBulkCopy2.csv
peterbae: Actually, I don't think we should add a newline here, because it's a CSV file. AFAIK if a CSV file only had 1 column and the file has a newline at the end, I think that would be treated as a NULL value, which might not always be what's being tested (I remember specifically not leaving a newline at the end of CSV files for this reason)

src/test/java/com/microsoft/sqlserver/jdbc/AlwaysEncrypted/BulkCopySendTemporalDataTypesAsStringAETest.java
ulvii: The comments seem to be off.

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopySendTemporalDataTypesAsStringTest.java
peterbae: sure, i'll add that to my tests too then.

src/test/java/com/microsoft/sqlserver/jdbc/AlwaysEncrypted/BulkCopySendTemporalDataTypesAsStringAETest.java
peterbae: Which comment for example?

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopySendTemporalDataTypesAsStringTest.java
ulvii: Understood.

Issue #1301, opened 2020-04-02
Batch fails when always encrypted is enabled in connection string


jezbera:
## Driver version
8.2.2.jre11 (older versions are affected, too)

## SQL Server version
Microsoft SQL Server 2016 (SP2) (KB4052908) - 13.0.5026.0 (X64)   Mar 18 2018 09:11:49   Copyright (c) Microsoft Corporation  Express Edition (64-bit) on Windows 10 Pro 10.0 <X64> (Build 18363: ) (Hypervisor) 

## Client Operating System
Windows/Linux

## JAVA/JVM version
11.0.6

## Table schema
_Table will be created by test._
`create table TEST2 (KEY1 numeric(19,0) not null, KEY2 numeric(19,0) not null, primary key (KEY1, KEY2))`

## Problem description
Prepare database with always encrypted feature. Update connection details in ExecuteBatchTest.java and run tests in attached gradle project.
1. Expected behaviour:
No column is encrypted, so it does not matter, whether columnEncryptionSetting is enabled or not.
2. Actual behaviour:
Only test with connection string containing "columnEncryptionSetting=Enabled" fails.
3. Error message/stack trace:
`java.lang.AssertionError
	at com.microsoft.sqlserver.jdbc.TDSCommand.onRequestComplete(IOBuffer.java:7397)
	at com.microsoft.sqlserver.jdbc.TDSWriter.writePacket(IOBuffer.java:4071)
	at com.microsoft.sqlserver.jdbc.TDSWriter.endMessage(IOBuffer.java:3185)
	at com.microsoft.sqlserver.jdbc.TDSCommand.startResponse(IOBuffer.java:7556)
	at com.microsoft.sqlserver.jdbc.SQLServerPreparedStatement.doExecutePreparedStatementBatch(SQLServerPreparedStatement.java:2786)
	at com.microsoft.sqlserver.jdbc.SQLServerPreparedStatement$PrepStmtBatchExecCmd.doExecute(SQLServerPreparedStatement.java:2675)
	at com.microsoft.sqlserver.jdbc.TDSCommand.execute(IOBuffer.java:7194)
	at com.microsoft.sqlserver.jdbc.SQLServerConnection.executeCommand(SQLServerConnection.java:2979)
	at com.microsoft.sqlserver.jdbc.SQLServerStatement.executeCommand(SQLServerStatement.java:248)
[...]

PR #1334, opened 2020-05-12
Fix multiple executeBatch issue when AE is enabled

lilgreenbird:
fixes issue [1301](https://github.com/microsoft/mssql-jdbc/issues/1301)


ulvii:
[PR1334.zip](https://github.com/microsoft/mssql-jdbc/files/4783663/PR1334.zip)

src/test/java/com/microsoft/sqlserver/jdbc/AlwaysEncrypted/PrecisionScaleTest.java
```diff
@@ -54,6 +54,12 @@
 
     static String numericPrecisionTable[][] = {{"Float", "float"}, {"Decimal", "decimal"}, {"Numeric", "numeric"}};
 
+    static String intTable[][] = {{"int", "int"}};
```
peterbae, line 57: This variable seems like it's never used?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -2790,7 +2791,7 @@ final void doExecutePreparedStatementBatch(PrepStmtBatchExecCmd batchCommand) th
                     // that repreparation is necessary.
                     ++numBatchesPrepared;
                     needsPrepare = doPrepExec(tdsWriter, batchParam, hasNewTypeDefinitions, hasExistingTypeDefinitions);
-                    if (needsPrepare || numBatchesPrepared == numBatches) {
+                    if (reqStarted || needsPrepare || numBatchesPrepared == numBatches) {
```
peterbae, line 2794: I think this solution is a roundabout solution to the problem (which might also cause other regressions). The reason why we are getting an assertion error is because in AE enabled cases, we're retrieving the parameter encryption metadata twice in the second iteration of execute batch. Not only is this second retrieval of parameter encryption metadata hurting performance, it had the side effect of making the driver think that the request had been completed (getting the parameter encryption metadata counts as a command, which sets the `requestComplete` (and other flags) to true, which causes assertion error when we get to the actual execution of the prepared statement). You can check this behavior in the image below from the SQL profiler:

image

If you were to compare this image to the non-AE path or the solution I've provided below, you'll see that the driver will no longer retrieve the paramter metadata twice.

My suggested solution is to instead call `encryptionMetadataIsRetrieved = true;` in line 2744 of this file, so that we only retrieve the param metadata once, which I believe is the correct way to solve this problem.

Let me know once you've confirmed if the fix will work (and tested it internally as well), so I can make a jar out of this PR and see if it solves the NPE that the user is experiencing from issue #1360.

peterbae:
Continued in PR #1378

PR #1336, opened 2020-05-14
Add delayLoadingLobs

rene-ye:
Adds a property delayLoadingLobs which will cause all LOB objects retrieved from the ResultSet to not be streamed.

The driver does this by setting a boolean on every LOB object when retrieved via a getter. This boolean is then used by the LOB object when users call stream getters.


src/main/java/com/microsoft/sqlserver/jdbc/ISQLServerDataSource.java
```diff
@@ -992,6 +992,21 @@
     void setClientKeyPassword(String password);
 
     /**
+     * Specifies the flag to load LOBs instead of stream them.
```
peterbae, line 995: instead of streaming them.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBlob.java
```diff
@@ -53,6 +53,12 @@
 
     private final String traceID;
 
+    private boolean loadLob = false;
+
+    void setStreaming(boolean b) {
+        loadLob = b;
```
peterbae, line 59: Can you refactor this `loadLob` variable name to `delayLoadingLobs` to maintain consistency? It will make it easier to search for its usage in the future if we were to come back to this feature

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResource.java
```diff
@@ -643,5 +643,7 @@ static String getResource(String key) {
             {"R_pvkParseError", "Could not read Private Key from PVK, check the password provided."},
             {"R_pvkHeaderError", "Cannot parse the PVK, PVK file does not contain the correct header."},
             {"R_clientCertError", "Reading client certificate failed. Please verify the location of the certificate."},
-            {"R_unassignableError", "The class specified by the {0} property must be assignable to {1}."}};
+            {"R_unassignableError", "The class specified by the {0} property must be assignable to {1}."},
+            {"R_delayLoadingLobsPropertyDescription",
+            "Boolean flag which indicates whether the driver will load LOB datatypes into memeory."}};
```
peterbae, line 648: We usually put descriptions along with other descriptions, so to not mix errors and descriptions together. Can you move this description to where the other descriptions are?

src/test/java/com/microsoft/sqlserver/jdbc/unit/lobs/LobsTest.java
```diff
@@ -9,12 +9,16 @@
 import static org.junit.jupiter.api.Assertions.assertTrue;
```
peterbae, line 9: Add DataSource getter/setter test cases.

src/test/java/com/microsoft/sqlserver/jdbc/unit/lobs/LobsTest.java
```diff
@@ -579,6 +583,221 @@ public void readMultipleBlobStreamsThenCloseRS() throws Exception {
...
+            Blob b = conn.createBlob();
+            Constants.RANDOM.nextBytes(data);
+            b.setBytes(1, data);
+            pstmt.setInt(1, 1);
+            pstmt.setBlob(2, b);
+            pstmt.executeUpdate();
+        }
+
+        String streamString = TestUtils.addOrOverrideProperty(connectionString, "delayLoadingLobs", "true");
+        String loadedString = TestUtils.addOrOverrideProperty(streamString, "delayLoadingLobs", "false");
```
peterbae, line 605: addOrOverrideProperty actually only appends the connection property at the end, so `loadedString` ends up having both delayLoadingLobs=true and delayedLoadingLobs=false in it. I tested to make sure that the test is working as intended (as in, loadedString should honor delayLoadingLobs=false), but you should be aware of this fyi.

src/test/java/com/microsoft/sqlserver/jdbc/unit/lobs/LobsTest.java
```diff
@@ -579,6 +583,221 @@ public void readMultipleBlobStreamsThenCloseRS() throws Exception {
         }
     }
 
+    @Test
```
peterbae, line 586: This is a public method so it needs javadoc.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBlob.java
peterbae: never mind, I realized that this variable is actually the opposite of `delayLoadingLobs`. I'd prefer if we had one type of boolean to represent the connection property (instead of having two booleans opposite of each other's values that are meant for the same purpose), but if others are fine with this, I am too.

src/test/java/com/microsoft/sqlserver/jdbc/unit/lobs/LobsTest.java
```diff
@@ -9,12 +9,16 @@
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.io.BufferedInputStream;
+import java.io.BufferedReader;
```
peterbae, line 12: I think you should add a case where you create a stream out of a ResultSet, close the resultset, then try to read from the stream. The tests below test that existing functionality is still supported, but doesn't verify the case that was intended to be fixed in this PR.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBlob.java
peterbae: Overall I think this setStreaming naming is confusing. The connection property `delayLoadingLobs` is true by default, which means we want to stream by default. Setting `delayLoadingLobs` to false ends up calling this method setStreaming to true, which makes the driver *not* stream (by calling getStringFromStream method). Unless I'm mistaken about the functionalities, we should rename the methods / variables to add more clarity to what the driver is doing.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBlob.java
```diff
@@ -143,6 +149,11 @@ private void checkClosed() throws SQLServerException {
     @Override
     public InputStream getBinaryStream() throws SQLException {
         checkClosed();
+        // If the LOB is currently streaming and the stream hasn't been read, read it.
+        if (null == value && !activeStreams.isEmpty() && loadLob) {
```
peterbae, line 153: I think a minor improvement here would be consolidate the two if statements that are on line 153 and 157 into one, then have a nested if statement that only checks for `loadLob` since the other two conditions are checked twice.

peterbae: Also applies to the SQLServerClob file.

TarasTielkes:
(I'm adding this comment as a user that follows the project, but does not actively participate - if this is not the right place for this comment, please let me know)

It would be good to document the _intent_ of this API. _Why_ would I, as a user of the driver, want to set the _loadLargeObjects_ property? What are the benefits, and what are the downsides? Interacting with large volumes of large BLOBs is a tricky area in my experience, so a clear description of what this API intends to facilitate would be a good idea, methinks.

I do agree with the comments from @peterbae regarding the naming, FWIW.

src/test/java/com/microsoft/sqlserver/jdbc/unit/lobs/LobsTest.java
rene-ye: Yeah the function doesn't erase the contents but our driver respects the last declared connection property for duplicates.

src/test/java/com/microsoft/sqlserver/jdbc/unit/lobs/LobsTest.java
rene-ye: testContinuousReading() tests the case that was intended to be fixed.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBlob.java
rene-ye: Can't be done, the point is to do this before the lobs create the streams, if we nest the if's, it'll result in the lobs loading the stream into a string and then attempting to return a streaming inputstream.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerLob.java
```diff
@@ -21,4 +21,13 @@
      * @throws SQLException
      */
     abstract void fillFromStream() throws SQLException;
+
+    protected boolean delayLoadingLob = true;
+    /**
+     * Provides functionality for the result set to set whether to load the LOB objects fully. Setting this property to
+     * TRUE will cause LOBs to be loaded into memory. The default behavior is FALSE.
+     */
+    void setDelayLoadingLob() {
```
peterbae, line 30: Shouldn't this setter take in a boolean value, so that the user can go back and forth between setting this to true or false?  The comment also seems to suggest this setter should be able to provide that functionality.

peterbae: Also since other `setDelayLoadingLobs`s have an `s` at the end, add an `s` to this method's name as well.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResource.java
```diff
@@ -643,5 +645,5 @@ static String getResource(String key) {
             {"R_pvkParseError", "Could not read Private Key from PVK, check the password provided."},
             {"R_pvkHeaderError", "Cannot parse the PVK, PVK file does not contain the correct header."},
             {"R_clientCertError", "Reading client certificate failed. Please verify the location of the certificate."},
-            {"R_unassignableError", "The class specified by the {0} property must be assignable to {1}."}};
+            {"R_unassignableError", "The class specified by the {0} property must be assignable to {1}."},};
```
peterbae, line 648: Remove this comma at the end

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerLob.java
rene-ye: Not necessary, this controls the option for the single LOB. The ResultSet will read the property from the Connection object and determine whether to call this API or not. The user is able to toggle it on and off with the current logic.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResource.java
rene-ye: I prefer to actually leave the commas for these lists in so when we add lines it's obvious where to add them, and it's more convenient to just copy+past the last line.

src/test/java/com/microsoft/sqlserver/jdbc/unit/lobs/LobsTest.java
peterbae, line 12: I don't think this file is formatted, can you format it?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBlob.java
```diff
@@ -143,6 +143,11 @@ private void checkClosed() throws SQLServerException {
     @Override
     public InputStream getBinaryStream() throws SQLException {
         checkClosed();
+        // If the LOB is currently streaming and the stream hasn't been read, read it.
+        if (!delayLoadingLob && null == value && !activeStreams.isEmpty()) {
```
lilgreenbird, line 147: null check for activeStream

peterbae: I think this is fine, it's impossible for that variable to be null, and many other parts of this class doesn't null check that variable.

lilgreenbird: it would be null if free is called before. That would be an error, but should not cause driver to crash on null exception

peterbae: Well...after free() the blob object is invalid, so the user shouldn't be using the blob after free() is called. Also, you'll see that on line 151 that this method wasn't doing a null check before anyways, but I'll leave it up to @rene-ye to make the call.

rene-ye: After the lob is freed it won't get past checkClosed().

ulvii:
Hi @TarasTielkes ,
Thanks for the suggestions. We will definitely document the property on MSDN once it is in production release. To give you a brief idea about it, as of 7.0.0 version of the driver, we introduced some changes to the way the driver handles LOB streaming and the new property is meant to provide backward compatibility with 6.4.0 version of the driver for some edge cases.

Issue #1270, opened 2020-03-02
inserting large-ish multipolygon w/ prepared statement using `setObject(...)` fails


mprins:
## Driver version
8.2.1.jre8

## SQL Server version

docker  versions of mssql 2017 and 2019:
- Microsoft SQL Server 2017 (RTM-CU17) (KB4515579) - 14.0.3238.1 (X64) 
	Sep 13 2019 15:49:57 
	Copyright (C) 2017 Microsoft Corporation
	Developer Edition (64-bit) on Linux (Ubuntu 16.04.6 LTS)
- Microsoft SQL Server 2019 (RTM-CU2) (KB4536075) - 15.0.4013.40 (X64) 
	Feb  3 2020 16:40:57 
	Copyright (C) 2019 Microsoft Corporation
	Developer Edition (64-bit) on Linux (Ubuntu 16.04.6 LTS) <X64>  

## Client Operating System
Ubuntu 19.10

## JAVA/JVM version
 
openjdk version "1.8.0_242"
OpenJDK Runtime Environment (build 1.8.0_242-8u242-b08-0ubuntu3~19.10-b08)
OpenJDK 64-Bit Server VM (build 25.242-b08, mixed mode)

## Table schema

```sql
CREATE DATABASE geomtest;
CREATE TABLE geomtest.dbo.geom (id NUMERIC(19,0) NOT NULL, geom geometry NOT NULL );
```

You can:

```sh
docker pull mcr.microsoft.com/mssql/server:2017-latest
docker run --rm -e 'ACCEPT_EULA=Y' -e 'MSSQL_SA_PASSWORD=YourStrong!Passw0rd' \
-p 1401:1433 --name sql1 -h sql1 \
-d mcr.microsoft.com/mssql/server:2017-latest

sqlcmd -S localhost,1401 -U SA -P 'YourStrong!Passw0rd' -Q "CREATE DATABASE geomtest;"
sqlcmd -S localhost,1401 -U SA -P 'YourStrong!Passw0rd' -Q "CREATE TABLE geomtest.dbo.geom (id NUMERIC(19,0) NOT NULL, geom geometry NOT NULL );"
```

## Problem description

When inserting a certain geometry using a prep. statement with `setObject(...)` a failure is generated, while inserting the geometry using `setSting(...)` + `geometry::STGeomFromText(...)` works. Using `setObject()` works for other geometries, validating the WKT in the database yields "24400: Valid" for the testdata.
[...]

PR #1337, opened 2020-05-15
Fix inserting large spatial datatypes using setObject or setGeometry/setGeography

peterbae:
Fixes issue #1270.

When the driver is inserting byte[] data into a column in the server's table, the driver changes the JdbcType of the column (effectively, we change what we send the data as in the INSERT query) if the byte array is larger than the 8000. For example, this means that if the driver was going to insert the data as geometry, it was instead sending varbinary(max) if the byte array that represents the geometry/geography data was larger than 8000. Excluding geometry/geography from being typecast into varbinary(max) solves this issue (geometry/geography data types don't have a size limit).


src/test/java/com/microsoft/sqlserver/jdbc/datatypes/SQLServerSpatialDatatypeTest.java
```diff
@@ -1142,6 +1142,63 @@ public void testWrongtype() throws SQLException {
...
+        beforeEachSetup();
+
+        Geometry geomWKT = Geometry.STGeomFromText(wktLargeGeometry, 0);
```
mprins, line 1157: I don't it matters much, but these coordinates are in SRID 28992

ulvii:
[PR1337.zip](https://github.com/microsoft/mssql-jdbc/files/4652810/PR1337.zip)
@mprins 
I attached the binaries to this PR, would you give them a try and let us know if your issues are resolved?

mprins:
@ulvii Thanks, I can/will give this a try at the end of the week at the earliest

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/SQLServerSpatialDatatypeTest.java
peterbae: Thanks. For the purpose of this test, it shouldn't make a difference.

peterbae:
Hi @mprins, did you have a chance to test out the changes?

mprins:
I've ran 2 projects with this updated driver and all tests appear to be passing, including that reported in #1270

PR #1338, opened 2020-05-20
Feature | Added support for Data Classification v2

lilgreenbird:


src/test/java/com/microsoft/sqlserver/jdbc/resultset/DataClassificationTest.java
```diff
@@ -10,13 +10,19 @@
 import java.sql.Statement;
```
ulvii, line 10: I think Azure now has DataClassification support. Could you remov Azure tags from tests and check if they pass?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDriver.java
```diff
@@ -455,6 +455,35 @@ public String toString() {
...
+    private SensitivityRank(int rank) {
+        this.rank = rank;
+    }
+
+    public int getValue() {
+        return rank;
+    }
+
+    static boolean isValid(int rank) throws SQLServerException {
+        for (SensitivityRank r : SensitivityRank.values()) {
```
ulvii, line 477: Make values() static array for better performance. See https://github.com/microsoft/mssql-jdbc/pull/1065 for reference.

src/main/java/com/microsoft/sqlserver/jdbc/StreamColumns.java
```diff
@@ -259,55 +259,73 @@ private SensitivityClassification processDataClassification(TDSReader tdsReader)
...
 
-        List<InformationType> informationTypes = new ArrayList<InformationType>(numInformationTypes);
-        for (int i = 0; i < numInformationTypes; i++) {
+        List<InformationType> informationTypes = new ArrayList<InformationType>(informationTypeCount);
+        for (int i = 0; i < informationTypeCount; i++) {
             informationTypes.add(readSensitivityInformationType(tdsReader));
         }
 
+        int sensitivityRank = SensitivityRank.NOT_DEFINED.getValue();
+        if (TDS.MAX_SUPPORTED_DATA_CLASSIFICATION_VERSION <= tdsReader.getServerSupportedDataClassificationVersion()) {
```
ulvii, line 278: If `getServerSupportedDataClassificationVersion` is 1, the driver will still read the rank. Don't we want to read rank only if the version is at least 2?

src/main/java/com/microsoft/sqlserver/jdbc/dataclassification/SensitivityProperty.java
```diff
@@ -19,10 +20,13 @@
      *        Label as received from SQL Server for this SensitivityProperty
      * @param informationType
      *        InformationType as received from SQL Server for this SensitivityProperty
+     * @param sensitivityRank
+     *        sensitivity rank as received from SQL Server for this SensitivityProperty
      */
-    public SensitivityProperty(Label label, InformationType informationType) {
+    public SensitivityProperty(Label label, InformationType informationType, int sensitivityRank) {
```
ulvii, line 26: This is a breaking change! I suggest to create another constructor instead.

src/main/java/com/microsoft/sqlserver/jdbc/StreamColumns.java
peterbae: No, it looks like if `getServerSupportedDataClassificationVersion` is 1, then we won't read the rank because 2 <= 1 is false?

lilgreenbird: yes, what @peterbae said

src/test/java/com/microsoft/sqlserver/jdbc/resultset/DataClassificationTest.java
lilgreenbird: ? yes it's supported and had been removed and tested. The only place that has these tags is in the negative test

Issue #1352, opened 2020-06-04
IS_AUTOINCREMENT & IS_GENERATEDCOLUMN returning the incorrect type (Azure DW)


tsegall:
## Driver version
8.2.0

## SQL Server version
Microsoft Azure SQL Data Warehouse - 10.0.11115.0 May 16 2020 02:43:26 Copyright (c) Microsoft Corporation

## Client Operating System
MacOS X 10.15.4 (although the problem is present on both Windows & Linux)

## JAVA/JVM version
OpenJDK Runtime Environment Corretto-8.202.08.2 (build 1.8.0_202-b08)

## Problem description
1. Expected Behavior
databaseMetaData.getColumns(catalog, schema, tableName, null))
should return the correct type for all columns

2. Actual behaviour:
The correct type is not returned for the columns IS_AUTOINCREMENT and IS_GENERATEDCOLUMN.  These are returned as type 4 (Int) when they should actually be a string.  This means that when you attempt to do a resultSet.getInt() you receive a ClassCastException.  The actual value should be YES or NO - so the issue is that the type being returned is Int.  Note: this is an Azure DW problem only and does not impact the other MS databases.

## Reproduction code
Have a code example that illustrates the issue - if that would help, but the bug is relatively easy to reproduce.

ulvii:
Hi @tsegall ,

> This means that when you attempt to do a resultSet.getInt() you receive a ClassCastException.

I am confused a bit, are you trying to retrieve `IS_AUTOINCREMENT`, `IS_GENERATEDCOLUMN` using  `resultSet.getInt()`? This will obviously result in `ClassCastException` the possible values for these columns are `YES/NO`.

> These are returned as type 4 (Int) when they should actually be a string.

I can't reproduce this behavior, the driver returns `YES/NO`, please share a table schema and JDBC repro code if you want us to investigate this further.

tsegall:
The issue is that the result of calling rsmd.getColumnType(i) on these two columns returns 4 and hence according to the JDBC specification you should use getInt() to retrieve their values.

I have attached sample code which illustrates this problem - if you run this sample code an another database (e.g. Oracle, Postgres, etc) it will work in addition the code will work on SQL Server (both hosted and on-premise.  It does not work on Azure DW.
Regards, Tim.

tiny.txt](https://github.com/microsoft/mssql-jdbc/files/4733613/tiny.txt)

ulvii:
Hi @tsegall ,
[...]

PR #1356, opened 2020-06-09
Fixed an issue with getColumns() not returning correct type for IS_AUTOINCREMENT and IS_GENERATEDCOLUMN against Azure Data Warehouse

ulvii:
- Modified the query for Azure DW to select 0/1 as YES/NO.
- Improved getColumns() test.

Fix for #1352

### DIFF
--- a/src/test/java/com/microsoft/sqlserver/jdbc/databasemetadata/DatabaseMetaDataTest.java
+++ b/src/test/java/com/microsoft/sqlserver/jdbc/databasemetadata/DatabaseMetaDataTest.java
@@ -24,6 +24,7 @@ import java.sql.ResultSet;
 import java.sql.ResultSetMetaData;
 import java.sql.SQLException;
 import java.sql.Statement;
+import java.sql.Types;
 import java.text.MessageFormat;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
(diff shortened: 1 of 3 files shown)


ulvii:
Jar for testing.
[mssql-jdbc-8.4.0-SNAPSHOT.jre8.zip](https://github.com/microsoft/mssql-jdbc/files/4749192/mssql-jdbc-8.4.0-SNAPSHOT.jre8.zip)

src/test/java/com/microsoft/sqlserver/jdbc/databasemetadata/DatabaseMetaDataTest.java
```diff
@@ -651,12 +653,64 @@ public void testGetColumn() throws SQLException {
...
                 int rowCount = 0;
                 while (resultSet.next()) {
                     for (int i = 1; i < rsmd.getColumnCount(); i++) {
-                        assertEquals(rsmd.getColumnName(i), getColumnsDWColumns.values().toArray()[i - 1]);
+                        String columnName = rsmd.getColumnName(i);
+                        Object value = resultSet.getObject(columnName);
+                        if (0 == rowCount) {
+                            Object expectedValue = firstRow.get(columnName);
+                            if (expectedValue instanceof String) {
+                                assertEquals(firstRow.get(columnName).toString().toLowerCase(),
```
peterbae, line 705: The fix is good (as confirmed by the customer), but this test isn't accurately testing the changes made by this PR. This new test will pass with or without the changes made in this PR, because when it's time to test the values of the `IS_AUTOINCREMENT` entry, the filter will change the `firstRow.get(columnName).toString()` from `0` to `NO`. To ensure that we've addressed the issue (so that we don't change these values in the future, for example), I think we should additionally check to make sure that `rsmd.getColumnType()` method on the two columns should return 12 (varchar) instead of 4 (int). Other than that, the changes look good to me.

ulvii: You are right, updated the test.

Issue #1280, opened 2020-03-06
STAsBinary is not returning in WKB format


humada05:
## Driver version

7.4.1.jre8 (but includes all versions up til now)

## Problem description

1. Expected behaviour:
`STAsBinary()` method should return the spacial object in WKB format, [as detailed here](https://en.wikipedia.org/wiki/Well-known_text_representation_of_geometry#Well-known_binary).

2. Actual behaviour:
`STasBinary()` returns the spacial object in CLR format.

Issue created from discussion in #1277.

peterbae:
Thanks for creating this issue @humada05. The team will look into this as discussed.

ulvii:
Potentially related to #1270.

peterbae:
Hi @humada05, I've created PR #1364 that fixes this issue. I'm attaching the jars here with the fix included. Could you try it out and let me know if STAsBinary is working as intended now? Thanks.
[github-1280.zip](https://github.com/microsoft/mssql-jdbc/files/4800948/github-1280.zip)

peterbae:
@humada05 were you able to find some time to test out my fix? Just double checking.

humada05:
Hey @peterbae, awesome! I'll be sure to test this out when i get some free time this week.

peterbae:
Fixed in #1364. Please let me know if you encounter any issues.

PR #1364, opened 2020-06-18
Fix STAsBinary to return WKB format instead of CLR

peterbae:
Fixes issue #1280.

STAsBinary is a method that's supported by SSMS to return a WKB (Well-Known binary) format of a Geometry/Geography instance. The driver also implemented this method when it first implemented spatial datatypes, but it was wrongly returning a CLR (Microsoft's internal representation of a Geometry/Geography instance, different from WKB) value instead. This PR adds new logic to parse existing spatial datatype instance into a WKB format.

The counterpart method to this method, STGeomFromWKB, (which takes in a WKB and returns a Geometry/Geography instance) will be fixed at a later date, and I've put a note in the Javadocs telling users to use deserialize(byte[]) instead which does the exact same thing as current STGeomFromWKB.


src/test/java/com/microsoft/sqlserver/jdbc/datatypes/SQLServerSpatialDatatypeTest.java
```diff
@@ -51,141 +52,1161 @@
     static boolean isDenaliOrLater = false;
 
     @Test
```
lilgreenbird, line 54: tests should have a header briefly describing what the tests are testing

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/SQLServerSpatialDatatypeTest.java
```diff
@@ -51,141 +52,1161 @@
...
+        Geometry geomWKT = Geometry.deserialize(geomCLR);
+        Geography geogWKT = Geography.deserialize(geogCLR);
+        assertEquals(geomWKT.asTextZM(), geoWKT);
+        assertEquals(geogWKT.asTextZM(), geoWKT);
+        assertTrue(Arrays.equals(geomWKT.STAsBinary(), geoWKB));
+        assertTrue(Arrays.equals(geogWKT.STAsBinary(), geoWKB));
+    }
+
+    @Test
+    public void testPoint2() throws SQLException {
```
lilgreenbird, line 71: all these testsBlah and testBlah2 tests are identical other than test data, can we refactor them to call a common test method instead?

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/SQLServerSpatialDatatypeTest.java
```diff
@@ -821,7 +1842,7 @@ public void testParse() throws SQLException {
 
     @Test
     @Tag(Constants.xAzureSQLDW)
-    public void testPoint() throws SQLException {
+    public void testSendBasicPoint() throws SQLException {
         beforeEachSetup();
```
lilgreenbird, line 1846: can we rename this, as initially I thought this was a hack to avoid @BeforeEach. But actually this isn't called before each tests.

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/SQLServerSpatialDatatypeTest.java
```diff
@@ -51,141 +52,1161 @@
     static boolean isDenaliOrLater = false;
```
lilgreenbird, line 53: add a negative test for DW to make sure it throws the proper error

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/SQLServerSpatialDatatypeTest.java
lilgreenbird: and, so if these table names are only used in the before each test which use them, then it shouldn't be in the @BeforeAll method right? (lines 2273-2275) Should move to this beforeEachSetup() method instead. And, can we do the escape quotes at the same time we're generating the names so we don't have to do it everywhere?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerSpatialDatatype.java
```diff
@@ -48,13 +48,32 @@
...
-    protected Segment segments[];
+    protected Figure figures[] = {};
+    protected Shape shapes[] = {};
+    protected Segment segments[] = {};
+
+    // WKB properties
+    protected byte[] wkb;
+    protected byte endian = 1; // little endian
+    protected int wkbType;
+    // Open Geospatial Consortium specifications
```
ulvii, line 67: Multi-line comment.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerSpatialDatatype.java
```diff
@@ -48,13 +48,32 @@
...
-    protected Figure figures[];
-    protected Shape shapes[];
-    protected Segment segments[];
+    protected Figure figures[] = {};
+    protected Shape shapes[] = {};
+    protected Segment segments[] = {};
+
+    // WKB properties
+    protected byte[] wkb;
+    protected byte endian = 1; // little endian
```
ulvii, line 65: Is the returned value little endian only?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerSpatialDatatype.java
```diff
@@ -184,23 +203,544 @@ protected void serializeToWkb(boolean excludeZMFromWKB, SQLServerSpatialDatatype
...
+                break;
+            case CURVEPOLYGON:
+                addCurvePolygonToBuffer(buf, numberOfFigures);
+                break;
+            case FULLGLOBE:
+                addFullGlobeToBuffer(buf);
+                break;
+            default:
+                // Shouldn't be possible to come here, existing Geometry/Geography doesn't support other cases
+                break;
```
ulvii, line 282: Need to throw an exception here instead.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerSpatialDatatype.java
ulvii: For feature when we implement STGeomFromWKB, we should keep in mind that both values are possible when user provides WKB.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerSpatialDatatype.java
```diff
@@ -184,23 +203,544 @@ protected void serializeToWkb(boolean excludeZMFromWKB, SQLServerSpatialDatatype
...
+                case CURVEPOLYGON:
+                    if (shapes[currentWKBShapeIndex].getFigureOffset() == -1) {
+                        addCurvePolygonToBuffer(buf, 0);
+                    } else {
+                        addCurvePolygonToBuffer(buf, calculateNumFiguresInThisShape(false));
+                    }
+                    currentWKBShapeIndex++;
+                    break;
+                default:
+                    // Shouldn't be possible to come here, existing Geometry/Geography doesn't support other cases
```
ulvii, line 479: Same here and in a few other default cases below, instead of ignoring unexpected values, we should throw and exception.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerSpatialDatatype.java
```diff
@@ -184,23 +203,544 @@ protected void serializeToWkb(boolean excludeZMFromWKB, SQLServerSpatialDatatype
...
+            remainingStructureCount--;
+        }
+    }
+
+    private int calculateNumPointsInThisFigure() {
+        if (figures.length == 0) {
+            return 0;
+        }
+
+        return (currentWKBFigureIndex == figures.length - 1)
```
ulvii, line 616: Can you add comment explaining what this block does?

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/SQLServerSpatialDatatypeTest.java
ulvii, line 71: testEmptyPoint() is better name?

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/SQLServerSpatialDatatypeTest.java
```diff
@@ -51,141 +52,1161 @@
...
+        Geometry geomWKT = Geometry.deserialize(geomCLR);
+        Geography geogWKT = Geography.deserialize(geogCLR);
+        assertEquals(geomWKT.asTextZM(), geoWKT);
+        assertEquals(geogWKT.asTextZM(), geoWKT);
+        assertTrue(Arrays.equals(geomWKT.STAsBinary(), geoWKB));
+        assertTrue(Arrays.equals(geogWKT.STAsBinary(), geoWKB));
+    }
+
+    @Test
+    public void testPolygon2() throws SQLException {
```
ulvii, line 133: testPolygonEmpty()

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerSpatialDatatype.java
peterbae: Yes. Returned value is little endian only (as SQL Server does). I will keep this in mind when the user provides WKB.

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/SQLServerSpatialDatatypeTest.java
```diff
@@ -51,141 +52,1161 @@
...
+    public void testGeometryCollection6() throws SQLException {
+        String geoWKT = "GEOMETRYCOLLECTION(LINESTRING EMPTY)";
+        byte[] geomCLR = hexStringToByteArray(
+                "100000000104000000000000000002000000FFFFFFFFFFFFFFFF0700000000FFFFFFFF02");
+        byte[] geogCLR = hexStringToByteArray(
+                "E61000000104000000000000000002000000FFFFFFFFFFFFFFFF0700000000FFFFFFFF02");
+        byte[] geoWKB = hexStringToByteArray("010700000001000000010200000000000000");
+        Geometry geomWKT = Geometry.deserialize(geomCLR);
+        Geography geogWKT = Geography.deserialize(geogCLR);
+        assertEquals(geomWKT.asTextZM(), geoWKT);
```
ulvii, line 396: In all these tests, expected value should come first

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerSpatialDatatype.java
peterbae: Absolutely.

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/SQLServerSpatialDatatypeTest.java
peterbae: Well...we can't. Geometry/Geography are not supported datatypes against DW. We wouldn't be able to test it because we can't create a table with those column types.

lilgreenbird: that's what I mean, add a negative test that runs on DW and check that it will fail with the appropriate unsupported error msg

peterbae: That's not a negative test, that's just an unsupported scenario that shouldn't be concerned with the driver.

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/SQLServerSpatialDatatypeTest.java
peterbae: I've added comments to tests instead.

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/SQLServerSpatialDatatypeTest.java
lilgreenbird: why not take the suggestion above to rename all the testBlah2 tests to more meaningful names?

peterbae: I wanted to stick to numberings because geometrycollection has up to 50 different ones, but I'll change these to "empty" at the end since you insist.

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/SQLServerSpatialDatatypeTest.java
```diff
@@ -1264,6 +2184,16 @@ private void testWkt(String geoWKT, String geoWKTSS) throws SQLException {
         return data;
     }
 
+    private void validateGeometryGeography(String geoWKT, byte[] geomCLR, byte[] geogCLR,
+            byte[] geoWKB) throws SQLServerException {
+        Geometry geomWKT = Geometry.deserialize(geomCLR);
+        Geography geogWKT = Geography.deserialize(geogCLR);
+        assertEquals(geoWKT, geomWKT.asTextZM());
+        assertEquals(geoWKT, geogWKT.asTextZM());
+        assertTrue(Arrays.equals(geomWKT.STAsBinary(), geoWKB));
```
ulvii, line 2193: Why can't you just remove assertTrue?

ulvii, line 2193: Here and in some the tests above, please make sure expected value is first.

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/SQLServerSpatialDatatypeTest.java
```diff
@@ -50,144 +51,1015 @@
...
+     */
+    @Test
+    public void testFullGlobe() throws SQLException {
         String geoWKT = "FULLGLOBE";
-        byte[] geogWKB = hexStringToByteArray("E61000000224000000000000000001000000FFFFFFFFFFFFFFFF0B");
-        Geography geogWKT = Geography.deserialize(geogWKB);
+        byte[] geogCLR = hexStringToByteArray("E61000000224000000000000000001000000FFFFFFFFFFFFFFFF0B");
+        byte[] geoWKB = hexStringToByteArray("017E000000");
+        Geography geogWKT = Geography.deserialize(geogCLR);
         assertEquals(geogWKT.asTextZM(), geoWKT);
```
ulvii, line 1056: Expected value first

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/SQLServerSpatialDatatypeTest.java
```diff
@@ -50,144 +51,1015 @@
...
+        assertTrue(Arrays.equals(geogWKT.STAsBinary(), geoWKB));
```
ulvii, line 1057: Can remove assertTrue

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/SQLServerSpatialDatatypeTest.java
peterbae: No, the assertTrue is the part where the STAsBinary method (the one I'm adding) is getting tested, unless I misunderstood your question.

peterbae: My bad, I forgot these ones.

peterbae: Actually, in this line, the expected value is coming first, right? I already swapped them this morning because you asked me to.

ulvii: `assertArrayEquals(geoWKB, geomWKT.STAsBinary());`

ulvii: geoWKB is the expected value

peterbae: I missed that one. I'll swap to assertArrayEquals as well.

PR #1368, opened 2020-06-24
Fix issue when tests throw a different error message when AE is on

rene-ye:
Tests throw a different error because the column name is invalid and cannot be passed into spde.

### DIFF
--- a/src/test/java/com/microsoft/sqlserver/jdbc/unit/lobs/LobsTest.java
+++ b/src/test/java/com/microsoft/sqlserver/jdbc/unit/lobs/LobsTest.java
@@ -41,7 +41,6 @@ import org.junit.platform.runner.JUnitPlatform;
 import org.junit.runner.RunWith;
 
 import com.microsoft.sqlserver.jdbc.RandomUtil;
-import com.microsoft.sqlserver.jdbc.SQLServerConnection;
 import com.microsoft.sqlserver.jdbc.SQLServerDataSource;
 import com.microsoft.sqlserver.jdbc.TestResource;
 import com.microsoft.sqlserver.jdbc.TestUtils;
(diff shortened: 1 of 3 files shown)


src/test/java/com/microsoft/sqlserver/jdbc/unit/lobs/LobsTest.java
```diff
@@ -201,7 +202,8 @@ else if (lobClass == Blob.class)
                         if (!verified) {
                             // Odd CharacterStream length will throw this exception
                             if (!e.getMessage().contains(TestResource.getResource("R_badStreamLength"))) {
-                                assertTrue(e.getMessage().contains(TestResource.getResource("R_streamReadError")));
+                                assertTrue(e.getMessage().contains(TestResource.getResource("R_streamReadError"))
+                                        || e.getMessage().contains(TestResource.getResource("R_aeStreamReadError")));
```
lilgreenbird, line 206: should we check if AE is enabled to see which error it should be?

src/test/java/com/microsoft/sqlserver/jdbc/unit/lobs/LobsTest.java
```diff
@@ -41,6 +41,7 @@
 import org.junit.runner.RunWith;
 
 import com.microsoft.sqlserver.jdbc.RandomUtil;
+import com.microsoft.sqlserver.jdbc.SQLServerConnection;
```
lilgreenbird, line 44: why?

src/test/java/com/microsoft/sqlserver/jdbc/unit/lobs/LobsTest.java
rene-ye: The only really good way is to check the connection string, and that's honestly pretty hacky and I don't want to add the extra logic when the test doesn't really care if AE is on or off.

lilgreenbird: yeah I did that in PR 1371, we should agree to do the same thing, if we're not going to check then I'll remove that from that PR too

rene-ye: I'm ok with not checking, it doesn't have to be consistent persay.

lilgreenbird: use ds.getColumnEncryptionSetting() as per discussion

PR #1379, opened 2020-07-02
Fixed javadoc issue

lilgreenbird:


peterbae:
Actually I just realized that this issue will be fixed in your PR 1373, so either remove this PR or remove it from the other PR, otherwise there will be a conflict.

PR #1386, opened 2020-07-09
Add dependency:purge-local-repository goal to maven

lilgreenbird:
purge and force re download of all dependencies


lilgreenbird:
Closing as this shouldn't be necessary it was an attempt to prevent a suspected glitch in maven repository. Will wait to see if it happens again.

PR #1387, opened 2020-07-10
Cleanup before running tests

lilgreenbird:
- clean up same name tables left over form previous runs if exist
- purge dependencies from maven to force download


src/test/java/com/microsoft/sqlserver/jdbc/databasemetadata/DatabaseMetaDataForeignKeyTest.java
```diff
@@ -53,16 +53,12 @@
 
     @BeforeAll
     public static void setupVariation() throws SQLException {
+        cleanup();
```
peterbae, line 56: I don't think this would do anything since the table names are randomly generated each test run. Why do we need this change?

lilgreenbird: this is so cleanup can be called in @AfterAll.  Issue it that these things get left behind when the test dies or is cancelled so you have to manually clean the up before test can be run again

lilgreenbird: I found the the generated names seem to be the same the old tables from previous runs were hanging around preventing subsequent runs and required manual clean up.  If they were different every time then that wouldn't have happened. This cleans them up and if they're different then it wouldn't do anything so doesn't hurt to try to clean up

PR #1388, opened 2020-07-15
Change | Enclave caching key

rene-ye:
Added database name to caching key. The new key format is `serverName` + `databaseName` + `attestationUrl`.

Removed an unused package privatate SQLServerConnection getter `getScatalog()` which returned the same value as the public getter `getCatalog`.

Changed cache key lookup to occur before calling sp_de because doing so after would cause a new connection to call sp_de with 3 parameters which would create another enclave session.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerAASEnclaveProvider.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerAASEnclaveProvider.java
@@ -46,15 +46,15 @@ public class SQLServerAASEnclaveProvider implements ISQLServerEnclaveProvider {
 
     private AASAttestationParameters aasParams = null;
     private AASAttestationResponse hgsResponse = null;
-    private String attestationURL = null;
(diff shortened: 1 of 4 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerAASEnclaveProvider.java
```diff
@@ -65,21 +65,24 @@ public void getAttestationParameters(String url) throws SQLServerException {
     public ArrayList<byte[]> createEnclaveSession(SQLServerConnection connection, String userSql,
             String preparedTypeDefinitions, Parameter[] params,
             ArrayList<String> parameterNames) throws SQLServerException {
+        // Check if the session exists in our cache
```
ulvii, line 68: Can you explain why this has been moved outside of the if block?

ulvii: nvm, looks like its in the description.

PR #1390, opened 2020-07-20
Check certificate expiry explicitly

rene-ye:
Changes for security items.


src/main/java/com/microsoft/sqlserver/jdbc/KeyStoreProviderCommon.java
```diff
@@ -30,8 +32,13 @@
...
     Key privateKey;
 
-    CertificateDetails(X509Certificate certificate, Key privateKey) {
+    CertificateDetails(X509Certificate certificate, Key privateKey) throws SQLServerException {
         this.certificate = certificate;
+        try {
+            certificate.checkValidity();
+        } catch (CertificateExpiredException | CertificateNotYetValidException e) {
+            SQLServerException.makeFromDriverError(null, this, e.getLocalizedMessage(), "0", false);
+        }
```
lilgreenbird, line 41: state should be ""?

PR #1395, opened 2020-07-28
Append extra error info to DLL exception message

lilgreenbird:


peterbae:
Does this need to be merged now? This will change the resources file and I'll have to modify each language file.

lilgreenbird:
will defer till post release

Issue #1391, opened 2020-07-23
SQLServerBulkCopy expects ISQLServerBulkData implementation to return sorted set


PavelPenkov:
## Driver version
8.2.2.jre11

## SQL Server version
Doesn't matter, it fails before making a query

## Client Operating System
Doesn't matter

## JAVA/JVM version
Amazon Corretto 11, but doesn't matter

## Table schema
Doesn't matter

## Problem description

1. Expected behaviour: Bulk copy to succeed.
2. Actual behaviour: Bulk copy fails before even sending anything to a server.
3. Error message/stack trace:
```
Exception in thread "main" com.microsoft.sqlserver.jdbc.SQLServerException: Column 5 is invalid.  Please check your column mappings.
	at com.microsoft.sqlserver.jdbc.SQLServerBulkCopy.validateColumnMappings(SQLServerBulkCopy.java:1814)
	at com.microsoft.sqlserver.jdbc.SQLServerBulkCopy.writeToServer(SQLServerBulkCopy.java:1617)
	at com.microsoft.sqlserver.jdbc.SQLServerBulkCopy.writeToServer(SQLServerBulkCopy.java:619)
	at example.Hello$.delayedEndpoint$example$Hello$1(Hello.scala:102)
	at example.Hello$delayedInit$body.apply(Hello.scala:66)
	at scala.Function0.apply$mcV$sp(Function0.scala:39)
	at scala.Function0.apply$mcV$sp$(Function0.scala:39)
	at scala.runtime.AbstractFunction0.apply$mcV$sp(AbstractFunction0.scala:17)
	at scala.App.$anonfun$main$1(App.scala:73)
	at scala.App.$anonfun$main$1$adapted(App.scala:73)
	at scala.collection.IterableOnceOps.foreach(IterableOnce.scala:553)
	at scala.collection.IterableOnceOps.foreach$(IterableOnce.scala:551)
	at scala.collection.AbstractIterable.foreach(Iterable.scala:920)
	at scala.App.main(App.scala:73)
	at scala.App.main$(App.scala:71)
	at example.Hello$.main(Hello.scala:66)
	at example.Hello.main(Hello.scala)

```
[...]

PR #1406, opened 2020-08-11
Fix | Sort column ordinal mapping during bulk copy

peterbae:
Fixes issue #1391.

Straightforward fix, sorts the 'columnOrdinals' object if the returned set happens to be unordered.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
@@ -1840,17 +1840,33 @@ public class SQLServerBulkCopy implements java.lang.AutoCloseable, java.io.Seria
                 // if no mapping is provided for csv file and metadata is missing for some columns throw error
                 if (null != serverBulkData) {
                     Set<Integer> columnOrdinals = serverBulkData.getColumnOrdinals();
-                    List<Integer> sortedList = new ArrayList<>(columnOrdinals);
-                    Collections.sort(sortedList);
-                    Iterator<Integer> columnsIterator = sortedList.iterator();
+                    Iterator<Integer> columnsIterator = columnOrdinals.iterator();
                     int j = 1;
+                    outerWhileLoop:
                     while (columnsIterator.hasNext()) {
                         int currentOrdinal = columnsIterator.next();
                         if (j != currentOrdinal) {
-                            MessageFormat form = new MessageFormat(SQLServerException.getErrString("R_invalidColumn"));
-                            Object[] msgArgs = {currentOrdinal};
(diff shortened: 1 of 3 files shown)


rene-ye (dismissed):

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
```diff
@@ -1839,7 +1840,9 @@ private void validateColumnMappings() throws SQLServerException {
                 // if no mapping is provided for csv file and metadata is missing for some columns throw error
                 if (null != serverBulkData) {
                     Set<Integer> columnOrdinals = serverBulkData.getColumnOrdinals();
-                    Iterator<Integer> columnsIterator = columnOrdinals.iterator();
+                    List<Integer> sortedList = new ArrayList<>(columnOrdinals);
+                    Collections.sort(sortedList);
+                    Iterator<Integer> columnsIterator = sortedList.iterator();
```
rene-ye, line 1845: If the `columnMetadata` Map objects implemented in the SQLServerBulkRecord classes were `TreeMap` objects instead of `Hashmap`, you wouldn't need to sort the keyset you retrieve here. However I'm not sure if the cost of lookups/inserts going from O(1) to O(logN) would be worth that trade off.

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/ISQLServerBulkRecordIssuesTest.java
```diff
@@ -253,6 +256,26 @@ public void afterEachTests() throws SQLException {
     }
 }
 
+
```
ulvii, line 259: Why is this in a different file?

rene-ye (approved):

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/ISQLServerBulkRecordIssuesTest.java
peterbae: This interface already has another (very similar) bulk data class used for test, so I think it makes sense to keep this code here.

ulvii: That class is only used inside the same test file. I would keep  the required classes for a test inside its own file.

peterbae: Ok, I've updated the test.

ulvii (approved):

rene-ye (approved):

lilgreenbird (approved):

Issue #1414, opened 2020-08-26
Bulkcopy fails when inserting non-unicode multibyte String.


kmzxy:
## Driver version
mssql-jdbc-8.4.0.jre11.jar

## SQL Server version
Microsoft SQL Server 2017 (RTM-GDR) (KB4505224) - 14.0.2027.2 (X64)   Jun 15 2019 00:26:19   Copyright (C) 2017 Microsoft Corporation  Developer Edition (64-bit) on Windows Server 2016 Standard 10.0 <X64> (Build 14393: ) (Hypervisor) 

## Client Operating System
Windows 10

## JAVA/JVM version
RedHat OpenJDK
openjdk version "11.0.8" 2020-07-14 LTS
OpenJDK Runtime Environment 18.9 (build 11.0.8+10-LTS)
OpenJDK 64-Bit Server VM 18.9 (build 11.0.8+10-LTS, mixed mode)

## Table schema
```
CREATE TABLE [dbo].[TEST](
	[name] [varchar](10) COLLATE Japanese_CS_AS_KS_WS NOT NULL
)
```

## Problem description
1. Expected behaviour:
  One row has been bulk-inserted.
2. Actual behaviour:
  SQL Exception occured.

3. Error message/stack trace:
  Exception in thread "main" com.microsoft.sqlserver.jdbc.SQLServerException: (... (Japanese message))
4. Any other details that can be helpful:
  In SQLBulkCopy.java, Line Around 2292  (commit-id 4e3ab6690557c000bc16042771020ef530b54bf2):

  > `2292: tdsWriter.writeShort((short) (colValueStr.length()));`

  I think the above code should be something like this:
    `                                         tdsWriter.writeShort((short) (colValueStr.getBytes(destColumnMetadata.get(destColOrdinal).collation.getCharset()).length));
`
  when destCollation is not null.
  Because colValueStr.length() is different from the transferred byte size by tdsWriter.writeBytes();
  
[Bulk.java.txt](https://github.com/microsoft/mssql-jdbc/files/5130296/Bulk.java.txt)

peterbae:
Thanks for reporting this issue @kmzxy, your fix seems good and I'll make a PR as soon as testing is concluded.

ulvii:
#1421 merged.

PR #1421, opened 2020-09-02
Fix | Fix bulkcopy failing when inserting non-unicode multibyte String

peterbae:
Fixes issue #1414.

We cannot assume that length = number of bytes when destination collation is not equal to null.

### DIFF
--- a/src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyRowSetTest.java
+++ b/src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyRowSetTest.java
@@ -36,6 +36,8 @@ public class BulkCopyRowSetTest extends AbstractTest {
 
     private static String tableName = AbstractSQLGenerator
             .escapeIdentifier(RandomUtil.getIdentifier("BulkCopyFloatTest"));
+    private static String tableName2 = AbstractSQLGenerator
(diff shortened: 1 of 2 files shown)


rene-ye (dismissed):

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyRowSetTest.java
```diff
@@ -71,18 +73,44 @@ public void testBulkCopyFloatRowSet() throws SQLException {
...
+            rsmd.setTableName(1, tableName2);
+
+            crs.setMetaData(rsmd);
+            crs.moveToInsertRow();
+            crs.updateString("c1", "ああ");
+            crs.insertRow();
+            crs.moveToCurrentRow();
+
+            bulkCopy.setDestinationTableName(tableName2);
+            bulkCopy.writeToServer(crs);
```
ulvii, line 95: Need to read from the table here to make sure the data is inserted correctly.

peterbae: Thanks, updated.

ulvii (approved):

rene-ye (approved):

lilgreenbird (approved):

PR #1423, opened 2020-09-02
Fix | Intermittent deadlock error by querying only the procedure created

rene-ye:
sp_sproc_columns_100 returns all stored procedures that match the criteria. There is a chance that stored procedures which match the criteria are being altered/updated during the test. If we only query for the stored procedure we created in the test by restricting the SP to use the function name, this won't happen as the SP names are unique.


peterbae (approved):

peterbae:
Good fix and analysis.

lilgreenbird (approved):

ulvii (approved):

PR #1430, opened 2020-09-17
Retain error information in SQLServerPreparedStatement.getMetaData exceptions

danburkert:
Fixes a bug in SQLServerPreparedStatement.buildExecuteMetaData which
caused the SqlState as returned from the server error to be discarded in
thrown exceptions. The SqlState information is useful for application
level error handling.

No tests are provided, as I can't get the test harness to run in my
local environment. However, here are steps to reproduce the 'bad'
exception with no included SqlState:

```java
Connection conn = null;     // Fill in with SqlServer connection.
conn.prepareStatement("s");
conn.getMetaData();         // Throws exception without SqlState.
```

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
@@ -1050,7 +1050,7 @@ public class SQLServerPreparedStatement extends SQLServerStatement implements IS
     }
 
     @Override
-    public final java.sql.ResultSetMetaData getMetaData() throws SQLServerException {
+    public final java.sql.ResultSetMetaData getMetaData() throws SQLServerException, SQLTimeoutException {
         loggerExternal.entering(getClassNameLogging(), "getMetaData");
         checkClosed();
         boolean rsclosed = false;
@@ -1063,7 +1063,7 @@ public class SQLServerPreparedStatement extends SQLServerStatement implements IS
             rsclosed = true;
         }
         if (resultSet == null || rsclosed) {
-            SQLServerResultSet emptyResultSet = (SQLServerResultSet) buildExecuteMetaData();
+            SQLServerResultSet emptyResultSet = buildExecuteMetaData();
             if (null != emptyResultSet)
                 rsmd = emptyResultSet.getMetaData();
         } else if (resultSet != null) {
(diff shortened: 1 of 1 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -1050,7 +1050,7 @@ else if (needsPrepare)
     }
 
     @Override
-    public final java.sql.ResultSetMetaData getMetaData() throws SQLServerException {
+    public final java.sql.ResultSetMetaData getMetaData() throws SQLServerException, SQLTimeoutException {
```
ulvii, line 1053: Is adding SQLTimeoutException necessary here?

danburkert: Yes.  `buildExecuteMetaData()` previously caught and wrapped the SQLTimeoutException in another exception with less info.  That's been changed so that it's no longer caught, and it gets thrown directly.  The JDBC API allows for throwing this exception type, so it shouldn't be rethrown with less info.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -1080,21 +1080,18 @@ else if (needsPrepare)
...
         String fmtSQL = userSQL;
 
-        ResultSet emptyResultSet = null;
+        SQLServerResultSet emptyResultSet = null;
         try {
             fmtSQL = replaceMarkerWithNull(fmtSQL);
             internalStmt = (SQLServerStatement) connection.createStatement();
             emptyResultSet = internalStmt.executeQueryInternal("set fmtonly on " + fmtSQL + "\nset fmtonly off");
-        } catch (SQLException sqle) {
+        } catch (SQLServerException sqle) {
```
peterbae, line 1091: We're only catching a subset of exceptions possibly thrown from this class (SQLException -> SQLServerException) now, but we should also catch the other exceptions that could be thrown from here.

danburkert: The point of this PR is to _not_ catch exceptions.  All exceptions are now propagated, except the specific empty result set exception which is handled.  Previously all exceptions were caught, then a useless exception containing no details was thrown through `SQLServerException.makeFromDriverError`.  This PR does not change the circumstances under which exceptions are thrown, it simply ensures that the original exception is propagated with the useful details, instead of throwing a different exception with no information.  This is critical so that applications can handle errors appropriately.

peterbae: I see, thanks for the explanation.

peterbae (approved):

ulvii:
Hi @danburkert ,
Would you sign the [Contributor License Agreement ](https://cla.opensource.microsoft.com/microsoft/mssql-jdbc?pullRequest=1430)?

danburkert:
> Hi @danburkert ,
> Would you sign the [Contributor License Agreement ](https://cla.opensource.microsoft.com/microsoft/mssql-jdbc?pullRequest=1430)?

done!

rene-ye (approved):

ulvii (approved):

lilgreenbird (approved):

ulvii:
Thanks for the contribution @danburkert .

PR #1431, opened 2020-09-17
Add maxResultBuffer property

sawkoj:
Implementation of new property - maxResultBuffer. Enables max bytes read during reading result set.  
Enable declare in two styles:
• as size of bytes (i.e. 100, 150M, 300K, 400G);
• as percent of max heap memory (i.e. 10p, 15pct, 20percent).  

It also validates if defined size isn't bigger than 90% of max heap memory. If so, it limits max bytes read to 90% of memory heap. If maxResultBuffer property isn't provided, driver work isn't changed.


rene-ye:
Hi @sawkoj, can you give some points regarding why this feature is necessary? The JDBC Statement APIs provide a similar "memory" control option in `setFetchSize` and `setMaxFieldSize`. Although this doesn't directly allow you to control memory, the intentions are the same. Can you give a use case where your proposed feature would be significantly better than the existing JDBC apis?

peterbae:
Thanks for the contribution @sawkoj, sorry for the delay in response. I see that this change allows the user to set a limit to how much data can be read from a resultset, but in what situation will this change be useful? So far, users haven't asked for such feature because they can choose to stream the data instead of storing the entire resultset in memory, which I think is more flexible than the driver throwing an error if the amount of data read goes over the limit specified in this buffer.

sawkoj:
Hi @rene-ye @peterbae , a purpose of the added property is to protect user from Out Of Memory Exception. Properties for “memory” control, which you have mentioned, won’t be enough in some cases. The most common case, I can think of, is when we have 2 queries which produce different result sets. Let’s call them A and B. A is large, contains thousands of rows, B’s size doesn’t matter. Upon executing A’s query, only first packet of A is kept in memory. We don’t iterate or do anything with A. Next, we call B query, to get B result set. In this moment, all remaining packets of A’s result set are being read to memory with first packet of B. Here OOM can occur and setting `setFetchSize` or `setMaxFieldSize` can prevent that.  

Setting these limits can be very inconvenient though. Setting `fetchSize` changes amount of packets that need to be send to read whole resultSet. It could be a way of controlling of memory usage, but this approach is very error prone (you have to calculate the number of  bytes for single row from a given query). If any column in query/table changes, you have to recompute number of rows to fetch if you want to handle memory using `setFetchSize`. The similar issue applies for setting `maxFieldSize`. You depend memory control on size of a single column, on change of query/table, you have to take that in consideration in code. 

By using maxResultBuffer you just put amount of memory that you want to be used and don’t worry about having OOM and stopping whole application.

ebrandsberg:
Adding to this.  This work is being done as part of a larger project at Heimdall Data, which provides a proxy for a variety of database types.  We are working to bring consistent controls over memory usage across multiple driver types.  This feature has already been included in the Postgres JDBC driver (https://github.com/pgjdbc/pgjdbc/pull/1657), as well as a fix submitted to the Oracle MySQL jdbc driver (https://bugs.mysql.com/bug.php?id=100415), so our next target was the SQL Server driver.  To completely understand the need, don't think about one connection, but when you have thousands of connections active, and you need to be able to ensure there is enough memory for all of them.  @sawkoj explained one very specific case, but there are other cases where (I believe) adaptive fetching doesn't work, in particular if retrieving large result sets from tables with columnstore indexes.  Cursors are not allowed in this case, and can trigger oom problems.

rene-ye:
Hi @sawkoj, I've been testing the PR and there is currently some undesirable behavior with the new feature. There are currently some scenarios where if the maxResultBuffer property wasn't specifically set, the property will default to -1, and when attempting to call `getMultiplier` while the property is -1, the driver will throw an error. New properties should be introduced with a default value where it won't impact the rest of the driver when either not specified or off. The error message also seems to need to work, currently I'm getting:

`com.microsoft.sqlserver.jdbc.SQLServerException: Invalid syntax: [Ljava.lang.Object;@628190ec in maxResultBuffer parameter at com.microsoft.sqlserver.jdbc.MaxResultBufferParser.throwNewInvalidMaxResultBufferParameterException(MaxResultBufferParser.java:120)`

The above error is easily reproducible if you run our unit tests provided in the repo, if you have any trouble setting it up, we can help with that. Some of the tests require special setup, and it's ok if you can't run them on your end, but the CIs should pass.

src/main/java/com/microsoft/sqlserver/jdbc/MaxResultBufferParser.java
```diff
@@ -0,0 +1,123 @@
...
+     * @return 'maxResultBuffer' property as number of bytes
+     * @throws SQLServerException
+     *         Is Thrown when maxResultProperty's syntax is wrong
+     */
+    public static long validateMaxResultBuffer(String input) throws SQLServerException {
+        String numberString;
+        long number = -1;
+
+        // check for null values and empty String "", if so return -1
+        if (StringUtils.isEmpty(input)) {
```
rene-ye, line 40: Temporary solution would be to change this line to

```java
if (StringUtils.isEmpty(input) || "-1".equalsIgnoreCase(input)) {
```

But we should ask whether it makes sense to allow users to even input something like -1 at all.

src/test/java/com/microsoft/sqlserver/testframework/AbstractTest.java
```diff
@@ -355,6 +355,8 @@ protected static ISQLServerDataSource updateDataSource(String connectionString,
                             break;
                         case Constants.SEND_TEMPORAL_DATATYPES_AS_STRING_FOR_BULK_COPY:
                             ds.setSendTemporalDataTypesAsStringForBulkCopy(Boolean.parseBoolean(value));
```
rene-ye, line 357: This needs a break at the end of the case statement, or it'll overflow into the next one.

src/test/java/com/microsoft/sqlserver/testframework/AbstractTest.java
```diff
@@ -355,6 +355,8 @@ protected static ISQLServerDataSource updateDataSource(String connectionString,
...
+                        case Constants.MAX_RESULT_BUFFER:
+                            ds.setMaxResultBuffer(value);
```
rene-ye, line 359: Add at break to the end of the this to future proof.

sawkoj:
Hi @rene-ye, thank you for your feedback. According to your suggestions, I changed the default value of maxResultBuffer property to "-1", so from now it should be parsed correctly whether value for the property is specified or not. Also, I fixed the error message (I was unnecessarily packaging it as Array). I refactored the parsing logic, so from now it's gonna throw an error upon providing negative values, tests for negative values had been added as well. Of course I didn't forget to add missing break statements in AbstractTest.java.

src/main/java/com/microsoft/sqlserver/jdbc/ICounter.java
```diff
@@ -0,0 +1,27 @@
...
+ * Microsoft JDBC Driver for SQL Server Copyright(c) Microsoft Corporation All rights reserved. This program is made
+ * available under the terms of the MIT License. See the LICENSE file in the project root for more information.
+ */
+
+package com.microsoft.sqlserver.jdbc;
+
+/**
+ * Interface for MaxResultBufferCounter
+ */
+public interface ICounter {
```
ulvii, line 11: Please make this package-private.

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -7344,6 +7364,23 @@ final boolean readingResponse() {
 
     protected ArrayList<byte[]> enclaveCEKs;
 
+    //Counter reference, so maxResultBuffer property can by acknowledged
+    private ICounter counter;
+
+    public ICounter getCounter() {
```
ulvii, line 7370: Package-private please.

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -7344,6 +7364,23 @@ final boolean readingResponse() {
...
+        return counter;
+    }
+
+    public void createCounter(ICounter previousCounter, Properties activeConnectionProperties) {
```
ulvii, line 7374: package-private

src/test/java/com/microsoft/sqlserver/jdbc/MaxResultBufferTest.java
```diff
@@ -0,0 +1,280 @@
...
+    }
+
+    private static void setResponseBufferingAdaptive(boolean adaptive) {
+        String value = adaptive ? "adaptive" : "full";
+        connectionString = TestUtils.addOrOverrideProperty(connectionString, "responseBuffering", value);
+        AbstractTest.updateDataSource(connectionString, ds);
+    }
+
+    private static void setMaxResultBuffer(String maxResultBuffer) {
+        connectionString = TestUtils.addOrOverrideProperty(connectionString, "maxResultBuffer", maxResultBuffer);
```
ulvii, line 277: Because `connectionString` is defined in `AbstractTest` and is static, this line will change the connection string for all the tests. Please use a local connection string .

src/test/java/com/microsoft/sqlserver/jdbc/MaxResultBufferTest.java
```diff
@@ -0,0 +1,280 @@
...
+        int leftLimit = 33;
+        int rightLimit = 126;
+        Random random = new Random();
+        return random.ints(leftLimit, rightLimit).limit(precision)
+                .collect(StringBuilder::new, StringBuilder::appendCodePoint, StringBuilder::append).toString();
+    }
+
+    private static void setResponseBufferingAdaptive(boolean adaptive) {
+        String value = adaptive ? "adaptive" : "full";
+        connectionString = TestUtils.addOrOverrideProperty(connectionString, "responseBuffering", value);
```
ulvii, line 272: Because connectionString is defined in AbstractTest and is static, this line will change the connection string for all the tests. Please use a local connection string .

src/test/java/com/microsoft/sqlserver/jdbc/MaxResultBufferTest.java
```diff
@@ -0,0 +1,280 @@
+/*
```
ulvii, line 1: Please add more test for different possible values of the connection property, both valid and invalid cases.

sawkoj:
Hi @ulvii, thank you for the feedback. According to your suggestions, I changed access modifiers in ICounter interface and IOBuffer class methods getCounter(), createCounter() as well. In MaxResultBufferTest I added an another private static field, which is used as local connection String. The original connection string is unaffected. Also I added more test cases to MaxResultBufferParserTest (valid and invalid ones). MaxResultBufferTest class was refactored too, I parameterized existing tests and I added different MaxResultBuffer property values to illustrate the change of behavior.

peterbae (dismissed):

peterbae:
The PR looks good to me. Once the other team members have reviewed & approved, we will re-run the CI tests and merge the PR.

src/main/java/com/microsoft/sqlserver/jdbc/MaxResultBufferCounter.java
```diff
@@ -0,0 +1,50 @@
...
+        }
+    }
+
+    public void resetCounter() {
+        counter = 0;
+    }
+
+    private void checkForMaxResultBufferOverflow(long number) throws SQLServerException {
+        if (number > maxResultBuffer) {
+            logger.log(Level.WARNING, "MaxResultBuffer exceeded: {0}. Property was set to {1}.",
```
ulvii, line 40: Please change to :
```
if (logger.isLoggable(Level.SEVERE)) {
    logger.log(Level.SEVERE, "MaxResultBuffer exceeded: {0}. Property was set to {1}.",
            new Object[] {number, maxResultBuffer});
}
```

src/main/java/com/microsoft/sqlserver/jdbc/MaxResultBufferParser.java
```diff
@@ -0,0 +1,141 @@
...
+        if (StringUtils.isEmpty(input) || input.equals("-1")) {
+            return number;
+        }
+
+        // check if input is number
+        if (!StringUtils.isEmpty(input) && input.matches("-?\\d+(\\.\\d+)?")) {
+            try {
+                number = Long.parseLong(input);
+            } catch (NumberFormatException e) {
+                logger.log(Level.INFO, ERROR_MESSAGE, new Object[] {input});
```
ulvii, line 49:
```
if (logger.isLoggable(Level.SEVERE)) {
    logger.log(Level.SEVERE, ERROR_MESSAGE, new Object[] {input});
}
```

src/main/java/com/microsoft/sqlserver/jdbc/MaxResultBufferParser.java
```diff
@@ -0,0 +1,141 @@
...
+            return adjustMemory(number, 1);
+        } else {
+            // check PERCENT_PHRASES
+            for (String percentPhrase : PERCENT_PHRASES) {
+                if (input.endsWith(percentPhrase)) {
+                    numberString = input.substring(0, input.length() - percentPhrase.length());
+                    try {
+                        number = Long.parseLong(numberString);
+                    } catch (NumberFormatException e) {
+                        logger.log(Level.INFO, ERROR_MESSAGE, new Object[] {input});
```
ulvii, line 61:
```
if (logger.isLoggable(Level.SEVERE)) {
    logger.log(Level.SEVERE, ERROR_MESSAGE, new Object[] {input});
}
```

src/main/java/com/microsoft/sqlserver/jdbc/MaxResultBufferParser.java
```diff
@@ -0,0 +1,141 @@
...
+            }
+
+            // check if prefix was supplied
+            long multiplier = getMultiplier(input);
+            numberString = input.substring(0, input.length() - 1);
+
+            try {
+                number = Long.parseLong(numberString);
+            } catch (NumberFormatException e) {
+                logger.log(Level.INFO, ERROR_MESSAGE, new Object[] {input});
```
ulvii, line 75:
```
if (logger.isLoggable(Level.SEVERE)) {
    logger.log(Level.SEVERE, ERROR_MESSAGE, new Object[] {input});
}
```

src/main/java/com/microsoft/sqlserver/jdbc/MaxResultBufferParser.java
```diff
@@ -0,0 +1,141 @@
...
+                multiplier = 1_000_000L;
+                break;
+            case 'G':
+                multiplier = 1_000_000_000L;
+                break;
+            case 'T':
+                multiplier = 1_000_000_000_000L;
+                break;
+            default:
+                logger.log(Level.INFO, ERROR_MESSAGE, new Object[] {input});
```
ulvii, line 109:
```
if (logger.isLoggable(Level.SEVERE)) {
    logger.log(Level.SEVERE, ERROR_MESSAGE, new Object[] {input});
}
```

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResource.java
```diff
@@ -652,5 +652,9 @@ static String getResource(String key) {
             {"R_unassignableError", "The class specified by the {0} property must be assignable to {1}."},
             {"R_InvalidCSVQuotes",
                     "Failed to parse the CSV file, verify that the fields are correctly enclosed in double quotes."},
-            {"R_TokenRequireUrl", "Token credentials require a URL using the HTTPS protocol scheme."},};
-};
+            {"R_TokenRequireUrl", "Token credentials require a URL using the HTTPS protocol scheme."},
+            {"R_maxResultBufferPropertyDescription",
```
ulvii, line 656: `R_maxResultBufferPropertyDescription` should describe what `maxResultBuffer` connection property does, please check  `R_useFmtOnlyPropertyDescription` for reference.  I'd suggest to create another String resource for `MaxResultBuffer property exceeded: {0}. MaxResultBuffer was set to: {1}.` and use `R_maxResultBufferPropertyDescription` for the actual description of the connection property.

sawkoj:
Hi @peterbae and @ulvii, thanks for feedback. Loggers were adjusted in MaxResultBufferParser and in MaxResultBufferCounter. I created additional resource: `R_maxResultBufferPropertyExceeded` when MaxResultBuffer is exceeded. Description of `R_maxResultBufferPropertyDescription` is updated too.

ulvii (approved):

lilgreenbird (approved):

rene-ye (approved):

peterbae (approved):

PR #1464, opened 2020-11-23
Added support for ActiveDirectoryInteractive authentication

lilgreenbird:
Add new type to authentication property. ie `authentication=ActiveDirectoryInteractive". Driver will attempt to silently authenticate if access token has been cached or will be presented with a browser popup to authenticate interactively.

Tokens will persist between applications unless it is cleared via new API:
public static void clearUserTokenCache() 

Note if MFA is enabled, authentication may require longer timeout which can be set via the setLoginTimeout API.


peterbae (approved):

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -3907,6 +3910,7 @@ int writeFedAuthFeatureRequest(boolean write, /* if false just calculates the le
                             workflow = TDS.ADALWORKFLOW_ACTIVEDIRECTORYINTEGRATED;
                             break;
                         case ActiveDirectoryMSI:
+                        case ActiveDirectoryInteractive:
```
ulvii, line 3913: Please create a separate enum for `ActiveDirectoryInteractive` to avoid confusion. See  `ADALWORKFLOW_ACTIVEDIRECTORYSERVICEPRINCIPAL` and `ADALWORKFLOW_ACTIVEDIRECTORYPASSWORD` for reference.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerMSAL4JUtils.java
```diff
@@ -119,6 +127,76 @@ static SqlFedAuthToken getSqlFedAuthTokenIntegrated(SqlFedAuthInfo fedAuthInfo,
...
+    private static IAccount getAccountByUsername(Set<IAccount> accounts, String username) {
+        if (!accounts.isEmpty()) {
+            for (IAccount account : accounts) {
+                if (account.username().equals(username)) {
+                    return account;
+                }
+            }
+        }
+        return null;
+    }
```
rene-ye, line 198: Are null checks not required for accounts, entries, or usernames? The following checks if `accounts` is null and not empty, before attempting to find the first element that matches the username. It also checks that the `IAccount` it just iterated over isn't null, the `username` within the `IAccount` isn't null, and then does the comparison. If it finds an element that matches the criteria, it'll assign that value to the return variable, if not, it'll assign `null` to the return variable.

```java
// Helper function to return account containing user name from set of accounts, or null if no match
private static IAccount getAccountByUsername(Set<IAccount> accounts, String username) {
    IAccount result = null;
    if (null != accounts && !accounts.isEmpty())
        result = accounts.parallelStream()
                .filter(a -> (null != a && null != a.username() && a.username().equals(username))).findFirst()
                .orElse(null);
    return result;
}
```

src/main/java/com/microsoft/sqlserver/jdbc/PersistentTokenCacheAccessAspect.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    /**
+     * Clears token cache. This will clear all account info so interactive login will be required on the next request to
+     * acquire an access token.
+     */
+    public static void clearUserTokenCache() {
+        if (null != cache && !cache.isEmpty()) {
+            cache = null;
+        }
```
rene-ye, line 60: This doesn't make sense? The class is a singleton and this is a static method. Shouldn't this be

```java
if (null != instance.cache && !instance.cache.isEmpty()) {
    instance.cache = null;
}
```

src/main/java/com/microsoft/sqlserver/jdbc/PersistentTokenCacheAccessAspect.java
```diff
@@ -0,0 +1,62 @@
...
+        if (null == instance) {
+            instance = new PersistentTokenCacheAccessAspect();
+        }
+        return instance;
+    }
+
+    /**
+     * Token cache in JSON format
+     */
+    private static String cache = null;
```
rene-ye, line 36: This object should not be static in the context of a Singleton class.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerMSAL4JUtils.java
```diff
@@ -16,37 +18,43 @@
...
 class SQLServerMSAL4JUtils {
 
+    static final String REDIRECTURI = "http://localhost";
+
     static final private java.util.logging.Logger logger = java.util.logging.Logger
             .getLogger("com.microsoft.sqlserver.jdbc.SQLServerMSAL4JUtils");
 
     static SqlFedAuthToken getSqlFedAuthToken(SqlFedAuthInfo fedAuthInfo, String user, String password,
             String authenticationString) throws SQLServerException {
         ExecutorService executorService = Executors.newFixedThreadPool(1);
```
rene-ye, line 49: This should be `Executors.newSingleThreadExecutor()`.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerMSAL4JUtils.java
```diff
@@ -119,6 +127,76 @@ static SqlFedAuthToken getSqlFedAuthTokenIntegrated(SqlFedAuthInfo fedAuthInfo,
         }
     }
 
+    static SqlFedAuthToken getSqlFedAuthTokenInteractive(SqlFedAuthInfo fedAuthInfo, String user,
+            String authenticationString) throws SQLServerException {
+        ExecutorService executorService = Executors.newFixedThreadPool(1);
```
rene-ye, line 132: Please use `Executors.newSingleThreadExecutor()`.

src/main/java/com/microsoft/sqlserver/jdbc/PersistentTokenCacheAccessAspect.java
```diff
@@ -0,0 +1,62 @@
...
+ * Access aspect for accessing the token cache.
+ * 
+ * MSAL token cache does not persist beyond lifetime of the application. This class implements the
+ * ITokenCacheAccessAspect interface to persist the token cache between application instances so subsequent
+ * authentications can use silent authentication if the user account is in the token cache.
+ * 
+ * @see <a href="https://aka.ms/msal4j-token-cache">https://aka.ms/msal4j-token-cache</a>
+ */
+class PersistentTokenCacheAccessAspect implements ITokenCacheAccessAspect {
+    private static PersistentTokenCacheAccessAspect instance = null;
```
rene-ye, line 22: if this is changed to:
`private static PersistentTokenCacheAccessAspect instance = new PersistentTokenCacheAccessAspect();`
we can avoid a few null checks.

src/main/java/com/microsoft/sqlserver/jdbc/PersistentTokenCacheAccessAspect.java
```diff
@@ -0,0 +1,62 @@
...
+    public synchronized void beforeCacheAccess(ITokenCacheAccessContext iTokenCacheAccessContext) {
+        if (null != cache && null != iTokenCacheAccessContext && null != iTokenCacheAccessContext.tokenCache()) {
+            iTokenCacheAccessContext.tokenCache().deserialize(cache);
+        }
+    }
+
+    @Override
+    public synchronized void afterCacheAccess(ITokenCacheAccessContext iTokenCacheAccessContext) {
+        if (null != iTokenCacheAccessContext && null != iTokenCacheAccessContext.tokenCache()
+                && iTokenCacheAccessContext.hasCacheChanged()) {
```
rene-ye, line 48: I would reorder this to 
```java
if (null != iTokenCacheAccessContext && iTokenCacheAccessContext.hasCacheChanged()
        && null != iTokenCacheAccessContext.tokenCache())
```
since we want to know if the cache has changed first, and then if it has, then we need to do a null check on the token cache so we can call `.serialize()`.

src/test/java/com/microsoft/sqlserver/jdbc/SQLServerConnectionTest.java
```diff
@@ -861,4 +864,18 @@ public void testConnectionPoolProxyWithLobs() throws SQLException, IOException {
...
+    @Test
+    public void testPersistentTokenCacheAccessAspect() throws SQLException {
+        TokenCacheAccessContext tokenCacheAccessContext = TokenCacheAccessContext.builder().clientId(null)
+                .tokenCache(new TokenCache()).account(null).hasCacheChanged(true).build();
+
+        PersistentTokenCacheAccessAspect persistentTokenAspect = PersistentTokenCacheAccessAspect.getInstance();
+        persistentTokenAspect.afterCacheAccess(tokenCacheAccessContext);
+        persistentTokenAspect.beforeCacheAccess(tokenCacheAccessContext);
+        PersistentTokenCacheAccessAspect.clearUserTokenCache();
+    }
```
rene-ye, line 880: This test currently exists for raw code coverage, and if the `if` blocks return false, it doesn't even really give codcov. Is there a meaningful way for us to verify whether the cache has been serialized/deserialized after calling the before/after CacheAccess methods?

src/main/java/com/microsoft/sqlserver/jdbc/PersistentTokenCacheAccessAspect.java
```diff
@@ -0,0 +1,62 @@
...
+    /**
+     * Clears token cache. This will clear all account info so interactive login will be required on the next request to
+     * acquire an access token.
+     */
+    public static void clearUserTokenCache() {
+        if (null != cache && !cache.isEmpty()) {
+            cache = null;
+        }
+    }
+}
```
rene-ye, line 62: Please make sure newlines exist at the end of every file.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerMSAL4JUtils.java
lilgreenbird: accounts is the set of accounts in cache, which could be and would be null on 1st connection.

src/test/java/com/microsoft/sqlserver/jdbc/SQLServerConnectionTest.java
lilgreenbird: that requires interactive testing, currently can only do timeout tests in junit. Added comment in header

rene-ye (approved):

rene-ye (approved):

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -2052,7 +2054,13 @@ else if (null != (trustStoreFileName = System.getProperty("javax.net.ssl.trustSt
 
     final int read(byte[] data, int offset, int length) throws SQLServerException {
         try {
+            // System.out.println("socket timeout:"+ tcpSocket.getSoTimeout());
```
ulvii, line 2057: Remove these changes

lilgreenbird: oops

ulvii (approved):

rene-ye (approved):

peterbae (approved):

ulvii (approved):

PR #1476, opened 2020-12-06
Fix | Avoid high mem alloc in TDSWriter#writeReader

### LINKED ISSUE #1475: TDSWriter.writeReader high memory allocations
## Driver version
8.4.1

## SQL Server version
15

## Client Operating System
Ubuntu 20

## JAVA/JVM version
11

## Table schema
N/A

## Problem description
There is alloc of char and byte arrays/buffers on every writeReader invocation:
https://github.com/microsoft/mssql-jdbc/blob/8c0d32ba4e3cac18e134699e861f52513a842123/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java#L4099
We use SQLServerBulkCopy to write data from file to SQL Server and it yields few GiB/s of mem allocations. 
1. Expected behaviour:
char/byte buffer is created once per TDSWriter and reused on subsequent writes
2. Actual behaviour:
char and byte arrays/buffers on every writeReader invocation
3. Error message/stack trace:

## Reproduction code
Call SQLServerBulkCopy#writeToServer many times

vyaslav:
This fixes #1475

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
@@ -3151,6 +3151,10 @@ final class TDSWriter {
     private ByteBuffer socketBuffer;
     private ByteBuffer logBuffer;
 
+    // Intermediate arrays
+    private char[] streamCharBuffer;
+    private byte[] streamByteBuffer;
+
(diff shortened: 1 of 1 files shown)


rene-ye (approved):

peterbae:
Thanks for the contribution @vyaslav, we've merged in your PR.

PR #1491, opened 2020-12-22
Connection.getMetaData().getProcedures(String,String,String)) missing 9th column

garydgregory:
Connection.getMetaData().getProcedures(String,String,String)) should return 9 columns but is missing the 9th column SPECIFIC_NAME.

While the underlying server's stored procedure only returns 8 columns, the JDBC API must return the 9th to be compliant with the API even if the column values are null. This bug makes is harder to write JDBC code that is vendor neutral. 

If you agree this is a bug, I can provide a fix.


peterbae:
Hi @garydgregory, that sounds like a good fix. If you could add the changes to this PR, the team will review it.

barryw-mssql:
We have looked at this PR and have found that it does not work with the existing code base. We want to get this  working but at this point we have not been able to dedicate the effort

PR #1498, opened 2021-01-15
Fix | Changes after 9.2.0 static analysis

rene-ye:
Fixes static code analysis issues since last release.


src/main/java/com/microsoft/sqlserver/jdbc/KeyVaultTokenCredential.java
```diff
@@ -29,6 +29,8 @@
  */
 @Immutable
 class KeyVaultTokenCredential implements TokenCredential {
+    private static final String NULL_VALUE = "R_NullValue";
```
lilgreenbird, line 32: if you do this why not define this to be the error string? 
i.e. private static final String NULL_VALUE = SQLServerException.getErrString("R_NullValue");

peterbae (approved):

lilgreenbird (approved):

Issue #1499, opened 2021-01-18
Batch fails when always encrypted is enabled in connection string and clearParameters was called


jezbera:
## Driver version
9.1.1.jre11-preview (8.4.1.jre11 is affected too)

## SQL Server version
Microsoft SQL Server 2016 (SP2) (KB4052908) - 13.0.5026.0 (X64) Mar 18 2018 09:11:49 Copyright (c) Microsoft Corporation Express Edition (64-bit) on Windows 10 Pro 10.0 (Build 18363: ) (Hypervisor)-->

## Client Operating System
Windows/Linux

## JAVA/JVM version
11.0.6

## Table schema
Table will be created by test.
create table TEST2 (KEY1 numeric(19,0) not null, KEY2 numeric(19,0) not null, primary key (KEY1, KEY2))

## Problem description
Prepare database with always encrypted feature. Update connection details in ExecuteBatchTest.java and run tests in attached gradle project.

1. Expected behaviour:
No column is encrypted, so it does not matter, whether columnEncryptionSetting is enabled or not.
1. Actual behaviour:
Only test with connection string containing "columnEncryptionSetting=Enabled" fails.
1. Error message/stack trace:
`Index 2 out of bounds for length 2
java.lang.ArrayIndexOutOfBoundsException: Index 2 out of bounds for length 2
	at com.microsoft.sqlserver.jdbc.SQLServerPreparedStatement.doExecutePreparedStatementBatch(SQLServerPreparedStatement.java:2764)
	at com.microsoft.sqlserver.jdbc.SQLServerPreparedStatement$PrepStmtBatchExecCmd.doExecute(SQLServerPreparedStatement.java:2675)
	at com.microsoft.sqlserver.jdbc.TDSCommand.execute(IOBuffer.java:7417)
	at com.microsoft.sqlserver.jdbc.SQLServerConnection.executeCommand(SQLServerConnection.java:3274)
	at com.microsoft.sqlserver.jdbc.SQLServerStatement.executeCommand(SQLServerStatement.java:247)
	at com.microsoft.sqlserver.jdbc.SQLServerStatement.executeStatement(SQLServerStatement.java:222)
	at com.microsoft.sqlserver.jdbc.SQLServerPreparedStatement.executeBatch(SQLServerPreparedStatement.java:2076)
	at ExecuteBatchTest.executeBatch(ExecuteBatchTest.java:83)`
1. Any other details that can be helpful:
[...]

PR #1509, opened 2021-01-29
Fix | Clear batch params when clearing params in pstmt

rene-ye:
For issue #1499.

### DIFF
--- a/src/test/java/com/microsoft/sqlserver/jdbc/preparedStatement/BatchExecutionWithNullTest.java
+++ b/src/test/java/com/microsoft/sqlserver/jdbc/preparedStatement/BatchExecutionWithNullTest.java
@@ -110,6 +110,45 @@ public class BatchExecutionWithNullTest extends AbstractTest {
     public void testAddbatch2() throws SQLException {
         testAddBatch2(getConnection());
     }
+    
+    @Test
+    public void testClearBatch() throws SQLException {
+        String batchTable = RandomUtil.getIdentifier("batchTable");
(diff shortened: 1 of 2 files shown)


src/test/java/com/microsoft/sqlserver/jdbc/preparedStatement/BatchExecutionWithNullTest.java
```diff
@@ -110,6 +110,45 @@ public void testAddbatch2AEOnConnection() throws SQLException {
...
+
+        for (int idx = 1; idx <= count; idx++) {
+            pstmt.setLong(1, base + idx);
+            pstmt.setLong(2, run);
+            pstmt.addBatch();
+        }
+
+        int[] rowCounts = pstmt.executeBatch();
+        for (int idx = 0; idx < rowCounts.length; idx++) {
+            System.out.printf("Row %2d %s successfully inserted.\n", idx + 1, rowCounts[idx] == 1 ? "was" : "was not");
```
aimiebell, line 149: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [FS: Format string should use %n rather than n (VA_FORMAT_STRING_USES_NEWLINE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#fs-format-string-should-use-n-rather-than-n-va-format-string-uses-newline).
This format string includes a newline character (\n). In format strings, it is generally  preferable to use %n, which will produce the platform-specific line separator.

lilgreenbird:
see #1578

PR #1519, opened 2021-02-18
Fix clientSecret is null or empty for ActiveDirectoryServicePrincipal in Azure container

lilgreenbird:
This fixes the issue where an IllegalArgumentException is thrown when ActiveDirectoryServicePrincipal is used in an Azure container. 

[mssql-jdbc-hotfix.zip](https://github.com/microsoft/mssql-jdbc/files/6005361/mssql-jdbc-hotfix.zip)


sdebruyn (approved):
confirmed that this fixes #1519

peterbae (approved):

rene-ye (approved):

PR #1522, opened 2021-02-19
Optimize the retrieval of Java8 dates/times

fleiber:
Retrieval of Java8 dates/times with calls such as `rs.getObject(columnIndex, LocalDate.class)` can be made much faster, mostly because many unnecessary allocations are done.

For instance, when a datetime column is read, the convertTemporalToObject function:
1. Creates a new LocalDateTime, which internally creates a new LocalDate and a new LocalTime
2. Adds days, which creates a new LocalDate, wrapped in a new LDT
3. Adds nanos, which after complex computations (mostly useless because it's intraday nanos) creates a new LocalTime, wrapped in a new LDT

That's 7 allocations, when only 3 (or 2 for dates) are really useful.

I've also extended the test to ignore the time component, because I see many cases in the open world were timeless dates are stored in needlessly large types such as datetime.

I've only tested it on the table I was working on, but it should be fairly straightforward, and reduced CPU usage by a factor of ~3 when reading a date column.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/DDC.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/DDC.java
@@ -21,7 +21,9 @@ import java.nio.ByteBuffer;
 import java.nio.CharBuffer;
 import java.nio.charset.Charset;
 import java.text.MessageFormat;
+import java.time.LocalDate;
 import java.time.LocalDateTime;
+import java.time.LocalTime;
 import java.util.Calendar;
 import java.util.GregorianCalendar;
 import java.util.Locale;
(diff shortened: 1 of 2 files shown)


tkyc:
Hi @fleiber , sorry for the delay in reviewing your PR. It may take additional time to get approved as we are investigating an AE failure that came about from the PR.

fleiber:
> Hi @fleiber , sorry for the delay in reviewing your PR. It may take additional time to get approved as we are investigating an AE failure that came about from the PR.

Cool news (the review, not the failure ^^), thanks!

v-zhangw (approved):

lilgreenbird (approved):

Issue #1530, opened 2021-03-03
Writing strings longer than 2^30-1 characters results in endless loop


micheljung:
If a String longer than 2^30-1 characters is written to the database, an integer overflows causes an endless loop.

peterbae:
Thanks @micheljung, the team will test this and merge your PR once we've verified it.

PR #1531, opened 2021-03-03
Fix integer overflow in TDSWriter.writeString()

micheljung:
Fixes GH-1530

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
@@ -3940,7 +3940,7 @@ final class TDSWriter {
         int charsCopied = 0;
         int length = value.length();
         while (charsCopied < length) {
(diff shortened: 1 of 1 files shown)


peterbae:
@micheljung could you add a test case to show that this fix will solve the issue?

peterbae:
I was hoping we could host the file somewhere, since the standard procedure for pushing a fix to our repository requires some form of testing (to make sure the fix is intended). However, I think this fix is pretty straightforward, so I think we don't need a test case for this one. We will review and merge this soon.

micheljung:
The file wouldn't need to be hosted, the string can be generated. My question is whether you want a unit test that consumes 1 GiB of heap just for a single string :)

peterbae:
No, I think it would be ok without the test this time.

lilgreenbird (approved):

peterbae (approved):

v-zhangw (approved):

karinazhou (approved):

PR #1535, opened 2021-03-09
Open Connection Retry

lilgreenbird:
Adds support for open connection retry to handle idle connection resiliency by adding 2 new connection properties similar to those in SqlClient: connectRetryCount and connectionRetryInterval.


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -1283,65 +1288,105 @@ void validateMaxSQLLoginName(String propName, String propValue) throws SQLServer
...
-                        // only for the TLS 1.2 exception case.
-                        // if the user does not specify a default timeout, default is 15 per spec
-                        loginTimeoutSeconds = SQLServerDriverIntProperty.LOGIN_TIMEOUT.getDefaultValue();
-
-                        String sPropValue = propsIn.getProperty(SQLServerDriverIntProperty.LOGIN_TIMEOUT.toString());
-                        if (null != sPropValue && sPropValue.length() > 0) {
-                            int sPropValueInt = Integer.parseInt(sPropValue);
-                            if (0 != sPropValueInt) { // Use the default timeout in case of a zero value
-                                loginTimeoutSeconds = sPropValueInt;
+        for (int connectRetryAttempt = 0;;) {
```
peterbae, line 1304: I was confused as to why we had a for loop inside another for loop. If we just need those two variables declared to be 0, we could do `for (int connectRetryAttempt = 0, tlsRetryAttempt = 0;;)` instead?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -1283,65 +1288,105 @@ void validateMaxSQLLoginName(String propName, String propValue) throws SQLServer
...
+                                                    + connectRetryInterval + ")s >= loginTimeout(" + loginTimeoutSeconds
+                                                    + ")s");
+                                }
+                                throw e;
+                            }
+
+                            // wait for connectRetryInterval before retry
+                            if (connectionlogger.isLoggable(Level.FINEST)) {
+                                connectionlogger.finest(toString() + "Connection failed on transient error "
+                                        + sqlServerError.getErrorNumber() + ". Wait for  connectRetryInterval("
```
peterbae, line 1375: Wait for  connectRetryInterval -> Wait for connectRetryInterval

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerError.java
```diff
@@ -7,10 +7,96 @@
...
 /**
  * SQLServerError represents a TDS error or message event.
  */
 public final class SQLServerError extends StreamPacket implements Serializable {
+
+    /**
+     * List SQL Server transient errors drivers will retry on from
+     * https://docs.microsoft.com/en-us/azure/azure-sql/database/troubleshoot-common-errors-issues#transient-faults-connection-loss-and-other-temporary-errors
+     */
+    enum TransientError {
```
peterbae, line 20: Why is 49920 not being handled as a transient error? It looks like 40020 should've been added on line 31, and 40020 should be 49920 instead according to the docs.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResource.java
```diff
@@ -208,6 +208,10 @@ static String getResource(String key) {
                     "The authentication scheme to be used for integrated authentication."},
             {"R_lockTimeoutPropertyDescription",
                     "The number of milliseconds to wait before the database reports a lock time-out."},
+            {"R_connectRetryCountPropertyDescription",
+                    "The number of reconnection attempts if there is a connection failure."},
+            {"R_connectRetryIntervalPropertyDescription",
+                    "The number of seconds between each connection retry attempt.."},
```
peterbae, line 214: .. -> .

peterbae (approved):

v-zhangw (approved):

karinazhou (approved):

PR #1559, opened 2021-04-06
Fix insertion of Localdatetime during bulkcopy

### LINKED ISSUE #1376: BulkCopy won't accept a LocalTime/LocalDateTime instances of 100ns precision
## Driver version
8.3.1.jre14-preview

## SQL Server version
Microsoft SQL Server 2019 (RTM-CU4) (KB4548597) - 15.0.4033.1 (X64)   Mar 14 2020 16:10:35   Copyright (C) 2019 Microsoft Corporation  Developer Edition (64-bit) on Linux (Ubuntu 18.04.4 LTS) <X64>

## Client Operating System
Windows 10 with Docker

## JAVA/JVM version
JDK 14

## Problem description
1. Expected behaviour:
The driver should be able to bulk insert LocalTime/LocalDateTime instances of up to 100ns precision e.g. `2020-07-01 12:23:45.1234567`
2. Actual behaviour:
Throws an exception.
3. Error message/stack trace:
```
Caused by: com.microsoft.sqlserver.jdbc.SQLServerException: Received an invalid column length from the bcp client for colid XX.
	at com.microsoft.sqlserver.jdbc.SQLServerException.makeFromDatabaseError(SQLServerException.java:262)
	at com.microsoft.sqlserver.jdbc.TDSTokenHandler.onEOF(tdsparser.java:283)
	at com.microsoft.sqlserver.jdbc.TDSParser.parse(tdsparser.java:129)
	at com.microsoft.sqlserver.jdbc.TDSParser.parse(tdsparser.java:37)
	at com.microsoft.sqlserver.jdbc.TDSParser.parse(tdsparser.java:26)
	at com.microsoft.sqlserver.jdbc.SQLServerBulkCopy.doInsertBulk(SQLServerBulkCopy.java:1589)
	at com.microsoft.sqlserver.jdbc.SQLServerBulkCopy$1InsertBulk.doExecute(SQLServerBulkCopy.java:662)
	at com.microsoft.sqlserver.jdbc.TDSCommand.execute(IOBuffer.java:7349)
	at com.microsoft.sqlserver.jdbc.SQLServerConnection.executeCommand(SQLServerConnection.java:3181)
	at com.microsoft.sqlserver.jdbc.SQLServerBulkCopy.sendBulkLoadBCP(SQLServerBulkCopy.java:696)
	at com.microsoft.sqlserver.jdbc.SQLServerBulkCopy.writeToServer(SQLServerBulkCopy.java:1654)
	at com.microsoft.sqlserver.jdbc.SQLServerBulkCopy.writeToServer(SQLServerBulkCopy.java:619)
```
The `SQLServerBulkCopy` class handles 1000ns precision fine e.g. `2020-07-01 12:23:45.123456` but not 100ns.  The reason is that the bulk insert code calls the `toString` method on LocalTime/LocalDateTime to convert the data before sending.  If you look at the docs for `LocalTime#toString` you can see it will output the ns portion with trailing zeros for each scientific unit...
```
HH:mm:ss.SSS
HH:mm:ss.SSSSSS
HH:mm:ss.SSSSSSSSS
```
so once LocalTime/LocalDateTime hits 100ns precision the `toString` method will add 2 trailing zeros e.g. `2020-07-01 12:23:45.123456700` which BCP doesn't seem to like.

The String conversion is done [here](https://github.com/microsoft/mssql-jdbc/blob/4e3ab6690557c000bc16042771020ef530b54bf2/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java#L2272)

A workaround is to pass the data as a String without trailing zeros instead of using the LocalTime/LocalDateTime types.

v-zhangw:
Fixes the issue where insertion of localdatetime/datetime with certain previsions fail due to trailing zero's being added by the toString function.
Fixes #1376


src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyISQLServerBulkRecordTest.java
```diff
@@ -118,6 +143,39 @@ public void testISQLServerBulkRecord() throws SQLException {
                 data.add(CurrentRow);
             }
         }
+        
+        BulkData(DBTable dstTable, boolean dtTest) {
```
karinazhou, line 147: Can we update the existing BulkData() to support the LocalDataTime test? And the `dtTest` param is not used. I see that it is used to distinguish it from existing BulkData() but there should be a better way to handle this.

src/test/java/com/microsoft/sqlserver/testframework/DBSchema.java
```diff
@@ -95,6 +95,19 @@
...
         }
     }
+    
+    /**
+     * 
+     * @param autoGenerateSchema
+     * @param alternateSchema
+     * @param dtSchema
+     */
+    DBSchema(boolean autoGenerateSchema, boolean alternateSchema, boolean dtSchema) {
```
karinazhou, line 105: The same problem where not all the parameters are used. The existing DBSchema(boolean autoGenerateSchema) has covered `sqlTypes.add(new SqlDateTime2());`.

src/test/java/com/microsoft/sqlserver/testframework/DBTable.java
```diff
@@ -108,7 +131,7 @@ private void addColumns() {
         columns = new ArrayList<>(totalColumns);
 
         // Add RowID column
-        SqlType sqlType = schema.getSqlType(1); // Type SqlInt
+        SqlType sqlType = new SqlInt(); // Type SqlInt
```
karinazhou, line 134: Why is this changed here?

v-zhangw: getSqlType(1) only works if one of the columns is int before adding the index column, therefore it fails when making a test which does not have to have an int column.

karinazhou: If that works before, we keep it working as it is. The newly added test should not break the existing logic.

v-zhangw: Does the existing logic work not because of good coding, but because the existing test cases happen to meet the broken logic's requirements?

karinazhou: The initial logic is not broken if you read the code flow where the `addColumns()` is called. The `sqlTypes.add(new SqlInt());` is executed before when `autoGenerateSchema` is true. And `addColumns()` and `addColumns(boolean)` are only called when  `autoGenerateSchema` is true. If this needs to be changed to `new SqlInt()`, you will need to change both places to keep the consistency.

karinazhou (dismissed):

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
```diff
@@ -2285,6 +2287,12 @@ else if (null != sourceCryptoMeta) {
                             writeNullToTdsWriter(tdsWriter, bulkJdbcType, isStreaming);
                         } else {
                             String colValueStr = colValue.toString();
+                            
+                            // Remove extra trailing zeros added from toString
+                            if (colValue instanceof LocalDateTime || colValue instanceof LocalTime) {
+                            	colValueStr = colValueStr.contains(".") ? colValueStr.replaceAll("0*$","") : colValueStr;
+                            }
```
lilgreenbird, line 2294: I think you need to run formatter on this

v-zhangw: Formatted ran. Commited updates.

peterbae (dismissed):

lilgreenbird (approved):

karinazhou (approved):

lilgreenbird (approved):

PR #1562, opened 2021-04-08
Fix to ignore computed columns during BulkCopy

alagrede:
Hello,

I'm facing a problem with the bulk API for batch insert with computed columns.
I'm using an SQL table with computed columns like
```
create table myTable (
  id nvarchar(100) not null, 
  json nvarchar(max) not null,
  vcol1 as json_value([json], '$.vcol1')
)
```
My insert query uses for bulk:
`insert into myTable (id, json) values (?, ?)`
The current behavior of the driver is to reject the query with:
`java.sql.BatchUpdateException: The column "vcol1" cannot be modified because it is either a computed column or is the result of a UNION operator.`

After investigating and testing this fix on my project, **there is no specific reason to reject the batch insert**. The current Bulk code simply reads and compares all the columns found in the table.
I updated the current code to ignore irrelevant computed columns when inserting in bulk and it works like a charm.

Currently, the relevant documentation does not mention any limitation on the computed columns (so the documentation is still true):
https://docs.microsoft.com/en-us/sql/connect/jdbc/use-bulk-copy-api-batch-insert-operation?view=sql-server-ver15

I also added a new unit test to verify and demonstrate this use case: `BatchExecutionWithBulkCopyTest#testComputedCols`

_EDIT: Code updated to be compatible with versions prior than SQL Server 2016_

Sincerely

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
@@ -1710,6 +1710,24 @@ public class SQLServerBulkCopy implements java.lang.AutoCloseable, java.io.Seria
     }
 
     /**
+     * True if computed cols are supported by database server
+     *
+     * @return true if computed cols are supported
+     */
+    private boolean isComputedColsSupported() {
+        if (connection.sqlServerVersion != null) {
+            try {
+                int majorVersion = Integer.parseInt(connection.sqlServerVersion.split("\\.")[0]);
+                if (majorVersion >= 13) { // SQL Server 2016
+                    return true;
+                }
+            } catch (NumberFormatException e) {
+            }
+        }
+        return false;
+    }
+
+    /**
      * Returns the column metadata for the destination table (and saves it for later)
      */
     private void getDestinationMetadata() throws SQLServerException {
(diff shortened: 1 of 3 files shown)


peterbae:
Thanks for the contribution @alagrede, the team will test this and get back to you with the results.

lilgreenbird:
hi @alagrede 

Apologies it took so long to get back to you on this, this had fallen through the cracks..

We had discussed this within our team we are ok to add this to the driver could you please resolve the conflicts in your branch?

Also, we would rather not add server version checks in the driver, I think it should be fine to just check for is_computed since if it was not supported that column would not be returned. 

Thank you for your contribution.

alagrede:
Hi @lilgreenbird
As requested, I resolved the conflict and remove the server version check.
Best regards,

lilgreenbird (dismissed):

tkyc (dismissed):

Jeffery-Wasty (dismissed):

Jeffery-Wasty (approved):

atul-delphix:
@lilgreenbird & @alagrede ,

This fix cause NPE with computed column (if that computed column is NOT column as per column-id), see the exception:


java.lang.NullPointerException: Cannot read field "columnName" because the return value of "java.util.Map.get(Object)" is null
        at com.microsoft.sqlserver.jdbc.SQLServerBulkCopy.validateColumnMappings(SQLServerBulkCopy.java:1907)
        at com.microsoft.sqlserver.jdbc.SQLServerBulkCopy.writeToServer(SQLServerBulkCopy.java:1668)
        at com.microsoft.sqlserver.jdbc.SQLServerBulkCopy.writeToServer(SQLServerBulkCopy.java:630)

Jeffery-Wasty:
Hi @atul-delphix,

Can you open a new issue describing the above problem? Thanks.

atul-delphix:
@Jeffery-Wasty created new issue https://github.com/microsoft/mssql-jdbc/issues/2606

PR #1563, opened 2021-04-09
Fix intermittent test failure

lilgreenbird:


src/test/java/com/microsoft/sqlserver/jdbc/connection/TimeoutTest.java
```diff
@@ -118,7 +118,7 @@ public void testDMLoginTimeoutNotApplied() {
     public void testConnectRetryBadServer() {
         long timerEnd = 0;
         long timerStart = System.currentTimeMillis();
-        int loginTimeout = 1;
+        int loginTimeout = 15;
```
karinazhou, line 121: Why this timeout is increased here?

src/test/java/com/microsoft/sqlserver/jdbc/SQLServerConnectionTest.java
```diff
@@ -773,13 +771,10 @@ public void testThreadInterruptedStatus() throws InterruptedException {
             public void run() {
                 SQLServerDataSource ds = new SQLServerDataSource();
 
-                ds.setURL(connectionString);
+                ds.setURL(connectionString + "serverName=invalidServerName" + UUID.randomUUID());
```
karinazhou, line 774: The `serverName` is set in the next statement. Why it is set twice?

src/test/java/com/microsoft/sqlserver/jdbc/connection/TimeoutTest.java
lilgreenbird: this caused an intermittent failure as depending on the network 1s is likely not enough time for socket to time out so it test will fail intermittently if it's checking to see if timeout is < 1s

karinazhou (approved):

v-zhangw (approved):

peterbae (approved):

Issue #1539, opened 2021-03-11
trustStorePassword is null when using applicationIntent=ReadOnly


jvtrigueros:
## Driver version
2.9.1

## SQL Server version
Microsoft SQL Server 2016 (SP2-CU12) (KB4536648) - 13.0.5698.0 (X64) 
     \tFeb 15 2020 01:47:30 
     \tCopyright (c) Microsoft Corporation
     \tEnterprise Edition: Core-based Licensing (64-bit) on Windows Server 2016 Datacenter 10.0 <X64> (Build 14393: )

## Client Operating System
RHEL 7

## JAVA/JVM version
AdoptOpenJDK 11

## Table schema
N/A

## Problem description

Using `applicationIntent=ReadOnly` and configuring FIPS mode using the [documentation](https://docs.microsoft.com/en-us/sql/connect/jdbc/fips-mode?view=sql-server-ver15#properties) fails to connect specifically when using a custom truststore via the `trustStore` and `trustStorePassword` properties.

1. Expected behaviour:

Successful connection to secondary replica when given the hostname of the primary replica.

2. Actual behaviour:

Fails to connect to secondary replica because an SSL session cannot be created.

3. Error message/stack trace:

```
                              com.zaxxer.hikari.HikariDataSource.<init>     HikariDataSource.java:   81
                               com.zaxxer.hikari.pool.HikariPool.<init>           HikariPool.java:  115
                        com.zaxxer.hikari.pool.HikariPool.checkFailFast           HikariPool.java:  561
                      com.zaxxer.hikari.pool.HikariPool.createPoolEntry           HikariPool.java:  476
                           com.zaxxer.hikari.pool.PoolBase.newPoolEntry             PoolBase.java:  206
                          com.zaxxer.hikari.pool.PoolBase.newConnection             PoolBase.java:  364
                  com.zaxxer.hikari.util.DriverDataSource.getConnection     DriverDataSource.java:  138
                   com.microsoft.sqlserver.jdbc.SQLServerDriver.connect      SQLServerDriver.java:  881
               com.microsoft.sqlserver.jdbc.SQLServerConnection.connect  SQLServerConnection.java: 1291
       com.microsoft.sqlserver.jdbc.SQLServerConnection.connectInternal  SQLServerConnection.java: 2265
                 com.microsoft.sqlserver.jdbc.SQLServerConnection.login  SQLServerConnection.java: 2418
[...]

PR #1565, opened 2021-04-15
Fixes trustStorePassword set to null when using applicationIntent=ReadOnly

lilgreenbird:
Fixed issue where trustSTorePassword is removed too early


jvtrigueros:
This PR fixes #1539 for me! 🙌

v-zhangw (approved):

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -4520,6 +4520,8 @@ final void processFedAuthInfo(TDSReader tdsReader, TDSTokenHandler tdsTokenHandl
 
         aadPrincipalSecret = "";
         activeConnectionProperties.remove(SQLServerDriverStringProperty.AAD_SECURE_PRINCIPAL_SECRET.toString());
+
+        activeConnectionProperties.remove(SQLServerDriverStringProperty.TRUST_STORE_PASSWORD.toString());
```
cheenamalhotra, line 4524: Any specific reason to call it inside "processFedAuthInfo"?
Can you confirm this is called when there is no FedAuthInfo?

lilgreenbird: you are right I put this in the same place as the hotfix, but this should b somewhere else....🤦‍♀️

v-zhangw (approved):

karinazhou (approved):

peterbae (approved):

Issue #1058, opened 2019-05-15
Add Replication property to mssql-jdbc driver


dougmei:
This is a feature request to add the "Replication" boolean property to the SQL Server connection in the mssql-jdbc driver so that it can be configured by setting the property in the url and/or in the connection properties, something like this:

db.driver=com.microsoft.sqlserver.jdbc.SQLServerDriver
db.url=jdbc:sqlserver://localhost:59706;databaseName=symchuck;**replication=true**
db.connection.properties=**replication=true**;

This property is available through the SQLConnectionSringBuilder in ADO.Net and it would be nice to have it available in jdbc as well.

rene-ye:
Hi @dougmei, the team will have a look to see what can be done.

PR #1566, opened 2021-04-15
Adding replication connection option

David-Engel:
The replication property simply tells the database that a connection is for replication and affects when the "NOT FOR REPLICATION" setting on triggers and constraints takes affect.

This is the feature requested in #1058.


src/test/java/com/microsoft/sqlserver/jdbc/connection/ReplicationTest.java
```diff
@@ -0,0 +1,89 @@
...
+@RunWith(JUnitPlatform.class)
+@Tag(Constants.xAzureSQLDW)
+public class ReplicationTest extends AbstractTest {
+
+    @Test
+    public void testReplication() throws SQLException {
+        String tableName = RandomUtil.getIdentifier("repl");
+        String triggerName = RandomUtil.getIdentifier("trig");
+        String escapedTableName = DBTable.escapeIdentifier(tableName);
+        String escapedTriggerName = DBTable.escapeIdentifier(triggerName);
```
lilgreenbird, line 40: don't use DBTable, that is old framework code that we are moving away from

src/test/java/com/microsoft/sqlserver/jdbc/connection/ReplicationTest.java
```diff
@@ -0,0 +1,89 @@
...
+        SQLServerDataSource ds = new SQLServerDataSource();
+        ds.setURL(connectionString);
+        ds.setReplication(false);
+
+        String sqlDropTable = "IF OBJECT_ID('" + TestUtils.escapeSingleQuotes(escapedTableName) + "', 'U') IS NOT NULL "
```
lilgreenbird, line 45: use TestUtils.dropTableIfExists?

src/test/java/com/microsoft/sqlserver/jdbc/connection/ReplicationTest.java
David-Engel: Is there a current utility method that replaces it?

lilgreenbird: oh just use AbstractSQLGenerator.escapeIdentifier(tableName). As an example, see SQLServerConnectTest.java

lilgreenbird (approved):

PR #1568, opened 2021-04-22
Fixed issue when redirected token contains named instance in servername

lilgreenbird:


peterbae (approved):

peterbae (dismissed):

karinazhou:
Could you please add some description about what the scenario the PR is for and what issue it is to fix? Since this does not come from any GitHub issues, we need some background for reference in the future.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -4365,13 +4365,20 @@ final void processEnvChange(TDSReader tdsReader) throws SQLServerException {
                     }
                 }
 
-                isRoutedInCurrentAttempt = true;
-                routingInfo = new ServerPortPlaceHolder(routingServerName, routingPortNumber, null, integratedSecurity);
+                int px = routingServerName.indexOf('\\');
```
karinazhou, line 4368: This will be checked again when we call ServerportPlaceHolder later. The routingInstanceName doesn't seem necessary which can be set up in `ServerPortPlaceHolder` constructor.

karinazhou (approved):

David-Engel (approved):

peterbae (approved):

v-zhangw (approved):

PR #1578, opened 2021-05-11
Fix | Clear batch params when clearing params in pstmt

karinazhou:
This PR is to port the changes in #1509. 
- Fix issue #1499.
- Fix test failure in `testClearBatch`.
- Add `testClearBatchAEOnConnection` for `columnEncryptionSetting=Enabled` scenario.

### DIFF
--- a/src/test/java/com/microsoft/sqlserver/jdbc/preparedStatement/BatchExecutionWithNullTest.java
+++ b/src/test/java/com/microsoft/sqlserver/jdbc/preparedStatement/BatchExecutionWithNullTest.java
@@ -134,7 +134,7 @@ public class BatchExecutionWithNullTest extends AbstractTest {
         testClearBatch(getConnection());
     }
     
-    public void testClearBatch(Connection conn) throws SQLException {
+    private void testClearBatch(Connection conn) throws SQLException {
         String batchTable = TestUtils
                 .escapeSingleQuotes(AbstractSQLGenerator.escapeIdentifier(RandomUtil.getIdentifier("batchTable")));
         String CREATE_TABLE_SQL = "create table " + batchTable + " (KEY1 numeric(19,0) not null, KEY2 numeric(19,0) not null, primary key (KEY1, KEY2))";
@@ -169,7 +169,7 @@ public class BatchExecutionWithNullTest extends AbstractTest {
 
         int[] rowCounts = pstmt.executeBatch();
         for (int idx = 0; idx < rowCounts.length; idx++) {
-            System.out.printf("Row %2d %s successfully inserted.\n", idx + 1, rowCounts[idx] == 1 ? "was" : "was not");
+            assertTrue(rowCounts[idx] == 1, "Row " + idx + " was not successfully inserted.");
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/test/java/com/microsoft/sqlserver/jdbc/preparedStatement/BatchExecutionWithNullTest.java
```diff
@@ -110,6 +110,68 @@ public void testAddbatch2AEOnConnection() throws SQLException {
...
+     * Test the same as testClearBatchAEOnConnection, with AE disabled
+     * 
+     * @throws SQLException
+     */
+    @Test
+    public void testClearBatch() throws SQLException {
+        testClearBatch(getConnection());
+    }
+    
+    public void testClearBatch(Connection conn) throws SQLException {
```
lilgreenbird, line 137: this can be private

src/test/java/com/microsoft/sqlserver/jdbc/preparedStatement/BatchExecutionWithNullTest.java
```diff
@@ -110,6 +110,68 @@ public void testAddbatch2AEOnConnection() throws SQLException {
...
+
+        for (int idx = 1; idx <= count; idx++) {
+            pstmt.setLong(1, base + idx);
+            pstmt.setLong(2, run);
+            pstmt.addBatch();
+        }
+
+        int[] rowCounts = pstmt.executeBatch();
+        for (int idx = 0; idx < rowCounts.length; idx++) {
+            System.out.printf("Row %2d %s successfully inserted.\n", idx + 1, rowCounts[idx] == 1 ? "was" : "was not");
```
lilgreenbird, line 172: remove debug printf and add check for expected value and fail test accordingly

src/test/java/com/microsoft/sqlserver/jdbc/preparedStatement/BatchExecutionWithNullTest.java
```diff
@@ -110,6 +110,68 @@ public void testAddbatch2AEOnConnection() throws SQLException {
...
+     * @throws SQLException
+     */
+    @Test
+    public void testClearBatch() throws SQLException {
+        testClearBatch(getConnection());
+    }
+    
+    public void testClearBatch(Connection conn) throws SQLException {
+        String batchTable = TestUtils
+                .escapeSingleQuotes(AbstractSQLGenerator.escapeIdentifier(RandomUtil.getIdentifier("batchTable")));
```
lilgreenbird, line 139: all the tests in this class use  `tableName` it creates that for every test in `@BeforeEach` so this gets created here too. Can this use the same table? Otherwise we need to update `testSetup` and `terminateVariation`. Also, usually we do a `dropTableIfExists` before table creation in case there's junk left from previous runs.

karinazhou: The table columns in the `testClearBatch` cases are different from the ones we use in other tests. It currently creates a new table when `testClearBatch` with and without AE run, and drops the table in the `finally` block. If we are going to use the same table name for each test, we need to move the table creation out of `@BeforeEach` and create each table in the corresponding test. What do you think?

src/test/java/com/microsoft/sqlserver/jdbc/preparedStatement/BatchExecutionWithNullTest.java
karinazhou: Done.

src/test/java/com/microsoft/sqlserver/jdbc/preparedStatement/BatchExecutionWithNullTest.java
karinazhou: Added comment based on conversation.

lilgreenbird (approved):

v-zhangw (approved):

PR #1591, opened 2021-06-03
Fix the hanging issue after running the clearBatch test

karinazhou:
After running the BatchExecutionWithNullTest.testClearBatch(), the following tests will encounter a hanging status. The server is locked by the test process.


karinazhou:
Close this PR since another PR #1593 is for the same purpose.

PR #1595, opened 2021-06-03
Send client process ID to server on Java 9+

David-Engel:
Since we can easily do it in Java 9 and up, we should. If catching NoClassDefFoundError is objectionable, I'm open to other suggestions.

You can validate the behavior by querying `sys.dm_exec_sessions` when a connection is active:
`select * from sys.dm_exec_sessions`

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
@@ -1216,6 +1216,21 @@ public class SQLServerConnection implements ISQLServerConnection, java.io.Serial
             .getLogger("com.microsoft.sqlserver.jdbc.Connection");
     private static String loggingClassName = "com.microsoft.sqlserver.jdbc.SQLServerConnection:";
 
+    /** Client process ID sent during login */
+    private static int pid = 0;
+
+    static {
+        long pidLong = 0;
+        try {
+            pidLong = ProcessHandle.current().pid();
+        } catch (NoClassDefFoundError e) { // ProcessHandle is Java 9+
+            if (loggerExternal.isLoggable(Level.FINER) && Util.isActivityTraceOn()) {
+                loggerExternal.finer(loggingClassName + " NoClassDefFoundError for ProcessHandle. ProcessId will be 0.");
+            }
+        }
+        pid = (pidLong > Integer.MAX_VALUE) ? 0 : (int)pidLong;
+    }
+
     /**
      * There are three ways to get a failover partner connection string, from the failover map, the connecting server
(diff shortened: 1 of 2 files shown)


lilgreenbird (approved):

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -1216,6 +1216,21 @@ final SQLCollation getDatabaseCollation() {
...
     private static String loggingClassName = "com.microsoft.sqlserver.jdbc.SQLServerConnection:";
 
+    /** Client process ID sent during login */
+    private static int pid = 0;
+
+    static {
+        long pidLong = 0;
+        try {
+            pidLong = ProcessHandle.current().pid();
+        } catch (NoClassDefFoundError e) { // ProcessHandle is Java 9+
```
lilgreenbird, line 1226: I like this, better than check java version !

lilgreenbird (approved):

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -1216,6 +1216,21 @@ final SQLCollation getDatabaseCollation() {
...
+            if (loggerExternal.isLoggable(Level.FINER) && Util.isActivityTraceOn()) {
+                loggerExternal.finer(loggingClassName + " NoClassDefFoundError for ProcessHandle. ProcessId will be 0.");
+            }
+        }
+        pid = (pidLong > Integer.MAX_VALUE) ? 0 : (int)pidLong;
```
karinazhou, line 1231: If NoClassDefFoundError is caught, what value will pidLong be? Is it still 0 r something unexpected like Integer.MAX_VALUE or Integer.MIN_VALUE?

lilgreenbird: should be 0? line 1223 initialized that to 0

karinazhou: So in what case, pidLong would be bigger than Integer.MAX_VALUE?

v-zhangw: In the case where the actual process ID is simply greater than the maximum integer value. This is theoretically possible as Windows process ID's are stored as DWORDs and could go up to 2^(32 − 1).

karinazhou (approved):

v-zhangw (approved):

cheenamalhotra:
Just curious.. should this call be wrapped under JDBC 43 support check?

PR #1620, opened 2021-07-23
Log the driver version + plat info on connect to aid in log analysis

David-Engel:
To aid in log analysis. I was tempted to log all system properties, but there are some sensitive ones in there, so I targeted the logging to relevant properties.

This is not a requirement to merge. I'm open to discussion. It just seems like this might help when we get issues where people leave out important details.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDriver.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDriver.java
@@ -903,6 +903,11 @@ public final class SQLServerDriver implements java.sql.Driver {
         return null;
     }
 
+    private final static String[] systemPropertiesToLog = new String[] { "java.specification.vendor",
+            "java.specification.version", "java.class.path", "java.class.version", "java.runtime.name",
+            "java.runtime.version", "java.vendor", "java.version", "java.vm.name", "java.vm.vendor", "java.vm.version",
+            "java.vm.specification.vendor", "java.vm.specification.version", "os.name", "os.version", "os.arch" };
+
     @Override
     public java.sql.Connection connect(String Url, Properties suppliedProperties) throws SQLServerException {
         loggerExternal.entering(getClassNameLogging(), "connect", "Arguments not traced.");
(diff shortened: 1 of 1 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDriver.java
```diff
@@ -903,11 +903,30 @@ static String getPropertyOnlyName(String name, Logger logger) {
...
         loggerExternal.entering(getClassNameLogging(), "connect", "Arguments not traced.");
         SQLServerConnection result = null;
 
+        if (loggerExternal.isLoggable(Level.FINE)) {
+            loggerExternal.log(Level.FINE, "Microsoft JDBC Driver " + SQLJdbcVersion.major + "." + SQLJdbcVersion.minor +
+                    "." + SQLJdbcVersion.patch + "." + SQLJdbcVersion.build + SQLJdbcVersion.releaseExt + " for SQL Server");
+            if (loggerExternal.isLoggable(Level.FINER)) {
+                for (String propertyKeyName : systemPropertiesToLog) {
+                    String propertyValue = System.getProperty(propertyKeyName);
+                    if (propertyValue != null && !propertyValue.isEmpty()) {
```
lilgreenbird, line 922:
```suggestion
                    if (null != propertyValue && !propertyValue.isEmpty()) {
```

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDriver.java
```diff
@@ -903,11 +903,30 @@ static String getPropertyOnlyName(String name, Logger logger) {
         return null;
     }
 
+    private final static String[] systemPropertiesToLog = new String[] { "java.specification.vendor",
+            "java.specification.version", "java.class.path", "java.class.version", "java.runtime.name",
+            "java.runtime.version", "java.vendor", "java.version", "java.vm.name", "java.vm.vendor", "java.vm.version",
+            "java.vm.specification.vendor", "java.vm.specification.version", "os.name", "os.version", "os.arch" };
```
cheenamalhotra, line 909: I'm not very sure if these System properties are "required" to be logged by the driver. Reason being, they are not owned/provided by the driver, if anyone wants to log, they could do themselves as they're readily available in JVM.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDriver.java
David-Engel: That style honestly drives me nuts. It just reads unnaturally to me. I've read the arguments for both styles, and there is no real advantage to one over the other in Java, that I'm aware of.

Our [Coding Guidelines](https://github.com/microsoft/mssql-jdbc/blob/dev/Coding_Guidelines.md) don't mention it, either. Is this an unspoken style the team follows?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDriver.java
David-Engel: The intent is to gather information that is sometimes left out during support but that affects driver functionality. I've encountered issues that can be traced to differences around these settings.

dibianalvarez (approved):

dibianalvarez (approved):

v-zhangw (approved):

lilgreenbird (approved):

PR #1626, opened 2021-07-29
Remove Java 9-specific class references from the Java 8 jar

David-Engel:


lilgreenbird (approved):

cheenamalhotra:
We need to be able to build with JDK 1.8 - it seems there's an override in SqlServerDatabaseMetadata that should be split in a separate *43 class respectively and excluded.

```
    @Override
    public boolean supportsSharding() throws SQLException {
        DriverJDBCVersion.checkSupportsJDBC43();
        checkClosed();
        return false;
    }
```

pom.xml
```diff
@@ -252,7 +252,12 @@
 						<configuration>
 							<excludes>
 								<exclude>**/com/microsoft/sqlserver/jdbc/SQLServerJdbc43.java</exclude>
+								<exclude>**/com/microsoft/sqlserver/jdbc/SQLServerConnection43.java</exclude>
```
cheenamalhotra, line 255: Please also include:
`<exclude>**/com/microsoft/sqlserver/jdbc/ISQLServerConnection43.java</exclude>`

pom.xml
```diff
@@ -252,7 +252,12 @@
...
 							</excludes>
+							<testExcludes>
```
cheenamalhotra, line 257: Please also include:
`<exclude>**/com/microsoft/sqlserver/jdbc/connection/RequestBoundaryMethodsTest.java</exclude>`

David-Engel:
> We need to be able to build with JDK 1.8 - it seems there's an override in SqlServerDatabaseMetadata that should be split in a separate *43 class respectively and excluded for jre8 build.

@cheenamalhotra I saw that when I tried building with JDK 1.8. Since that class wasn't mentioned in the original issue, I thought I could skip it and hope the reference checking that is happening at load time in their scenario is not evaluating overrides.

cheenamalhotra:
Yes, it's resolved on runtime as Java 8 doesn't have the API to override so it ignores but we should consider Eclipse environment errors too, and exclude all that contributes to errors. I would say this code is still present that's not applicable to jre8 and could be removed with no risk.

lilgreenbird (approved):

PR #1639, opened 2021-08-12
Fix NPE with Statement.closeOnCompletion()

David-Engel:
This is the test from #1518 plus the fix for the underlying issue.

(Bonus: Fix for the javadoc warning.)

### DIFF
--- a/src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/StatementTest.java
+++ b/src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/StatementTest.java
@@ -7,6 +7,7 @@ package com.microsoft.sqlserver.jdbc.unit.statement;
 import static org.junit.Assert.assertArrayEquals;
 import static org.junit.Assert.fail;
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.io.StringReader;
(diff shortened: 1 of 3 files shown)


lilgreenbird (dismissed):

v-zhangw (approved):

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/StatementTest.java
```diff
@@ -2052,6 +2053,40 @@ public void testClosedResultSet() throws Exception {
...
+            try (Connection conn = getConnection(); Statement stmt = conn.createStatement()) {
+
+                // enable isCloseOnCompletion
+                try {
+                    stmt.closeOnCompletion();
+                } catch (Exception e) {
+                    fail(TestResource.getResource("R_unexpectedException") + e.getMessage());
+                }
+
+                SQLServerResultSet rs = (SQLServerResultSet) stmt.executeQuery("SELECT 1");
```
lilgreenbird, line 2072: use try with resource on rs

David-Engel: It's been a while since I submitted this PR, but I have a vague recollection that this was intentional to ensure rs was not cleaned up before the subsequent stmt.executeQuery() runs. I feel like I changed the code in the original PR to using try with resources and couldn't get it to fail before... I can try again. I'll add a comment if it really needs to be done without the try with resources.

David-Engel: Done.

lilgreenbird (approved):

PR #1485, opened 2020-12-16
BulkCopy won't accept a LocalTime/LocalDateTime instances of 100ns precision; issue-1376

MariusVladoiu:
if the date is a instance of LocalDateTime, then limit the nanoseconds to a maximum 7 digits; because mssql database accept only a maximum of 7 digit nanoseconds.

PR #1640, opened 2021-08-13
Fix conversion of LocalDateTime and LocalTime to String in Bulk Copy

joechu1:
When writing a LocalDateTime with bulk copy, the driver may return an error when second-of-minute and nano-of-second is 0:

```
com.microsoft.sqlserver.jdbc.SQLServerException: Conversion failed when converting date and/or time from character string. Msg 241, Level 16, State 1, Conversion failed when converting date and/or time from character string.
```

When time units to the right of the minute are all 0's, the LocalDateTime.toString() will simplify the string format to **uuuu-MM-dd'T'HH:mm**, which SQL Server is not able to process due to the missing seconds. This PR changes the string conversion to use the DateTimeFormatter and ensures that it includes the necessary padding.

For example:
```
LocalDateTime                  = 2021-08-13 00:00:00

Current - toString()           = 2021-08-13T00:00              // Returns error
New     - DateTimeFormatter    = 2021-08-13T00:00:00           // It works!
```

Note that this is very similar to PR #1485.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
@@ -2292,22 +2292,25 @@ public class SQLServerBulkCopy implements java.lang.AutoCloseable, java.io.Seria
                         if (null == colValue) {
                             writeNullToTdsWriter(tdsWriter, bulkJdbcType, isStreaming);
                         } else {
-                            String colValueStr = colValue.toString();
 
+                            String colValueStr;
                             if (colValue instanceof LocalDateTime) {
                                 colValueStr = ((LocalDateTime)colValue).format(DateTimeFormatter.ofPattern("uuuu-MM-dd'T'HH:mm:ss.SSSSSSS"));
                             }
                             else if (colValue instanceof LocalTime) {
                                 colValueStr = ((LocalTime)colValue).format(DateTimeFormatter.ofPattern("HH:mm:ss.SSSSSSS"));
                             }
+                            else {
+                                colValueStr = colValue.toString();
+                            }
 
                             if (unicodeConversionRequired(bulkJdbcType, destSSType)) {
-                                int stringLength = colValue.toString().length();
+                                int stringLength = colValueStr.length();
                                 byte[] typevarlen = new byte[2];
                                 typevarlen[0] = (byte) (2 * stringLength & 0xFF);
                                 typevarlen[1] = (byte) ((2 * stringLength >> 8) & 0xFF);
                                 tdsWriter.writeBytes(typevarlen);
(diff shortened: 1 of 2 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
```diff
@@ -2294,10 +2294,11 @@ else if (null != sourceCryptoMeta) {
...
-                            // Remove extra trailing zeros added from toString
-                            if (colValue instanceof LocalDateTime || colValue instanceof LocalTime) {
-                                colValueStr = colValueStr.contains(".") ? colValueStr.replaceAll("0*$", "")
-                                                                        : colValueStr;
+                            if (colValue instanceof LocalDateTime) {
+                                colValueStr = ((LocalDateTime)colValue).format(DateTimeFormatter.ofPattern("uuuu-MM-dd'T'HH:mm:ss.SSSSSSS"));
+                            }
+                            else if (colValue instanceof LocalTime) {
+                                colValueStr = ((LocalTime)colValue).format(DateTimeFormatter.ofPattern("HH:mm:ss.SSSSSSS"));
                             }
```
David-Engel, line 2302: Small optimization: I would move colValue.toString() into an else block so that it doesn't run unnecessarily.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
```diff
@@ -2294,10 +2294,11 @@ else if (null != sourceCryptoMeta) {
...
 
                             if (unicodeConversionRequired(bulkJdbcType, destSSType)) {
```
David-Engel, line 2304: Also, since you are in here, the below references to colValue.toString() (lines 2304-2310) can be updated to colValueStr to avoid additional toString() calls.

David-Engel (commented):
We should also extend testBulkCopyDateTimePrecision added in #1559 to include data that this fix is covering.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
joechu1: Done

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
joechu1: Done

AbductedNoah (dismissed):

David-Engel (approved):

lilgreenbird (approved):

lilgreenbird:
hi @joechu1 
Thank you for your contribution, apologies this took so long, we were busy with other higher priorities issues.

PR #1643, opened 2021-08-19
Add 42108 and 42109 to transient errors list

David-Engel:
Also fixing some build and javadoc warnings

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerLexer.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerLexer.java
@@ -20,7 +20,7 @@ import org.antlr.v4.runtime.dfa.DFA;
 @SuppressWarnings({"all", "warnings", "unchecked", "unused", "cast"})
 class SQLServerLexer extends Lexer {
     static {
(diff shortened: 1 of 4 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerLexer.java
```diff
@@ -20,7 +20,7 @@
 @SuppressWarnings({"all", "warnings", "unchecked", "unused", "cast"})
 class SQLServerLexer extends Lexer {
     static {
-        RuntimeMetaData.checkVersion("4.7.2", RuntimeMetaData.VERSION);
+        RuntimeMetaData.checkVersion("4.9.2", RuntimeMetaData.VERSION);
```
lilgreenbird, line 23: what's this for? this doesn't have anything to do with transient errors?

David-Engel: That's to fix the build warning we've been getting since bumping the ANTLR dependency from 4.7.2 to 4.9.2.

cheenamalhotra (approved):

PR #1665, opened 2021-09-29
Updated MD4 based on Bouncy Castle implementation

lilgreenbird:


David-Engel (approved):
Approved - assuming no tests break...  😄

PR #1687, opened 2021-11-11
Fixes #1590 - [FEATURE REQUEST] Ability to explicitly specify conversion to datetime or datetime2 

### LINKED ISSUE #1590: [FEATURE REQUEST] Ability to explicitly specify conversion to datetime or datetime2
We have an application that uses SQL Server 2019 and the current schema uses `datetime` datatype for tracking timestamps. Since the database compatibility level is set to 2019, the JDBC driver is converting all timestamps into `datetime2` datatypes which is causing a number of bugs in our application.

We are currently using a third party JDBC driver (Progress® DataDirect Connect® Series for JDBC™) which offers a `DateTimeInputParameterType` connection string which you can use on newer databases to have it declare parameters as `datetime` instead of `datetimeoffset`. However, we are soon planning on moving our application stack and want to use the Microsoft SQL Server Driver as it is officially support in the new application server we plan on using.

I would like to see a similar connection string added to the JDBC driver which would allow you to specify how the JDBC driver handlers timestamps without having to lower the compatibility level of the database.

Add a new connection string (e.g. `timestampEncoding`) which can be used to determine how timestamp should be encoded.  For example, you might have the following options:

Option | Description
------- | ------------
DATETIMEOFFSET | always use DATETIMEOFFSET.
DATETIME2 | always use DATETIME2. Good if the user knows they only use DATETIME2 datatypes in their schema (current behavior).
DATETIME | always use DATETIME (old behavior). Good if the user is dealing with a legacy schema and is not willing to upgrade.
QUERYDB | the new code, but not compatible with encrypted connections. Good for mixed schemas.

See https://github.com/microsoft/mssql-jdbc/issues/443#issuecomment-849720800

Alternatives we've considered:

* Converting the schema to use `datetime2`. We will probably eventually do this, but in a large multi-tenant application stack this is going to take a lot of work. We have a lot of functionality driven by timestamps, so this refactoring will take time.
* Casting values to `datetime`. Timely to implement and would require another round of refactoring when we final do migrate to `datetime2` values.
* Lowering the database compatibility level. Unfortunately, we are using some SQL Server 2017 functionality, so in order to do this we would need to remove this functionality. We would also lose query plan performance improvements implemented in newer versions of SQL Server.
* Forking the current code and using our fork. We'd like to avoid forking the code if possible, so we can stay on the official code path.

## Additional context
none

n/a

dswitzer:
I've implemented a new connection string (i.e. `datetimeParameterType`) which fixes #1590 by allowing users to specify the datatype to use for date/timestamp parameters.

This pull request adds the ability to define the `datetimeParameterType` connection string with any of the following values:

Option | Description
------- | ------------
datetime2| always use DATETIME2. Good if the user knows they only use DATETIME2 datatypes in their schema (current behavior, this is the default).
datetime| always use DATETIME. Good if the user is dealing with a legacy schema and is not willing to upgrade.
datetimeoffset| always use DATETIMEOFFSET. 

By default, the value defaults to `datetime2`, which is the same behavior as today. Setting this value to any of the above will cause any SQL version of Katmai or higher to always use that datatype for date/time values when generating parameters. For older versions of SQL Server that do not support `datetime2` or `datetimeoffset`, then `datetime` is always used (just as it has been working).

I've added unit test coverage for the changes.

There should be no behavioral change for users who do not set the `datetimeParameterType` connection string value. It should work as it has been. However, when supporting older databases running in SQL Server 2016 or higher compatibility mode, the `datetimeParameterType` can be set to `datetime` to support databases only using that data type without worrying about casting issues in SQL Server with conversion between `datetime2` and `datetime`.


v-zhangw:
Hello @dswitzer,
Could you please resolve the merge conflicts in this PR? Then we can run the pipeline tests to check for acceptance.

dswitzer:
@v-zhangw ,

Ok, I believe the merge should be fixed, some of the checks are failing, but they not in portions of the code that should have been affected by my changes.

dswitzer:
@v-zhangw

Just checking up to see if there's anything more you need from me.

pacham7:
@lilgreenbird @v-zhangw, this pull request would be really useful, as updating to a later driver would unblock migration to jdk17 for us (ssl issues with existing very old driver). Is there any chance someone can take a look at it?

dswitzer:
It should be pretty easy to resolve the conflicts, I just do not have write access to the repository to do it.

AlBundy33:
@dswitzer not sure but I think you have to update your forked branch and fix the conflicts there.

dswitzer:
@AlBundy33,

I'd be happy to update the fork, but the origin code keeps changing so fast, that the PR keeps becoming invalidated. It wasn't worth the time to keep fixing the code when there's no movement on them approving the PR. 

If they're willing to actually move on the PR, I'm happy to fix the conflicts.

I just don't want to keep spending time to fix my PR for no reason.

AlBundy33:
@v-zhangw @lilgreenbird @Jeffery-Wasty what do you think about the pr?

Jeffery-Wasty:
Hi @AlBundy33,

The issue and associated PR are on our radar and we'll look into both of them as soon as we are able to.

lilgreenbird:
Hi @dswitzer 

Apologies this has been put off for so long. We had determined that the #1590 needed more investigation and then we have had problems juggling this as this a feature request so it kept getting bumped by other higher priority issues. We are now getting back to looking into this again. Once you get this PR up to date we will run this thru our test lab to verify.

dswitzer:
I've updated the pull request. The failures do not appear to be related to my changes.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerLexer.java
```diff
@@ -20,7 +20,7 @@
 @SuppressWarnings({"all", "warnings", "unchecked", "unused", "cast"})
 class SQLServerLexer extends Lexer {
     static {
-        RuntimeMetaData.checkVersion("4.9.3", RuntimeMetaData.VERSION);
+        RuntimeMetaData.checkVersion("4.9.2", RuntimeMetaData.VERSION);
```
lilgreenbird, line 23: why is this change necessary?

dswitzer: It shouldn't be. Mostly likely just a merge conflict. That's not something I purposefully changed.

src/main/java/com/microsoft/sqlserver/jdbc/Parameter.java
```diff
@@ -625,24 +625,18 @@ private void setTypeDefinition(DTV dtv) {
                          * generic type info can be used as before.
                          */
                         if (userProvidesScale) {
-                            param.typeDefinition = con
-                                    .isKatmaiOrLater() ? (SSType.DATETIME2.toString() + "(" + outScale + ")")
-                                                       : (SSType.DATETIME.toString());
+                            param.typeDefinition = getDatetimeDataType(con, outScale);
```
lilgreenbird, line 628: this check still need to be there for !isKatmaiOrLater()as DATATIME2 is not supported before

same with line 632

lilgreenbird: perhaps we should log an error if it's set to datatime2 and pre 2008

lilgreenbird (commented):
- please use our code formatter on all the changes.
- there seems to be an issue with the Mac CI pipeline line please ignore that
- fyi I'm in the process of running this in our test lab. As this adds a new API we will also need this to be approved by our product manager who's away this week, I'll ask him to review this next week.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResource.java
```diff
@@ -213,6 +214,7 @@ protected Object[][] getContents() {
         {"R_hostNameInCertificatePropertyDescription", "The host name to be used when validating the SQL Server Secure Sockets Layer (SSL) certificate."},
         {"R_replicationPropertyDescription", "This setting tells the server if the connection is used for replication."},
         {"R_sendTimeAsDatetimePropertyDescription", "Determines whether to use the SQL Server datetime data type to send java.sql.Time values to the database."},
+        {"R_datetimeParameterTypePropertyDescription", "Determines the datatype to use for datetime and timestamp values. Valid options are: datetime, datetime2 or datetimeoffset"},
```
lilgreenbird, line 217: we don't usually add property value options in the error strings as there may be changed

src/main/java/com/microsoft/sqlserver/jdbc/Parameter.java
```diff
@@ -902,6 +896,45 @@ private void setTypeDefinition(DTV dtv) {
...
+                }
+            }
+
+            /*
+                For older versions of SQL server and for some reason the datetimeParameterType
+                connection property cannot be determined, we fall back to the "datetime"
+                format.
+            */
+            return SSType.DATETIME.toString();
+        }
```
lilgreenbird, line 936: I'm wondering how this should be handled, if one explicitly sets data2 but it's pre SQL Server 2008. Should this silently change to datetime or error or just log a warning? Don't make any changes here yet I'll have to check with our product manager next week to see which he prefers.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResource.java
dswitzer: I was just trying to follow the pattern for what I saw in the code for the `sendTimeAsDatetime` option, as they seemed closely related.

dswitzer:
@lilgreenbird,

> * please use our code formatter on all the changes.

When I run the _Format Document_ command from VS Code, it's making a lot of changes to the files in places of the code I did not change—especially in the comments. 

Is that what you're wanting me to do? 

If not, if you could provide me with explicit instructions on what you need me to do, I would appreciate it very much!

src/main/java/com/microsoft/sqlserver/jdbc/Parameter.java
dswitzer: For pre-Katmai, I wouldn't do anything here. Maybe there should be a warning in the log if you try to set the value saying the parameter is unsupported.

For Katmai, I wasn't sure how you generally approach things like this. I was going to fault tolerance and trying to minimize any backwards compatibility issues. I think it would make some sense to throw a hard error if an invalid value is supplied, but a warning message certainly makes sense.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -6880,6 +6897,25 @@ public void setSendTimeAsDatetime(boolean sendTimeAsDateTimeValue) {
...
+            String errorMessage = "The timestamp encoding value (i.e. " + datetimeParameterTypeValue.toString() + ") must be: datetime, datetime2 or datetimeoffset.";
+            SQLServerException newe = new SQLServerException(errorMessage, null);
+            throw newe;
+        }
+        datetimeParameterType = DatetimeType.valueOfString(datetimeParameterTypeValue);
+    }
+
+    private boolean isValidDatetimeParameterType(String datetimeParameterTypeValue) {
+        return (datetimeParameterTypeValue.equals("datetime") || datetimeParameterTypeValue.equals("datetime2") || datetimeParameterTypeValue.equals("datetimeoffset"));
+    }
```
David-Engel, line 6917: You can run all this through the `DatetimeType.valueOfString(datetimeParameterValueType)` method to validate the input parameter and eliminate the error message instead of hard coding more strings here.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerLexer.java
David-Engel: This should be reverted.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResource.java
David-Engel: It looks like we haven't been consistent in the past. I'm good with including the possible values as it's helpful to the user.

David-Engel (commented):
I'm good with the public-facing aspects of this change (public APIs, connection string additions). I had one javadoc-related suggestion and I commented on a few code issues, too.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
lilgreenbird: this will be updated in a subsequent PR

src/main/java/com/microsoft/sqlserver/jdbc/Parameter.java
```diff
@@ -902,6 +896,45 @@ private void setTypeDefinition(DTV dtv) {
...
+                            datatype += "(" + scale + ")";
+                        }
+                        return datatype;
+                    case "datetimeoffset":
+                        datatype = SSType.DATETIMEOFFSET.toString();
+                        if (scale != null){
+                            datatype += "(" + scale + ")";
+                        }
+                        return datatype;
+                    case "datetime":
```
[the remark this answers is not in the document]
lilgreenbird: this will be updated in a subsequent PR

lilgreenbird (approved):

tkyc (approved):

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerLexer.java
lilgreenbird:
```suggestion
        RuntimeMetaData.checkVersion("4.9.3", RuntimeMetaData.VERSION);

```

lilgreenbird (approved):

tkyc (approved):

AlBundy33:
@lilgreenbird is this already available in a release?

Jeffery-Wasty:
Hi @AlBundy33,

This is planned to be included in the 12.2.0 release, scheduled for the end of this month.

PR #1693, opened 2021-11-19
Deprecate and Rename Service principal connection properties

lilgreenbird:
- deprecate AADSecurePrincipalId/AADSecurePrincipalSecret connection properrties and replace with username/password
- remove getAADSecretPrincipalId for security reasons


tkyc:
Just out of the loop, what were the security reasons to deprecate getAADSecretPrincipalId?

v-zhangw (approved):

tkyc (approved):

PR #1696, opened 2021-11-24
Fixed ActiveDirectoryInteractive authentication timeout issue

lilgreenbird:
Use 10 * loginTimeout timeout value which is consistent with SqlClient behaviour

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
@@ -870,6 +870,8 @@ public class SQLServerConnection implements ISQLServerConnection, java.io.Serial
         return sessionRecovery;
     }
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -870,6 +870,8 @@ IdleConnectionResiliency getSessionRecovery() {
         return sessionRecovery;
     }
 
+    static boolean isWindows;
```
v-zhangw, line 873: What is this for?

tkyc (approved):

v-zhangw (approved):

tkyc:
Was just reading the files, and noticed this alot in project. So the convention for null checks is "null == someVar"? And not the other way around? eg. "someVar == null" or "someVar != null"

PR #1699, opened 2021-11-25
Add timeout error for tests

lilgreenbird:
slow Azure servers could result in timeout errors


tkyc:
Looks good, but maybe add a comment somewhere incase we ever forget why we are doing this

tkyc (approved):

PR #1711, opened 2021-12-10
Updated javadoc for SQLServerBulkCSVFileRecord

### LINKED ISSUE #1691: SQLServerBulkCSVFileRecord incorrectly splits line when given delimiters with special regex characters
## Driver version
9.2.1.jre15

## JAVA/JVM version
JDK 15

## Problem description
Expected behaviour: Importing data from a csv file with a delimiter containing a special regex character (for example ("|") should work

I have a csv file where the delimiter is the pipe character (`|`):
```csv
FOO1 | BAR1 | 1
FOO2 | BAR2 | 2
```

When trying to import the data using `SQLServerBulkCSVFileRecord`, import fails because the lines are not split correctly:
```java
        try (
                var connection = DriverManager.getConnection(connectionUrl);
                var bulkCopy = new SQLServerBulkCopy(connection)
        ) {
            # delimiter = |
            var fileRecord = new SQLServerBulkCSVFileRecord(csvSourcePath, null, "|",false);
            ...

            bulkCopy.setDestinationTableName(destinationTable);
        }
```

I looked into `SQLServerBulkCSVFileRecord::getRowData` and this clearly happens because the method `String::split` is used to split the line. However special regex characters are not treated as delimiters if not escaped, so the line is not split correctly.

This can be fixed by one of the following:
1. Implementing special handling for delimiters passed to the constructor of `SQLServerBulkCSVFileRecord`
2. Passing the responsibility to the user, and explicitly state in the javadoc that special regex chars should be escaped.

I think the second approach is better, but I would be happy to provide a pull request for either approach.

v-zhangw:
Java doc updated as per user ask. SQLServerBulkCSVFileRecord now states that regex characters should be escaped.
Fixes #1691

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCSVFileRecord.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCSVFileRecord.java
@@ -75,11 +75,11 @@ public class SQLServerBulkCSVFileRecord extends SQLServerBulkRecord implements j
      * Constructs a simple reader to parse data from a delimited file with the given encoding.
      * 
      * @param fileToParse
-     *        File to parse data from, please ensure special regex characters are properly escaped.
+     *        File to parse data from.
      * @param encoding
      *        Charset encoding to use for reading the file, or NULL for the default encoding.
      * @param delimiter
-     *        Delimiter to used to separate each column
+     *        Delimiter to used to separate each column. Regex characters must be escaped with double backslashes.
      * @param firstLineIsColumnNames
      *        True if the first line of the file should be parsed as column names; false otherwise
      * @throws SQLServerException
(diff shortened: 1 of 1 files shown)


tkyc (approved):

lilgreenbird (approved):

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCSVFileRecord.java
```diff
@@ -75,7 +75,7 @@
      * Constructs a simple reader to parse data from a delimited file with the given encoding.
      * 
      * @param fileToParse
-     *        File to parse data from
+     *        File to parse data from, please ensure special regex characters are properly escaped.
```
David-Engel, line 78: Does this mean the file name can contain regex characters? I'm confused by this change.

v-zhangw: This should be under the delimiter param, will change

tkyc (approved):

lilgreenbird (approved):

PR #1719, opened 2021-12-17
Added new connection string property prepareMethod to toggle use of sp_prepare

tkyc:


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResource.java
```diff
@@ -192,6 +192,7 @@ static String getResource(String key) {
         {"R_failoverPartnerPropertyDescription", "The name of the failover server used in a database mirroring configuration."},
         {"R_packetSizePropertyDescription", "The network packet size used to communicate with SQL Server."},
         {"R_encryptPropertyDescription", "Determines if Secure Sockets Layer (SSL) encryption should be used between the client and the server."},
+        {"R_useSpPrepExecPropertyDescription", "Determines if sp_prepexec should used."},
```
lilgreenbird, line 195: should BE used?

otbutz:
I'd rather take the approach of jTDS as it allows to add different strategies in the future without breaking compatibility.

Jeffery-Wasty (dismissed):

src/main/java/com/microsoft/sqlserver/jdbc/ISQLServerConnection.java
```diff
@@ -278,6 +278,25 @@ CallableStatement prepareCall(String sql, int nType, int nConcur, int nHold,
      */
     void setEnablePrepareOnFirstPreparedStatementCall(boolean value);
 
+    /**
+     * Returns the behavior for a specific connection instance. Returns one of the following:
+     * 1. prepxec, if sp_prepexec is set (default behavior)
```
lilgreenbird, line 283: typo

src/main/java/com/microsoft/sqlserver/jdbc/ISQLServerConnection.java
```diff
@@ -278,6 +278,25 @@ CallableStatement prepareCall(String sql, int nType, int nConcur, int nHold,
...
+     * 2. prepare, if sp_prepare is set
+     *
+     * @return Returns current setting for prepareMethod connection property.
+     */
+    String getPrepareMethod();
+
+    /**
+     * Sets the behavior for the prepare method. Only the following string values are permitted:
+     * 1. prepexec, for use of sp_prepexec (default behavior)
```
lilgreenbird, line 292: I don't think we should list this everywhere if we ever need to add a new option this would need to be updated in so many places. Should just reference a @link to PrepareMethod in SQLServerDriverPropertyInfo

lilgreenbird: also pls fix spelling typo in line 304

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -123,6 +123,8 @@
     /** Current limit for this particular connection. */
     private Boolean enablePrepareOnFirstPreparedStatementCall = null;
```
lilgreenbird, line 124: why does this need to be Boolean?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDriver.java
```diff
@@ -334,10 +335,47 @@ public String toString() {
...
+enum PrepareMethod {
+    PREPEXEC("prepexec"), //sp_prepexec, default prepare method
+    PREPARE("prepare");
+
+    private final String value;
+
+    private static final Map<String, PrepareMethod> PREPARE_METHODS = Map.of(
+            "prepexec", PREPEXEC,
+            "prepare", PREPARE
+    );
```
lilgreenbird, line 347: why is this necessary when we're just mapping to the toString() value?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -1050,36 +1078,62 @@ private boolean reuseCachedHandle(boolean hasNewTypeDefinitions, boolean discard
...
-            // Second execution, use prepared statements since we seem to be re-using it.
-            else if (needsPrepare)
-                buildPrepExecParams(tdsWriter);
-            else
+            } else if (needsPrepare) { // Second execution, use prepared statements since we seem to be re-using it.
+                if (isPrepExecEnabled) { // If true, we're using sp_prepexec.
+                    buildPrepExecParams(tdsWriter);
+                } else { // Otherwise, we're using sp_prepare instead of sp_prepexec.
+                    isSpPrepareExecuted = true;
+                    // If we're preparing for a statement in a batch we just need to call sp_prepare b/c in the
```
lilgreenbird, line 1103: nitpick: can we spell out because

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -1050,36 +1078,62 @@ private boolean reuseCachedHandle(boolean hasNewTypeDefinitions, boolean discard
...
         return needsPrepare;
     }
 
+    /**
+     * Executes sp_prepare to prepare a parameterized statement and sets the prepared statement handle
+     *
+     * @param tdsWriter TDS writer to write sp_prepare params to
+     * @throws SQLServerException
+     */
+    private void doPrep(TDSWriter tdsWriter, TDSCommand command) throws SQLServerException {
```
lilgreenbird, line 1131: I find this a bit confusing. We have a setting isPrepExecEnabled but we always call doPrepExec regardless. We check this flag inside the doPrepExec method and call doPrep. Perhaps it would be more clear to refactor and check flag if prepexec then doPrepExec otherwise doPrep?

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/PreparedStatementTest.java
```diff
@@ -43,8 +45,9 @@
 @RunWith(JUnitPlatform.class)
 public class PreparedStatementTest extends AbstractTest {
 
-    final String tableName = RandomUtil.getIdentifier("#update1");
-    final String tableName2 = RandomUtil.getIdentifier("#update2");
+    final String tableName = RandomUtil.getIdentifier("test1");
+    final String tableName2 = RandomUtil.getIdentifier("test2");
+    final String tableName3 = RandomUtil.getIdentifier("test3");
```
lilgreenbird, line 50: I know this had stupid names before too but can we name them to something more meaningful so we know what the test the table is used for if find them left in the database.

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/PreparedStatementTest.java
```diff
@@ -70,6 +73,70 @@ private int executeSQLReturnFirstInt(SQLServerConnection conn, String sql) throw
...
+        String sql = "insert into " + AbstractSQLGenerator.escapeIdentifier(tableName3)
+                + " (c1_nchar, c2_int) values (?, ?)";
+
+        try (SQLServerConnection con = (SQLServerConnection) getConnection()) {
+            con.setPrepareMethod("prepare"); // Use sp_prepare rather than sp_prepexec
+
+            executeSQL(con, "create table " + AbstractSQLGenerator.escapeIdentifier(tableName3)
+                    + " (c1_nchar nchar(512), c2_int integer)");
+
+            PreparedStatement ps = con.prepareStatement(sql);
```
lilgreenbird, line 97: leak, use try with resources. (multiple occurence)

lilgreenbird:
- can we also add a test that test the property in connection string? 
- we should add this property to updateDataSource in AbstractTest
- should add tests for this property when AE is enabled

lilgreenbird (changes requested):
- can we also add a test that test the property in connection string? (instead of just datasource)
- we should add this property to updateDataSource in AbstractTest as well
- we should also add tests for this property when AE is enabled

v-zhangw (approved):

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
tkyc: So that is related to this feature here: https://github.com/microsoft/mssql-jdbc/wiki/PreparedStatement-metadata-caching

It's kinda confusing because it uses the "prepare" wording. So if that is enabled and depending on the prepareMethod, the driver will call sp_prepexec or sp_prepare  -> sp_execute for the first execution

lilgreenbird (approved):

v-zhangw (approved):

Jeffery-Wasty (approved):

otbutz:
Should this be added to the docs at https://docs.microsoft.com/en-us/sql/connect/jdbc/improving-performance-and-reliability-with-the-jdbc-driver ?

Jeffery-Wasty:
Hi @otbutz,

We're adding this under [Prepared statement metadata caching for the JDBC driver](https://docs.microsoft.com/en-us/sql/connect/jdbc/prepared-statement-metadata-caching-for-the-jdbc-driver?view=sql-server-ver16). The proposed changes are currently under review by the MS Docs team, but expect the docs to be updated soon.

PR #1758, opened 2022-03-01
Added a check to prevent Implicit narrowing conversion in compound assignment

Jeffery-Wasty:
(Update 3/2/2022) While the below approach works, given that the use of `readSkipBytes` has changed, its no longer necessary to pass in a `long`. Thus the changes can be reduced to ensuring all involved types are `int`, avoiding any information loss.

-----------------------------------------------------------------------------------------------------------------------------------------

As seen here: https://github.com/microsoft/mssql-jdbc/security/code-scanning/1?query=ref%3Arefs%2Fheads%2Fmain, there is an implicit cast taking place in IOBuffer.java, that may lead to a loss of information. The solution was to first check if the long value `bytesToSkip` can be converted without information loss, and only allow addition to `payloadOffset` if this is true.

The reverse, that is widening `payloadOffset` to long, to accommodate `bytesToSkip` is not possible because of the many uses `payloadOffset` has throughout the file. Lines 7043 and 7049 in `IOBuffer.java` are good examples of this. Extensive reformatting is needed to change `payloadOffset`, while `bytesToSkip` can be checked with ease.

The function used to check if `bytesToSkip` is small enough, `isSafeToConvert`, uses a switch function even though it only has one case, this is to allow future cases to be added, if needed. Additionally, other values, besides long, can be passed in to convert, provided they are cast beforehand.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
@@ -1076,7 +1076,7 @@ final class TDSChannel implements Serializable {
 
         /**
          * Poll the stream to verify connectivity.
-         * 
+         *
          * @return true if the stream is readable.
          * @throws IOException
          *         If an I/O exception occurs.
@@ -1815,7 +1815,7 @@ final class TDSChannel implements Serializable {
 
     /**
      * Enables SSL Handshake.
-     * 
+     *
      * @param host
      *        Server Host Name for SSL Handshake
      * @param port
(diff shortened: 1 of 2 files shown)


tkyc:
I might be missing something here, but isn't an alternative like changing the "valueLength" param to int and then keeping the rest of the types in the method to int possible also? I only see the "readSkipBytes" method used in one place and it is taking in an int only, never a long

Jeffery-Wasty:
> 

You know what, I think you're absolutely right. I'm looking at the initial implementation, and it may have been necessary to pass in a long in that case, but that's no longer needed. This simplifies things greatly.

tkyc (approved):

lilgreenbird (approved):

v-zhangw (approved):

PR #1766, opened 2022-03-10
Add Configurable IPv6 Support

v-zhangw:
A new connection property IPAddressPreference is added to specify the IP preference by the client application when doing TCP connection. There are three valid types:
• IPv4First
This is the default preference. The driver will traverse IPv4 addresses first. If none of them can be connected successfully, it continues to try IPv6 addresses if there are any.
• IPv6First
The driver will traverse IPv6 addresses first. If none of them can be connected successfully, it continues to try IPv4 addresses if there are any.
• UsePlatformDefault
The driver will traverse all the IP addresses in their initial orders from DNS resolution.


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDriver.java
```diff
@@ -601,6 +602,8 @@ public String toString() {
                     SQLServerDriverStringProperty.DOMAIN.getDefaultValue(), false, null),
             new SQLServerDriverPropertyInfo(SQLServerDriverStringProperty.SERVER_NAME.toString(),
                     SQLServerDriverStringProperty.SERVER_NAME.getDefaultValue(), false, null),
+            new SQLServerDriverPropertyInfo(SQLServerDriverStringProperty.IPADDRESSPREFERENCE.toString(),
+                    SQLServerDriverStringProperty.IPADDRESSPREFERENCE.getDefaultValue(), false, null),
```
tkyc, line 606: The last param shouldn't be null right? It should be a string array containing the permissible IPAddressPreferences?

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -2609,16 +2617,20 @@ Socket findSocket(String hostName, int portNumber, int timeoutInMilliSeconds, bo
...
+                    return getSocketByIPvPreference(hostName, portNumber, timeoutInMilliSeconds, IPAddressPreference);
+                    // return getDefaultSocket(hostName, portNumber, timeoutInMilliSeconds);
                 }
             }
 
             // inetAddrs is only used if useParallel is true or TNIR is true. Skip resolving address if that's not the
             // case.
             if (useParallel || useTnir) {
                 // Ignore TNIR if host resolves to more than 64 IPs. Make sure we are using original timeout for this.
+                // coffee add new function here
```
Jeffery-Wasty, line 2633: What's the point of this comment?

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -749,17 +752,20 @@ void resetPooledConnection() {
 
     /**
      * Opens the physical communications channel (TCP/IP socket and I/O streams) to the SQL Server.
+     * 
+     * @param IPAddressPreference
```
Jeffery-Wasty, line 756: Starting on line 2922 of this file, you add descriptions for your params, but not here. It should be one or the other (I would assume we would prefer to have descriptions for all params).

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -749,17 +752,20 @@ void resetPooledConnection() {
...
-            boolean isTnirFirstAttempt, int timeoutMillisForFullTimeout) throws SQLServerException {
+            boolean isTnirFirstAttempt, int timeoutMillisForFullTimeout,
+            String IPAddressPreference) throws SQLServerException {
         if (logger.isLoggable(Level.FINER))
             logger.finer(this.toString() + ": Opening TCP socket...");
 
         SocketFinder socketFinder = new SocketFinder(traceID, con);
         channelSocket = tcpSocket = socketFinder.findSocket(host, port, timeoutMillis, useParallel, useTnir,
-                isTnirFirstAttempt, timeoutMillisForFullTimeout);
+                isTnirFirstAttempt, timeoutMillisForFullTimeout, IPAddressPreference);
```
lilgreenbird, line 768: coding standard, vars don't start with caps

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -2904,19 +2916,71 @@ private SocketFactory getSocketFactory() throws IOException {
...
+     * @throws IOException
+     * @throws SQLServerException
+     */
+    private Socket getSocketByIPvPreference(String hostName, int portNumber, int timeoutInMilliSeconds,
+            String IPAddressPreference) throws IOException, SQLServerException {
+        InetSocketAddress addr = null;
+        InetAddress addresses[] = InetAddress.getAllByName(hostName);
+        Queue<InetAddress> addrq = null;
+
+        if (IPAddressPreference.equals("IPv6First")) {
```
lilgreenbird, line 2940: don't hardcode, these should be added to SQLServerDriver

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -2925,9 +2989,30 @@ private Socket getDefaultSocket(String hostName, int portNumber, int timeoutInMi
             InetSocketAddress cacheEntry = SQLServerConnection.getDNSEntry(hostName);
             addr = (null != cacheEntry) ? cacheEntry : addr;
         }
+
         return getConnectedSocket(addr, timeoutInMilliSeconds);
     }
 
+    private Queue<InetAddress> getIPv6Addresses(InetAddress[] addresses) {
```
lilgreenbird, line 2996: the names seem a bit misleading, this is actually returning an ipv6AdressQueue

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResource.java
```diff
@@ -502,6 +503,7 @@ static String getResource(String key) {
         {"R_crClientUnrecoverable","The connection is broken and recovery is not possible. The connection is marked by the client driver as unrecoverable. No attempt was made to restore the connection."},
         {"R_crClientSSLStateNotRecoverable", "The server did not preserve SSL encryption during a recovery attempt, connection recovery is not possible."},
         {"R_crCommandCannotTimeOut", "Request failed to time out and SQLServerConnection does not exist"},
+        {"R_InvalidIPAddressPreference", "Invalid option for connection string option IPAddressPreference, the available options are: IPv4First, IPv6First and UsePlatformDefault"},
```
lilgreenbird, line 506: add {0} on what the invalid option was, makes it easier to debug failures. Also, we don't usually list all the valid options here in the error string, in case we need to add/change anything

src/test/java/com/microsoft/sqlserver/jdbc/connection/ConnectionTest.java
```diff
@@ -43,4 +43,32 @@ public void testConnections() throws SQLException {
...
         }
     }
+    
+    @Test
+    public void testConnectWithIPAddressPreference () throws SQLException {
+        SQLServerDataSource ds = new SQLServerDataSource();
+        ds.setURL(connectionString);
+        ds.setIPAddressPreference("IPv4First");
+        
+        try (Connection con = ds.getConnection()) {
```
lilgreenbird, line 53: [nitpick] there seems to be a lot of unnecessary/random spacing here can we group the tests together so it's easier to read?

src/test/java/com/microsoft/sqlserver/jdbc/connection/ConnectionTest.java
```diff
@@ -43,4 +43,32 @@ public void testConnections() throws SQLException {
...
+        }
+        
+        ds.setIPAddressPreference("UsePlatformDefault");
+        
+        try (Connection con = ds.getConnection()) {
+        }
+        
+        ds.setIPAddressPreference("Bogus");
+
```
lilgreenbird, line 67: there should be a R_expectedFailPassed here

src/test/java/com/microsoft/sqlserver/jdbc/connection/ConnectionTest.java
```diff
@@ -43,4 +43,32 @@ public void testConnections() throws SQLException {
...
+        
+        try (Connection con = ds.getConnection()) {
+        }
+        
+        ds.setIPAddressPreference("Bogus");
+        
+        try (Connection con = ds.getConnection()) {
+        }
+        catch (Exception e){
+            assert e.getMessage().equals("Invalid option for connection string option IPAddressPreference, the available options are: IPv4First, IPv6First and UsePlatformDefault");
```
lilgreenbird, line 71: we should be comparing against the string in the driver, use TestResource.getResource. Check other tests to see how this is done

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDriver.java
```diff
@@ -378,6 +378,7 @@ static PrepareMethod valueOfString(String value) throws SQLServerException {
     SELECT_METHOD("selectMethod", "direct"),
     DOMAIN("domain", ""),
     SERVER_NAME("serverName", ""),
+    IPADDRESSPREFERENCE("IPAddressPreference", "IPv4First"),
```
lilgreenbird, line 381: [nitpick] IPADDRESS_PREFERENCE seems a bit easier to read?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDataSource.java
```diff
@@ -577,6 +577,14 @@ public void setServerName(String serverName) {
     public String getServerName() {
         return getStringProperty(connectionProps, SQLServerDriverStringProperty.SERVER_NAME.toString(), null);
     }
+    
+    public void setIPAddressPreference(String IPAddressPreference) {
```
lilgreenbird, line 581: javadocs!! same below

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -2904,19 +2916,71 @@ private SocketFactory getSocketFactory() throws IOException {
...
+    private Socket getSocketByIPvPreference(String hostName, int portNumber, int timeoutInMilliSeconds,
+            String IPAddressPreference) throws IOException, SQLServerException {
+        InetSocketAddress addr = null;
+        InetAddress addresses[] = InetAddress.getAllByName(hostName);
+        Queue<InetAddress> addrq = null;
+
+        if (IPAddressPreference.equals("IPv6First")) {
+            // Try to connect to IPv6 Addresses
+            addrq = getIPv6Addresses(addresses);
+            while (addrq.peek() != null) {
```
lilgreenbird, line 2943: we should do some fine or finest level logging here

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -2904,19 +2916,71 @@ private SocketFactory getSocketFactory() throws IOException {
...
+                addr = new InetSocketAddress(addrq.poll(), portNumber);
+                if (!addr.isUnresolved())
+                    return getConnectedSocket(addr, timeoutInMilliSeconds);
+            }
+            // No IPv6 Addresses can be connected to, try IPv4 addresses now
+            addrq = getIPv4Addresses(addresses);
+            while (addrq.peek() != null) {
```
lilgreenbird, line 2950: it seems these are all doing the same thing, can we refactor and move to the helper method instead?

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -2904,19 +2916,71 @@ private SocketFactory getSocketFactory() throws IOException {
...
+     *        Port number
+     * @param timeoutInMilliSeconds
+     *        Timeout
+     * @param IPAddressPreference
+     *        IPv4First (default), IPv6First or UsePlatformDefault
+     * @return Connected Socket
+     * @throws IOException
+     * @throws SQLServerException
+     */
+    private Socket getSocketByIPvPreference(String hostName, int portNumber, int timeoutInMilliSeconds,
```
lilgreenbird, line 2934: why do we need to change the name of this method just because 1 param has been added? This method gets the default (ie not custom) socket by hostname, portnum timeout and ipPreference. I think it's better to keep the orig name as it's more descriptive

v-zhangw: I think the behavior has changed enough that "default" is not descriptive enough. getDefaultSocket only ever did one thing regardless of inputs while this function will change order it gets prioritizes sockets based on parameters.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDataSource.java
v-zhangw: Added.

src/test/java/com/microsoft/sqlserver/jdbc/connection/ConnectionTest.java
v-zhangw: Changes made as asked.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDriver.java
v-zhangw: Agreed, changes made.

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
v-zhangw: Agreed, changes made.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResource.java
v-zhangw: Agreed, changes made.

src/test/java/com/microsoft/sqlserver/jdbc/connection/ConnectionTest.java
v-zhangw: Agreed, changes made.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDriver.java
v-zhangw: Agreed, changes made.

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
v-zhangw: I think that can potentially result in a large amount of logging for systems with many IP addresses resolving. The cases of no ip addresses resolving will be caught by the log entry below.

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
v-zhangw: Agreed, changes made.

src/test/java/com/microsoft/sqlserver/jdbc/connection/ConnectionTest.java
v-zhangw: Agreed, changes made.

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
v-zhangw: Agreed, changes made.

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -2609,16 +2617,20 @@ Socket findSocket(String hostName, int portNumber, int timeoutInMilliSeconds, bo
                 // MSF is false. TNIR could be true or false. DBMirroring could be true or false.
                 // For TNIR first attempt, we should do existing behavior including how host name is resolved.
                 if (useTnir && isTnirFirstAttempt) {
-                    return getDefaultSocket(hostName, portNumber, SQLServerConnection.TnirFirstAttemptTimeoutMs);
+                    return getSocketByIPvPreference(hostName, portNumber, SQLServerConnection.TnirFirstAttemptTimeoutMs,
+                            IPAddressPreference);
+                    // return getDefaultSocket(hostName, portNumber, SQLServerConnection.TnirFirstAttemptTimeoutMs);
```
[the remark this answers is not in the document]
v-zhangw: Old commented out code, removed.

tkyc (approved):

Jeffery-Wasty (approved):

tkyc (dismissed):

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResource.java
```diff
@@ -502,6 +503,7 @@ static String getResource(String key) {
         {"R_crClientUnrecoverable","The connection is broken and recovery is not possible. The connection is marked by the client driver as unrecoverable. No attempt was made to restore the connection."},
         {"R_crClientSSLStateNotRecoverable", "The server did not preserve SSL encryption during a recovery attempt, connection recovery is not possible."},
         {"R_crCommandCannotTimeOut", "Request failed to time out and SQLServerConnection does not exist"},
+        {"R_InvalidIPAddressPreference", "Invalid option for connection string option IPAddressPreference: {0}"},
```
lilgreenbird, line 506: sentence is awkward...can you check the other strings for consistency. Usually they are something like invalid <blah> specified: {0}

src/test/java/com/microsoft/sqlserver/jdbc/TestResource.java
```diff
@@ -200,5 +200,6 @@ public static String formatErrorMsg(String resource) {
             {"R_AKVProviderNull", "The Azure key store provider is null."},
             {"R_objectNullOrEmpty", "The {0} is null or empty."},
             {"R_cekDecryptionFailed", "Failed to decrypt a column encryption key using key store provider: {0}."},
-            {"R_connectTimedOut", "connect timed out"}};
+            {"R_connectTimedOut", "connect timed out"},
+            {"R_InvalidIPAddressPreference", "Invalid option for connection string option IPAddressPreference: {0}"}};
```
lilgreenbird, line 204: why is this necessary?

src/test/java/com/microsoft/sqlserver/jdbc/connection/ConnectionTest.java
```diff
@@ -42,4 +46,28 @@ public void testConnections() throws SQLException {
...
+        ds.setIPAddressPreference("UsePlatformDefault");
+        try (Connection con = ds.getConnection()) {
+        }
+        ds.setIPAddressPreference("Bogus");
+        try (Connection con = ds.getConnection()) {
+            fail(TestResource.getResource("R_expectedFailPassed"));
+        }
+        catch (Exception e){
+            MessageFormat form = new MessageFormat(TestResource.getResource("R_InvalidIPAddressPreference"));
+            Object[] msgArgs1 = {"Bogus"};
```
lilgreenbird, line 69: why? check other tests how comparisons are done

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -3214,10 +3223,11 @@ private InetSocketAddress connectHelper(ServerPortPlaceHolder serverInfo, int ti
 
         // if the timeout is infinite slices are infinite too.
         tdsChannel = new TDSChannel(this);
+        String IPAddressPreference = activeConnectionProperties.getProperty(SQLServerDriverStringProperty.IPADDRESS_PREFERENCE.toString());
 
         InetSocketAddress inetSocketAddress = tdsChannel.open(serverInfo.getParsedServerName(),
                 serverInfo.getPortNumber(), (0 == timeOutFullInSeconds) ? 0 : timeOutSliceInMillis, useParallel,
-                useTnir, isTnirFirstAttempt, timeOutsliceInMillisForFullTimeout);
+                useTnir, isTnirFirstAttempt, timeOutsliceInMillisForFullTimeout, IPAddressPreference);
```
lilgreenbird, line 3230: prev comment re lower case vars (upper case for classes only) not addressed, please go thru all

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDriver.java
```diff
@@ -257,6 +257,43 @@ static SSLProtocol valueOfString(String value) throws SQLServerException {
 }
 
 
+enum IPvAddressPreferenceEnum {
```
lilgreenbird, line 260: remove "Enum", please see how others are named and and be consistent

lilgreenbird (changes requested):
- in planning this property is referred to as "IPAddressPreference" but there is inconsistent use of "IPvAddressPreference" all over the place (both in enum and vars". Please find out what this should be and be consistent
- please go thru all changes to adhere to coding standard - var names should start with lower case only Class names start with cap

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -2907,19 +2917,86 @@ private SocketFactory getSocketFactory() throws IOException {
...
+                if (!addr.isUnresolved())
+                    return getConnectedSocket(addr, timeoutInMilliSeconds);
+                break;
+            case UsePlatformDefault:
+                for (InetAddress address : addresses) {
+                    addr = new InetSocketAddress(address, portNumber);
+                    if (!addr.isUnresolved())
+                        return getConnectedSocket(addr, timeoutInMilliSeconds);
+                }
+                break;
```
lilgreenbird, line 2994: missing default

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -2907,19 +2917,86 @@ private SocketFactory getSocketFactory() throws IOException {
...
+     * Helper function which traverses through queue of InetAddresses to find a resolved one
+     * 
+     * @param addrq
+     *        Address queue
+     * @param portNumber
+     *        Port Number
+     * @return First resolved queue or unresolved queue if none found
+     * @throws IOException
+     */
+    private InetSocketAddress traverseAddressQueue(Queue<InetAddress> addrq, int portNumber) throws IOException {
```
lilgreenbird, line 2930: this gets address from Q right? so perhaps getIpAdressFromQueue is more descriptive?

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -2907,19 +2917,86 @@ private SocketFactory getSocketFactory() throws IOException {
...
+        InetSocketAddress addr = null;
+        InetAddress addresses[] = InetAddress.getAllByName(hostName);
+        Queue<InetAddress> addrq = null;
+
+        IPvAddressPreferenceEnum IPvPref = IPvAddressPreferenceEnum.valueOfString(IPAddressPreference);
+
+        switch (IPvPref) {
+            case IPv6First:
+                // Try to connect to IPv6 Addresses
+                addrq = getIPv6AddressQueue(addresses);
```
lilgreenbird, line 2966: this local var seems unnecessary

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -2907,19 +2917,86 @@ private SocketFactory getSocketFactory() throws IOException {
...
+                        return getConnectedSocket(addr, timeoutInMilliSeconds);
+                }
+                break;
+        }
+
         // Note that Socket(host, port) throws an UnknownHostException if the host name
         // cannot be resolved, but that InetSocketAddress(host, port) does not - it sets
         // the returned InetSocketAddress as unresolved.
-        InetSocketAddress addr = new InetSocketAddress(hostName, portNumber);
         if (addr.isUnresolved()) {
```
lilgreenbird, line 3000: addr can be null here

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -2907,19 +2917,86 @@ private SocketFactory getSocketFactory() throws IOException {
...
+
+        switch (IPvPref) {
+            case IPv6First:
+                // Try to connect to IPv6 Addresses
+                addrq = getIPv6AddressQueue(addresses);
+                addr = traverseAddressQueue(addrq, portNumber);
+                if (addr.isUnresolved())
+                    return getConnectedSocket(addr, timeoutInMilliSeconds);
+                // No IPv6 Addresses can be connected to, try IPv4 addresses now
+                addrq = getIPv4AddressQueue(addresses);
```
lilgreenbird, line 2971: prev comment was not addressed - 
these are all doing the same thing, why do we need to get q, traverse q all over the place? can be refactored to just a getIpAddress which checks preference option and gets the addr from the whatever q.

src/test/java/com/microsoft/sqlserver/jdbc/connection/ConnectionTest.java
```diff
@@ -42,4 +46,28 @@ public void testConnections() throws SQLException {
...
+        }
+        ds.setIPAddressPreference("UsePlatformDefault");
+        try (Connection con = ds.getConnection()) {
+        }
+        ds.setIPAddressPreference("Bogus");
+        try (Connection con = ds.getConnection()) {
+            fail(TestResource.getResource("R_expectedFailPassed"));
+        }
+        catch (Exception e){
+            MessageFormat form = new MessageFormat(TestResource.getResource("R_InvalidIPAddressPreference"));
```
lilgreenbird, line 68: this error is not correct and totally hides the real error. Please check how other tests do this

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
v-zhangw: Yes. Changed.

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
v-zhangw: Added

src/test/java/com/microsoft/sqlserver/jdbc/connection/ConnectionTest.java
v-zhangw: Added earlier check

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
v-zhangw: Should no longer be any IP'v' s anywhere

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResource.java
v-zhangw: Changed to be similar to other outputs.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDriver.java
v-zhangw: Changed.

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -749,17 +752,21 @@ void resetPooledConnection() {
 
     /**
      * Opens the physical communications channel (TCP/IP socket and I/O streams) to the SQL Server.
+     * 
+     * @param iPAddressPreference
+     *        Preferred type of IP address to use first; valid options are IPv4First, IPv6First and UsePlatformDefault
```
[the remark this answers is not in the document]
v-zhangw: Changed.

src/test/java/com/microsoft/sqlserver/jdbc/TestResource.java
v-zhangw: To check for correct error message in fail case.

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
v-zhangw: We could refactor all the functionality to exist in an new overload of getConnectedSocket, but I think decomposing the logic into smaller functions is better than having one big 100 line function.

tkyc:
Hmm, for filtering of the addresses to either ipv6 or ipv4, possibly do it in one pass using 2 member lists instead of local lists? If it makes sense to do so with the implementation. Does the address list change often?

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -2928,8 +3018,29 @@ private Socket getDefaultSocket(String hostName, int portNumber, int timeoutInMi
             InetSocketAddress cacheEntry = SQLServerConnection.getDNSEntry(hostName);
             addr = (null != cacheEntry) ? cacheEntry : addr;
         }
+
         return getConnectedSocket(addr, timeoutInMilliSeconds);
     }
+    
+    private ArrayList<InetAddress> getIPv6AddressList(InetAddress[] addresses) {
```
tkyc, line 3025: Just to clarify, instead of using these getIPvXAddressList methods and creating new arraylist objects everytime, can't we just do one pass through of the InetAddresses and filter them once using member/instance arraylists?

For example:

private ArrayList IPv6AddressList;
private ArrayList IPv4AddressList;

Then in some appropriate place do the filtering like this:
       for (InetAddress addr : addresses) {
            if (addr instanceof Inet6Address) {
                IPv6AddressList.add(addr);
            } else {
                IPv4AddressList.add(addr);
        }

src/test/java/com/microsoft/sqlserver/jdbc/connection/ConnectionTest.java
v-zhangw: I've split them up and used the recommended check now.

src/test/java/com/microsoft/sqlserver/jdbc/TestResource.java
v-zhangw: Removed and changed as recommended.

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
v-zhangw: This makes sense. Will adjust.

tkyc (approved):

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
v-zhangw: Changed as recommended by Terry

tkyc (approved):

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDriver.java
```diff
@@ -257,6 +257,43 @@ static SSLProtocol valueOfString(String value) throws SQLServerException {
 }
 
 
+enum IPAddressPreference {
+    IPv4First("IPv4First"),
+    IPv6First("IPv6First"),
+    UsePlatformDefault("UsePlatformDefault"),;
```
Jeffery-Wasty, line 263: The comma before the semi-colon is unnecessary.

Jeffery-Wasty (approved):

lilgreenbird (dismissed):

Jeffery-Wasty (dismissed):

tkyc (dismissed):

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDriver.java
v-zhangw: removed

Jeffery-Wasty (approved):

tkyc (approved):

PR #1768, opened 2022-03-10
Ensure that batchParamValues is cleared in all cases when using bulk update mode

### LINKED ISSUE #1767: Batch not cleared on failure when using bulk update
## Driver version
All (tested with 9.2.1 but same code is present in latest `main` branch).

## Problem description
The normal behaviour of `executeBatch` is to clear the pending batch of operations, regardless of whether the batch execution completes successfully, succeeds for some rows and fails for others, or fails entirely with an exception:

https://github.com/microsoft/mssql-jdbc/blob/d9a07bd748317bf82b9b0a734ddc623407400268/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java#L2174-L2176

(the code to reset `batchParamValues` to `null` is in a `finally` block and fires in all cases, even when an exception is thrown).

However, in the case where the bulk update API is used, `batchParamValues` is _not_ cleared in the case of an exception:

https://github.com/microsoft/mssql-jdbc/blob/d9a07bd748317bf82b9b0a734ddc623407400268/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java#L2113-L2122

(`batchParamValues = null` is not in a `finally` so does not run if any of the `bcOperation` methods were to throw an exception).  This means that subsequent `addBatch` calls following a failure will _append_ to the existing batch rather than starting a new one, and the next `executeBatch` will re-submit the previous batch of rows along with the new data.

## Expected behavior
The logic should be the same in all cases - `executeBatch` should _always_ clear the batch state so subsequent `addBatch` calls begin a new batch.

## Actual behavior
As described above - in the bulk update case the batch is not cleared on failure.

Since `clearBatch` is idempotent, a workaround would be for user code to do an explicit `stmt.clearBatch()` call (in a `finally`) after every call to `executeBatch()`.

ianroberts:
Closes #1767

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
@@ -2103,23 +2103,26 @@ public class SQLServerPreparedStatement extends SQLServerStatement implements IS
                         batchRecord.addColumnMetadata(i, c.getColumnName(), jdbctype, ti.getPrecision(), ti.getScale());
                     }
 
-                    SQLServerBulkCopy bcOperation = new SQLServerBulkCopy(connection);
-                    SQLServerBulkCopyOptions option = new SQLServerBulkCopyOptions();
-                    option.setBulkCopyTimeout(queryTimeout);
-                    bcOperation.setBulkCopyOptions(option);
-                    bcOperation.setDestinationTableName(tableName);
-                    bcOperation.setStmtColumnEncriptionSetting(this.getStmtColumnEncriptionSetting());
-                    bcOperation.setDestinationTableMetadata(rs);
-                    bcOperation.writeToServer(batchRecord);
-                    bcOperation.close();
-                    updateCounts = new int[batchParamValues.size()];
-                    for (int i = 0; i < batchParamValues.size(); ++i) {
-                        updateCounts[i] = 1;
-                    }
+                    try {
+                        SQLServerBulkCopy bcOperation = new SQLServerBulkCopy(connection);
+                        SQLServerBulkCopyOptions option = new SQLServerBulkCopyOptions();
+                        option.setBulkCopyTimeout(queryTimeout);
+                        bcOperation.setBulkCopyOptions(option);
+                        bcOperation.setDestinationTableName(tableName);
+                        bcOperation.setStmtColumnEncriptionSetting(this.getStmtColumnEncriptionSetting());
+                        bcOperation.setDestinationTableMetadata(rs);
+                        bcOperation.writeToServer(batchRecord);
+                        bcOperation.close();
+                        updateCounts = new int[batchParamValues.size()];
+                        for (int i = 0; i < batchParamValues.size(); ++i) {
+                            updateCounts[i] = 1;
+                        }
 
-                    batchParamValues = null;
-                    loggerExternal.exiting(getClassNameLogging(), "executeBatch", updateCounts);
-                    return updateCounts;
+                        loggerExternal.exiting(getClassNameLogging(), "executeBatch", updateCounts);
+                        return updateCounts;
+                    } finally {
+                        batchParamValues = null;
+                    }
                 }
             }
         } catch (SQLException e) {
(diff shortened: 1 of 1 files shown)


Jeffery-Wasty:
Hi @ianroberts,

Thank you for the report and for the potential fix in the pull request. We'll take a look at both and get back to you as soon as we can.

Jeffery-Wasty (approved):

Jeffery-Wasty (changes requested):
@ianroberts We'll be taking a closer look at this at a later date.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -2103,23 +2103,26 @@ public int[] executeBatch() throws SQLServerException, BatchUpdateException, SQL
...
-                    bcOperation.setDestinationTableName(tableName);
-                    bcOperation.setStmtColumnEncriptionSetting(this.getStmtColumnEncriptionSetting());
-                    bcOperation.setDestinationTableMetadata(rs);
-                    bcOperation.writeToServer(batchRecord);
-                    bcOperation.close();
-                    updateCounts = new int[batchParamValues.size()];
-                    for (int i = 0; i < batchParamValues.size(); ++i) {
-                        updateCounts[i] = 1;
-                    }
+                    try {
```
Jeffery-Wasty, line 2106: Simply adding `batchParamValues = null` to a finally clause should be enough, no need for an additional `try` block.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -2103,23 +2103,26 @@ public int[] executeBatch() throws SQLServerException, BatchUpdateException, SQL
...
+                        for (int i = 0; i < batchParamValues.size(); ++i) {
+                            updateCounts[i] = 1;
+                        }
 
-                    batchParamValues = null;
-                    loggerExternal.exiting(getClassNameLogging(), "executeBatch", updateCounts);
-                    return updateCounts;
+                        loggerExternal.exiting(getClassNameLogging(), "executeBatch", updateCounts);
+                        return updateCounts;
+                    } finally {
```
Jeffery-Wasty, line 2123: This can be moved to 2133, `batchParamValues = null` should be executed in all cases so it makes sense for it to be in a `finally` block belonging to the initial `try` block.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -2260,23 +2263,26 @@ public long[] executeLargeBatch() throws SQLServerException, BatchUpdateExceptio
...
-                    bcOperation.setDestinationTableName(tableName);
-                    bcOperation.setStmtColumnEncriptionSetting(this.getStmtColumnEncriptionSetting());
-                    bcOperation.setDestinationTableMetadata(rs);
-                    bcOperation.writeToServer(batchRecord);
-                    bcOperation.close();
-                    updateCounts = new long[batchParamValues.size()];
-                    for (int i = 0; i < batchParamValues.size(); ++i) {
-                        updateCounts[i] = 1;
-                    }
+                    try {
```
Jeffery-Wasty, line 2266: Same comments as above.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -2260,23 +2263,26 @@ public long[] executeLargeBatch() throws SQLServerException, BatchUpdateExceptio
...
+                        for (int i = 0; i < batchParamValues.size(); ++i) {
+                            updateCounts[i] = 1;
+                        }
 
-                    batchParamValues = null;
-                    loggerExternal.exiting(getClassNameLogging(), "executeLargeBatch", updateCounts);
-                    return updateCounts;
+                        loggerExternal.exiting(getClassNameLogging(), "executeLargeBatch", updateCounts);
+                        return updateCounts;
+                    } finally {
```
Jeffery-Wasty, line 2283: Same as above.

Jeffery-Wasty (changes requested):

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
ianroberts: Alternative approach could be to remove both the additional inner `try` I've added here _and_ the one at line 2296 in the "old style batch insert" branch, and instead wrap the whole method body in an outer try/finally that ensures `batchParamValues` is cleared in all cases immediately before the method returns.  That might be clearer overall.

ianroberts: I've made a separate PR #1869 for the alternative approach, which is also rebased against the latest main.

Jeffery-Wasty:
Let's just keep all changes in one PR, [#1869 ](https://github.com/microsoft/mssql-jdbc/pull/1869), even if they're 'alternate scenarios' we can just choose which commit we'd like to merge.

Issue #1777, opened 2022-03-26
Insertion of NULLs into GUID column in bulk mode


andreitokar:
## Driver version
9.4.1, 10.2.0

## SQL Server version
Microsoft SQL Server 2016 (SP2-CU15) (KB4577775) - 13.0.5850.14 (x64)  
Sep 17 2020 22:12:45
Copyright (C) 2019 Microsoft Corporation
Enterprise Edition (64-bit) on Windows Server 2016 Standard 10.0 <X64> (Build 14393: )

## Client Operating System
Windows 10

## JAVA/JVM version
1.8.0

## Table schema
CTEATE TABLE test (u_id UNIQUEIDENTIFIER);
INSERT INTO test(u_id) VALUES (?);

## Problem description
Unable to insert NULL value into UNIQUEIDENTIFIER column in bulk insert mode (;useBulkCopyForBatchInsert=true), neither with setString(1,null), nor with setNull(1, VARCHAR).

## Expected behavior
record inserted into table

## Actual behavior
Exception is thrown

## Error message/stack trace
Data type guid is not supported in bulk copy.

v-zhangw:
Hi @andreitokar,

Thanks for reaching out, we will be taking a look shortly.

v-zhangw:
Hello @andreitokar,

Could you provide some more context on what you are trying to do?
Are you trying to insert a null value with bulkcopy or inserting a null value via some other method while useBulkCopyForBatchInsert=true is set?

It would help a lot to see the java stack trace with the JDBC exception and a small java app reproducing the problem.

andreitokar:
Hi @v-zhangw. Yes, i was trying to insert a bunch of records by using PreparedStatement INSERT in a batch mode. Target table has a column of type UNIQUEIDENTIFIER and some rows should have NULLs in this column (pls. see "TABLE SCHEMA" section above for example of table definition and INSERT statement). To improve performance, I would like to use bulkCopy mode, but if any of the values for ? parameter are NULL, batch execution fails with exception listed in "Error Message" section.

Due to an obvious omission, exception is thrown [here](https://github.com/microsoft/mssql-jdbc/blob/82a1c5dae01cecb8cbaab3e4ef5a7420d44f1981/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java#L2048)   
[...]

PR #1778, opened 2022-03-26
bulk mode insert with NULL GUID value (#1777)

andreitokar:
Fix for #1777

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
@@ -2002,6 +2002,7 @@ public class SQLServerBulkCopy implements java.lang.AutoCloseable, java.io.Seria
             boolean isStreaming) throws SQLServerException {
 
         switch (srcJdbcType) {
+            case microsoft.sql.Types.GUID:
             case java.sql.Types.CHAR:
             case java.sql.Types.NCHAR:
             case java.sql.Types.VARCHAR:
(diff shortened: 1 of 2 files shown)


v-zhangw (approved):

Jeffery-Wasty (dismissed):

tkyc (dismissed):

lilgreenbird (approved):

lilgreenbird (changes requested):
hi @andreitokar
thank you for the submission however could you please add a test for this?

andreitokar:
@lilgreenbird,
The test would be a trivial addition of data to some CSV file, i.e. BulkCopyCSVTestInput.csv, but unfortunately existing testing framework does not support GUID type, and I do not fill comfortable to add such support due to a lack of knowledge about various aspects of SqlType sub-classing (coercions etc).

v-zhangw:
Hi @andreitokar 
An easier way of adding this test would be to add it to BatchExecutionWithBulkCopy.java
Either as an addition to testNullOrEmptyColumns or as it's own test case.

v-zhangw:
Something like 
`    @Test
    public void testNullGuid () throws Exception {
        String valid = "insert into " + AbstractSQLGenerator.escapeIdentifier(tableName) + " (c1) values (?)";
        try (Connection connection = PrepUtil.getConnection(connectionString + ";useBulkCopyForBatchInsert=true;");
                SQLServerPreparedStatement pstmt = (SQLServerPreparedStatement) connection.prepareStatement(valid);
                Statement stmt = (SQLServerStatement) connection.createStatement();) {
        pstmt.setNull(1, microsoft.sql.Types.GUID);
        pstmt.executeBatch();
        }
    }`

andreitokar:
Thank you @v-zhangw . Of course, I could have done something like that, but just did not want to blow test code needlessly, since test for a dozen of other types is already in place, so I though, that the right way would be to include GUID support into the testing framework. But either way, it is a test.

Jeffery-Wasty (approved):

tkyc (approved):

tkyc (approved):

v-zhangw (approved):

PR #1779, opened 2022-03-28
Add option for NONE attestation protocol

Jeffery-Wasty:
Similar to what Microsoft.Data.SqlClient added in their 4.1 release, we need to support no attestation in the JDBC driver. This reduces security but allows users to use VBS secure enclaves in environments where attestation services are unavailable. The driver should consume the information from the server about the secure enclave but not use the attestation service to validate it. 

Files changed:

- *New* **SQLNoneEnclaveProvider.java** - New filed added for the NONE protcol option. Based off of SQLAASEnclaveProvider.java with token/URL validation removed.
- **SQLServerDriver.java** - Added NONE option to protocol options.
- **SQLServerConnection.java** - Pointed the driver towards SQLNoneEnclaveProvider in the case of NONE attestation option. As well, added checks for protocol type:
    - A missing attestation URL is only a problem if the protocol is not NONE.
    - A protocol of NONE must use secure enclaves.
- **EnclavePackageTest.java** - Pointed NONE protocol to be tested against VBS enclaves, as well as making sure the protocol cannot be none for tests run in this file (will return a false negative).

_This PR also removed a build warning by adding a comment missing from a variable in SQLServerConnection_


tkyc (approved):

v-zhangw (approved):

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerNoneEnclaveProvider.java
```diff
@@ -0,0 +1,249 @@
...
+            }
+        }
+    }
+
+    private ArrayList<byte[]> describeParameterEncryption(SQLServerConnection connection, SQLServerStatement statement,
+            String userSql, String preparedTypeDefinitions, Parameter[] params,
+            ArrayList<String> parameterNames) throws SQLServerException {
+        ArrayList<byte[]> enclaveRequestedCEKs = new ArrayList<>();
+        try (PreparedStatement stmt = connection.prepareStatement(connection.enclaveEstablished() ? SDPE1 : SDPE2)) {
+            try (ResultSet rs = connection.enclaveEstablished() ? executeSDPEv1(stmt, userSql,
```
lilgreenbird, line 114: eclipse is showing potential mem leaks here

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerNoneEnclaveProvider.java
```diff
@@ -0,0 +1,249 @@
...
+        response.get(publicKeySig, 0, DHPKSsize);
+
+        if (0 != response.remaining()) {
+            SQLServerException.makeFromDriverError(null, this,
+                    SQLServerResource.getResource("R_EnclaveResponseLengthError"), "0", false);
+        }
+    }
+
+    void validateDHPublicKey(byte[] nonce) throws SQLServerException, GeneralSecurityException {
+        if (this.enclaveType == 2) {
```
lilgreenbird, line 242: don't use hardcoded. I think we need a valueOfString or something like that in SQLServerDrivers?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -1970,17 +1970,26 @@ Connection connectInternal(Properties propsIn,
 
                     if (enclaveAttestationProtocol.equalsIgnoreCase(AttestationProtocol.HGS.toString())) {
                         this.enclaveProvider = new SQLServerVSMEnclaveProvider();
-                    } else {
-                        // If it's a valid Provider & not HGS, then it has to be AAS
+                    } else if (enclaveAttestationProtocol.equalsIgnoreCase(AttestationProtocol.NONE.toString())) {
+                        this.enclaveProvider = new SQLServerNoneEnclaveProvider();
+                    }else {
```
lilgreenbird, line 1975: hmm seems this is a bad assumption esp now there there are 3 options, should check for AAS and if not, throw error

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerNoneEnclaveProvider.java
```diff
@@ -0,0 +1,249 @@
...
+    byte[] getNonce() {
+        return nonce;
+    }
+}
+
+
+@SuppressWarnings("unused")
+class NoneAttestationResponse extends BaseAttestationResponse {
+
+    private static final Hashtable<String, JWTCertificateEntry> certificateCache = new Hashtable<>();
```
lilgreenbird, line 194: remove, this is not used anywhere

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerNoneEnclaveProvider.java
```diff
@@ -0,0 +1,249 @@
...
+                throw new SQLServerException(SQLServerException.getErrString("R_UnableRetrieveParameterMetadata"), null,
+                        0, e);
+            }
+        }
+        return enclaveRequestedCEKs;
+    }
+}
+
+
+class NoneAttestationParameters extends BaseAttestationRequest {
```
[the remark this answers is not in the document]
Jeffery-Wasty: Based this class off of SQLServerAASEnclaveProvider, and forgot to remove some stuff. Fixed.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerNoneEnclaveProvider.java
Jeffery-Wasty: Fixed.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
Jeffery-Wasty: There was a check above for if its a valid protocol, but I agree it would look better to check for each valid protocol and then throw an error for no match. Moved the code block to do this.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerNoneEnclaveProvider.java
Jeffery-Wasty: Based this class off of SQLServerAASEnclaveProvider, and forgot to remove some stuff. Fixed.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerNoneEnclaveProvider.java
Jeffery-Wasty: This is a false positive, ResultSet is AutoClosable. Added a resource warning suppression to the method.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerNoneEnclaveProvider.java
```diff
@@ -0,0 +1,256 @@
...
+        response.get(DHpublicKey, 0, DHPKsize);
+        response.get(publicKeySig, 0, DHPKSsize);
+
+        if (0 != response.remaining()) {
+            SQLServerException.makeFromDriverError(null, this,
+                    SQLServerResource.getResource("R_EnclaveResponseLengthError"), "0", false);
+        }
+    }
+
+    void validateDHPublicKey(byte[] nonce) throws SQLServerException, GeneralSecurityException {
```
lilgreenbird, line 248: this is only used here should be private

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -1963,24 +1963,31 @@ Connection connectInternal(Properties propsIn,
...
+
+                        // An attestation protocol that is not NONE requires a URL
                         || (null != enclaveAttestationProtocol && !enclaveAttestationProtocol.isEmpty()
+                                && !enclaveAttestationProtocol.equalsIgnoreCase(AttestationProtocol.NONE.toString())
                                 && (null == enclaveAttestationUrl || enclaveAttestationUrl.isEmpty()))
+
+                        // An attestation protocol also requires column encryption
                         || (null != enclaveAttestationUrl && !enclaveAttestationUrl.isEmpty()
                                 && (null != enclaveAttestationProtocol || !enclaveAttestationProtocol.isEmpty())
                                 && (null == columnEncryptionSetting || !isColumnEncryptionSettingEnabled()))) {
```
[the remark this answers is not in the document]
tkyc: yeah, maybe it's not necessary since there's already comments

tkyc (approved):

lilgreenbird (approved):

tkyc (approved):

v-zhangw (approved):

PR #1780, opened 2022-03-29
Add encrypt utility to obfuscate password strings

lilgreenbird:
This allows storing of the trustStorePassword in encrypted form since it is needed for Idle Connection Resiliency when recovering connections.


tkyc (approved):

tkyc:
Hmm, I was wondering if we should make a directory to contain all the util files, I think there were a few newly created util files in recent PRs. Nothing wrong with the status quo, this is just more of an opinion/thought I had.

Jeffery-Wasty (approved):

Issue #1781, opened 2022-03-29
TimeoutPoller leaks memory when query frequency is high


johnny-egeland:
## Driver version
com.microsoft.sqlserver:mssql-jdbc:10.2.0.jre17

## SQL Server version
Microsoft SQL Server 2019 (RTM-CU9) (KB5000642) - 15.0.4102.2 (X64) 
	Jan 25 2021 20:16:12 
	Copyright (C) 2019 Microsoft Corporation
	Developer Edition (64-bit) on Windows Server 2019 Datacenter 10.0 <X64> (Build 17763: ) (Hypervisor)

## Client Operating System
Ubuntu 20.04.4 LTS

## JAVA/JVM version
openjdk version "17.0.2" 2022-01-18
OpenJDK Runtime Environment (build 17.0.2+8-86)
OpenJDK 64-Bit Server VM (build 17.0.2+8-86, mixed mode, sharing)

## Table schema
Not applicable

## Problem description
Any "TimeoutCommands" entered into the "TimeoutPoller" is retained in the poller until the entire Timeout period has elapsed. The command is not removed when it's completed or closed. In our application the default timeout period is (or was) 1800 seconds, and the ArrayList retaining these commands grew very large when the SQL query volume was high. On our culprit server, the memory usage peaked at 17 Gb when the VM ran out of memory.

I do realize that the timeout can be shortened dramatically on my side. But I still don't think it's a good idea to retain commands to a fixed timeout period, as this means I need to "tune" this in order to get an acceptable middle line between working properly; a.e. long running queries are not canceled, while I don't run out of memory due to completed queries (as most of them complete quickly). They should be removed when they are complete.

## Expected behavior
The poller should check if the command is complete, and remove it from the poller if so. Suggested patch to the code (not testet, nor verified if it is thread-safe and race condition free):
```
diff --git a/src/main/java/com/microsoft/sqlserver/jdbc/TimeoutPoller.java b/src/main/java/com/microsoft/sqlserver/jdbc/TimeoutPoller.java
index 6c53d4d7..ab69c836 100644
--- a/src/main/java/com/microsoft/sqlserver/jdbc/TimeoutPoller.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/TimeoutPoller.java
@@ -61,7 +61,10 @@ final class TimeoutPoller implements Runnable {
                     while (timeoutCommandIterator.hasNext()) {
[...]

PR #1782, opened 2022-03-29
Remove Commands from TimeoutPoller when they have Completed

johnny-egeland:
The TimeoutPoller previously did not release completed commands, and retained the instance in the command array until the entire timeout period has elapsed. Then it would unnecessarily call interrupt() on the command, before removing it from the TimeoutPoller.

This had the side effect that object being added to the TimeoutPoller would not be Garbage Collected until after the entire timeout period. If the query timeout was sufficiently long, it could potentially result in out-of-memory exceptions when the frequency of queries is high. Managing this by reducing the query timeout becomes an unnecessary "tuning" problem, where you choose between memory usage and the possibility of cancelling fully working but long lasting queries.

This commit fixes this issue by removing completed commands from the TimeoutPoller regardless of being timed out or not. This insures the command instances can be garbage collected. The issue has been reported here:

https://github.com/microsoft/mssql-jdbc/issues/1781


lilgreenbird:
closing as this is obsoleted by #1794

PR #1789, opened 2022-04-04
Remove timeout commands from TimeoutPoller on command completion.

johnny-egeland:
This is an alternative fix to the TimeoutPoller memory leak (Issue #1781), caused by commands being retained in the poller until timeout even if the commands complete.

Link to issue:
https://github.com/microsoft/mssql-jdbc/issues/1781

The idea here is simply to remove the command from the TimeoutPoller on command completion. This is an alternative solution  to the following pull-request:
https://github.com/microsoft/mssql-jdbc/pull/1782

Note that the two requests are not in conflict with each other, and should be thread safe together. However, I think only one of them should be applied, as it will render the other redundant.


lilgreenbird:
closing as this is obsoleted by #1794

PR #1813, opened 2022-04-21
Avoid String use with secure strings

David-Engel:
Since Strings are immutable in Java and tend to stay around in memory for potential re-use in the JVM, it's best practice to avoid them for sensitive data. It's recommended to use char arrays instead and clear/overwrite them once they're no longer needed. This limits the amount of time sensitive data remains in memory.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SecureStringUtil.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SecureStringUtil.java
@@ -2,7 +2,7 @@ package com.microsoft.sqlserver.jdbc;
 
 import java.security.SecureRandom;
 import java.text.MessageFormat;
-import java.util.Base64;
+import java.util.Arrays;
 
 import javax.crypto.Cipher;
 import javax.crypto.KeyGenerator;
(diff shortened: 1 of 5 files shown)


v-zhangw (approved):

lilgreenbird (approved):

tkyc (approved):

Jeffery-Wasty (approved):

src/main/java/com/microsoft/sqlserver/jdbc/SecureStringUtil.java
```diff
@@ -88,24 +88,26 @@ private SecureStringUtil() throws SQLServerException {
...
      * 
      * @throws SQLServerException
      *         if error
      */
-    String getEncryptedString(String str) throws SQLServerException {
+    byte[] getEncryptedBytes(char[] chars) throws SQLServerException {
+        if (chars == null)
+            return null;
         SecureRandom random = new SecureRandom();
         random.nextBytes(iv);
```
mmimica, line 103: This is very fragile. Out of order encrypt-decrypt calls will not work because iv is stored in instance variable.

PR #1845, opened 2022-06-13
Cache Parameter Metadata

Jeffery-Wasty:
Cache sp_describe_parameter_encryption calls to improved performance by reducing trips to the server. For now, supports AE without secure enclaves, with enclave support (v3) to be added in a later update.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/AE.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/AE.java
@@ -7,9 +7,9 @@ package com.microsoft.sqlserver.jdbc;
 
 import java.io.Serializable;
 import java.util.ArrayList;
-import java.util.concurrent.ConcurrentHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 
 /**
@@ -237,30 +237,33 @@ class CryptoMetadata {
  */
 class CryptoCache {
     /**
-    * The cryptocache stores both result sets returned from sp_describe_parameter_encryption calls. CEK data in cekMap,
-    * and parameter data in paramMap. 
-    */
+     * The cryptocache stores both result sets returned from sp_describe_parameter_encryption calls. CEK data in cekMap,
+     * and parameter data in paramMap.
+     */
     private final ConcurrentHashMap<String, Map<Integer, CekTableEntry>> cekMap = new ConcurrentHashMap<>(16);
-    private final ConcurrentHashMap<String, ConcurrentHashMap<String, CryptoMetadata>> paramMap 
-        = new ConcurrentHashMap<>(16);
+    private ConcurrentHashMap<String, ConcurrentHashMap<String, CryptoMetadata>> paramMap = new ConcurrentHashMap<>(16);
 
-    public ConcurrentHashMap<String, ConcurrentHashMap<String, CryptoMetadata>> getParamMap() {
+    ConcurrentHashMap<String, ConcurrentHashMap<String, CryptoMetadata>> getParamMap() {
         return paramMap;
     }
-    
-    public Map<Integer, CekTableEntry> getEnclaveEntry(String enclaveLookupKey) {
+
+    void replaceParamMap(ConcurrentHashMap<String, ConcurrentHashMap<String, CryptoMetadata>> newMap) {
+        paramMap = newMap;
+    };
+
+    Map<Integer, CekTableEntry> getEnclaveEntry(String enclaveLookupKey) {
         return cekMap.get(enclaveLookupKey);
     }
-    
-    public ConcurrentHashMap<String, CryptoMetadata> getCacheEntry(String cacheLookupKey) {
+
+    ConcurrentHashMap<String, CryptoMetadata> getCacheEntry(String cacheLookupKey) {
         return paramMap.get(cacheLookupKey);
     }
 
-    public void addParamEntry(String key, ConcurrentHashMap<String, CryptoMetadata> value) {
+    void addParamEntry(String key, ConcurrentHashMap<String, CryptoMetadata> value) {
         paramMap.put(key, value);
     }
 
-    public void removeParamEntry(String cacheLookupKey) {
+    void removeParamEntry(String cacheLookupKey) {
         paramMap.remove(cacheLookupKey);
     }
 }
--- a/src/main/java/com/microsoft/sqlserver/jdbc/AE.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/AE.java
@@ -236,8 +236,12 @@ class CryptoMetadata {
  * Represents a cache of all queries for a given enclave session.
  */
 class CryptoCache {
-    private HashMap<String, Map<Integer, CekTableEntry>> cekMap = new HashMap<>();
-    private HashMap<String, HashMap<String, CryptoMetadata>> paramMap = new HashMap<>();
+    /**
+    * The cryptocache stores both result sets returned from sp_describe_parameter_encryption calls. CEK data in cekMap,
+    * and parameter data in paramMap.
+    */
+    private final HashMap<String, Map<Integer, CekTableEntry>> cekMap = new HashMap<>(16);
+    private final HashMap<String, HashMap<String, CryptoMetadata>> paramMap = new HashMap<>(16);
 
     public HashMap<String, Map<Integer, CekTableEntry>> getCekMap() {
         return cekMap;
@@ -246,13 +250,11 @@ class CryptoCache {
     public HashMap<String, HashMap<String, CryptoMetadata>> getParamMap() {
         return paramMap;
     }
-
-    // Returns a list of parameters with associated metadata, for a given enclave cache.
+    
     public Map<Integer, CekTableEntry> getEnclaveEntry(String enclaveLookupKey) {
         return cekMap.get(enclaveLookupKey);
     }
-
-    // Returns a list of parameters with associated metadata, for a given enclave cache.
+    
     public HashMap<String, CryptoMetadata> getCacheEntry(String cacheLookupKey) {
         return paramMap.get(cacheLookupKey);
     }
--- a/src/main/java/com/microsoft/sqlserver/jdbc/AE.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/AE.java
@@ -232,7 +232,7 @@ class CryptoMetadata {
 }
 
 
-/*
+/**
  * Represents a cache of all queries for a given enclave session.
  */
 class CryptoCache {
@@ -241,7 +241,8 @@ class CryptoCache {
      * and parameter data in paramMap.
      */
     private final ConcurrentHashMap<String, Map<Integer, CekTableEntry>> cekMap = new ConcurrentHashMap<>(16);
-    private ConcurrentHashMap<String, ConcurrentHashMap<String, CryptoMetadata>> paramMap = new ConcurrentHashMap<>(16);
+    private ConcurrentHashMap<String, ConcurrentHashMap<String, CryptoMetadata>> paramMap = 
+        new ConcurrentHashMap<>(16);
 
     ConcurrentHashMap<String, ConcurrentHashMap<String, CryptoMetadata>> getParamMap() {
         return paramMap;
@@ -249,7 +250,7 @@ class CryptoCache {
 
     void replaceParamMap(ConcurrentHashMap<String, ConcurrentHashMap<String, CryptoMetadata>> newMap) {
         paramMap = newMap;
-    };
+    }
 
     Map<Integer, CekTableEntry> getEnclaveEntry(String enclaveLookupKey) {
         return cekMap.get(enclaveLookupKey);
--- a/src/main/java/com/microsoft/sqlserver/jdbc/AE.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/AE.java
@@ -7,6 +7,7 @@ package com.microsoft.sqlserver.jdbc;
 
 import java.io.Serializable;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
 
 
@@ -226,6 +227,30 @@ class CryptoMetadata {
 
     boolean isAlgorithmInitialized() {
         return null != cipherAlgorithm;
+    }
+}
+
+
+/*
+ * Represents a cache of all queries for a given enclave session.
+ */
+class CryptoCache {
+    private HashMap<String, HashMap<String, CryptoMetadata>> map = new HashMap<>();
+
+    public HashMap<String, HashMap<String, CryptoMetadata>> getMap() {
+        return map;
+    }
+
+    public HashMap<String, CryptoMetadata> getEntry(String key) {
+        return map.get(key);
+    }
+
+    public void addEntry(String key, HashMap<String, CryptoMetadata> value) {
+        map.put(key, value);
+    }
+
+    public void remove(String key) {
+        map.remove(key);
     }
 }
 
--- a/src/main/java/com/microsoft/sqlserver/jdbc/AE.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/AE.java
@@ -7,7 +7,7 @@ package com.microsoft.sqlserver.jdbc;
 
 import java.io.Serializable;
 import java.util.ArrayList;
-import java.util.HashMap;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.List;
 import java.util.Map;
 
@@ -238,16 +238,13 @@ class CryptoMetadata {
 class CryptoCache {
     /**
     * The cryptocache stores both result sets returned from sp_describe_parameter_encryption calls. CEK data in cekMap,
-    * and parameter data in paramMap.
+    * and parameter data in paramMap. 
     */
-    private final HashMap<String, Map<Integer, CekTableEntry>> cekMap = new HashMap<>(16);
-    private final HashMap<String, HashMap<String, CryptoMetadata>> paramMap = new HashMap<>(16);
+    private final ConcurrentHashMap<String, Map<Integer, CekTableEntry>> cekMap = new ConcurrentHashMap<>(16);
+    private final ConcurrentHashMap<String, ConcurrentHashMap<String, CryptoMetadata>> paramMap 
+        = new ConcurrentHashMap<>(16);
 
-    public HashMap<String, Map<Integer, CekTableEntry>> getCekMap() {
-        return cekMap;
-    }
-
-    public HashMap<String, HashMap<String, CryptoMetadata>> getParamMap() {
+    public ConcurrentHashMap<String, ConcurrentHashMap<String, CryptoMetadata>> getParamMap() {
         return paramMap;
     }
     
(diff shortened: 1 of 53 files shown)


src/test/java/com/microsoft/sqlserver/jdbc/parametermetadata/ParameterMetaDataCacheTest.java
```diff
@@ -0,0 +1,248 @@
...
+            fail(e.getMessage());
+        }
+        return System.currentTimeMillis() - timer;
+    }
+
+    private void createCMK(String cmkName, String keyPath) throws SQLException {
+        try (Statement stmt = connection.createStatement()) {
+            String sql = " if not exists (SELECT name from sys.column_master_keys where name='" + cmkName + "')"
+                    + " begin" + " CREATE COLUMN MASTER KEY " + cmkName + " WITH (KEY_STORE_PROVIDER_NAME = '"
+                    + Constants.AZURE_KEY_VAULT_NAME + "', KEY_PATH = '" + keyPath + "')" + " end";
```
tkyc, line 178: Should keystore provider be windows key store provider? eg. MSSQL_CERTIFICATE_STORE

src/main/java/com/microsoft/sqlserver/jdbc/AE.java
```diff
@@ -230,6 +232,51 @@ boolean isAlgorithmInitialized() {
...
+ */
+class CryptoCache {
+    /**
+    * The cryptocache stores both result sets returned from sp_describe_parameter_encryption calls. CEK data in cekMap,
+    * and parameter data in paramMap.
+    */
+    private final HashMap<String, Map<Integer, CekTableEntry>> cekMap = new HashMap<>(16);
+    private final HashMap<String, HashMap<String, CryptoMetadata>> paramMap = new HashMap<>(16);
+
+    public HashMap<String, Map<Integer, CekTableEntry>> getCekMap() {
```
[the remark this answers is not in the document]
lilgreenbird: also I don't see this getCekMap used anywhere?

src/main/java/com/microsoft/sqlserver/jdbc/SQLQueryMetadataCache.java
```diff
@@ -0,0 +1,281 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+
+/**
+ * Implements a cache for query metadata returned from sp_describe_parameter_encryption calls. Adding, removing, and
+ * reading from the cache is handled here, with the location of the cache being in the EnclaveSession.
+ * 
+ */
+
```
lilgreenbird, line 18: [nitpick] don't need blank line here

src/main/java/com/microsoft/sqlserver/jdbc/SQLQueryMetadataCache.java
```diff
@@ -0,0 +1,281 @@
...
+    *        Enclave session containing the cryptocache
+    * @param connection
+    *        SQLServerConnection
+    * @param stmt
+    *        SQLServer statement used to retrieve keys to find correct cache
+    * @param cekList
+    *        The list of CEKs (from the first RS) that is also added to the cache as well as parameter metadata
+    * @return true, if the query metadata has been added correctly
+*/
+    public static boolean addQueryMetadata(Parameter[] params, ArrayList<String> parameterNames, EnclaveSession session,
```
lilgreenbird, line 129: these shouldn't be public

src/main/java/com/microsoft/sqlserver/jdbc/SQLQueryMetadataCache.java
```diff
@@ -0,0 +1,281 @@
...
+                            metaData.getEncryptionType().getValue(), metaData.getNormalizationRuleVersion());
+                }
+                if (cryptoCopy != null && !cryptoCopy.isAlgorithmInitialized()) {
+                    String paramName = parameterNames.get(i);
+                    metadataMap.put(paramName, cryptoCopy);
+                } else {
+                    return false;
+                }
+            } catch (SQLServerException e) {
+                e.printStackTrace();
```
lilgreenbird, line 161: we don't do printStackTrace in the driver, catch the exception and throw an error with the appropriate error message. See how it's done elsewhere

src/main/java/com/microsoft/sqlserver/jdbc/SQLQueryMetadataCache.java
```diff
@@ -0,0 +1,281 @@
...
+            }
+        }
+
```
v-zhangw, line 164: Nitpick. Extra blank lines in a bunch of places.

src/test/java/com/microsoft/sqlserver/jdbc/parametermetadata/ParameterMetaDataCacheTest.java
```diff
@@ -0,0 +1,248 @@
...
+    private final String cmkName = "my_cmk";
+    private final String cekName = "my_cek_1";
+    private final String cekNameAlt = "my_cek_2";
+
+    @BeforeAll
+    public static void setupTests() throws Exception {
+        setConnection();
+    }
+
+    public void tableSetup() throws SQLServerException {
```
lilgreenbird, line 48: this is just in a test, but there's no reason why this needs to be public and it's good practice not to declare something public when it's not necessary

src/test/java/com/microsoft/sqlserver/jdbc/parametermetadata/ParameterMetaDataCacheTest.java
```diff
@@ -0,0 +1,248 @@
...
+
+    /**
+     * 
+     * Tests caching of parameter metadata by running a query to be cached, another to replace parameter information,
+     * then the first again to measure the difference in time between the two runs.
+     * 
+     * @throws SQLServerException
+     */
+    @Test
+    @Tag(Constants.xAzureSQLDW)
```
lilgreenbird, line 75: why are we skipping DW everywhere? doesn't this work for DW? we should excludeSQLv11, 12, 14 tho since AE is only supposed after that

src/main/java/com/microsoft/sqlserver/jdbc/AE.java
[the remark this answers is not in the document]
Jeffery-Wasty: Both resolved. It was initially public as I thought it was needed for a test, not the case. getCekMap was added as an analogue to getParamMap, but its not needed.

src/main/java/com/microsoft/sqlserver/jdbc/SQLQueryMetadataCache.java
Jeffery-Wasty: Removed.

src/main/java/com/microsoft/sqlserver/jdbc/SQLQueryMetadataCache.java
Jeffery-Wasty: All methods set to default. They were initially public as I thought they would be needed in tests, but that didn't turn out to be the case.

src/main/java/com/microsoft/sqlserver/jdbc/SQLQueryMetadataCache.java
Jeffery-Wasty: Changed the catch to throw actual error messages.

src/test/java/com/microsoft/sqlserver/jdbc/parametermetadata/ParameterMetaDataCacheTest.java
Jeffery-Wasty: Resolved.

src/main/java/com/microsoft/sqlserver/jdbc/SQLQueryMetadataCache.java
Jeffery-Wasty: Resolved.

src/main/java/com/microsoft/sqlserver/jdbc/SQLQueryMetadataCache.java
```diff
@@ -0,0 +1,278 @@
...
+     * @param stmt
+     *        The SQLServer statement, whose returned metadata we're checking
+     * @return true, if the metadata for the query can be retrieved
+     * 
+     */
+    static boolean getQueryMetadata(Parameter[] params, ArrayList<String> parameterNames,
+            EnclaveSession session, SQLServerConnection connection, SQLServerStatement stmt)throws SQLServerException {
+        
+        if (connection.activeConnectionProperties
+                .getProperty(SQLServerDriverStringProperty.COLUMN_ENCRYPTION.toString()).equalsIgnoreCase("Disabled")) {
```
lilgreenbird, line 45: this seems unnecessary? we would never even be here if this was disabled

src/main/java/com/microsoft/sqlserver/jdbc/AE.java
```diff
@@ -230,6 +232,51 @@ boolean isAlgorithmInitialized() {
...
+    
+    public Map<Integer, CekTableEntry> getEnclaveEntry(String enclaveLookupKey) {
+        return cekMap.get(enclaveLookupKey);
+    }
+    
+    public HashMap<String, CryptoMetadata> getCacheEntry(String cacheLookupKey) {
+        return paramMap.get(cacheLookupKey);
+    }
+
+    public void addCekEntry(String key, Map<Integer, CekTableEntry> value) {
```
lilgreenbird: these don't need to be public

src/main/java/com/microsoft/sqlserver/jdbc/SQLQueryMetadataCache.java
```diff
@@ -0,0 +1,278 @@
...
+                                                    ("R_cryptoCacheInaccessible"), null, 0, false);
+            }
+        }
+        
+        // If the size of the cache exceeds the threshold, set that we are in trimming and trim the cache accordingly.
+        int cacheSizeCurrent = session.getCryptoCache().getParamMap().size();
+        if (cacheSizeCurrent > cacheSize + cacheTrimThreshold) {
+            try {
+                int entriesToRemove = cacheSizeCurrent - cacheSize;
+                ConcurrentHashMap<String, ConcurrentHashMap<String, CryptoMetadata>> newMap = new ConcurrentHashMap<>();
```
David-Engel, line 172: I see below that we are copying into the newMap from the oldMap, but I don't see where we are replacing oldMap with newMap in the CryptoCache. Am I just missing it?

src/test/java/com/microsoft/sqlserver/jdbc/parametermetadata/ParameterMetaDataCacheTest.java
```diff
@@ -0,0 +1,253 @@
...
+    private SQLServerColumnEncryptionKeyStoreProvider registerProvider(
+            SQLServerColumnEncryptionKeyStoreProvider provider) throws SQLServerException {
+        Map<String, SQLServerColumnEncryptionKeyStoreProvider> map1 = 
+            new HashMap<String, SQLServerColumnEncryptionKeyStoreProvider>();
+        map1.put(provider.getName(), provider);
+        SQLServerConnection.registerColumnEncryptionKeyStoreProviders(map1);
+        return provider;
+    }
+
+    private String bytesToHexString(byte[] b, int length) {
```
lilgreenbird, line 243: there's one in TestUtils already

src/main/java/com/microsoft/sqlserver/jdbc/SQLQueryMetadataCache.java
```diff
@@ -0,0 +1,272 @@
...
+    /**
+     * 
+     * Copy the enclave CEKs so they can be later used to retry secure enlave queries.
+     * 
+     * @param keysToBeSentToEnclave
+     *        The CEKs sent to the enclave cryptocache
+     * @return
+     *        A copy of the CEKs, this is what is actually added to the cryptocache
+     */
+    private static Map<Integer, CekTableEntry> copyEnclaveKeys(Map<Integer, CekTableEntry> keysToBeSentToEnclave) {
```
lilgreenbird, line 257: remove - this is not used anywhere

src/main/java/com/microsoft/sqlserver/jdbc/SQLQueryMetadataCache.java
```diff
@@ -0,0 +1,272 @@
+/*
+ * Microsoft JDBC Driver for SQL Server Copyright(c) Microsoft Corporation All rights reserved. This program is made
+ * available under the terms of the MIT License. See the LICENSE file in the project root for more information.
+ */
+package com.microsoft.sqlserver.jdbc;
+
+import java.util.AbstractMap;
+import java.util.ArrayList;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.HashMap;
```
lilgreenbird, line 10: not used

src/main/java/com/microsoft/sqlserver/jdbc/SQLQueryMetadataCache.java
```diff
@@ -0,0 +1,272 @@
...
+    * @param connection
+    *        The connection from which database name is retrieved
+    * @return
+    *        A key value pair containing cache lookup key and enclave lookup key
+    */
+    private static AbstractMap.SimpleEntry<String, String> getCacheLookupKeys(
+            SQLServerStatement statement, SQLServerConnection connection) {
+        final int sqlIdentifierLength = 128;
+        
+        if (connection == null) {
```
lilgreenbird, line 225: when would connection be null? I would think something bad has happened wondering if we should fail or at least log a warning?

src/main/java/com/microsoft/sqlserver/jdbc/SQLQueryMetadataCache.java
```diff
@@ -0,0 +1,272 @@
...
+
+/**
+ * Implements a cache for query metadata returned from sp_describe_parameter_encryption calls. Adding, removing, and
+ * reading from the cache is handled here, with the location of the cache being in the EnclaveSession.
+ * 
+ */
+class SQLQueryMetadataCache {
+
+    final static int cacheSize = 2000; // Size of the cache in number of entries
+    final static int cacheTrimThreshold = 300; // Threshold above which to trim the cache
```
lilgreenbird, line 23: typically we capitalize static finals

src/main/java/com/microsoft/sqlserver/jdbc/ISQLServerEnclaveProvider.java
```diff
@@ -279,6 +280,12 @@ default void processSDPEv1(String userSql, String preparedTypeDefinitions, Param
                 }
             }
         }
+
+        // If using AEv1, add to cache
```
lilgreenbird, line 284: can we change to describing this instead of using referring to internal AEv1?

src/main/java/com/microsoft/sqlserver/jdbc/SQLQueryMetadataCache.java
```diff
@@ -0,0 +1,272 @@
...
+                    return false;
+                }
+            } catch (SQLServerException e) {
+                throw new SQLServerException(null, SQLServerException.getErrString
+                                                    ("R_cryptoCacheInaccessible"), null, 0, false);
+            }
+        }
+        
+        // If the size of the cache exceeds the threshold, set that we are in trimming and trim the cache accordingly.
+        int cacheSizeCurrent = session.getCryptoCache().getParamMap().size();
```
lilgreenbird, line 160: can we save the cryptoCache instead so we don't have to call getCryptoCache() so many times?

src/main/java/com/microsoft/sqlserver/jdbc/SQLQueryMetadataCache.java
```diff
@@ -0,0 +1,272 @@
...
+                }
+                if (cryptoCopy != null && !cryptoCopy.isAlgorithmInitialized()) {
+                    String paramName = parameterNames.get(i);
+                    metadataMap.put(paramName, cryptoCopy);
+                } else {
+                    return false;
+                }
+            } catch (SQLServerException e) {
+                throw new SQLServerException(null, SQLServerException.getErrString
+                                                    ("R_cryptoCacheInaccessible"), null, 0, false);
```
lilgreenbird, line 155: error string not defined, multiple occurrences. Also, please add the exception error message to the error message

lilgreenbird: this made me realized there are no negative testing as this would've caused a crash. Can we add some negative tests to go thru the error paths

src/main/java/com/microsoft/sqlserver/jdbc/SQLQueryMetadataCache.java
Jeffery-Wasty: This will be used for secure enclaves in the future. Thoughts on adding an `unused` tag temporarily?

src/test/java/com/microsoft/sqlserver/jdbc/parametermetadata/ParameterMetaDataCacheTest.java
```diff
@@ -0,0 +1,253 @@
...
+@Tag(Constants.xSQLv12)
+@Tag(Constants.xSQLv14)
+public class ParameterMetaDataCacheTest extends AbstractTest {
+    private static final String firstTable = "firstTable";
+    private static final String secondTable = "secondTable";
+    private final String cmkName = "my_cmk";
+    private final String cekName = "my_cek_1";
+    private final String cekNameAlt = "my_cek_2";
+    private final String sampleData = "testData";
+    private final String sampleData2 = "testData2";
```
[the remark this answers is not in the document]
Jeffery-Wasty: Was able to get it working this time, resolved.

src/main/java/com/microsoft/sqlserver/jdbc/SQLQueryMetadataCache.java
Jeffery-Wasty: Correct, there's no case it would be null. If it was null we wouldn't even be able to call the function, this has been removed. I initially added it as I planned to use addQueryMetadata in testing, but decided against it.

src/main/java/com/microsoft/sqlserver/jdbc/SQLQueryMetadataCache.java
Jeffery-Wasty: Resolved

src/main/java/com/microsoft/sqlserver/jdbc/SQLQueryMetadataCache.java
Jeffery-Wasty: Added tests for both cases - addQueryMetadata and getQueryMetadata, to assert that the the tests fail when they should + the error returned is as expected.

src/main/java/com/microsoft/sqlserver/jdbc/SQLQueryMetadataCache.java
```diff
@@ -0,0 +1,272 @@
...
+    * @param statement
+    *        The SQLServer statement used to construct part of the keys
+    * @param connection
+    *        The connection from which database name is retrieved
+    * @return
+    *        A key value pair containing cache lookup key and enclave lookup key
+    */
+    private static AbstractMap.SimpleEntry<String, String> getCacheLookupKeys(
+            SQLServerStatement statement, SQLServerConnection connection) {
+        final int sqlIdentifierLength = 128;
```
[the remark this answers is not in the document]
Jeffery-Wasty: This wasn't checking for max length. It was using max length to pad DB name to keep keys at a consistent length. However, I'm not sure that's necessary, so I've removed it entirely.

lilgreenbird (approved):

David-Engel (dismissed):
LGTM

tkyc (approved):

lilgreenbird (dismissed):

v-zhangw (approved):

Jeffery-Wasty:
Hi @alex-bochkov,

We'll be releasing documentation regarding new features shortly, in the meantime I can answer the questions you have included:

**1. What's the cache TTL?**

The cache doesn't have an independent TTL, rather the enclave session it belongs to has a TTL of eight hours.

**2. Is it possible to disable cache if needed? For example, if we need to make schema changes and make sure the app won't break in the middle.**

It's not possible to disable the cache at the moment. This is something we can potentially add a later time, however current efforts are towards extending caching for secure enclave scenarios.

Issue #1853, opened 2022-06-24
Parameter order in connection string makes a difference


Chrriis:
## Driver version
11.1.2

## Problem description
When creating report for bug #1852, I simplified my my properties by putting them all into the connection string. I noticed that certain connections were not working depending on the order of parameters.

Here is a standalone test case that illustrate the problem, switching the first 2 parameters:
```Java
public class TestMsSqlJdbcConnectionStringParameterOrder {
    public static void main(String[] args) throws Exception {
        Class.forName(com.microsoft.sqlserver.jdbc.SQLServerDriver.class.getName());
        String connectionString1 = "jdbc:sqlserver://databaseName=xxxxx;serverName=localhost;userName=xxxxx;password=xxxxx;instanceName=SQLEXPRESS;packetSize=16384;prepareMethod=prepare;trustServerCertificate=true";
        String connectionString2 = "jdbc:sqlserver://serverName=localhost;databaseName=xxxxx;userName=xxxxx;password=xxxxx;instanceName=SQLEXPRESS;packetSize=16384;prepareMethod=prepare;trustServerCertificate=true";
        try {
            Connection conn = DriverManager.getConnection(connectionString1);
            conn.close();
            System.out.println("Connection 1: OK");
        } catch (SQLException e) {
            System.out.println("Connection 1: FAILED");
            e.printStackTrace();
        }
        try {
            Connection conn = DriverManager.getConnection(connectionString2);
            conn.close();
            System.out.println("Connection 2: OK");
        } catch (SQLException e) {
            System.out.println("Connection 2: FAILED");
            e.printStackTrace();
        }
    }
}
```

Simply replace the "xxxxx" by real data. The first call prints OK, the second call fails with "java.net.UnknownHostException: serverName=localhost" after a timeout of about 15s.

tkyc:
Hey @Chrriis that is strange. I recall from an example I saw (trying to find it again, it's been awhile), when explicitly defining all the connection properties like you have eg. `jdbc:sqlserver://<prop>=<value>;......`, you need to include a `;` in front of the initial property like so `jdbc:sqlserver://;serverName=localhost;....`. I was able to connect using `connectionString2` after making this change. It's weird that it wasn't the case for `connectionString1`. Definitely something is inconsistent here.

I'll need to look into this some more.

Chrriis:
[...]

PR #1856, opened 2022-06-27
Added validation for serverName field

tkyc:
Added validation for server name field for https://github.com/microsoft/mssql-jdbc/issues/1853


v-zhangw (approved):

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResource.java
```diff
@@ -107,6 +107,7 @@ protected Object[][] getContents() {
         {"R_statementMustBeExecuted", "The statement must be executed before any results can be obtained."},
         {"R_modeSuppliedNotValid", "The supplied mode is not valid."},
         {"R_errorConnectionString", "The connection string contains a badly formed name or value."},
+        {"R_errorServerName", "The server name {0} is badly formed."},
```
lilgreenbird, line 110: in looking at the other error string relating to connection properties, this should be "The serverName connection property {0} is badly formed."

Jeffery-Wasty (approved):

src/main/java/com/microsoft/sqlserver/jdbc/Util.java
```diff
@@ -34,6 +36,8 @@ final class Util {
 
     final static String ACTIVITY_ID_TRACE_PROPERTY = "com.microsoft.sqlserver.jdbc.traceactivity";
 
+    final static Pattern SERVER_NAME_PATTERN = Pattern.compile("((?=^.{1,253}$)(^(((?!-)[a-zA-Z0-9-'\"_]{1,}(?<!-))|((?!-)[a-zA-Z0-9-'\"_]{1,}(?<!-)\\.)+[a-zA-Z]{2,})$))|(\\d+\\.?)+");
```
David-Engel, line 39: Can we include a comment above this line that describes what rules the regex is trying to enforce?

My concern is we might end up breaking something like [IDN support](https://en.wikipedia.org/wiki/Internationalized_domain_name). Could we simplify this check by simply checking for an `=` in the serverName field, which would indicate the user is really trying to specify a property and is the potentially common error we are guarding against?

lilgreenbird: yeah it's not too bad since we only parse this once but regex is really slow. As for IDN, we have tests for that and also we have fuzz tests for IDN you can run

Jeffery-Wasty (dismissed):

src/test/java/com/microsoft/sqlserver/jdbc/SQLServerConnectionTest.java
```diff
@@ -1007,4 +1008,31 @@ public void testParameterMetadataAccessDenial() throws SQLException {
...
         }
     }
+
+    @Test
+    public void testServerNameField() throws SQLException {
+        String subProtocol = "jdbc:sqlserver://";
+        int indexOfFirstDelimiter = connectionString.indexOf(";");
+        int indexOfLastDelimiter = connectionString.lastIndexOf(";");
+
+        String[] serverNameAndPort = connectionString.substring(17, indexOfFirstDelimiter).split(":");
```
lilgreenbird, line 1018: instead of hardcoding 17,  subProtocol.length()?

src/test/java/com/microsoft/sqlserver/jdbc/SQLServerConnectionTest.java
```diff
@@ -1007,4 +1008,31 @@ public void testParameterMetadataAccessDenial() throws SQLException {
...
+
+        try (SQLServerConnection conn = (SQLServerConnection) DriverManager.getConnection(emptyServerNameField)) {
+        }
+
+        try (SQLServerConnection conn = (SQLServerConnection) DriverManager.getConnection(invalidServerNameField)) {
+        } catch (SQLException e) {
+            MessageFormat form = new MessageFormat(SQLServerException.getErrString("R_errorServerName"));
+            assertEquals(form.format(new Object[] {loginTimeout}), e.getMessage(),
+                    TestResource.getResource("R_wrongExceptionMessage"));
+        }
```
lilgreenbird, line 1036: you can use TestUtils.formatErrorMsg

lilgreenbird (approved):

David-Engel (approved):

Issue #1505, opened 2021-01-28
Kill session in SQLServer has executeQuery continue without any Exception on the client


mreuvers:
## Driver version
8.4.1.jre11

## SQL Server version
Microsoft SQL Server 2019 (RTM-CU8-GDR) (KB4583459) - 15.0.4083.2 (X64) \n	Nov  2 2020 18:35:09 \n	Copyright (C) 2019 Microsoft Corporation\n	Developer Edition (64-bit) on Linux (Ubuntu 18.04.5 LTS) <X64>

## Client Operating System
MacOS Mojave (I expect the issue no matter which OS).

## JAVA/JVM version
openjdk version "11.0.5" 2019-10-15
OpenJDK Runtime Environment AdoptOpenJDK (build 11.0.5+10)
OpenJDK 64-Bit Server VM AdoptOpenJDK (build 11.0.5+10, mixed mode)

## Table schema
Not relevant, reproduction case attached.

## Problem description
Long running query client (preparedStatement.executeQuery()) is waiting for it's ResultSet. In SQLServer find the relevant session id, and kill it using: `KILL <IDHERE>`
1. Expected behaviour: SQLException at the client
2. Actual behaviour: Continues as if nothing was wrong (no results at least)
3. Error message/stack trace: None, not even warnings.

## JDBC trace logs
Not really relevant, but when issuing the KILL sessionid ->
2021-01-27 13:25:53.68 spid70      Process ID 55 was killed by hostname <HostName>, host process ID 0.

## Reproduction code
Maven reproduction case attached. Please read below for full instructions.

## Reproduction case

- Install and setup jdk from https://adoptopenjdk.net, pick one of: 11 LTS + HOTSPOT (or OPENJ9)
- Install  and setup maven from https://maven.apache.org 
- Docker (makes this easier, or you'll have to setup the SQLServer version mentioned above manually and change the test code to match your credentials/url):
  - Install and setup docker 
  - Install and setup docker-compose 
- Extract the .zip
- Open a shell and "cd" into the docker directory
   - `docker-compose up -d`  (see settings.env for details), it sets up the latest docker SQLServer 19 image and binds to port 1433
- Go back to the root directory again and run: `mvn clean package` This will run the test (see Main.java for details)

What you'll see is that it prints "Test started..." and then "hangs", at this point the long running query started.
- Go to SQLServer and execute the following query to find the session of this query:
```
SELECT r.session_id,
       st.TEXT AS batch_text,
[...]

PR #1857, opened 2022-06-29
Added check for DONE_ERROR status token and throw appropriate exception

### LINKED ISSUE #1846: [QUESTION] Exception is not thrown as expected when the session is killed in SQL database
## Question
when using this library in Spark to run the SQL query and get the data into a Spark dataframe, if the session is accidentally killed from SQL database side, there's no reset error thrown as expected, but the count result has indicated that the data is not read completely.

### **steps to reproduce:**
(1)  use this library spark to run the SQL query and get the data into a Spark dataframe, code shown as below:
```
%scala
val token = "<token>"

val jdbcHostname = "xinrandatabseserver.database.windows.net"
val jdbcDatabase = "xinranSQLDatabase"
val jdbcPort = 1433
val jdbcUrl = "jdbc:sqlserver://%s:%s;databaseName=%s;encrypt=true;trustServerCertificate=false;hostNameInCertificate=*.database.windows.net".format(jdbcHostname, jdbcPort, jdbcDatabase)+ ";accessToken="


import java.util.Properties
val connectionProperties = new Properties()
val driverClass = "com.microsoft.sqlserver.jdbc.SQLServerDriver"
connectionProperties.setProperty("Driver", driverClass)
connectionProperties.setProperty("accesstoken", token)

val sql_pushdown = "(select UNITS from payment_balance_new) emp_alias"
val df_stripe_dispute = spark.read.option("connectRetryCount", 200).option("numPartitions",1).jdbc(url=jdbcUrl, table=sql_pushdown, properties=connectionProperties)
df_stripe_dispute.count()
```

(2) kill the session from the SQL database side by "kill [SPID]"

(3) check the count() result, which is less than the actual one, but no error exception is thrown, the command is executed successfully.

### **Need help on**
In general Spark connects to SqlServer and reads the data using the below code where **rs** is **com.microsoft.sqlserver.jdbc.SQLServerResultSet**, so whenever you killed the session rs.next() returned false but didn't throw the exception.

```
override protected def getNext(): InternalRow = {
if (rs.next()) {
inputMetrics.incRecordsRead(1)
var i = 0
while (i < getters.length) {
getters(i).apply(rs, mutableRow, i)
if (rs.wasNull) mutableRow.setNullAt(i)
i = i + 1
}
mutableRow
} else {
finished = true
null.asInstanceOf[InternalRow]
}
}
```

Based on the code analysis of com.microsoft.sqlserver.jdbc.SQLServerResultSet.next() there is a possibility that it returns **false** when the connection is broken, You can refer the code here https://github.com/microsoft/mssql-jdbc/blob/main/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResultSet.java#L1011-L1107

**Could you help check why the exception is not thrown as expected? Thanks!**

lilgreenbird:
This fixes #1846 and #1505 where a DONE_ERROR status from the server resulting from a killed session was ignored by the driver.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
@@ -180,6 +180,14 @@ final class TDS {
     static final int TDS_SQLRESCOLSRCS = 0xa2;
     static final int TDS_SQLDATACLASSIFICATION = 0xa3;
 
+    // DONE status https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-tds/3c06f110-98bd-4d5b-b836-b1ba66452cb7
+    static final int DONE_FINAL = 0x0000;
+    static final int DONE_MORE = 0x0001;
+    static final int DONE_COUNT = 0x0010;
+    static final int DONE_ERROR = 0x0002;
+    static final int DONE_ATTN = 0x0020;
+    static final int DONE_SRVERROR = 0x0100;
+
     // FedAuth
     static final byte TDS_FEATURE_EXT_FEDAUTH = 0x02;
     static final int TDS_FEDAUTH_LIBRARY_SECURITYTOKEN = 0x01;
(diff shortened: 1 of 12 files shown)


Jeffery-Wasty (approved):

tkyc (dismissed):

David-Engel (approved):
Assuming the test stability is worked out, I'm good with the changes, from a funtional perspective.

tkyc (approved):

Jeffery-Wasty (approved):

v-zhangw (approved):

PR #1858, opened 2022-06-30
Fix: Initialization vector cannot be reused

David-Engel:
Prepend iv onto the encrypted bytes so that each iv is unique and preserved with its encrypted bytes for later decryption.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SecureStringUtil.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SecureStringUtil.java
@@ -31,9 +31,6 @@ final class SecureStringUtil {
     /* authentication tag length in bits */
     static final int TAG_LENGTH = 16;
 
-    /* initialization vector */
-    byte[] iv;
-
     /** secret key for encryption/decryption */
     SecretKeySpec secretKey;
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/SecureStringUtil.java
```diff
@@ -127,13 +128,19 @@ byte[] getEncryptedBytes(char[] chars) throws SQLServerException {
...
+        System.arraycopy(bytes, IV_LENGTH, encryptedBytes, 0, bytes.length - IV_LENGTH);
+
         GCMParameterSpec ivParamSpec = new GCMParameterSpec(TAG_LENGTH * 8, iv);
 
         byte[] plainText = null;
         try {
             decryptCipher.init(Cipher.DECRYPT_MODE, secretKey, ivParamSpec);
 
-            plainText = decryptCipher.doFinal(bytes);
+            plainText = decryptCipher.doFinal(encryptedBytes);
```
mmimica, line 143: You can avoid array copy with `decryptCipher.doFinal(bytes, IV_LENGTH, bytes.length - IV_LENGTH);`

lilgreenbird (approved):

Jeffery-Wasty (approved):

tkyc (approved):

Issue #1843, opened 2022-06-03
issue due to rounding of timestamps


AlBundy33:
## Driver version
10.2.0

## SQL Server version
Microsoft SQL Server 2019 (RTM-CU15) (KB5008996) - 15.0.4198.2 (X64) 
	Jan 12 2022 22:30:08 
	Copyright (C) 2019 Microsoft Corporation
	Developer Edition (64-bit) on Linux (Ubuntu 20.04.3 LTS) <X64>

## Client Operating System
Windows 10 but also affects Ubuntu

## JAVA/JVM version
OpenJDK 11 but also affects Oracle Java 8

## Description

We've recently tried to switch from jTDS to mssql-jdbc-driver which seems to work pretty good in most cases.
But there seems to be an issue due to the rounding of timestamps in mssql-jdbc because if we write a tiemstamp with System.currentTimeMillis() in the database, read the value from the database and try to update all records with timestamp >= the previous read tiemstamp this doesn't work for some values.

## Test-case

I hope this test-case will describe the problem better :-)
-> This test-case is similar to some logic in our application that currently fails in our integration tests.

I've used com.microsoft.sqlserver.jdbc.preparedStatement.SetObjectTest as base - not sure if this is the right place.

```
    @Test
    public void testTimestampRounding()
        throws SQLException
    {
      try (Connection con = getConnection())
      {
        final String tableName = "sometesttable";
        try (Statement stmt = con.createStatement())
        {
          stmt.executeUpdate("CREATE TABLE " + AbstractSQLGenerator.escapeIdentifier(tableName)
              + " (id INT PRIMARY KEY, ts datetime)");

          try
          {
            final long millis = System.currentTimeMillis();
            for (int i = 0; i < 100; i++)
            {
              // 1. create a timestamp
              final Timestamp preparedTimestamp = new Timestamp(millis + i);
              // 2. insert timestamp
              try (PreparedStatement pstmt = con.prepareStatement(
                  "INSERT INTO " + AbstractSQLGenerator.escapeIdentifier(tableName) + " (id, ts) VALUES (?, ?)"))
              {
                pstmt.setInt(1, i);
                pstmt.setObject(2, preparedTimestamp);
                pstmt.executeUpdate();
              }

              // 3. read timestamp from db
              Timestamp timestampFromDatabase = null;
              try (PreparedStatement select = con.prepareStatement("select ts from " + AbstractSQLGenerator.escapeIdentifier(tableName)
[...]

PR #1865, opened 2022-07-12
Feat specify datetime type

### LINKED ISSUE #1590: [FEATURE REQUEST] Ability to explicitly specify conversion to datetime or datetime2

v-zhangw:
An update on @dswitzer 's original PR https://github.com/microsoft/mssql-jdbc/pull/1687 which fixes #1590 and #1843.

Provides the option to allow users to specify the SQL Server datatype to use for date/timestamp parameters.


lilgreenbird:
close, this needs more work

PR #1866, opened 2022-07-13
Cache parameter metadata secure

Jeffery-Wasty:
Extends previous cache parameter metadata function for secure enclaves. Secure enclaves must support retrying the enclave when the session is no longer valid (Always Encrypted version 3).


src/main/java/com/microsoft/sqlserver/jdbc/ParameterMetaDataCache.java
```diff
@@ -240,4 +243,68 @@ private static AbstractMap.SimpleEntry<String, String> getCacheLookupKeys(SQLSer
...
+     * Key data in cekMap, and parameter data in paramMap.
+     */
+    private final ConcurrentHashMap<String, Map<Integer, CekTableEntry>> cekMap = new ConcurrentHashMap<>(16);
+    private ConcurrentHashMap<String, ConcurrentHashMap<String, CryptoMetadata>> paramMap = new ConcurrentHashMap<>(16);
+
+    ConcurrentHashMap<String, ConcurrentHashMap<String, CryptoMetadata>> getParamMap() {
+        return paramMap;
+    }
+
+    void replaceParamMap(ConcurrentHashMap<String, ConcurrentHashMap<String, CryptoMetadata>> newMap) {
```
tkyc, line 287: Just a question related to the cache trimming. When we trim the cache, how do we know we are removing stale entries?

src/main/java/com/microsoft/sqlserver/jdbc/ParameterMetaDataCache.java
```diff
@@ -40,8 +40,8 @@ class ParameterMetaDataCache {
      * @return true, if the metadata for the query can be retrieved
      * 
      */
-    static boolean getQueryMetadata(Parameter[] params, ArrayList<String> parameterNames, CryptoCache cache,
-            SQLServerConnection connection, SQLServerStatement stmt) throws SQLServerException {
+    boolean getQueryMetadata(Parameter[] params, ArrayList<String> parameterNames, SQLServerConnection connection,
```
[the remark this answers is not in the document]
Jeffery-Wasty: Absolutely right. Comment was made in the first go around, and session is no longer used. I think other comments might also be out of date.

src/main/java/com/microsoft/sqlserver/jdbc/ParameterMetaDataCache.java
Jeffery-Wasty: We're not exactly removing stale entries, just the ~300 oldest, which will have occurred 2000 queries ago. If 2000 queries have been executed since the first one, I think it would be safe to assume that first one is no longer cached. I can't think of an alternative, only expanding the cache size, but that would just have memory concerns.

tkyc: Hmm, I'm little worried about entry trim then, because we remove entries by iterating from the first entry. Maps do not preserve order of insertion.

src/test/java/com/microsoft/sqlserver/jdbc/AlwaysEncrypted/ParameterMetaDataCacheTest.java
```diff
@@ -0,0 +1,138 @@
...
+
+
+/**
+ * Tests for caching parameter metadata in sp_describe_parameter_encryption calls
+ */
+@RunWith(JUnitPlatform.class)
+@Tag(Constants.xSQLv11)
+@Tag(Constants.xSQLv12)
+@Tag(Constants.xSQLv14)
+public class ParameterMetaDataCacheTest extends AESetup {
```
tkyc, line 34: Just a reminder to run the coco pipeline if you haven't. Disregard if you did.

src/main/java/com/microsoft/sqlserver/jdbc/ParameterMetaDataCache.java
```diff
@@ -31,17 +31,14 @@ class ParameterMetaDataCache {
...
      * @param connection
      *        The SQLServer connection
      * @param stmt
      *        The SQLServer statement, whose returned metadata we're checking
      * @return true, if the metadata for the query can be retrieved
-     * 
      */
-    static boolean getQueryMetadata(Parameter[] params, ArrayList<String> parameterNames, CryptoCache cache,
-            SQLServerConnection connection, SQLServerStatement stmt) throws SQLServerException {
+    boolean getQueryMetadata(Parameter[] params, ArrayList<String> parameterNames, SQLServerConnection connection,
```
David-Engel, line 40: I would keep existing static methods static. It may only be a small perf difference, but we definitely care about small perf changes in the driver.

src/main/java/com/microsoft/sqlserver/jdbc/ParameterMetaDataCache.java
David-Engel: Good catch. We should have used ConcurrentLinkedHashMap in the original implementation to ensure order is preserved.

src/test/java/com/microsoft/sqlserver/jdbc/AlwaysEncrypted/ParameterMetaDataCacheTest.java
```diff
@@ -0,0 +1,138 @@
...
+            assertTrue(1 - (secondRun / firstRun) > threshold);
+        }
+    }
+
+    /**
+     * 
+     * Tests that the enclave is retried when using secure enclaves (assuming the server supports this). This is done by
+     * executing a query generating metadata in the cache, changing the column encryption type to make the metadata
+     * stale, and running the query again. The query should fail, but retry and pass. Currently disabled as secure
+     * enclaves are not supported.
```
v-zhangw, line 83: Stale text?

src/main/java/com/microsoft/sqlserver/jdbc/ParameterMetaDataCache.java
Jeffery-Wasty: I ended up changing them back. Running tests with the methods as static and not, I couldn't see a difference, so I'll go with static since its cleaner.

src/test/java/com/microsoft/sqlserver/jdbc/AlwaysEncrypted/ParameterMetaDataCacheTest.java
Jeffery-Wasty: Last line removed. Test is enabled.

src/main/java/com/microsoft/sqlserver/jdbc/ParameterMetaDataCache.java
Jeffery-Wasty: Yup thanks for that, and actually, that exposed a possibility I didn't test for. I added a test for trimming which should have been there in the first place.

src/test/java/com/microsoft/sqlserver/jdbc/AlwaysEncrypted/ParameterMetaDataCacheTest.java
```diff
@@ -0,0 +1,160 @@
...
+     * stale, and running the query again. The query should fail, but retry and pass.
+     * 
+     * @throws Exception
+     */
+    @Test
+    @Tag(Constants.xSQLv11)
+    @Tag(Constants.xSQLv12)
+    @Tag(Constants.xSQLv14)
+    @Tag(Constants.reqExternalSetup)
+    @Tag(Constants.AEv2)
```
Jeffery-Wasty, line 136: Similar to above, this will only work when the server supports retry (AEv3) which are enclave servers do not, so the tag is also set here to exclude.

src/main/java/com/microsoft/sqlserver/jdbc/ParameterMetaDataCache.java
```diff
@@ -54,15 +54,14 @@ static boolean getQueryMetadata(Parameter[] params, ArrayList<String> parameterN
...
         for (int i = 0; i < params.length; i++) {
-            boolean found = metadataMap.containsKey(parameterNames.get(i));
             CryptoMetadata foundData = metadataMap.get(parameterNames.get(i));
 
             /*
-             * If ever the map doesn't contain a parameter, the cache entry cannot be used. If there is data found, it
-             * should never have the initialized algorithm as that would contain the key. Clear all metadata that has
-             * already been assigned in either case.
+             * It's fine for the cache to not contain a parameter. This means the parameter uses a plaintext encryption
+             * and its CryptoMetadata was never formed. It is NOT okat for there to be found data with an initialized
```
lilgreenbird, line 61: typo also when would this ever happen is this an error condition should we fail or at least log a warning or something?

Jeffery-Wasty: I can add a warning. I don't think its a failure as its working correctly; cache parameter encryption values, no encryption, nothing to cache. But I agree with logging a warning.

src/main/java/com/microsoft/sqlserver/jdbc/ParameterMetaDataCache.java
```diff
@@ -8,19 +8,22 @@
...
  */
-class ParameterMetaDataCache {
-
-    static final int CACHE_SIZE = 2000; // Size of the cache in number of entries
-    static final int CACHE_TRIM_THRESHOLD = 300; // Threshold above which to trim the cache
+public class ParameterMetaDataCache {
 
+    static int CACHE_SIZE = 2000; // Size of the cache in number of entries
+    static int MAX_WEIGHTED_CAPACITY = 2300; // Size of cache + threshold, above which we trim.
+    static int CACHE_HITS = 0; // Record of the number of cache hits, mainly for testing
```
lilgreenbird, line 25: this gets incremented but is never used in the driver anywhere

Jeffery-Wasty: I used it in testing. I used it to monitor whether the cache is working correctly, as a successful `get` will increment `CACHE_HITS`

src/main/java/com/microsoft/sqlserver/jdbc/ParameterMetaDataCache.java
```diff
@@ -156,12 +158,11 @@ static boolean addQueryMetadata(Parameter[] params, ArrayList<String> parameterN
...
+                    if (cryptoCopy.isAlgorithmInitialized()) {
+                        return false;
+                    }
                     String paramName = parameterNames.get(i);
                     metadataMap.put(paramName, cryptoCopy);
-                } else {
-                    return false;
                 }
             } catch (SQLServerException e) {
                 MessageFormat form = new MessageFormat(SQLServerException.getErrString("R_CryptoCacheInaccessible"));
```
lilgreenbird, line 168: I'm thinking we should add more info to this error to help with debugging since I'm not sure if the exception would contain all the info we want.

Jeffery-Wasty: Both `R_CryptoCacheInaccessible` errors would originate from the same source: making a new CryptoMetaData with an invalid column encryption type, so I just passed that as an additional parameter to the final output.

src/main/java/com/microsoft/sqlserver/jdbc/ParameterMetaDataCache.java
```diff
@@ -172,13 +173,15 @@ static boolean addQueryMetadata(Parameter[] params, ArrayList<String> parameterN
...
         // If the size of the cache exceeds the threshold, set that we are in trimming and trim the cache accordingly.
         int cacheSizeCurrent = cache.getParamMap().size();
-        if (cacheSizeCurrent > CACHE_SIZE + CACHE_TRIM_THRESHOLD) {
+        if (cacheSizeCurrent > MAX_WEIGHTED_CAPACITY) {
             int entriesToRemove = cacheSizeCurrent - CACHE_SIZE;
-            ConcurrentHashMap<String, ConcurrentHashMap<String, CryptoMetadata>> newMap = new ConcurrentHashMap<>();
-            ConcurrentHashMap<String, ConcurrentHashMap<String, CryptoMetadata>> oldMap = cache.getParamMap();
+            ConcurrentLinkedHashMap<String, ConcurrentLinkedHashMap<String, CryptoMetadata>> newMap = new Builder<String, ConcurrentLinkedHashMap<String, CryptoMetadata>>()
+                    .maximumWeightedCapacity(MAX_WEIGHTED_CAPACITY).build();
+            ConcurrentLinkedHashMap<String, ConcurrentLinkedHashMap<String, CryptoMetadata>> oldMap = cache
```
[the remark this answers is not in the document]
Jeffery-Wasty: Yeah, a lot of this is code from the original implementation that I sort of built upon. You're right, this can be improved.

tkyc (dismissed):

src/main/java/com/microsoft/sqlserver/jdbc/ParameterMetaDataCache.java
Jeffery-Wasty: Ok this was dumb. Removed and resolved.

src/test/java/com/microsoft/sqlserver/jdbc/AlwaysEncrypted/ParameterMetaDataCacheTest.java
```diff
@@ -0,0 +1,163 @@
...
+            long firstRun = timedCharUpdate(charValues);
+            populateNumeric(numericValues);
+            long secondRun = timedCharUpdate(charValues);
+
+            double threshold = 0.15;
+            // AEv2 does not support caching
+            if (!TestUtils.isAEv2(con)) {
+                assertTrue(1 - (secondRun / firstRun) > threshold);
+            }
+            con.close();
```
lilgreenbird, line 74: this close isn't necessary but I'm seeing warnings in eclipse on other leaks, please take a look

lilgreenbird (dismissed):

tkyc (approved):

lilgreenbird (dismissed):

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -855,6 +855,9 @@ boolean getServerSupportsColumnEncryption() {
     ColumnEncryptionVersion getServerColumnEncryptionVersion() {
         return serverColumnEncryptionVersion;
     }
+    
+    /** whether the server supports retrying a connection on failure */
+    private boolean serverSupportsRetry = false;
```
David-Engel, line 860: Minor NIT: The variable name should be more specific. Maybe serverSupportsEnclaveRetry?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -7673,6 +7678,10 @@ final synchronized void removeOpenStatement(ISQLServerStatement st) {
     boolean isAEv2() {
         return (aeVersion >= TDS.COLUMNENCRYPTION_VERSION2);
     }
+    
+    boolean isRetrySupported() {
```
David-Engel, line 7682: NIT: Similar to above. This seems too generic for what it actually applies to.

Jeffery-Wasty: I've changed this to doesServerSupportEnclaveRetry()

v-zhangw (approved):

lilgreenbird (approved):

v-zhangw (approved):

tkyc (approved):

PR #1869, opened 2022-07-14
Ensure that batchParamValues is cleared in all cases when executing a batch

### LINKED ISSUE #1767: Batch not cleared on failure when using bulk update

ianroberts:
This is an alternative implementation to #1768 which moves the `batchParamValues = null` to a new top-level try/finally block.

Closes #1767

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
@@ -2042,112 +2042,111 @@ public class SQLServerPreparedStatement extends SQLServerStatement implements IS
         checkClosed();
         discardLastExecutionResults();
 
-        int updateCounts[];
-
-        localUserSQL = userSQL;
-
         try {
-            if (this.useBulkCopyForBatchInsert && isInsert(localUserSQL)) {
-                if (null == batchParamValues) {
-                    updateCounts = new int[0];
-                    loggerExternal.exiting(getClassNameLogging(), "executeBatch", updateCounts);
-                    return updateCounts;
-                }
+            int updateCounts[];
 
-                // From the JDBC spec, section 9.1.4 - Making Batch Updates:
-                // The CallableStatement.executeBatch method (inherited from PreparedStatement) will
-                // throw a BatchUpdateException if the stored procedure returns anything other than an
-                // update count or takes OUT or INOUT parameters.
-                //
-                // Non-update count results (e.g. ResultSets) are treated as individual batch errors
-                // when they are encountered in the response.
-                //
-                // OUT and INOUT parameter checking is done here, before executing the batch. If any
-                // OUT or INOUT are present, the entire batch fails.
-                for (Parameter[] paramValues : batchParamValues) {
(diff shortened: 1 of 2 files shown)


Jeffery-Wasty:
Hi @ianroberts,

Just a comment. We haven't forgotten about this, but for the time being priority is being given to preparing for our upcoming release. Expect to see this addressed shortly thereafter.

Jeffery-Wasty:
I've cleaned up the tests slightly, but otherwise this perfectly reproduces, resolves, and confirms resolution of, the issue. Thank you so much!

Jeffery-Wasty (dismissed):

Jeffery-Wasty:
Almost ready, can we ask that you run the files through your ide, formatting using the format files included at the root. We think there may be some minor spacing issues.

ianroberts:
I don't have Eclipse (I'm an IntelliJ user) but I was able to apply the `mssql-jdbc_formatter.xml` format using https://code.revelc.net/formatter-maven-plugin/ - it changed 117 files in total but most of them were just cases where it stripped trailing spaces off lines (mostly blank lines in the middle of Javadoc comments).  It made one indentation change to the test file I touched in this PR so I've committed that change now.

Jeffery-Wasty:
Thank you so much. With that I can ask the rest of the team to review, and we can finally get this merged.

Jeffery-Wasty (approved):

tkyc:
We'll need to run the PR through some of our internal pipelines.

Jeffery-Wasty (approved):

lilgreenbird (approved):

tkyc (approved):

PR #1870, opened 2022-07-15
Added TDS version 8 for strict mode

lilgreenbird:


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -6038,7 +6038,10 @@ final boolean complete(LogonCommand logonCommand, TDSReader tdsReader) throws SQ
         int dataLen = 0;
 
         // Denali --> TDS 7.4, Katmai (10.0) & later 7.3B, Prelogin disconnects anything older
-        if (serverMajorVersion >= 11) {
+        // TDS version 8 if strict mode
+        if (serverMajorVersion >= 17 && encryptOption.compareToIgnoreCase(EncryptOption.Strict.toString()) == 0) {
```
David-Engel, line 6042: Don't limit this to a specific server version.

1. This will certainly be extended to Azure SQL databases in the future.
2. I suspect (I don't know for sure) TDS 8 might even be backported to previous SQL versions in the future if TLS 1.2 is no longer considered secure before those versions fall out of support.

I would probably rename VER_SQL2022 to VER_TDS80, too.

tkyc (dismissed):

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -150,9 +150,11 @@ private ExtendedSocketOptions() {}
 
 
 final class TDS {
-    // TDS protocol versions
-    static final String VER_TDS80 = "tds/8.0"; // TLS-first connections
+    // application protocol
+    static final String PROTOCOL_TDS80 = "tds/8.0"; // TLS-first connections
 
+    // TDS versions
+    static final int VER_TDS80 = 0x8000005; // TDS 8.0
```
Jeffery-Wasty, line 157: All the others use the SQL Server codename, should TDS8 follow suit (and use `Dallas`)?

Jeffery-Wasty: Hmm actually, this might be fine since there's only one version, not multiple like 7.2, 7.4. If an 8.x comes out we can change it, otherwise it should be fine.

Jeffery-Wasty (approved):

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
Jeffery-Wasty: I see, thanks for the explanation!

David-Engel (approved):

tkyc (approved):

Jeffery-Wasty (approved):

PR #1875, opened 2022-07-21
Added fine logging for access token expiry

tkyc:
I also saw a test with unsued imports so I removed those in this PR as well.


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerMSAL4JUtils.java
```diff
@@ -58,6 +58,11 @@ static SqlFedAuthToken getSqlFedAuthToken(SqlFedAuthInfo fedAuthInfo, String use
                     .build());
 
             final IAuthenticationResult authenticationResult = future.get();
+
+            if (logger.isLoggable(Level.FINE)) {
+                logger.fine(logger.toString() + "Access token expires on the following date: " + authenticationResult.expiresOnDate());
```
lilgreenbird, line 63: should change to FINEST level

Jeffery-Wasty (approved):

lilgreenbird (approved):

PR #1878, opened 2022-07-27
Fix sending TDS8 version during login

David-Engel:
Also fix logging from unit tests

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
@@ -6224,7 +6224,7 @@ public class SQLServerConnection implements ISQLServerConnection, java.io.Serial
         tdsWriter.writeShort((short) 0);
         tdsWriter.writeShort((short) 0);
 
-        if (tdsVersion >= TDS.VER_YUKON) {
+        if (tdsVersion >= TDS.VER_YUKON || tdsVersion == TDS.VER_TDS80) {
             // TDS 7.2: Password change
             tdsWriter.writeShort((short) 0);
             tdsWriter.writeShort((short) 0);
--- a/src/test/java/com/microsoft/sqlserver/testframework/AbstractTest.java
+++ b/src/test/java/com/microsoft/sqlserver/testframework/AbstractTest.java
@@ -121,9 +121,6 @@ public abstract class AbstractTest {
      */
     @BeforeAll
     public static void setup() throws Exception {
-        // Invoke fine logging...
-        invokeLogging();
-
         // get Properties from config file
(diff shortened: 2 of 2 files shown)


tkyc (dismissed):

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -6224,7 +6224,7 @@ final boolean complete(LogonCommand logonCommand, TDSReader tdsReader) throws SQ
         tdsWriter.writeShort((short) 0);
         tdsWriter.writeShort((short) 0);
 
-        if (tdsVersion >= TDS.VER_YUKON) {
+        if (tdsVersion >= TDS.VER_YUKON || tdsVersion == TDS.VER_TDS80) {
```
Jeffery-Wasty, line 6227:
```suggestion
        if (tdsVersion >= TDS.VER_TDS80) {
```
Since TDS.VER_TDS80 is our new lowest version, can we simplify to the above? TDS versions listed below:

```
    static final int VER_TDS80 = 0x8000000; // TDS 8.0
    static final int VER_DENALI = 0x74000004; // TDS 7.4
    static final int VER_KATMAI = 0x730B0003; // TDS 7.3B(includes null bit compression)
    static final int VER_YUKON = 0x72090002; // TDS 7.2
    static final int VER_UNKNOWN = 0x00000000; // Unknown/uninitialized
```

David-Engel: Ugh. I just noticed there are other places we are checking tdsVersion >= CONSTANT so the new TDS80 version int breaks all that logic. More changes are needed.

David-Engel: Also, it's not that simple. Older TDS versions from server responses not listed could look like these:
```
            // Server responds:
            // 0x07000000 -> 7.0         // Notice server response format is different for bwd compat
            // 0x07010000 -> 2000 RTM     // Notice server response format is different for bwd compat
            // 0x71000001 -> 2000 SP1
            // 0x72xx0002 -> 2005 RTM
```
The JDBC driver just got lucky that its range of supported TDS versions happened to align nicely over the past 2 decades. It's much more [complex in SqlClient, for example](https://github.com/dotnet/SqlClient/blob/v4.1.0/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs#L4009).

I don't think I want to refactor the logic significantly at this point in time.

Jeffery-Wasty: I see okay, thanks!

lilgreenbird (approved):

Jeffery-Wasty (approved):

Issue #1896, opened 2022-08-12
Canceling "insert into" statement closes connection


thomek:
## Driver version
11.2.0

## SQL Server version
Microsoft SQL Server 2017 (RTM-GDR) (KB4505224) - 14.0.2027.2 (X64)   Jun 15 2019 00:26:19   Copyright (C) 2017 Microsoft Corporation  Developer Edition (64-bit) on Windows 10 Enterprise 2016 LTSB 10.0 <X64> (Build 14393: ) (Hypervisor) 

## Client Operating System
OS Name	Microsoft Windows 10 Enterprise 2016 LTSB
Version	10.0.14393 Build 14393
System Type	x64-based PC

## JAVA/JVM version
Eclipse Adoptium jdk-17.0.1.12-hotspot

## Table schema
```
create table test_table (column_name bit)
```

## Problem description
1. Prepare an "insert into" table `Statement`
2. Call `Statement.executeBatch`
3. Start to iterate over the statement's `ResultSet`
4. Cancel the statement via a concurrent thread
5. Iterating over the statement's results may now fail with an exception
6. If iterating over the results does not fail with an exception, then commit the transaction

## Expected behavior
a. Either an exception in step 5. telling me that the query was canceled or
b. a successful commit of the transaction in step 6.

The behavior may depend on timing. For each run I expect to see either a. or b..

## Actual behavior
I frequently get:

c. the commit in step 6. fails because the connection got closed

## Error message/stack trace
These are the last lines from the console output as attached under "JDBC trace logs" when executing the test provided in "Any other details that can be helpful":

```
delayInMilliseconds=2, numberOfCommands=2: canceled
Aug 12, 2022 8:52:14 AM com.microsoft.sqlserver.jdbc.TDSParser throwUnexpectedTokenException
SEVERE: TDSReader@34 (ConnectionID:1 ClientConnectionId: 9b4d1828-a2de-495b-84ca-39cd32b9900e): batch completion: Encountered unexpected TDS_RETURN_VALUE (0xAC)
Aug 12, 2022 8:52:14 AM com.microsoft.sqlserver.jdbc.TDSReader throwInvalidTDSToken
SEVERE: TDSReader@34 (ConnectionID:1 ClientConnectionId: 9b4d1828-a2de-495b-84ca-39cd32b9900e) got unexpected value in TDS response at offset:18
Aug 12, 2022 8:52:14 AM com.microsoft.sqlserver.jdbc.SQLServerException logException
[...]

PR #1897, opened 2022-08-15
Fixed query cancellation bug that intermittently occurs in batch queries

tkyc:
Fixes https://github.com/microsoft/mssql-jdbc/issues/1896

doExecutePreparedStatement() isn't affected because it doesn't execute TDS requests one after the other eg. It doesn't do batching.


src/test/java/com/microsoft/sqlserver/jdbc/TestResource.java
```diff
@@ -201,5 +201,6 @@ protected Object[][] getContents() {
             {"R_objectNullOrEmpty", "The {0} is null or empty."},
             {"R_cekDecryptionFailed", "Failed to decrypt a column encryption key using key store provider: {0}."},
             {"R_connectTimedOut", "connect timed out"},
+            {"R_queryCancelled", "The query was canceled."},
```
Jeffery-Wasty, line 204: Spelling consistency (cancelled vs canceled).

Jeffery-Wasty (approved):

lilgreenbird (approved):

PR #1900, opened 2022-08-22
Update SQLServerStatement.java

### LINKED ISSUE #1899: [JDBC]Data inconsistency caused by JDBC calling stored procedures
[JDBC]Data inconsistency caused by JDBC calling stored procedures
## Driver version
All versions.

## SQL Server version
All versions.

## Problem description
I have a table and a stored procedure as bellow:

CREATE TABLE [dbo].[TEST_USER](
	[ID] [bigint] NOT NULL,
	[USER_NAME] [nvarchar](50) NULL,
	[USER_PWD] [nvarchar](50) NULL,
 CONSTRAINT [PK_TEST_ACTIVITY] PRIMARY KEY CLUSTERED 
(
	[ID] ASC
)WITH (PAD_INDEX = OFF, STATISTICS_NORECOMPUTE = OFF, IGNORE_DUP_KEY = OFF, ALLOW_ROW_LOCKS = ON, ALLOW_PAGE_LOCKS = ON) ON [PRIMARY]
) ON [PRIMARY]
GO

-- Create unique key for the user_name field.
CREATE UNIQUE NONCLUSTERED INDEX [IX_TEST_ACTIVITY] ON [dbo].[TEST_USER]
(
	[USER_NAME] ASC
)WITH (PAD_INDEX = OFF, STATISTICS_NORECOMPUTE = OFF, SORT_IN_TEMPDB = OFF, IGNORE_DUP_KEY = OFF, DROP_EXISTING = OFF, ONLINE = OFF, ALLOW_ROW_LOCKS = ON, ALLOW_PAGE_LOCKS = ON) ON [PRIMARY]
GO

CREATE PROCEDURE [dbo].[PROC_MS_JDBC_TEST]
AS   
BEGIN   
  SET XACT_ABORT ON  
  BEGIN TRANSACTION  
	INSERT INTO TEST_USER(ID,USER_NAME,[USER_PWD])
	VALUES(1,'unique_error_name','1');
       -- The second insert statement will raise an exception.
	INSERT INTO TEST_USER(ID,USER_NAME,[USER_PWD])
	VALUES(2,'unique_error_name','2');	  
 COMMIT  

 END 

My jdbc code like this:

public class SQLServerJdbcTest {

    public static void main(String[] args)  {
        try {

            Connection connection = DriverManager.getConnection("jdbc:sqlserver://xxx.xxx.x.xx:1433;database=xx;encrypt=false;", "xx", "xxxxxxx");

            CallableStatement stat = null;
            try{
                stat = connection.prepareCall("{call PROC_MS_JDBC_TEST}");
                stat.executeUpdate();
                System.out.println("{call PROC_MS_JDBC_TEST} command executed successfully.");
            }catch (Exception ex){
                ex.printStackTrace();
            } finally {
                try {
                    if (stat != null) {
                        stat.close();
                    }
                } catch (SQLException e) {
                    e.printStackTrace();
                } catch (NullPointerException e){
                    e.printStackTrace();
                }
            }
        }catch (Exception ex){
            ex.printStackTrace();
        }
    }
}

## Expected behavior
In my java code,an exception that from the database will be catched.
## Actual behavior
Executing the java code above,i can not get any exception.  In other words, the Java code is executed successfully, but the database actually has an exception. The behavior of JDBC is somewhat strange and may cause inconsistent data of the system.

Through debugging the trace code, I found that the message returned by the stored procedure to the client contains multiple results. Whether the JDBC client throws an error depends on the message parsing. The parsing is handled through a while loop. When the JDBC code does not detect the database exception in the condition of exiting the loop for the first time, the client code will not catch the exception.
In the stored procedure, you can avoid returning multiple results to the JDBC client by adding a “SET NOCOUNT ON” statement.If we do not encounter similar problems, no one knows that we must deal with them in this way.

winchannelBj:
fixed #1899  When executing the stored procedure, force the return of true in the dodone method of the nextresult class to allow the message parsing to continue.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerStatement.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerStatement.java
@@ -1520,7 +1520,7 @@ public class SQLServerStatement implements ISQLServerStatement {
                     if (EXECUTE_BATCH != executeMethod) {
                         // Always return update counts from statements executed in stored procedure calls.
                         if (null != procedureName)
-                            return false;
+                            return true;
 
                         // Always return all update counts from statements executed through Statement.execute()
                         if (EXECUTE == executeMethod)


Jeffery-Wasty:
Associated issue #1899 marked as resolved, PR is not needed. Closing.

Issue #1489, opened 2020-12-22
Math operations with decimal data type returning incorrect values when using PreparedStatement and BigDecimal objects


pmiguensdenodo:
## Driver version
mssql-jdbc-7.2.2

## SQL Server version
2016

## Client Operating System
Windows

## JAVA/JVM version
Java8 / Java 11

## Table schema
create table test_decimal (
	test_column decimal(10,5)
)
insert into test_decimal values (99999.12345)

## Problem description
When using prepared statements with parameters (?) and setting BigDecimal objects for those parameters, the result of math operations is unexpected, truncating the value to the precision of the BigDecimal object that was set.

The follwing code  projects:

1. the original test_column, which has precision 10 and scale 5
2. a subtraction between test_column and a parameter ?. The output of this operation should have the same precision and scale as test_column.
3. a case, which has a first then with a ? and a else branch with test_column. The output of this case should be test_column.

       try (PreparedStatement stmt = connection.prepareStatement("
               SELECT test_column,
                (test_column - ?) as subtraction_example, 
                case test_column when 0 then ? else test_column end as case_example from test_decimal
        ")) {

            BigDecimal value = new BigDecimal("1.5");
            stmt.setObject(1, value);

            MathContext mc = new MathContext(2);
            BigDecimal value2 = new BigDecimal(1.5, mc);
            stmt.setObject(2, value2);

            try (ResultSet rs = stmt.executeQuery()) {
                rs.next();
                System.out.println("test_column: " + rs.getObject(1));
                System.out.println("subtraction_example:" + rs.getObject(2));
                System.out.println("case_example:" + rs.getObject(3));
            }
        } catch (Exception e) {
            e.printStackTrace();
        }

The ? are replaced with BigDecimal objects with precision = 2 and scale = 1.
The output of the code is:
test_column: 99999.12345
subtraction_example:99997.6
case_example:99999.1

It should be:
test_column: 99999.12345
subtraction_example:99997.62345
case_example:99999.12345

Note that the expected values are obtained with this query 
 SELECT test_column,
                (test_column - 1.5) as subtraction_example, 
                case test_column when 0 then 1.5 else test_column end as case_example from test_decimal

lilgreenbird:
hi @pmiguensdenodo ,

we will investigate and get back to you, thanks

lilgreenbird:
hi @pmiguensdenodo ,

[...]

Issue #942, opened 2019-01-24
[BUG] update decimal type field. the original value round up


seektuzi:
## Driver version
mssql-jdbc-6.4.0.jre8.jar

## SQL Server version
mssql 2016
## Client Operating System
redhat 7.0

## JAVA/JVM version
java version "1.8.0",

## Table schema
default(dbo | public)

## Problem description
i have one script to update some decimal columns. 
the script is like below:
Update MyTable set r1_total=r1_total+'8.68000' ,r2_total=r2_total+'0' , ...  where key = MyKey
 both r1_total,r2_total are decimal(31,5).
once i this script was executed.  the r2_total was changed from 7.68000 to 8.00000. 

1. Expected behaviour:
if add '0'.  the value shouldn't be changed. 
2. Actual behaviour:
after add '0'. the original value round up
3. Error message/stack trace:
Null
4. Any other details that can be helpful:
Null
## JDBC trace logs
Null
## Reproduction code
Null

seektuzi:
Hi @peterbae, thanks for you reply. as the java code is from the vendor. so i can not sure i can get the java code or not. but  as tested, we found some more information.  
in this case. the original value in db for r2_total is 7.68.  and the field definition is decimal(31,5).  but once we do update operation like update table set r2_total+'0'. the value will become 8. if my update operation is update table set r2_total+'0.000000' . the value will be 7.68. it is working as design.  it is surprised me that the precision is depend on the value we pass. but is not according to the field definition in DB.

peterbae:
Hi @seektuzi, I'm still not fully sure how this code is being executed using our driver (through statement/pstmt, etc), but it sounds like the behavior is intended (also this sounds like database behavior, instead of our driver's behavior). I'm going to close this issue for now, but please feel free to re-open it if you need further assistance.

peterbae:
Hi @seektuzi, we looked into this issue a bit more, and we were able to reproduce the behavior. However, we think that this is an intended behavior as it was only reproducible when we were explicitly passing a BigDecimal value to the server. Using a BigDecimal value makes the driver explicitly send a decimal(p,s) value to the driver, and the SQL Server will round up the existing value if the user passes a decimal value with a lower precision than the precision of the existing value in the table accordingly.

[...]

PR #1912, opened 2022-09-03
BigDecimal Precision/Scale Fix

Jeffery-Wasty:
For operations with `BigDecimal`, precision and scale of 38 and 1 respectively (`decimal(38,1)`) are used, as opposed to the actual values for the object. This leads to loss of accuracy in transactions involving `BigDecimal`. This changes the `param.typeDefinition` to match the precision from the `BigDecimal` object. Four tests to confirm the fix for subtraction, addition, multiplication, and division operations are also present. This fixes issues [1489](https://github.com/microsoft/mssql-jdbc/issues/1489) and [942](https://github.com/microsoft/mssql-jdbc/issues/942).

Parameter.java
- 535 - 540: Given a `BigDecimal` object, whose value is not `null`, and whose precision is larger than the present scale (default `1`), assigns `param.typeDefinition` a precision based on the object, as opposed to the maximum value (`38`) previously used.

StatementTest.java
- 1483 - 1657: Tests confirming the fix works for operations involving subtraction, addition, multiplication, and division.


src/main/java/com/microsoft/sqlserver/jdbc/Parameter.java
```diff
@@ -531,8 +532,14 @@ private void setTypeDefinition(DTV dtv) {
                         if (userProvidesPrecision) {
                             param.typeDefinition = "decimal(" + valueLength + "," + scale + ")";
```
lilgreenbird, line 533: I know this isn't in the PR but why is "decimal" hardcoded instead of using SSType like the others?

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/StatementTest.java
```diff
@@ -1479,6 +1480,181 @@ public void testResultSetErrors() throws Exception {
...
+        /**
+         * Tests result of math operation in prepared statement using subtraction
+         * 
+         * @throws SQLException
+         *         when an error occurs
+         */
+        @Test
+        public void testMathBigDecimalSubtraction() throws SQLException {
+            try (Connection con = getConnection(); Statement stmt = con.createStatement()) {
+                stmt.executeUpdate("create table " + AbstractSQLGenerator.escapeIdentifier(tableName)
```
lilgreenbird, line 1492: instead of calling AbstractSQLGenerator.escapeIdentifier everywhere can we just do this once where tableName is defined

src/main/java/com/microsoft/sqlserver/jdbc/Parameter.java
Jeffery-Wasty: Not sure, it may have been easier to concatenate that way initially. I can swap it out for SSType.DECIMAL.toString() (there are only 8 occurrences like the above).

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/StatementTest.java
Jeffery-Wasty: I don't see why not. Every instance of `tableName` uses it in the context of `AbstractSQLGeneratior.escapeIdentifier(tableName)`, so this would help for clarity sake. I'll try it out.

Jeffery-Wasty:
I've applied the formatting change with `tableName` to other similar variables, where the only use case was `AbstractSQLGenerator.escapeIdentifier`. This can be rolled back if not appropriate for this PR.

tkyc (approved):

lilgreenbird (approved):

PR #1923, opened 2022-09-22
Fix cache account name casing issue

kassak:
See https://youtrack.jetbrains.com/issue/DBE-13085
The problem is that server reply is cached and user input is searched
We've been able to track the situation where user entered `JohnnyCash@folsom.org` and cache contained `johnnycash@folsom.org` -> cache miss
I've done the quickfix by using case-insensitive comparison.
I'm not aware if the logins are truly case-insensitive. Feel free to make the fix of your own :)

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerMSAL4JUtils.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerMSAL4JUtils.java
@@ -168,6 +168,16 @@ class SQLServerMSAL4JUtils {
             // try to acquire token silently if user account found in cache
             try {
                 Set<IAccount> accountsInCache = pca.getAccounts().join();
+                if (logger.isLoggable(Level.FINE)) {
+                    StringBuilder acc = new StringBuilder();
+                    if (accountsInCache != null) {
+                        for (IAccount account : accountsInCache) {
+                            if (acc.length() != 0) acc.append(", ");
+                            acc.append(account.username());
+                        }
+                    }
+                    logger.fine(logger.toString() + "Accounts in cache = " + acc + ", size = " + (accountsInCache == null ? null : accountsInCache.size()) + ", user = " + user);
+                }
                 if (null != accountsInCache && !accountsInCache.isEmpty() && null != user && !user.isEmpty()) {
                     IAccount account = getAccountByUsername(accountsInCache, user);
                     if (null != account) {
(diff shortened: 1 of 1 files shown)


Jeffery-Wasty:
We'll take a look into this, thank you.

Jeffery-Wasty:
Hi @kassak,

An update. We're not 100% sure on the behavior of MSAL4J and case-sensitivity. Depending on the answer we receive from the team, this may or may not be a JDBC issue. We'll keep you updated on the status of the issue.

kassak:
I'd like to note that the problem may lay deeper:
1. You provide the email with some casing
2. Server responds with the token, with the email with another casing
So the server may do case-insensitive matching or transform the case in the response

Jeffery-Wasty:
The reply from MSAL is that the items cached *are* case-sensitive, and so we shouldn't be looking up with equalsIgnoreCase. You explicitly describe an issue where you looked up `JohnnyCash@folsom.org` and cache contained `johnnycash@folsom.org`. How do you know this is the problem? Reading the thread, I don't see where you mention this is the problem.

Jeffery-Wasty:
If there is some sort of case mismatch between driver and msal, we can look into it more. But for now (1) it doesn't appear to be an MSAL problem, and (2) we need to confirm this caused the original issue mentioned in the thread [MFA Azure support causes multiple browser tabs to open]

kassak:
While investigating the issue, I've created a patched jar with extended logging (also in my patch)
Got this output (user email is replaced with, dummy, but the case is preserved)
First time login:
> Accounts in cache = , size = 0, user = user@example.com
Further attempts
> Accounts in cache = User@example.com, size = 1, user = user@example.com

So the cached e-mail, thus the one returned by server is `User@example.com` and the one provided by user is `user@example.com`

Jeffery-Wasty:
We haven't been able to look at this during the last week. We'll try looking at this as soon as we can. I'm still a bit confused on how this links to the original issue you posted. Is it that, because of the casing, users are not being authenticated properly, and this causes a window pop up asking them to reauthenticate?

kassak:
Nope, on the contrary. The user authenticates ok, using CamelCase login, but the reply contains it in lower case. So on the next connection that reply is not found in cache and authentication is requested once again. Making cache case insensitive helped our users

Jeffery-Wasty:
Okay, thank you for clearing that up.

Jeffery-Wasty:
I wasn't able to do this on my end, can you remove the logging? Just the casing fix would be all that is needed here.

lilgreenbird:
hi @kassak 

we can not have this change in the driver as usernames are case sensitive in linux/mac they could be different users. I have confirmed this is a bug (see https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/578) in the MSAL library we will wait for this to be fixed there.

lilgreenbird:
re-opening PR as Azure ActiveDirectory user names are not case sensitive and MSAL library converts to lower case

tkyc (approved):

Jeffery-Wasty (approved):

PR #1925, opened 2022-09-22
fix #1924

### LINKED ISSUE #1924: When errors occur in a batch, only last one is visible, masking the original cause of the failure
When batch execution fails, we usually want to see the _first_ failed statement. The subsequent failures can be caused by the first one, but in the exception thrown the (irrelevant) information of only the last failure is available. 

Note the following line:

https://github.com/microsoft/mssql-jdbc/blob/05ef7c07d184e9f8b659c351f68599b1a94132e3/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerStatement.java#L1879

Our scenario:

1. We are using `SET IDENTITY_INSERT .. ON` in one batch with `INSERT`s to fill a table with `IDENTITY` fields.
2. Insertions fails somewhere in the middle of the batch because of constraint violation.
3. But what we see in exception is "Cannot insert explicit value for identity column", which has nothing to do with the real cause of the error.

Can we change the logic of error handling so that the first failure in the batch is saved and is used when building `BatchUpdateException`?

inponomarev:
A proposal to fix  #1924

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerStatement.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerStatement.java
@@ -1838,9 +1838,9 @@ public class SQLServerStatement implements ISQLServerStatement {
             for (int batchNum = 0; batchNum < batchSize; batchNum++)
                 updateCounts[batchNum] = Statement.EXECUTE_FAILED;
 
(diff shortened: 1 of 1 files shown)


Jeffery-Wasty:
Thank you for your submission. We've chosen to not move forward with this proposal.

Issue #1154, opened 2019-10-07
[FEATURE REQUEST] Make mssql jdbc loom friendly


cogman:
## Is your feature request related to a problem? If so, please give a short summary of the problem and how the feature would resolve it

Currently, sql server uses the synchronized keyword in several places.  Unfortunately, loom may (or may not) support the synchronized keyword pausing a fiber, instead it may end up pausing the fiber carrier thread (it is an open question).

See: https://cr.openjdk.java.net/~rpressler/loom/Loom-Proposal.html

and for where those places are

https://github.com/microsoft/mssql-jdbc/search?q=synchronized&unscoped_q=synchronized

## Describe the preferred solution
There is an easy solution here, instead of using the synchronized keyword, use j.u.c locks such as https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/locks/ReentrantLock.html 

These APIs will be guaranteed to pause fibers when they are hit.  Further, blocking calls such as Socket IO will also cause fibers to pause (rather than blocking the carrier thread).

## Reference Documentations/Specifications

https://github.com/microsoft/mssql-jdbc/search?q=synchronized&unscoped_q=synchronized
 https://cr.openjdk.java.net/~rpressler/loom/Loom-Proposal.html
 https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/locks/ReentrantLock.html

peterbae:
Hi @cogman, thanks for contacting us. It seems like Project Loom is still under active development. We'll keep this issue open for now, and let us know when we can confirm that the synchronized keyword will have a problem with loom, then we'll look into it.

cogman:
@peterbae Will do.

I opened this issue primarily because oracle is starting to advertise their driver as being "Fiber ready"

https://mail.openjdk.java.net/pipermail/jdbc-spec-discuss/2019-September/000529.html

What that means is really just "They've removed synchronized".

I'm watching the project pretty closely so I'll let ya'll know if this becomes more concrete.

AlbozDroid:
Any update on this? Project Loom is now in preview in JDK 16.

cogman:
Project loom is currently building with JDK 16.  There's still not a final release date.

As far as I've seen, it's still the case that synchronized blocks aren't handled.

jono-coder:
I can do this, but if the files I edit are changed from a formatting perspective (as well as the relevant code changes), is that an issue?

Can someone please give me some info ito how best to proceed?

[...]

PR #1931, opened 2022-10-05
Loom friendly synchronization.

jono-coder:
See: #1154

Replaced the traditional "synchronized(foo)" with Reentrant locks as Loom, which is a preview feature in jdk-19, "breathes" more easily with a Lock which can be parked, and performance is increased.


jono-coder:
@microsoft-github-policy-service agree [company=""]

jono-coder:

Jeffery-Wasty:
@jono-coder 

Not sure what's going on with the bot here, can you try replying with `@microsoft-github-policy-service agree company=""` instead?

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -3150,6 +3174,7 @@ final class SocketConnector implements Runnable {
     // a counter used to give unique IDs to each connector thread.
     // this will have the id of the thread that was last created.
     private static long lastThreadID = 0;
+    private static final Lock LOCK = new ReentrantLock();
```
cogman, line 3177: This could probably all be replaced and simplified with an `AtomicLong`

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -3207,15 +3232,20 @@ public String toString() {
...
+            if (lastThreadID == Long.MAX_VALUE) {
+                if (logger.isLoggable(Level.FINER))
+                    logger.finer("Resetting the Id count");
+                lastThreadID = 1;
+            } else {
+                lastThreadID++;
+            }
+            return lastThreadID;
+        } finally {
+            LOCK.unlock();
```
cogman, line 3247: All of this could be tidied up with `AtomicLong#updateAndGet`

cogman (approved):
Looks like you found a bunch of great cases in here.  Great work!

David-Engel:
Judging by the timeframes, it looks like the CLA bot might have been affected by a GitHub service issue:
https://github.com/microsoft/GitOps/issues/703

I'd try replying to the bot once more and see if it works now. After that, I'd try closing and re-opening this issue (close/re-open button - don't need to re-create the PR) and see if the license/cla integration gets reset and replying starts working.

jono-coder:

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
jono-coder: Fair enough.  I did this in another commit.

jono-coder:
@microsoft-github-policy-service agree company=""

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -5,28 +5,16 @@
...
-import java.io.InputStream;
-import java.io.OutputStream;
-import java.io.Reader;
-import java.io.Serializable;
-import java.io.UnsupportedEncodingException;
+import com.microsoft.sqlserver.jdbc.SQLServerConnection.FedAuthTokenCommand;
+import com.microsoft.sqlserver.jdbc.dataclassification.SensitivityClassification;
+
+import javax.net.SocketFactory;
+import javax.net.ssl.*;
```
tkyc, line 12: Just a convention with the driver project, correct all wildcard imports back to explicit imports.

jono-coder: Done.

tkyc (dismissed):

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDataTable.java
```diff
@@ -49,24 +51,34 @@ public SQLServerDataTable() throws SQLServerException {
...
      * 
      * @return an iterator on the rows of the data table.
      */
-    public synchronized Iterator<Entry<Integer, Object[]>> getIterator() {
-        if ((null != rows) && (null != rows.entrySet())) {
-            return rows.entrySet().iterator();
+    public Iterator<Entry<Integer, Object[]>> getIterator() {
+        lock.lock();
+        try {
+            if (null != rows) {
```
Jeffery-Wasty, line 75: Why is the check for `rows.entrySet() not equal to null` removed here?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
```diff
@@ -1087,23 +1093,31 @@ private ResultSet executeSPFkeys(String[] procParams) throws SQLException, SQLTi
...
-                    getImportedKeysDWColumns.put(11, DELETE_RULE);
-                    getImportedKeysDWColumns.put(12, FK_NAME);
-                    getImportedKeysDWColumns.put(13, PK_NAME);
-                    getImportedKeysDWColumns.put(14, DEFERRABILITY);
+
+            LinkedHashMap<Integer, String> importedKeysDWColumns = getImportedKeysDWColumns;
+            if (null == importedKeysDWColumns) {
+                LOCK.lock();
+                try {
+                    importedKeysDWColumns = getImportedKeysDWColumns;
```
Jeffery-Wasty, line 1101: Bit confused about what's happening here. We grabbed the value, entered the block if its null, and grab it again. Is it possible that the value has changed by this point? If not, then I don't see the point of 1101 and 1102. Why isn't something like this possible?

`LOCK.lock()`
`try {`
`    --Lines 1091 - 1106--`
`} finally {`
`    LOCK.unlock()`
`}`

Like you have for the other cases.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerStatement.java
```diff
@@ -2410,79 +2423,98 @@ public final String getResponseBuffering() throws SQLServerException {
...
+        }
     }
 
-    synchronized boolean hasColumnEncryptionKeyStoreProvidersRegistered() {
-        return null != statementColumnEncryptionKeyStoreProviders
-                && statementColumnEncryptionKeyStoreProviders.size() > 0;
+    boolean hasColumnEncryptionKeyStoreProvidersRegistered() {
+        lock.lock();
+        try {
+            return null != statementColumnEncryptionKeyStoreProviders && statementColumnEncryptionKeyStoreProviders.size() > 0;
```
Jeffery-Wasty, line 2490: This doesn't exceed line length (80 chars)?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerXAConnection.java
```diff
@@ -29,13 +29,16 @@ public final class SQLServerXAConnection extends SQLServerPooledConnection imple
...
      * the connections it issues to applications, not the driver. These instances can and must commit/rollback
      */
-    private SQLServerXAResource XAResource;
-    
-    /** physical connection */
+    private volatile SQLServerXAResource XAResource;
+
+    /**
+     * physical connection
+     */
```
Jeffery-Wasty, line 36: Can be a single line comment like line 39 below.

src/main/java/com/microsoft/sqlserver/jdbc/SharedTimer.java
```diff
@@ -106,14 +110,20 @@ public void removeRef() {
      * When the caller is finished with the SharedTimer it must be released via {@link#removeRef}
      */
     public static SharedTimer getTimer() {
-        synchronized (lock) {
-            if (instance == null) {
-                // No shared object exists so create a new one
-                instance = new SharedTimer();
+        SharedTimer result = instance;
```
Jeffery-Wasty, line 113: Similar comment to above, why can't this just be wrapped in a try / catch like the other changes?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerXAConnection.java
```diff
@@ -101,21 +104,32 @@ public final class SQLServerXAConnection extends SQLServerPooledConnection imple
...
     @Override
-    public synchronized XAResource getXAResource() throws java.sql.SQLException {
+    public XAResource getXAResource() throws java.sql.SQLException {
         // All connections handed out from this physical connection have a common XAResource
         // for transaction control. IE the XAResource is one to one with the physical connection.
-
-        if (XAResource == null)
-            XAResource = new SQLServerXAResource(getPhysicalConnection(), physicalControlConnection, toString());
-        return XAResource;
+        SQLServerXAResource result = XAResource;
```
Jeffery-Wasty, line 110: Similar to above, why can't this just be wrapped in a try / catch block like the other changes.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDataTable.java
jono-coder: The change probably "came out in the wash".  It's a redundant check as rows.entrySet() can never be null.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
jono-coder: You are correct.  This is an error.  I'll commit a change for this.

jono-coder: Well, it's not technically incorrect, but this is using double-checked locking (using a volatile).  Perhaps you don't like the idiom?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerStatement.java
jono-coder: I'll commit a change.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerXAConnection.java
jono-coder: I'll commit a revert.

src/main/java/com/microsoft/sqlserver/jdbc/SharedTimer.java
jono-coder: Refer to https://en.wikipedia.org/wiki/Double-checked_locking

"Note the local variable "localRef", which seems unnecessary. The effect of this is that in cases where helper is already initialized (i.e., most of the time), the volatile field is only accessed once (due to "return localRef;" instead of "return helper;"), which can improve the method's overall performance by as much as 40 percent."

One could use the VarHandles stuff for even better performance but that's even trickier.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerXAConnection.java
jono-coder: Refer to https://en.wikipedia.org/wiki/Double-checked_locking

"Note the local variable "localRef", which seems unnecessary. The effect of this is that in cases where helper is already initialized (i.e., most of the time), the volatile field is only accessed once (due to "return localRef;" instead of "return helper;"), which can improve the method's overall performance by as much as 40 percent."

One could use the VarHandles stuff for even better performance but that's even trickier.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerXAResource.java
```diff
@@ -211,73 +211,89 @@ public String toString() {
...
+                case SQLServerXAResource.XA_ROLLBACK_EX:
+                    CS = controlConnection.prepareCall("{call master..xp_sqljdbc_xa_rollback_ex(?, ?, ?, ?, ?, ?)}");
+                    break;
+                case SQLServerXAResource.XA_FORGET_EX:
+                    CS = controlConnection.prepareCall("{call master..xp_sqljdbc_xa_forget_ex(?, ?, ?, ?, ?, ?)}");
+                    break;
+                default:
+                    assert false : "Bad handle request:" + number;
+                    break;
+            }
```
lilgreenbird, line 259: are these tabs? please make sure to use the code [formatter](https://github.com/microsoft/mssql-jdbc/blob/main/mssql-jdbc_formatter.xml)

jono-coder: I'm not seeing tabs and am using the mssql-jdbc formatter.  AFAIK, I've never used tabs for any coding.

Jeffery-Wasty: GitHub makes it hard to tell if they are space or tabs, thanks for confirming that these are spaces.

lilgreenbird (approved):

Jeffery-Wasty (approved):

Issue #1930, opened 2022-10-04
[FEATURE REQUEST] Managed Identity dev experience improvements


scottaddie:
As discussed with @David-Engel in an Oct. 3rd meeting, leverage the Azure Identity library to address the following items:

- [ ] Replace Managed Identity token caching logic introduced in https://github.com/microsoft/mssql-jdbc/pull/1825. The Azure Identity library's 1.7.0 release, expected to GA in November, already supports in-memory caching via MSAL in the `ManagedIdentityCredential`. See https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/identity/azure-identity/CHANGELOG.md#170-beta1-2022-09-20.
- [ ] Expand the number of supported Azure compute services by relying on the Azure Identity library. For the list the Azure Identity library supports, see https://github.com/Azure/azure-sdk-for-java/tree/main/sdk/identity/azure-identity#managed-identity-support. Related to https://github.com/microsoft/mssql-jdbc/issues/1844.
- [ ] Improve the local dev experience by supporting both `DefaultAzureCredential` (related to https://github.com/microsoft/mssql-jdbc/issues/1881) and `IntelliJCredential` from the Azure Identity library.

g2vinay:
When Replace Managed Identity token caching logic, following needs to be taken care of:

1. The logic of Token TTL to refresh tokens needs to be retained when using Managed Identity credential directly.
2. When working with Managed Identity credential, for each cache key combination a new instance of Managed Identity Credential will be required.

Issue #1881, opened 2022-07-29
[QUESTION] Is DefaultAzureCredential supported


saragluna:
## Question
By looking at this https://github.com/microsoft/mssql-jdbc#azure-key-vault-and-azure-active-directory-authentication-dependencies, seems like the `azure-identity` is included when talking to Azure
AD. I am wondering whether the [DefaultAzureCredential](https://www.rahulpnath.com/blog/defaultazurecredential-from-azure-sdk/)  is supported.

stliu:
`DefaultAzureCredential` is not supported in the JDBC connection stage I think, actually by looking into the code, seems azure-identity is only used by the KeyVault Client auth.

From the [current code](https://github.com/microsoft/mssql-jdbc/blob/main/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java#L5425), I think the auth methods supported by the ms SQL jdbc driver are fixed, it is impossible to extend w/o changing the existing code.

I'm thinking to make it extensible, we can do the following change:

1. the [authentication property](https://docs.microsoft.com/en-us/sql/connect/jdbc/connecting-using-azure-active-directory-authentication?view=sql-server-ver16), we can make it accept one more type of value, which is a class FQDN name.
2. create an new interface, for example `FedAuthTokenResolver`
3. in [this code](https://github.com/microsoft/mssql-jdbc/blob/main/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java#L5425), at the end of that long **if-else** check, we add one more check to see if `authenticationString` is a class name which is on the classpath and implements the `FedAuthTokenResolver`

tkyc:
That's correct, at the moment the JDBC driver doesn't support `DefaultAzureCredential`.

Since the semester is coming to an end, and is soon to be concluded with a GA release, I'll bring this up with the team in our planning.

Thanks @stliu for the implementation insights.

saragluna:
Hi @tkyc and @lilgreenbird, any updates on this feature request? Will this be included in this semester?

tkyc:
Hey @saragluna, unfortunately this feature request didn't make it in this semester. We'll update you once it's out of the backlog.

tkyc:
Hey all, an upcoming preview with this feature will be released on Nov 3rd.

saragluna:
@tkyc thanks. Could you help point us to the doc where we can follow how to use the new feature?

tkyc:
[...]

Issue #1844, opened 2022-06-04
[QUESTION] Dependencies and Configuration needed for database connectivity using Manage Identity in AKS


dhananjay12:
## Question
We have a spring app that we want to deploy on Azure AKS connecting to Azure SQL server. While regular JDBC URL along with username/password works fine, we wanted to use `manage-identity` instead.  Our DevOps team implemented `user-assigned managed identity` on azure and a service account that injects the following env variable to each pod:
```
AZURE_CLIENT_ID: 
AZURE_TENANT_ID: 
AZURE_FEDERATED_TOKEN_FILE: 
AZURE_AUTHORITY_HOST: 
```

From spring app perspective we added `azure-idenitty` dependencies [within the app] (https://github.com/dhananjay12/learn-azure/blob/main/employee-service/pom.xml#L34-L42).

In addition we are adding following env variables to test:
```
SPRING_DATASOURCE_URL: jdbc:sqlserver://<sql-server>.database.windows.net:1433;database=<test-db>;encrypt=true;trustServerCertificate=false;hostNameInCertificate=*.database.windows.net;loginTimeout=30;Authentication=ActiveDirectoryMSI;
spring.jpa.hibernate.ddl-auto: create-drop
SPRING_DATASOURCE_DRIVER_CLASS_NAME: com.microsoft.sqlserver.jdbc.SQLServerDriver
SPRING_JPA_PROPERTIES_HIBERNATE_DIALECT: org.hibernate.dialect.SQLServer2012Dialect
```

However we are getting following exception:
```
2022-06-04 11:47:02.768 DEBUG 1 --- [           main[] c.m.s.jdbc.internals.SQLServerDriver     : Property:serverName Value:<sql-server>.database.windows.net
2022-06-04 11:47:02.771 DEBUG 1 --- [           main[] c.m.s.jdbc.internals.SQLServerDriver     : Property:portNumber Value:1433
2022-06-04 11:47:02.780 DEBUG 1 --- [           main[] c.m.s.jdbc.internals.SQLServerDriver     : Property:databaseName Value:<test-db>
2022-06-04 11:47:02.780 DEBUG 1 --- [           main[] c.m.s.jdbc.internals.SQLServerDriver     : Property:encrypt Value:true
2022-06-04 11:47:02.780 DEBUG 1 --- [           main[] c.m.s.jdbc.internals.SQLServerDriver     : Property:trustServerCertificate Value:false
[...]

PR #1936, opened 2022-10-18
Managed Identity dev experience improvements

tkyc:
PR Resolves the following:
- https://github.com/microsoft/mssql-jdbc/issues/1930
- https://github.com/microsoft/mssql-jdbc/issues/1881
- https://github.com/microsoft/mssql-jdbc/issues/1844

Deprecated the following:
- https://github.com/microsoft/mssql-jdbc/pull/1825

New authentication property values:
- `authentication=ActiveDirectoryManagedIdentity`
- `authentication=DefaultAzureCredential`

New environment variables:
- `INTELLIJ_KEEPASS_PATH`
- `ADDITIONALLY_ALLOWED_TENANTS` (comma delimited list of additionally allowed tenant IDs, used with `DefaultAzureCredential`)

To use the IntellijCredential with the driver, supply the connection string property `authentication=DefaultAzureCredential` and set the `INTELLIJ_KEEPASS_PATH` environment variable to the path of the keepass database.


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerSecurityUtility.java
```diff
@@ -319,191 +316,91 @@ static void verifyColumnMasterKeyMetadata(SQLServerConnection connection, SQLSer
...
-        }
+        Optional<AccessToken> accessTokenOptional = mic
+                .getToken(tokenRequestContext)
+                .blockOptional();
 
-        String identityHeader = System.getenv("IDENTITY_HEADER");
-        if (null == identityHeader || identityHeader.trim().isEmpty()) {
-            identityHeader = System.getenv("MSI_SECRET");
+        if (!accessTokenOptional.isPresent()) {
+            if (connectionlogger.isLoggable(java.util.logging.Level.FINE)) {
```
David-Engel, line 352: If there is no result, no error, and no token, we'll need to throw our own exception. We can either throw it here or in the calling method. The calling method (or this one) should never return a null result.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerSecurityUtility.java
```diff
@@ -319,191 +316,91 @@ static void verifyColumnMasterKeyMetadata(SQLServerConnection connection, SQLSer
...
-                        throw new SQLServerException(SQLServerException.getErrString("R_MSITokenFailureUnexpected"),
-                                null);
-                    }
-                }
-            } finally {
-                if (connection != null) {
-                    connection.disconnect();
-                }
+        if (!accessTokenOptional.isPresent()) {
+            if (connectionlogger.isLoggable(java.util.logging.Level.FINE)) {
```
David-Engel, line 405: Similar to the ManagedIdentity method, if we get no result and no error, we need to throw our own exception. We should never return a null result when getFedAuthToken is called.

David-Engel (commented):
There is one other change we need to make to align the JDBC driver with other drivers. We should deprecate the msiClientId get/set methods (but they should continue to work. Comment: "Use the getUser/setUser method instead.") This will include changing the Authentication=ActiveDirectoryMSI validation to allow specifying User in the connection string. If both msiClientId and User are specified, User should override msiClientId.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDriver.java
```diff
@@ -67,7 +67,8 @@ enum SqlAuthentication {
     ActiveDirectoryIntegrated,
     ActiveDirectoryMSI,
```
lilgreenbird, line 68: rename this to ActiveDirectoryManagedIdentity since that is the current feature name

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -199,6 +199,7 @@ final class TDS {
     static final byte ADALWORKFLOW_ACTIVEDIRECTORYINTEGRATED = 0x02;
     static final byte ADALWORKFLOW_ACTIVEDIRECTORYMSI = 0x03;
```
lilgreenbird, line 200: rename to ADALWORKFLOW_ACTIVEDIRECTORYMANAGEDIDENTITY

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerSecurityUtility.java
```diff
@@ -319,191 +319,109 @@ static void verifyColumnMasterKeyMetadata(SQLServerConnection connection, SQLSer
...
+                .getToken(tokenRequestContext)
+                .blockOptional();
 
-        String identityHeader = System.getenv("IDENTITY_HEADER");
-        if (null == identityHeader || identityHeader.trim().isEmpty()) {
-            identityHeader = System.getenv("MSI_SECRET");
-        }
+        if (!accessTokenOptional.isPresent()) {
+            if (connectionlogger.isLoggable(java.util.logging.Level.FINE)) {
+                connectionlogger.fine("Access token is not present.");
```
lilgreenbird, line 353: we're gonna throw, what's the point to log?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -7487,14 +7503,15 @@ public void setStatementPoolingCacheSize(int value) {
             parameterMetadataCache.setCapacity(value);
     }
 
+    @Deprecated
     @Override
     public int getMsiTokenCacheTtl() {
```
lilgreenbird, line 7508: perhaps we should add a comment why this is deprecated?

lilgreenbird (changes requested):
please run formatter looks like a lot of these files aren't formatted

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerSecurityUtility.java
```diff
@@ -319,191 +319,109 @@ static void verifyColumnMasterKeyMetadata(SQLServerConnection connection, SQLSer
...
 
-        /*
-         * isAzureFunction is used for identifying if the current client application is running in a Virtual Machine
-         * (without Managed Identity environment variables) or App Service/Function (with Managed Identity environment
-         * variables) as the APIs to be called for acquiring MSI Token are different for both cases.
-         */
-        boolean isAzureFunction = null != identityEndpoint && !identityEndpoint.isEmpty() && null != identityHeader
-                && !identityHeader.isEmpty();
+            throw new SQLServerException(SQLServerException
+                    .getErrString("R_ManagedIdentityTokenAcquisitionFail"), null);
```
lilgreenbird, line 357: can we add more info to the error message to make it easier to debug when there's a failure

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerSecurityUtility.java
```diff
@@ -319,191 +319,109 @@ static void verifyColumnMasterKeyMetadata(SQLServerConnection connection, SQLSer
...
+     * Get Managed Identity Authentication token through the DefaultAzureCredential
+     *
+     * @param resource
+     *        token resource
+     * @param msiClientId
+     *        Managed Identity or User Assigned Managed Identity
+     * @return fedauth token
+     * @throws SQLServerException
+     */
+    static SqlFedAuthToken getDefaultAzureCredAuthToken(String resource, String msiClientId) throws SQLServerException {
```
lilgreenbird, line 377: lets not use the terminalogy "msi" anymore since it's confusing.  Use new terminology "managedIdentity" or "mi" everywhere for consistency

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -5567,6 +5564,25 @@ private SqlFedAuthToken getFedAuthToken(SqlFedAuthInfo fedAuthInfo) throws SQLSe
 
                 // Break out of the retry loop in successful case.
                 break;
+            } else if (authenticationString.equalsIgnoreCase(SqlAuthentication.DefaultAzureCredential.toString())) {
+                if (!msalContextExists()) {
```
lilgreenbird, line 5568: I know this was there before, but can we just refactor this so we don't have to dup this now 5 times

src/main/java/com/microsoft/sqlserver/jdbc/ISQLServerConnection.java
```diff
@@ -409,17 +409,16 @@ CallableStatement prepareCall(String sql, int nType, int nConcur, int nHold,
     String getIPAddressPreference();
 
     /**
-     * Gets the time-to-live for the the cached MSI token
-     *
-     * @return time-to-live for the cached MSI token
+     * Deprecated. Time-to-live is no longer supported for the cached Managed Identity tokens.
+     * This method is a no-op for backwards compatibility only.
```
David-Engel, line 413:
```suggestion
     * This method will always return 0 and is for backwards compatibility only.
```

src/main/java/com/microsoft/sqlserver/jdbc/ISQLServerDataSource.java
```diff
@@ -932,18 +933,24 @@ public interface ISQLServerDataSource extends javax.sql.CommonDataSource {
     void setUseBulkCopyForBatchInsert(boolean useBulkCopyForBatchInsert);
 
     /**
+     * This method is deprecated. Use {@link ISQLServerDataSource#setUser(String user)} instead.
+     *
      * Sets the client id to be used to retrieve access token from MSI EndPoint.
```
David-Engel, line 938:
```suggestion
     * Sets the client id to be used to retrieve the access token for a user-assigned Managed Identity.
```

src/main/java/com/microsoft/sqlserver/jdbc/ISQLServerDataSource.java
```diff
@@ -932,18 +933,24 @@ public interface ISQLServerDataSource extends javax.sql.CommonDataSource {
...
      * 
      * @param msiClientId
      *        Client ID of User Assigned Managed Identity
```
David-Engel, line 941:
```suggestion
     *        Client ID of the user-assigned Managed Identity.
```

src/main/java/com/microsoft/sqlserver/jdbc/ISQLServerDataSource.java
```diff
@@ -1208,17 +1215,16 @@ public interface ISQLServerDataSource extends javax.sql.CommonDataSource {
...
      */
+    @Deprecated
     void setMsiTokenCacheTtl(int timeToLive);
 
     /**
-     * Gets the time-to-live for the the cached MSI token
-     *
-     * @return time-to-live for the cached MSI token
+     * Deprecated. Time-to-live is no longer supported for the cached Managed Identity tokens.
+     * This method is a no-op for backwards compatibility only.
```
David-Engel, line 1226:
```suggestion
     * This method will always return 0 and is for backwards compatibility only.
```

David-Engel (commented):
Partial review

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDriver.java
```diff
@@ -524,8 +530,7 @@ enum SQLServerDriverIntProperty {
     STATEMENT_POOLING_CACHE_SIZE("statementPoolingCacheSize", SQLServerConnection.DEFAULT_STATEMENT_POOLING_CACHE_SIZE),
     CANCEL_QUERY_TIMEOUT("cancelQueryTimeout", -1),
     CONNECT_RETRY_COUNT("connectRetryCount", 1, 0, 255),
-    CONNECT_RETRY_INTERVAL("connectRetryInterval", 10, 1, 60),
-    MSI_TOKEN_CACHE_TTL("msiTokenCacheTtl", 3600, 0, Integer.MAX_VALUE);
```
David-Engel, line 528: If we remove all the connection string-related code for msiTokenCacheTtl, will the driver complain about an unknown option? Or does it ignore unknown options? @lilgreenbird Do you know off the top of your head?

(If it doesn't ignore them, we would not want it to error until we actually remove the deprecated setting.)

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerSecurityUtility.java
David-Engel: Similarly, do we have more info if the Optional task fails or doesn't return a result? I don't know the API well...

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerSecurityUtility.java
```diff
@@ -319,191 +319,109 @@ static void verifyColumnMasterKeyMetadata(SQLServerConnection connection, SQLSer
...
-        // Loop while maxRetry reaches its limit
-        while (retry <= maxRetry) {
-            HttpURLConnection connection = null;
+    /**
+     * Get Managed Identity Authentication token through the DefaultAzureCredential
+     *
+     * @param resource
+     *        token resource
+     * @param msiClientId
+     *        Managed Identity or User Assigned Managed Identity
```
David-Engel, line 373:
```suggestion
     *        Client ID of the user-assigned Managed Identity

```

tkyc:
@David-Engel For some reason there's no "reply" button for your comment. I'll just quote you and reply here.

>Similarly, do we have more info if the Optional task fails or doesn't return a result? I don't know the API well...

If the Optional task fails, it will throw its own error. The resulting exception, since we're using the Azure Identity credentials, will be a `CredentialUnavailableException`.

A `CredentialUnavailableException` indicates the following:
1. Error in the provided information for the credential eg. wrong client ID of the user-assigned Managed Identity
2. Connection was dropped mid query for the token

Howerver, there isn't any more available information other than a null token if the Optional succeeds (eg. request for token went through but a null token was returned). In this case, as you suggested (since we shouldn't return null), we'll throw our own error.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerSecurityUtility.java
David-Engel: Not sure. From the sounds of it, we don't have anything more available. The current iteration looks good.

David-Engel (approved):

Jeffery-Wasty (approved):

lilgreenbird (approved):

PR #1965, opened 2022-11-11
Fix connecting to the wrong SQLServer host

### LINKED ISSUE #1964: The driver connects to the wrong SQLServer host
## Driver version
11.2.1

## Client Operating System
Linux

## Problem description
When the database driver is used to connect to different SQLServer hosts, it can happen that the TCP connection is established to the wrong host. This is due to a race-condition in com.microsoft.sqlserver.jdbc.SocketFinder#getSocketByIPPreference. The change was introduced in #1766.

com.microsoft.sqlserver.jdbc.SocketFinder#addressList is `static`, and this state is shared across all `SQLServerConnection` instances, even if connected to different SQLServer hosts. A race condition when mutating this state in `com.microsoft.sqlserver.jdbc.SocketFinder#getSocketByIPPreference` can lead to the connection being connected to an unexpected host.

mmimica:
Fix #1964 by not having a static field at all. It was cleared each time anyway.


lilgreenbird:
thanks for the contribution @mmimica, could you please add a test for this?

lilgreenbird (approved):

Jeffery-Wasty (approved):

tkyc (approved):

PR #1969, opened 2022-11-16
More fixes for SonarQube warnings and coding standards

lilgreenbird:
address warnings from SonarQube scan and also some changes to adhere to coding standasrd

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
@@ -3274,6 +3274,9 @@ final class SocketConnector implements Runnable {
  */
 final class TDSWriter {
     private static Logger logger = Logger.getLogger("com.microsoft.sqlserver.jdbc.internals.TDS.Writer");
+
+    private static final String UNEXPECTED_SSTYPE = "Unexpected SSType: ";
+
     private final String traceID;
 
     final public String toString() {
@@ -6017,7 +6020,7 @@ final class TDSWriter {
         assert con.isKatmaiOrLater();
 
         assert SSType.DATE == ssType || SSType.TIME == ssType || SSType.DATETIME2 == ssType
-                || SSType.DATETIMEOFFSET == ssType : "Unexpected SSType: " + ssType;
+                || SSType.DATETIMEOFFSET == ssType : UNEXPECTED_SSTYPE + ssType;
 
         // First, for types with a time component, write the scaled nanos since midnight
         if (SSType.TIME == ssType || SSType.DATETIME2 == ssType || SSType.DATETIMEOFFSET == ssType) {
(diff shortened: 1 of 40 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -1403,7 +1402,7 @@ public int getPort() {
         }
 
         @Override
-        public int getReceiveBufferSize() throws SocketException {
+        public synchronized int getReceiveBufferSize() throws SocketException {
```
tkyc, line 1405: Need to replace `synchronized` with Reentrant locks, or is this case an exception?

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -3502,7 +3495,7 @@ void writeByte(byte value) throws SQLServerException {
                 if (dataIsLoggable)
                     logBuffer.put(value);
                 else
-                    ((Buffer) logBuffer).position(((Buffer) logBuffer).position() + 1);
+                    logBuffer.position(logBuffer.position() + 1);
```
tkyc, line 3498: Are braces needed for the if/else? I can't recall if there were checks that complained about missing braces.

Jeffery-Wasty: Not needed, but I believe that's our code style. I'm surprised SonarQube didn't complain about it.

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
lilgreenbird: this is an override of Serializable

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
```diff
@@ -270,6 +270,11 @@ private void checkClosed() throws SQLServerException {
 
     private static final String IS_GENERATEDCOLUMN = "IS_GENERATEDCOLUMN";
     private static final String IS_AUTOINCREMENT = "IS_AUTOINCREMENT";
+    private static final String SELECT = "SELECT";
+    private static final String WHERE = " WHERE ";
+    private static final String WHERE_CLAUSE = " WHERE 0 = 1";
```
Jeffery-Wasty, line 275: This is not a helpful name, when reading where it's used. E.g: 
image
Can this be renamed WHERE_ZERO_IS_ONE

src/main/java/com/microsoft/sqlserver/jdbc/Parameter.java
```diff
@@ -505,34 +502,35 @@ private void setTypeDefinition(DTV dtv) {
                             // so, here, if the decimal parameter is encrypted and it is null and it is not outparameter
                             // then we set precision as the default precision instead of max precision
                             if (!isOutput()) {
-                                param.typeDefinition = "decimal(" + SQLServerConnection.defaultDecimalPrecision + ", "
-                                        + scale + ")";
+                                param.typeDefinition = SSType.DECIMAL.toString() + "("
+                                        + SQLServerConnection.defaultDecimalPrecision + ", " + scale + ")";
```
Jeffery-Wasty, line 506: I think this was a typo in the original implementation, and it should be `","` instead of `", "`. I don't see the extra space in any other param.typeDefinition.

Jeffery-Wasty (approved):

tkyc (approved):

PR #1970, opened 2022-11-16
Added access token callback tests

tkyc:


Jeffery-Wasty (dismissed):

David-Engel:
Cursory scan looks good. 👍 I'll let @lilgreenbird be the second reviewer, though. 😄

src/test/java/com/microsoft/sqlserver/jdbc/SQLServerConnectionTest.java
```diff
@@ -1030,4 +1031,64 @@ public void testServerNameField() throws SQLException {
...
+    @Tag(Constants.xSQLv14)
+    @Tag(Constants.xSQLv15)
+    @Tag(Constants.xAzureSQLDW)
+    @Tag(Constants.reqExternalSetup)
+    @Test
+    public void testDSPooledConnectionAccessTokenCallback() throws Exception {
+        SQLServerConnectionPoolDataSource ds = new SQLServerConnectionPoolDataSource();
+        AbstractTest.updateDataSource(connectionString, ds);
+        ds.setUser("");
+        ds.setPassword("");
```
lilgreenbird, line 1046: is this necessary?

tkyc: Yep, the connection string on the pipelines by default have the user and password set. Access token call back doesn't require a user and password. If user and password is set and access token call back is used, it will fail/throw exception.

lilgreenbird: I see, so this is just to get around setup in our test lab, not that it's required interface to use this? If that's the case, then why not just use TestUtils.addorOverrideProperty as in other tests?

src/test/java/com/microsoft/sqlserver/jdbc/SQLServerConnectionTest.java
```diff
@@ -1030,4 +1031,64 @@ public void testServerNameField() throws SQLException {
...
+        String conn2ID;
+
+        // When token expires after first connection, it should create a new connection to get a new token.
+        // Connection ID should not be the same.
+        TestUtils.expireTokenToggle = true;
+        pc = (SQLServerPooledConnection) ds.getPooledConnection();
+        try (Connection conn1 = pc.getConnection()) {}
+        conn1ID = TestUtils.getConnectionID(pc);
+        // Sleep until token expires
+        Thread.sleep(120000);
```
lilgreenbird, line 1088: token expiry is always 2mins?

src/test/java/com/microsoft/sqlserver/jdbc/SQLServerConnectionTest.java
tkyc: I'll switch to using that

src/test/java/com/microsoft/sqlserver/jdbc/SQLServerConnectionTest.java
tkyc: Oops, it's suppose to be 12 minutes.

tkyc:
Closing to restart mac-ci pipeline.

PR #1977, opened 2022-11-28
More code clean up to fix SonarQube issues

lilgreenbird:
also some coding standard fixes


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerStatement.java
```diff
@@ -329,6 +329,11 @@ void decrResultSetCount() {
     static final int EXECUTE_BATCH = 4;
     static final int EXECUTE_QUERY_INTERNAL = 5;
 
+    /**
+     * internal class name used in tracing
+     */
+    private static final String INTERNAL_CLASSNAME = "SQLServerStatement";
```
tkyc, line 335: We can just use `SQLServerStatement.class` no? No need to define a string

tkyc (approved):

Jeffery-Wasty (approved):

Issue #1971, opened 2022-11-18
SQLServerPreparedStatement uses a wrong handle after server restart - may cause wrong/dangerous statement execution 


sergiuo:
## Driver version
10.2.1

## SQL Server version
Microsoft SQL Server 2019 (RTM-CU18) (KB5017593) - 15.0.4261.1 (X64)   Sep 12 2022 15:07:06   

## Client Operating System
Windows 10

## JAVA/JVM version
11.0.10

## Table schema
none

## Problem description

SQLServerPreparedStatement.retryBasedOnFailedReuseOfCachedHandle is not properly detecting if the handle is still valid. 
After DB server restart server handle IDs are reset and the numbering starts from 1. So the handle ID from cache may refer to a completely different statement after server restart. The result is unpredictable and may lead to data loss/damaging.

----------sample code-----------------------
public class ZopaSqlTest {

    public static void main(String[] args) throws SQLException, IOException {
        String statement1 = "select SYSTEM_USER where SYSTEM_USER not in (?)";
        String statement2 = "select SYSTEM_USER where SYSTEM_USER not in (?, ?)";
        try(Connection connection = DriverManager.getConnection("jdbc:sqlserver://localhost:1433;databaseName=master;sendStringParametersAsUnicode=false;disableStatementPooling=false;statementPoolingCacheSize=999999;encrypt=false", "username", "password")){

            runPrepared2(connection, statement1, "1");
            runPrepared2(connection, statement2, "1", "2");
            runPrepared2(connection, statement1, "1");
            runPrepared2(connection, statement2, "1", "2");
            System.out.println("Restart DB Server and press enter...");
            System.in.read();
            runPrepared2(connection, statement2, "1", "2");
            runPrepared2(connection, statement1, "1");//here the statement2 is executed instead of statement1
        }
    }

    private static void runPrepared2(Connection connection, String sql, String... params) throws SQLException {
        try(PreparedStatement preparedStatement = connection.prepareStatement(sql)) {
            for (int i = 0; i < params.length; i++) {
                preparedStatement.setString(i + 1, params[i]);
            }
            preparedStatement.execute();
        }
    }

}
-----------------------------------

## Expected behavior

After restart the server handle ID should not be used. 

## Actual behavior

The the server handle ID should is used, a wrong statement is executed and a "missing parameter" exception is thrown. 

## Error message/stack trace

[...]

PR #1989, opened 2022-12-05
Fixed Idle Connection recovery so that unprocessedResponseCount isn't over decremented

tkyc:
The `unprocessedResponseCount` needs to be decremented only when `doneToken.isFinal` is true.

Fixes https://github.com/microsoft/mssql-jdbc/issues/1971 also.


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResultSet.java
```diff
@@ -393,8 +393,6 @@ boolean onDone(TDSReader tdsReader) throws SQLServerException {
                     SQLServerException.makeFromDriverError(stmt.connection, stmt, form.format(msgArgs), null, false);
                 }
 
-                stmt.connection.getSessionRecovery().decrementUnprocessedResponseCount();
```
tkyc, line 396: This decrement is unnecessary because we are throwing an exception now on doneToken errors. In the old code, when we didn't explicitly throw an exception, we would decrement only on a `DONE_ERROR` eg. tds status flag of 0x0002.

lilgreenbird: this makes me think we need to add that back to the other onDone methods in Statement and CallableStatement as well, and, add tests for them

src/test/java/com/microsoft/sqlserver/jdbc/resiliency/BasicConnectionTest.java
```diff
@@ -14,18 +14,15 @@
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
+import java.util.UUID;
 
 import javax.sql.PooledConnection;
 
-import com.microsoft.sqlserver.jdbc.SQLServerPooledConnection;
+import com.microsoft.sqlserver.jdbc.*;
```
lilgreenbird, line 21: don't do this

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResultSet.java
lilgreenbird: can you add a comment here so we don't think it's missed when we look at this in the future

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -3996,6 +3996,9 @@ boolean executeCommand(TDSCommand newCommand) throws SQLServerException {
                                         false);
                             }
                             try {
+                                if (null != preparedStatementHandleCache)
+                                    preparedStatementHandleCache.clear();
```
Jeffery-Wasty, line 4000: Nitpick, brackets.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResultSet.java
```diff
@@ -385,16 +385,21 @@ boolean onDone(TDSReader tdsReader) throws SQLServerException {
                 // following the column metadata indicates an empty result set.
                 rowCount = 0;
 
+                // decrementUnprocessedResponseCount() outside the "if" is not necessary here. It will over decrement if added.
+
                 short status = tdsReader.peekStatusFlag();
                 if ((status & TDS.DONE_ERROR) != 0 || (status & TDS.DONE_SRVERROR) != 0) {
```
lilgreenbird, line 391: isError()?

lilgreenbird (dismissed):

Jeffery-Wasty (approved):

Jeffery-Wasty (approved):

lilgreenbird (approved):

PR #1994, opened 2022-12-09
Added additional Idle Connection Resiliency resultset tests

tkyc:


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResultSet.java
```diff
@@ -5377,6 +5379,7 @@ boolean onDone(TDSReader tdsReader) throws SQLServerException {
                 StreamDone doneToken = new StreamDone();
                 doneToken.setFromTDS(tdsReader);
                 if (doneToken.isFinal() && doneToken.isError()) {
+                    stmt.connection.getSessionRecovery().decrementUnprocessedResponseCount();
```
David-Engel, line 5382: We could simply move lines 5390-5392 above 5381.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResultSet.java
```diff
@@ -389,6 +389,8 @@ boolean onDone(TDSReader tdsReader) throws SQLServerException {
                     SQLServerException.makeFromDriverError(stmt.connection, stmt, form.format(msgArgs), null, false);
                 }
 
+                // decrementUnprocessedResponseCount() is not necessary here. It will over decrement if added.
```
David-Engel, line 392: I agree that we would not want to blindly decrement here. But would this onDone for ClientCursorInitializer ever get called in the context of a final done token? Or do other onDone methods always handle the final done tokens? (I admit, I don't even know where ClientCursorInitializer is used.)

David-Engel (commented):
I really like the added test coverage. Good job. In a search of the code, though, we have many places where it _looks_ like we are handling done tokens (onDone methods). Are we covering those scenarios and correctly tracking the count across them?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResultSet.java
tkyc: Yes, it does get called when there's an error for a final done. My mistake, it looks like I added a decrement in the other onDone in SQLServerResultSet, but not here when I was cleaning up my PR in the end to push.

The decrement is not necessary outside the "if" as I found that from debugging the tests it would over decrement. The decrement is accounted for in the other onDone when there is no error.

Jeffery-Wasty (dismissed):

tkyc:
> I really like the added test coverage. Good job. In a search of the code, though, we have many places where it _looks_ like we are handling done tokens (onDone methods). Are we covering those scenarios and correctly tracking the count across them?

Yes, from manually testing and stepping through the code paths, to me all the decrements look accounted for.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResultSet.java
```diff
@@ -381,8 +381,15 @@ boolean onDone(TDSReader tdsReader) throws SQLServerException {
                 // following the column metadata indicates an empty result set.
                 rowCount = 0;
 
+                // decrementUnprocessedResponseCount() outside the "if" is not necessary here. It will over decrement if added.
+
                 short status = tdsReader.peekStatusFlag();
                 if ((status & TDS.DONE_ERROR) != 0 || (status & TDS.DONE_SRVERROR) != 0) {
```
lilgreenbird, line 387: isError()?

Jeffery-Wasty (dismissed):

David-Engel (approved):

lilgreenbird (approved):

Issue #2042, opened 2023-01-16
NoClassDefFoundError: jdk/net/ExtendedSocketOptions during connection


dleborgne:
## Driver version
10.2.3

## SQL Server version
Microsoft SQL Azure (RTM) - 12.0.2000.8 

## Client Operating System
CentOS Linux release 7.9.2009 (Core)

## JAVA/JVM version
OpenJDK Runtime Environment (Red_Hat-11.0.17.0.8-2.el7_9)

## Table schema
n/a

## Problem description
When starting on WildFly 24.0.1, using mssql-jdbc-10.2.3.jre11.jar, our Hibernate-based application is unable to connect to the database. Application gets an IJ031084: Unable to create connection error
Caused by: java.lang.NoClassDefFoundError: jdk/net/ExtendedSocketOptions

## Error message/stack trace
```
Caused by: java.lang.NoClassDefFoundError: jdk/net/ExtendedSocketOptions
        at deployment.mssql-jdbc-10.2.3.jre11.jar//com.microsoft.sqlserver.jdbc.DriverJDBCVersion.setSocketOptions(SQLServerJdbc43.java:66)
        at deployment.mssql-jdbc-10.2.3.jre11.jar//com.microsoft.sqlserver.jdbc.TDSChannel.open(IOBuffer.java:702)
        at deployment.mssql-jdbc-10.2.3.jre11.jar//com.microsoft.sqlserver.jdbc.SQLServerConnection.connectHelper(SQLServerConnection.java:3180)
        at deployment.mssql-jdbc-10.2.3.jre11.jar//com.microsoft.sqlserver.jdbc.SQLServerConnection.login(SQLServerConnection.java:2833)
        at deployment.mssql-jdbc-10.2.3.jre11.jar//com.microsoft.sqlserver.jdbc.SQLServerConnection.connectInternal(SQLServerConnection.java:2671)
        at deployment.mssql-jdbc-10.2.3.jre11.jar//com.microsoft.sqlserver.jdbc.SQLServerConnection.connect(SQLServerConnection.java:1640)
        at deployment.mssql-jdbc-10.2.3.jre11.jar//com.microsoft.sqlserver.jdbc.SQLServerDriver.connect(SQLServerDriver.java:936)
```

## Any other details that can be helpful
* Works fine with mssql-jdbc-10.2.1.jre11.jar
* Same error with mssql-jdbc-10.2.2.jre11.jar

lilgreenbird:
[...]

PR #2043, opened 2023-01-17
Check for jdk.net.ExtendedSocketOptions and ignore errors

lilgreenbird:
This addresses issue #2042

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
@@ -603,6 +603,13 @@ final class TDSChannel implements Serializable {
 
     private static final Logger logger = Logger.getLogger("com.microsoft.sqlserver.jdbc.internals.TDS.Channel");
 
+    /**
+     * From jdk.net.ExtendedSocketOption for setting TCP keep-alive options
+     */
+    private static Method socketSetOptionMethod = null;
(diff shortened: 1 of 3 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -731,6 +734,39 @@ final InetSocketAddress open(String host, int port, int timeoutMillis, boolean u
...
+    private void setSocketOptions(Socket tcpSocket, TDSChannel channel) throws IOException {
+        Method setOptionMethod = null;
+        SocketOption<Integer> keepIdleOption = null;
+        SocketOption<Integer> keepIntervalOption = null;
+
+        try {
+            setOptionMethod = Socket.class.getMethod("setOption", SocketOption.class, Object.class);
+            Class<?> clazz = Class.forName("jdk.net.ExtendedSocketOptions");
+            keepIdleOption = (SocketOption<Integer>) clazz.getDeclaredField("TCP_KEEPIDLE").get(null);
+            keepIntervalOption = (SocketOption<Integer>) clazz.getDeclaredField("TCP_KEEPINTERVAL").get(null);
```
David-Engel, line 747: These steps should be static that we do once so that we don't have to do the reflection lookup every time we make a socket connection.

Jeffery-Wasty (approved):

tkyc (approved):

PR #2064, opened 2023-02-04
Move from compendium to jdbc spec artifact

laeubi:


Jeffery-Wasty:
This will likely be closed, and 2065 and 2066 will be the PRs we consider.

Jeffery-Wasty:
This PR is not needed.

PR #2065, opened 2023-02-04
Upgrade osgi jdbc spec add properties

laeubi:
This is based on:
- https://github.com/microsoft/mssql-jdbc/pull/2064

PR #2066, opened 2023-02-04
Upgrade osgi jdbc add tck

laeubi:
This includes:
- https://github.com/microsoft/mssql-jdbc/pull/2064
- https://github.com/microsoft/mssql-jdbc/pull/2065

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -66,7 +66,7 @@
 		<azure-security-keyvault-keys.version>4.5.3</azure-security-keyvault-keys.version>
 		<azure-identity.version>1.7.0</azure-identity.version>
 		<msal.version>1.13.3</msal.version>
-		<org.osgi.compendium.version>5.0.0</org.osgi.compendium.version>
+		<osgi.jdbc.version>1.1.0</osgi.jdbc.version>
 		<antlr-runtime.version>4.9.3</antlr-runtime.version>
 		<com.google.code.gson.version>2.9.0</com.google.code.gson.version>
 		<bcprov-jdk15on.version>1.70</bcprov-jdk15on.version>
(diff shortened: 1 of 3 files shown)


laeubi:
@Jeffery-Wasty I don't really understand how to get more details about the testfailure can you give me a hint?

Jeffery-Wasty:
@laeubi, we'll have to look into this. Previously the errors were caused by `eclipse.osgi` dependency, but now I'm not seeing that in our tests, I'll get back to you with more information. Regardless, we'll look at the upgrade to `osgi.service` and the new properties and see if we can get that merged first, those two changes look fine. Thank you for your work on this.

Jeffery-Wasty:
Can you please confirm for us the necessity and use of the test? Is implementation of this test required by spec, or is this just a confirmation that the properties have been added correctly. It looks like the issues are all caused by the added test, so if it can be omitted, that would help us.

laeubi:
It is not strictly required, this just ensures that the implementation is fulfilling the spec.
So if it can even be omitted, it would still be interesting *why* this fails (I can't really get an idea from the checks output), and I was not able to run the tests locally.

Jeffery-Wasty:
Thanks for the explanation. We're still looking into it, its not readily apparent from the logs why this is failing, but we'll keep you updated on progress regarding this pr.

lilgreenbird (commented):
Why are there multiple PRs on this? Does this obsolete #2064 and #2065 if so please close those PRs. Can you please add some details and references to why this change is necessary.

Jeffery-Wasty:
Hi @laeubi,

The issue is that the test is that its using Junit5, and when this is invoked, causes issues in other tests in the project. If the tests can be rewritten only using features of Junit4 or below, that will allow it to pass and be included as part of the changes.

Comparing the test styles from this PR and #2017, I think the one in 2017 fits the project better, as opposed to the additional test information in the pom. Let's move forward with 2017 being the only PR for this change. We'll close the others, but thank you for this layered approach, it really helped in narrowing down on the problem.

Issue #2071, opened 2023-02-09
Connection.rollback method does not work for the Azure Synapse "serverless" database


LonwoLonwo:
## Driver version
Provide the JDBC driver version - the latest

## SQL Server version
Database: Azure Synapse "serverless"
Server: Microsoft SQL Server 12.00.2531, Microsoft Azure SQL Data Warehouse - 10.0.15318.0

## Client Operating System
Windows 10

## Problem description

One of our users has problem with data transfer operation in the NOT auto-commit mode in our program because we use the standard Connection method - rollback.

https://github.com/dbeaver/dbeaver/issues/18820

The issue that he has during this method execution:

```java
com.microsoft.sqlserver.jdbc.SQLServerException:  '@@TRANCOUNT' is not supported.
	at com.microsoft.sqlserver.jdbc.SQLServerException.makeFromDatabaseError(SQLServerException.java:262)
	at com.microsoft.sqlserver.jdbc.TDSTokenHandler.onEOF(tdsparser.java:283)
	at com.microsoft.sqlserver.jdbc.TDSParser.parse(tdsparser.java:129)
	at com.microsoft.sqlserver.jdbc.TDSParser.parse(tdsparser.java:37)
	at com.microsoft.sqlserver.jdbc.TDSParser.parse(tdsparser.java:26)
	at com.microsoft.sqlserver.jdbc.SQLServerConnection$1ConnectionCommand.doExecute(SQLServerConnection.java:3316)
	at com.microsoft.sqlserver.jdbc.TDSCommand.execute(IOBuffer.java:7418)
	at com.microsoft.sqlserver.jdbc.SQLServerConnection.executeCommand(SQLServerConnection.java:3274)
	at com.microsoft.sqlserver.jdbc.SQLServerConnection.connectionCommand(SQLServerConnection.java:3320)
	at com.microsoft.sqlserver.jdbc.SQLServerConnection.rollback(SQLServerConnection.java:3516)
```

As I can see in the [documentation](https://learn.microsoft.com/en-us/sql/t-sql/functions/trancount-transact-sql?view=sql-server-ver16)

> This syntax is not supported by serverless SQL pool in Azure Synapse Analytics.

The user uses not the latest driver, but  I do not see in the code any conditions for the serverless databases:

[...]

PR #2075, opened 2023-02-15
Updated supportsTransactions method

tkyc:
Update for https://github.com/microsoft/mssql-jdbc/issues/2071

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
@@ -2171,6 +2171,14 @@ public final class SQLServerDatabaseMetaData implements java.sql.DatabaseMetaDat
     @Override
     public boolean supportsTransactions() throws SQLServerException {
         checkClosed();
+        try {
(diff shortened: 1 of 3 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
```diff
@@ -2171,6 +2171,14 @@ public boolean supportsTransactionIsolationLevel(int level) throws SQLServerExce
     @Override
     public boolean supportsTransactions() throws SQLServerException {
         checkClosed();
+        try {
+            connection.connectionCommand("SELECT @@TRANCOUNT", "SQLServerDatabaseMetaData.supportsTransactions");
+        } catch (SQLServerException e) {
+            if (e.getMessage().equals("'@@TRANCOUNT' is not supported.")) {
```
lilgreenbird, line 2177: you need to add a string in resource file for this as it may be localized

lilgreenbird: also, can we just do this once, similar to how we set isAzureblah

lilgreenbird (approved):

Jeffery-Wasty (approved):

PR #2076, opened 2023-02-15
Changed dependency from 'osgi.compendium' to 'osgi.service.jdbc' for build.gradle

Jeffery-Wasty:
Was missing from #2017.


stbischof:
why is it in `implementation` and not `compileOnly` ?

Issue #1432, opened 2020-09-17
Microsoft SQL Server Resource Adaptor(JDBC Driver) returning XAException with error code XAER_RMERR in case of fail-over of SQL Server AG from primary node to standby node while committing an XA Transaction


pa1484:
## Driver version

SQL Server JDBC Driver 7.0 & 8.2

## SQL Server version

SQL Server AG 2017

## Client Operating System

Windows Server 2016, version 10.0

## JAVA/JVM version

IBM J9 VM
## Problem description

During a fail-over test for one of the product when using SQL server AG as back end, we found that there are few XA transactions in SQL Server DTC are in in-doubt state.

After analyzing the trace files, we found that the SQL Server JDBC driver was returning an XAException with an error code of XAER_RMERR on the XAResource.commit due to SQL Server AG fail-over. Websphere application server treats XAER_RMERR as a non-retriable condition as per the XA specification. 

We tested with a diagnostic patch on Websphere application server to treat XAER_RMERR as a XAER_RMFAIL(which is retriable) and recreated the scenario.Due to the diagnostic patch Websphere retries the commits which are eventually successful and we no longer finding in-doubt transactions in SQL server DTC.

1. Expected behaviour:
SQL Server Resource Adaptor(JDBC Driver) should return an XAException with a retriable error code such as XAER_RMFAIL in case fail-over of SQL Server AG from primary node to standby node while committing an XA Transaction.
2. Actual behaviour:
SQL Server Resource Adaptor(JDBC Driver) returning XAException with error code XAER_RMERR in case fail-over of SQL Server AG from primary node to standby node while committing an XA Transaction
3. Error message/stack trace:
Below is the trace that shows XAException with XAER_RMERR error code returned by SQLServer JDBC driver during fail-over.

[7/30/20 17:20:37:254 CST] 00000369 WSRdbXaResour >  commit Entry
                                 com.ibm.ws.rsadapter.spi.WSRdbXaResourceImpl@5e889f7e
                                 WSRdbManagedConnectionImpl@3a76217e
                                 com.ibm.ws.tx.jta.XidImpl@e2696786
                                 {XidImpl: formatId(57415344), gtrid_length(36), bqual_length(54),
data(000001739f05fd25000000017aa729486c90e9918373568c6d645a1d91c7913943fe1a86000001739f05fd25000000017aa729486c90e9918373568c6d645a1d91c7913943fe1a86000000010000000000000000000000000001)}
[...]

PR #2078, opened 2023-02-17
XA error correction

tkyc:
Fix for https://github.com/microsoft/mssql-jdbc/issues/1432

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerXAResource.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerXAResource.java
@@ -24,6 +24,7 @@ import javax.transaction.xa.XAException;
 import javax.transaction.xa.XAResource;
 import javax.transaction.xa.Xid;
 
+import com.microsoft.sqlserver.jdbc.SQLServerError.TransientError;
 
 /**
  * Implements Transaction id used to recover transactions.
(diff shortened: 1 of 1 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerXAResource.java
```diff
@@ -751,9 +752,24 @@ private XAReturnValue DTC_XA_Interface(int nType, Xid xid, int xaFlags) throws X
...
+            if (xaLogger.isLoggable(Level.FINER))
+                xaLogger.finer(toString() + " exception:" + ex);
+
+            if (ex.getMessage().equals(SQLServerException.getErrString("R_noServerResponse"))
+                    || TransientError.isTransientError(ex.getSQLServerError())) {
+                XAException e = new XAException(ex.toString());
+                e.errorCode = XAException.XAER_RMFAIL;
+                throw e;
+            }
+
```
lilgreenbird, line 772: this makes me wonder about other errors that we should not RMERR on. I see similar issues in Oracle as well. I'm wondering if we should check severity?

David-Engel: As discussed offline, we expect the statement we call to return an XA status (including any XA errors). Any other error the driver receives, like a server-level error, it would be difficult to discern/translate those into the correct RMERR/RMFAIL status. So going with the conservative change here and only changing the specific behavior we believe should be changed from RMERR to RMFAIL. Other cases will remain the same (RMERR).

Jeffery-Wasty (approved):

lilgreenbird (approved):

Issue #2084, opened 2023-02-28
SharedTimer race condition


mmimica:
## Driver version
12.2.0.jre11

## Problem description
There is a race-condition in `SharedTimer` class.

The class holds a singleton instance of a `ScheduledThreadPoolExecutor` and keeps a reference count. It disposes the executor instance when reference count reaches 0. Because of ineffective use of double-lock check pattern in SharedTimer#getTimer the reference counter is prone to race-conditions.

The problem escalates with dawn of "Connection Resiliency" feature which is enabled by default. It can be mitigated by setting `connectRetryCount` connection property to 0.

mmimica:
Introduced by #1931

Jeffery-Wasty:
Hi @mmimica,

Thank you for finding this, we'll take a look at the issue and the PR and get back to you on how we'd like to proceed.

jubax:
I think I stumbled over this problem in production since upgrading from `11.2.2` to `12.2.0`. See the stack trace below:
```
java.util.concurrent.RejectedExecutionException: Task java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask@297df4d2[Not completed, task = java.util.concurrent.Executors$RunnableAdapter@40b78b8e[Wrapped task = com.microsoft.sqlserver.jdbc.TDSTimeoutTask@445ae498]] rejected from java.util.concurrent.ScheduledThreadPoolExecutor@6aafff45[Terminated, pool size = 0, active threads = 0, queued tasks = 0, completed tasks = 0]
	at java.util.concurrent.ThreadPoolExecutor$AbortPolicy.rejectedExecution(ThreadPoolExecutor.java:2065) ~[?:?]
	at java.util.concurrent.ThreadPoolExecutor.reject(ThreadPoolExecutor.java:833) ~[?:?]
	at java.util.concurrent.ScheduledThreadPoolExecutor.delayedExecute(ScheduledThreadPoolExecutor.java:340) ~[?:?]
	at java.util.concurrent.ScheduledThreadPoolExecutor.schedule(ScheduledThreadPoolExecutor.java:562) ~[?:?]
	at com.microsoft.sqlserver.jdbc.SharedTimer.schedule(SharedTimer.java:144) ~[mssql-jdbc-12.2.0.jre11.jar:?]
[...]

PR #2085, opened 2023-02-28
Fix shared timer race #2084

mmimica:


src/main/java/com/microsoft/sqlserver/jdbc/SharedTimer.java
```diff
@@ -111,20 +111,16 @@ public void removeRef() {
      * When the caller is finished with the SharedTimer it must be released via {@link#removeRef}
      */
     public static SharedTimer getTimer() {
-        SharedTimer result = instance;
```
Jeffery-Wasty, line 114: Thank you for the changes. For now, we'd like to resolve the race condition, but keep everything else the same, so keeping the extra SharedTimer variable etc. So, the final PR will have getAndIncrement() moved as well as the extra test. If you're able to do that for us, we can move forward with merging this.

mmimica: You can't just move the `getAndIncrement()`.
I can keep the local var, but it makes no sense:
```
    public static SharedTimer getTimer() {
        LOCK.lock();
        try {
            SharedTimer result = instance;
            if (result == null) {
                instance = result = new SharedTimer();
            }
            result.refCount.getAndIncrement();
            return result;
        } finally {
            LOCK.unlock();
        }
    }
```

Like this? If you insist, fine.

Jeffery-Wasty: Maybe I'm misunderstanding. The race condition comes about because the reference count has not been placed correctly in regard to the locking, so the only thing that needs to be fixed here is its placement. We still want the extra null check before the locking, and we still want to abstract away from instance and compare everything to result as instance is volatile. Below is what I'm looking for, just the refCount logic being moved. I would help you out here and push directly to the PR, but I'm unable to do so, pushing to user remotes doesn't always work.

```public static SharedTimer getTimer() {
        SharedTimer result = instance;
        if (result == null) {
            LOCK.lock();
            try {
                result = instance;
                if (result == null) {
                    instance = result = new SharedTimer();
                }
                result.refCount.getAndIncrement();
            } finally {
                LOCK.unlock();
            }
        }
        return result;
    }
```

mmimica: Your code doesn't increment the counter when the instance exists.

Jeffery-Wasty: You're right, I copied over the wrong thing, I meant to increase the refCount for instance.

```public static SharedTimer getTimer() {
        SharedTimer result = instance;
        if (result == null) {
            LOCK.lock();
            try {
                result = instance;
                if (result == null) {
                    instance = result = new SharedTimer();
                }
                instance.refCount.getAndIncrement();
            } finally {
                LOCK.unlock();
            }
        }
        return result;
    }
```

mmimica: Still the same...

Jeffery-Wasty:
After reviewing the PR again and talking with the team, I agree with the code changes, but not the test. It's not reliable enough, passing, at most, 80% of the time. We'll continue looking into ways we can test this that are more reliable.

jubax (approved):
Not sure if this actually waits for an approval and if I'm the right person to approve, but still: Looks fine 👍 

I was able to reproduce the race condition (an NPE was thrown) with the test before the fix and can verify that the NPE is gone after the fix is applied.

I also wrote an extended version of the test, using a latch to make sure all threads start at the same time. But as it turned out your test works fine for reproducing the issue.

(Due to security manager issues I just made a local copy of `SharedTime` and used that.)

Jeffery-Wasty:
Hi @jubax,

Yes, we were just making sure the test was passing consistently. I was having issues on my end, but that has been resolved. Everything looks fine, we'll look at getting this merged.

Jeffery-Wasty (approved):

lilgreenbird (approved):

PR #2100, opened 2023-03-27
Allow failover partner to be tried in case of socket timeout

Jeffery-Wasty:
A socket timeout exception when starting connection led to connection close before failover was being attempted. This allows the failover to be tried once, with the timeout exception being returned if failover can't be connected to.


tkyc (dismissed):

lilgreenbird (approved):

Jeffery-Wasty:
\azp

lilgreenbird (approved):

tkyc (approved):

PR #2104, opened 2023-03-31
Resolve high thread count when using findSocketUsingThreading

Jeffery-Wasty:
findSocketUsingThreading can reach high thread counts for some users. This was due to an uncaught exception.


src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -2455,6 +2456,14 @@ Socket findSocket(String hostName, int portNumber, int timeoutInMilliSeconds, bo
         try {
             InetAddress[] inetAddrs = null;
 
+            // Used in testing to force multiple InetAddress and thus findSocketUsingThreading.
+            // Cleanest way to implement this is for this test condition to be the first one checked.
+            if (hostName.equals("testServerName")) {
```
tkyc, line 2461: Hmm, maybe we should look into using flags to gate code specifically for testing. Maybe something like a JVM arg `-DtestMode`.

Jeffery-Wasty: Do we already do this anywhere?

David-Engel: Agreed. This doesn't feel right to me...

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -2950,7 +2959,16 @@ private void findSocketUsingThreading(InetAddress[] inetAddrs, int portNumber,
...
-                    parentCondition.await(timeRemaining, TimeUnit.MILLISECONDS);
+                    try {
+                        parentCondition.await(timeRemaining, TimeUnit.MILLISECONDS);
+                    } catch (InterruptedException ie) {
+                        // This uncaught exception was what was casuing the issues we saw previously with many
+                        // threads being created
+                        if (logger.isLoggable(Level.FINER)) {
+                            logger.finer(this.toString() + " Current thread has been interrupted.");
+                        }
+                        Thread.currentThread().interrupt();
```
David-Engel, line 2970: The log isn't necessary and the interrupt() will leave the driver in the same state we are trying to prevent with the catch...
```suggestion
```

Jeffery-Wasty: I thought the issue was with that the uncaught exception was causing problems, not that the thread is being interrupted.

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
Jeffery-Wasty: Ok, I'll take a look into this.

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
David-Engel: The issue is that the thread has been interrupted (that's the signal that is used to get parentCondition.await() to continue), unless we specifically catch the interrupt here (catch InterruptedException), which resets the thread interrupted status and keeps it from propagating further. If we don't catch (and thus reset) the InteruptedException, all subsequent Thread.sleep() calls just fall through. That skipping of all the sleeps is what triggers the thread storm since the outer connection retries no longer pause in between each try.

David-Engel: I see the original comment I proposed is completely different now, too. It should be changed back. We aren't ignoring the interrupt. We are expecting it and clearing it via the catch block.

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
lilgreenbird: don't put test code into the test driver, do this in the tests

Jeffery-Wasty: I agree, I'll revisit prior methods of testing to see if there's something I missed.

src/test/java/com/microsoft/sqlserver/jdbc/SQLServerConnectionTest.java
```diff
@@ -934,6 +935,59 @@ public void run() {
...
+                   }
+               }
+           });
+           SQLServerDataSource ds = new SQLServerDataSource();
+           ds.setServerName("localhost");
+           Connection conn = ds.getConnection();
+           Thread.sleep(5000);
+       } catch (Exception e) {
+           if (!(e instanceof SQLServerException)) {
+               System.out.println(e);
```
Jeffery-Wasty, line 972: Left this in by mistake, will remove.
```suggestion

```

tkyc (approved):

src/test/java/com/microsoft/sqlserver/jdbc/SQLServerConnectionTest.java
```diff
@@ -934,6 +935,58 @@ public void run() {
...
+           Thread.sleep(5000);
+       } catch (Exception e) {
+           if (!(e instanceof SQLServerException)) {
+               fail(TestResource.getResource("R_unexpectedException") + e.getMessage());
+           }
+       } finally {
+           if (executor != null) {
+               executor.shutdownNow();
+           }
+           Thread.sleep(10000);
```
lilgreenbird, line 978: why?

src/test/java/com/microsoft/sqlserver/jdbc/SQLServerConnectionTest.java
```diff
@@ -934,6 +935,58 @@ public void run() {
...
+           ds.setServerName("localhost");
+           Connection conn = ds.getConnection();
+           Thread.sleep(5000);
+       } catch (Exception e) {
+           if (!(e instanceof SQLServerException)) {
+               fail(TestResource.getResource("R_unexpectedException") + e.getMessage());
+           }
+       } finally {
+           if (executor != null) {
+               executor.shutdownNow();
```
lilgreenbird, line 976: why not shutdown()?

src/test/java/com/microsoft/sqlserver/jdbc/SQLServerConnectionTest.java
```diff
@@ -934,6 +935,58 @@ public void run() {
...
+   public void testThreadCountWhenFindingSocket() throws InterruptedException {
+       ExecutorService executor = null;
+       ManagementFactory.getThreadMXBean().resetPeakThreadCount();
+
+       try {
+           executor = Executors.newSingleThreadExecutor(r -> new Thread(r, ""));
+           executor.submit(() -> {
+               try {
+                   SQLServerDataSource ds = new SQLServerDataSource();
+                   ds.setServerName("localhost");
```
lilgreenbird, line 957: we shouldn't assume there's a local server running, test should not fail if no local server

src/test/java/com/microsoft/sqlserver/jdbc/SQLServerConnectionTest.java
```diff
@@ -934,6 +935,58 @@ public void run() {
...
+       } finally {
+           if (executor != null) {
+               executor.shutdownNow();
+           }
+           Thread.sleep(10000);
+       }
+
+       // At this point, thread count has returned to normal. If the peak was more
+       // than 2 times the current, this is an issue and the test should fail.
+       int acceptableMax = 2 * ManagementFactory.getThreadMXBean().getThreadCount();
```
lilgreenbird, line 983: how do we determine this acceptable value?

tkyc (approved):

David-Engel (approved):

lilgreenbird (approved):

PR #2109, opened 2023-04-12
Embed auth dll revert

tkyc:


lilgreenbird (approved):

Jeffery-Wasty (approved):

cboehme:
Is there any information available why that feature was reverted? It looked really promising and I was hoping that it could  solve my problems with using integrated authentication in the liquibase-maven-plugin.

tkyc:
@cboehme There are some licensing issues with embedding the DLL as it uses proprietary code.

cboehme:
Thank you @tkyc, that's a pity. 
Wouldn't it be possible to distribute the DLL in a separate jar which uses the same license as the DLL? That way the DLL could still be loaded as a resource from the main driver jar but is not bundled with the MIT licensed driver code. Developers would then only have to include both jars in their applications to enable integrated authentication. This would save them fiddling with java.library.path.

tkyc:
@cboehme Thanks for the suggestion. That would work, but I'll need to bring that up to management since it's their decision to make. We're still exploring solutions and nothing has been settled yet, but when we do I'll let you know. It might be awhile as right now this isn't a priority.

PR #2113, opened 2023-04-14
Change bigDecimal type definition to have proper precision for numbers between 0 and 1

Jeffery-Wasty:
Changed the way precision and scaling are decided for big decimal to account for a missed case.


tkyc (approved):

Jeffery-Wasty:
Sure thing.

Jeffery-Wasty:
Remaking this instead of messing around with rebase.

PR #2116, opened 2023-05-01
Fix for bigDecimal values between 0 and 1 having too high of a precision

Jeffery-Wasty:
Changes how precision and scaling are computed for bigDecimal values to account for a big where precision is 1 too high for numbers between 0 and 1.


src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/StatementTest.java
```diff
@@ -1426,6 +1426,32 @@ public void testFailedToResumeTransaction() throws Exception {
             }
         }
 
+        /**
```
lilgreenbird, line 1429: can we test all boundary cases here eg. < 1 and > 1 with precision 38, > 38, etc

lilgreenbird: also, test cstmt as in other tests

src/main/java/com/microsoft/sqlserver/jdbc/Parameter.java
```diff
@@ -531,19 +531,14 @@ private void setTypeDefinition(DTV dtv) {
                             param.typeDefinition = SSType.DECIMAL.toString() + "(" + valueLength + "," + scale + ")";
                         }
                     } else {
-                        BigDecimal bigDecimal = null;
+                        BigDecimal bigDecimal;
                         if (dtv.getJavaType() == JavaType.BIGDECIMAL
                                 && null != (bigDecimal = (BigDecimal) dtv.getSetterValue())) {
```
lilgreenbird, line 536: since I have comments in other places, can we move this assignment out of the 'if'

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/StatementTest.java
Jeffery-Wasty: cstmt introduces rounding to 4 decimal places. For cases with maximum precision, I'm sticking with pstmt, for the cases with less than maximum, I've switched to cstmt.

Test cases include: max precision between 0 and 1, max with number greater than 1, less than max precision for a number between 0 and 1, less than max for a number greater than 1.

src/main/java/com/microsoft/sqlserver/jdbc/Parameter.java
Jeffery-Wasty: Moved.

lilgreenbird (approved):

Issue #2110, opened 2023-04-13
lockTimeout doesn't work if applications run on Azure environment


nanahoshi:
## Driver version
12.2.0.jre11

## SQL Server version
Microsoft SQL Azure (RTM) - 12.0.2000.8 
	Mar  8 2023 17:58:50 
	Copyright (C) 2022 Microsoft Corporation

## Client Operating System
Ubuntu 18.04 LTS
Ubuntu 20.04 LTS
## JAVA/JVM version
openjdk version "11.0.18" 2023-01-17 LTS
OpenJDK Runtime Environment Microsoft-7208460 (build 11.0.18+10-LTS)
OpenJDK 64-Bit Server VM Microsoft-7208460 (build 11.0.18+10-LTS, mixed mode)

## Table schema
N/A

## Problem description
The behaviour of 'lockTimeout' is NOT the same on 'from within Azure' and 'from outside Azure'.

**From within Azure / connection policy = redirect**
ref: https://learn.microsoft.com/en-us/azure/azure-sql/database/connectivity-architecture?view=azuresql#connectivity-from-within-azure

The 'lockTimeout' is ignored.

**Outside Azure / connection policy = proxy**
ref: https://learn.microsoft.com/en-us/azure/azure-sql/database/connectivity-architecture?view=azuresql#connectivity-from-outside-of-azure

The 'lockTimeout' is applied.

We use the below connection string on both sides.

`jdbc:sqlserver://<dummy>.database.windows.net:1433;database=<dummy>;encrypt=true;trustServerCertificate=false;hostNameInCertificate=*.database.windows.net;user=<dummy>;password=<dummy>;lockTimeout=5000;`

I'm not sure if the other options are applied or not.

## Expected behavior
Even if the connection policy is redirect mode, JDBC shall send a packet `set lock_timeout 5000`.
below query returns '5000'
```sql
SELECT @@LOCK_TIMEOUT AS [Lock Timeout]
```

## Actual behavior
If the connection policy is redirect mode, JDBC doesn't send the packet `set lock_timeout 5000`.
below query returns '-1'
```sql
SELECT @@LOCK_TIMEOUT AS [Lock Timeout]
```
## Error message/stack trace
N/A
## Any other details that can be helpful
I tested it with simple class

```java
package jp.example;

import java.io.IOException;
import java.io.InputStream;
import java.sql.Connection;
import java.sql.DriverManager;
import java.sql.ResultSet;
import java.sql.SQLException;
[...]

PR #2117, opened 2023-05-02
Fix for applying lockTimeout when redirect mode is set on azure DB

tkyc:
https://github.com/microsoft/mssql-jdbc/issues/2110


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -1465,7 +1465,20 @@ final void initResettableValues() {
         transactionIsolationLevel = Connection.TRANSACTION_READ_COMMITTED;// default isolation level
         maxFieldSize = 0; // default: 0 --> no limit
         maxRows = 0; // default: 0 --> no limit
-        nLockTimeout = -1;
+
```
lilgreenbird, line 1468: can we refactor this so we don't dup the code in another place

lilgreenbird (approved):

Jeffery-Wasty (approved):

Issue #2124, opened 2023-05-07
Faulty `DATETIME` `ticksSinceMidnight` overflow check


Marcono1234:
## Driver version
4ff4e7ed160c90ae76f9322bc946f88fd17fdd83

## Problem description
The `Integer.MAX_VALUE < ticksSinceMidnight` check here is faulty:
https://github.com/microsoft/mssql-jdbc/blob/4ff4e7ed160c90ae76f9322bc946f88fd17fdd83/src/main/java/com/microsoft/sqlserver/jdbc/dtv.java#L3592

`ticksSinceMidnight` is an `int`, so it cannot be `> Integer.MAX_VALUE`. Maybe the solution here is to change `ticksSinceMidnight` to `long` (and adjust the calculation from `* 10` to `* 10L` to avoid overflow there).

I am not familiar with this code though and am not planning to submit a pull request.

tkyc:
Thanks for bringing this up. We'll take a look.

PR #2129, opened 2023-05-17
Removed useless check

tkyc:
https://github.com/microsoft/mssql-jdbc/issues/2124


tkyc:
Looks like it passed sonarqube. I'm ok with just leaving it as is.

Jeffery-Wasty (approved):

Issue #2130, opened 2023-05-18
JDBC Driver still throwing XAException with error code XAER_RMERR on AG failover.


djmatthews:
## Driver version
The JDBC driver version is 12.3.0 Preview Release

## SQL Server version
SQL Server AG 2017

## Client Operating System
Windows Server 2016, version 10.0

## JAVA/JVM version
Java version = 1.8.0_171, Java Runtime Version = 8.0.5.17 - pxa6480sr5fp17ifix-20180726_01(SR5 FP17+IJ08001)

## Problem description
This is a follow on from https://github.com/microsoft/mssql-jdbc/issues/1432
Despite having the fix for 1432 (https://github.com/microsoft/mssql-jdbc/pull/2078) the same behaviour occurred ie the JDBC driver reported an XAER_RMERR when it should have reported an XAER_RMFAIL resulting in indoubt transactions remaining in SQLServer during failover from AG primary to secondary.
## Expected behavior
JDBC driver should have thrown an XAException with errorCode XAER_RMFAIL
## Actual behavior
JDBC driver throws XAException with errorCode XAER_RMERR
## Error message/stack trace
Complete error message and stack trace.

Here is a callstack showing the XAException on the XA commit:

[5/15/23 20:05:33:883 UTC] 00000160 WSRdbXaResour <  commit Exit
                                 javax.transaction.xa.XAException: com.microsoft.sqlserver.jdbc.SQLServerException: Connection reset
	at com.microsoft.sqlserver.jdbc.SQLServerXAResource.dtc_XA_interface(SQLServerXAResource.java:773)
	at com.microsoft.sqlserver.jdbc.SQLServerXAResource.commit(SQLServerXAResource.java:839)
	at com.ibm.ws.rsadapter.spi.WSRdbXaResourceImpl.commit(WSRdbXaResourceImpl.java:361)
	at com.ibm.ejs.j2c.XATransactionWrapper.commit(XATransactionWrapper.java:496)
	at com.ibm.tx.jta.impl.JTAXAResourceImpl.commit(JTAXAResourceImpl.java:308)
	at com.ibm.tx.jta.impl.RegisteredResources.deliverOutcome(RegisteredResources.java:1644)
	at com.ibm.tx.jta.impl.RegisteredResources.distributeOutcome(RegisteredResources.java:2003)
	at com.ibm.tx.jta.impl.RegisteredResources.distributeCommit(RegisteredResources.java:2315)
[...]

PR #2132, opened 2023-05-24
Added case for XAER_RMFAIL

tkyc:
#2130

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerXAResource.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerXAResource.java
@@ -5,6 +5,7 @@
 
 package com.microsoft.sqlserver.jdbc;
 
+import java.net.SocketException;
 import java.sql.CallableStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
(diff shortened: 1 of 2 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerXAResource.java
```diff
@@ -943,4 +945,62 @@ private static int nextResourceID() {
...
+        }
+
+        // Add as needed here for future XAER_RMFAIL exceptions
+        if (root instanceof SocketException) {
+            switch (err) {
+                case CONN_RESET:
+                    return true;
+                default:
+                    return false;
+            }
```
lilgreenbird, line 1000: just return err == CONN_RESET

lilgreenbird: or actually just
return (root instanceof SocketException) && err == CONN_RESET

lilgreenbird (approved):

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerXAResource.java
Jeffery-Wasty: I this was made a switch instead of a simple return to make it easier to add future cases. Is it better to future proof, or just make the changes in the future when we eventually need to?

Jeffery-Wasty (approved):

PR #2136, opened 2023-06-01
Change Activity ID behavior:

David-Engel:
- Activity ID stays the same for the life of the process (not per thread)
- Always send Activity ID to the server in PRELOGIN
- Increment the sequence for each new connection (not each command)


tkyc (approved):

lilgreenbird (approved):

Jeffery-Wasty (approved):

FlyingWalross:
@David-Engel 

Hi quick question about this, the changes in #2254 seem to reverse the changes made in #465 (see also #314) that fixed a ClassLoader leak that was a result of the ActivityCorrelator ThreadLocal never being removed from the thread. Now the ThreadLocal seems to have been added back:

```diff
final class ActivityCorrelator {

+    private static ThreadLocal<ActivityId> t_ActivityId = new ThreadLocal<ActivityId>() {
+        @Override
+        protected ActivityId initialValue() {
+            return new ActivityId();
```

Since this ThreadLocal is now being created again, I am seeing a ClassLoader leak when unloading a webapp using the `mssql-jdbc-12.6.0.jre8.jar` driver:

```
SEVERE: The web application [...] created a ThreadLocal with key of type [com.microsoft.sqlserver.jdbc.ActivityCorrelator$1] (value [com.microsoft.sqlserver.jdbc.ActivityCorrelator$1@29887163]) and a value of type [com.microsoft.sqlserver.jdbc.ActivityId] (value [...]) but failed to remove it when the web application was stopped.
```

The same does not happen with the `mssql-jdbc-12.4.0.jre8.jar` driver, which was released before this change. I am properly deregistering the driver on webapp unload, but I just cannot seem to get rid of this ClassLoader leak. Using a profiler it shows the GC root of the webapp ClassLoader to be the ActivityCorrelator in the ThreadLocal map of the `main` thread.

Do you or anyone familiar with this issue have any ideas on how I could solve this? Unfortunately we frequently need to reload webapps and leaking a few hundred mb of memory due to the ClassLoader leak is not really an option, so we are currently having to stick to the older driver version.

Any help would be much appreciated, thanks!

lilgreenbird:
hi @FlyingWalross 

thanks for reporting this we will work on addressing this issue.

FlyingWalross:
@lilgreenbird Thanks for the quick response and fix!

PR #2148, opened 2023-06-14
avoid ClassCastException for floats

wuttke:
prior to this change, a ClassCastException occured if foreign (third-party) JDBC drivers return a Java Double and not a Float for JDBC FLOAT columns

actually, one could argue to even pass ((Number)colValue).doubleValue() to tdsWriter.writeDouble (for max  precision)

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
@@ -2167,8 +2167,7 @@ public class SQLServerBulkCopy implements java.lang.AutoCloseable, java.io.Seria
                         if (bulkNullable) {
                             tdsWriter.writeByte((byte) 0x08);
                         }
-                        float floatValue = ((Number)colValue).floatValue();
-                        tdsWriter.writeDouble(floatValue);
+                        tdsWriter.writeDouble(((Number) colValue).floatValue());
                     }
                     break;
 
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
@@ -2167,7 +2167,8 @@ public class SQLServerBulkCopy implements java.lang.AutoCloseable, java.io.Seria
                         if (bulkNullable) {
                             tdsWriter.writeByte((byte) 0x08);
                         }
-                        tdsWriter.writeDouble((float) colValue);
+                        float floatValue = ((Number)colValue).floatValue();
+                        tdsWriter.writeDouble(floatValue);
                     }
                     break;
 


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
```diff
@@ -2167,7 +2167,8 @@ else if (null != sourceCryptoMeta) {
                         if (bulkNullable) {
                             tdsWriter.writeByte((byte) 0x08);
                         }
-                        tdsWriter.writeDouble((float) colValue);
+                        float floatValue = ((Number)colValue).floatValue();
+                        tdsWriter.writeDouble(floatValue);
```
Jeffery-Wasty, line 2171: Can you make that change you suggested? While it is cleaner, it doesn't make sense to define a variable that is only used on the immediate following line.
```suggestion
                        tdsWriter.writeDouble(((Number)colValue).floatValue());
```

tkyc:
Thanks for the contribution. We'll take a look.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
Jeffery-Wasty, line 2171: Please combine these lines as follows:
```suggestion
                        tdsWriter.writeDouble(((Number) colValue).floatValue());
```

Jeffery-Wasty (dismissed):

Jeffery-Wasty:
Hi @wuttke, 

After reading over this again. While the change is small, I don't want to make any change that isn't necessary, and I'm not sure how you came across the exception in this part of the code. The code is being covered by our tests.

Can you give us the repro code and repro conditions used to generate the error?

If we're not clear on exactly what issue this is fixing, we'll have to disregard the changes.

Thanks!

wuttke:
This change fixes an exception that occurs if the JDBC provider of a second database returns a Double object if ResultSet.getObject is called on a SQL Type FLOAT column.

Jeffery-Wasty:
Can you give an example of a provider that does this? Also, why is a double being returned in this case? This may be an issue in the 3rd-party provider's output, not our input. If you can point us where to look, we can learn more about this issue.

As well, the use case, why is a second JDBC provider being used here along the MSSQL-JDBC driver? Is it possible to restrict the workflow to only use one driver?

wuttke:
I use the SAP MaxDB driver to migrate from MaxDB to MSSQL. For optimal performance, we use bulk inserts. The Double object in the ResultSet comes from this driver.

Jeffery-Wasty:
I still don't fully understand. The issue is that our driver is returning a double in this scenario, as opposed to a float. First of all, I'm not sure this is an issue, you will have to show us some proof of this being a problem. Second, I don't see how the proposed changes resolve the issue. Even with the changes, we're still appending a double value to the TDS stream. Please let me know if I'm misunderstanding something here, but to move forward with this issue, we'll need some proof that the driver is causing problems.

Jeffery-Wasty:
Ok, thank you, I think I understand this better now. We'll be taking another look at this after our GA release, which is scheduled for July 31st. For now this will be moved to our backlog.

Jeffery-Wasty:
Hi @wuttke,

I had a discussion regarding this PR with the rest of the JDBC team. This is the current situation, as I understand it:

- Our driver is being used in conjunction with an external driver
- This external driver is returning a `Double` value, for the SQL type `Float`
- Because we simply cast before writing to stream, this can produce a `ClassCastException`
- The solution you propose is to convert the input to a `Number` object, and take the `floatValue` from that

This does not seem like a change we want to incorporate into the driver. The problem seems to be with the 3rd party driver, and it returning the wrong value for the SQL type. Our suggestion is to contact the development team for this driver instead and have the issue be resolved on their end. If it turns out that a solution on their end is not possible, then we can continue discussion on this topic, but for the time-being work on this PR from our end will be halted.

wuttke:
Well, not much to add. I disagree and think m change is an improvement but you are in charge :-)

Jeffery-Wasty:
Is there a mistake in the process flow I described above? We're thinking about closing this because at this time we believe the issue is with this "third party driver" and accommodating this driver would lead to changes in the driver we're not comfortable with. Please let us know if there is something we're missing and we can continue discussion.

Issue #2146, opened 2023-06-13
[FEATURE REQUEST] provide localSocketAddress property 


bmarwell:
## Is your feature request related to a problem? If so, please give a short summary of the problem and how the feature would resolve it

When default routing is not sufficient, asking the driver to use a specific interface (either via interface name or IP or dns name) is useful.

## Describe the preferred solution

Add a JDBC driver property localSocketAddress.
Allowed values

* (must) IPv4 or IPv6 address
* (must) dns name
* (very optional) interface name

## Describe alternatives you've considered

./.
None available,  I am using Open Liberty.
Most drivers do support this already.

## Additional context

See below

## Reference Documentations/Specifications

* PostgreSQL: https://github.com/pgjdbc/pgjdbc/issues/2081
* MySQL: https://dev.mysql.com/doc/connectors/en/connector-j-connp-props-networking.html#cj-conn-prop_localSocketAddress
* Oracle: https://docs.oracle.com/cd/E17952_01/connector-j-8.0-en/connector-j-connp-props-networking.html#cj-conn-prop_localSocketAddress
* IBM DB2 request: https://ibm-data-and-ai.ideas.ibm.com/ideas/DB24LUW-I-1592

## Reference Implementation

Unknown, but a lot of drivers already support this exact JDBC property.

tkyc:
We'll visit this in planning and let you know if we'll take up work on this.

PR #2152, opened 2023-06-20
localSocketAddress implementation

tkyc:
Adds new `localSocketAddress` connection string property. Valid values are IPv4/IPv6 addresses or a hostname.

https://github.com/microsoft/mssql-jdbc/issues/2146

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
@@ -2890,6 +2890,13 @@ final class SocketFinder {
         if (addr.isUnresolved())
             throw new java.net.UnknownHostException();
         selectedSocket = getSocketFactory().createSocket();
+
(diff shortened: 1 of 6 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -2890,6 +2890,17 @@ private Socket getConnectedSocket(InetSocketAddress addr, int timeoutInMilliSeco
         if (addr.isUnresolved())
             throw new java.net.UnknownHostException();
         selectedSocket = getSocketFactory().createSocket();
+
+        String localSocketAddress = this.conn.activeConnectionProperties.getProperty(SQLServerDriverStringProperty.LOCAL_SOCKET_ADDRESS.toString());
+
+        if (null != localSocketAddress && !localSocketAddress.isEmpty()) {
+            selectedSocket.bind(new InetSocketAddress(InetAddress.getByName(localSocketAddress), 0));
```
machavan, line 2897: Do we validate this user specified loicalSocketAddress as a valid IP somewhere before coming here ? If not would it be good to handle an exception thrown by InetSocketAddress/bind call and raise an error in that case ?

PR #2154, opened 2023-06-28
Execute Stored Procedures Directly

tkyc:
Continued work from https://github.com/microsoft/mssql-jdbc/pull/547.

Points of interest:

1. In order to use parameter names and indexes interchangeably for acquiring/setting parameter eg. `cstmt.getXXX` or `cstmt.registerOutputParameter`, the driver will need to continue to use metadata query to get the parameter names when the new CS property `useFlexibleCallableStatements=true` (default behaviour).

2. xp_sqljdbc_xa calls have a special case registering their output parameters for CHAR and BINARY types as non-PLP. Prior to this PR, because cstmts calls were wrapped the RPC parameters sent to the server were defined within the `sp_execute` calls and so it wasn't necessary to do so.

3. To read the output parameter return values, parsing the tds response is deferred to the `getXXX` calls. So, if a user executes a cstmt but doesn't get/read the output params, the return values in the tds response will be in limbo/waiting for resolution. In such case, on statement close we will now skip these return values in the tds response which is safe to do.

4. Prior to the PR, it was possible to get the return value from sprocs as a byte array. This shouldn't be possible and this is corrected in this PR. It will throw a casting exception.

5. To allow for unordered setting/registering of parameters, we'll need to continue to use a metadata query for the parameter names (this goes along with point 1 when `useFlexibleCallableStatements=true`). 

6. If `outParams` are null then likely the statement was closed, so we throw an exception.

7. Potentially we can skip building the type definitions as this is no longer needed since we send the RPC params directly. However, this breaks AE and so needs more investigation.

8. For user defined functions (UDF) to work, we need to know when the TDS's response return value is for a sproc or a UDF. To do so, new logic was added to parse the TDS's response to determine this.

9. On the main branch for datetime (datetime, datetime2), the RPC call would always send the parameter as datetime2(7). However, since the cstmts are wrapped, the real param definition was set in the `sp_executesql/sp_prepexec` by building the type definitions in `buildParamTypeDefinitions` which are affected by the `datetimeParameterType` CS property. Since we are executing cstmts directly, datetime (datetime, datetime2) will now do the RPC with the exact datetime parameter definition. Meaning, depending on the `datetimeParameterType` CS property, datetime2 will be sent in the RPC as datetime2(3) exactly by default (which was the prior behaviour done by/indirectly set in the build type definitions) or sent plainly as a datetime.

10. New CS property `useFlexibleCallableStatements`. Default behaviour is `useFlexibleCallableStatements=true`, and in such case the driver will continue to use `sp_sproc_columns`. Otherwise, when `useFlexibleCallableStatements=false`, setting/registering parameters for callable statements require the following:

- a) Setting/Registering callable statement parameters must be exclusively by index or exclusively by parameter name. A mix of index and parameter names will not work.
- b) Setting/Registering callable statement parameters must be in same order as the sproc definition. Setting the parameters in an adhoc order will not work.
- c) Sprocs with cursorable methods will continue as is with sp_sproc_calls regardless of each new CS prop scenario.


src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -5611,7 +5611,7 @@ void writeRPCByteArray(String sName, byte[] bValue, boolean bOut, JDBCType jdbcT
 
         writeRPCNameValType(sName, bOut, tdsType);
 
-        if (usePLP) {
+        if (usePLP && !isNonPLP) {
             // Handle Yukon v*max type header here.
             writeVMaxHeader(nValueLen, bValueNull, collation);
```
lilgreenbird, line 5616: is this comment still relevant?! Yukon hasn't been supported for a while

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCallableStatement.java
```diff
@@ -73,12 +69,25 @@ public class SQLServerCallableStatement extends SQLServerPreparedStatement imple
...
 
     /** atomic integer */
     AtomicInteger ai = new AtomicInteger(0);
 
+    /**
+     * Enum to check if the callablestatement is a setter or getter method. This enum is used in method findColumn where
+     * we get the parameter/column by name
+     *
+     */
+    enum CallableStatementGetterSetterMethod {
```
lilgreenbird, line 86: coding standard - enums should be in caps

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCallableStatement.java
```diff
@@ -73,12 +69,25 @@ public class SQLServerCallableStatement extends SQLServerPreparedStatement imple
     /** Currently active Stream Note only one stream can be active at a time */
     private transient Closeable activeStream;
 
+    /** Checks if return values is already accessed in stored procedure */
+    private boolean returnValueIsAccessed = false;
```
lilgreenbird, line 73: rename to isReturnValueAccessed to be consistent with other booleans

pom.xml
```diff
@@ -43,13 +43,14 @@
...
-			kerberos -  - - - - For tests using Kerberos authentication (excluded by default)
+			kerberos  - - - - - For tests using Kerberos authentication (excluded by default)
+			DTC - - - - - - - - For tests using MSDTC authentication (excluded by default)
 			reqExternalSetup - For tests requiring external setup (excluded by default)
 			clientCertAuth - - For tests requiring client certificate authentication 
 			setup (excluded by default) - - - - - - - - - - - - - - - - - - - - - - - 
 			- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - 
 			Default testing enabled with SQL Server 2019 (SQLv15) -->
-		<excludedGroups>xSQLv12,xSQLv15,NTLM,MSI,reqExternalSetup,clientCertAuth,fedAuth,kerberos</excludedGroups>
+		<excludedGroups>xSQLv12,xSQLv15,NTLM,MSI,reqExternalSetup,clientCertAuth,fedAuth,kerberos,DTC</excludedGroups>
```
lilgreenbird, line 53: we discussed not adding yet another new exclude group

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCallableStatement.java
```diff
@@ -25,11 +25,7 @@
 import java.sql.Timestamp;
 import java.text.MessageFormat;
 import java.time.LocalDateTime;
-import java.util.Calendar;
-import java.util.HashMap;
-import java.util.TreeMap;
-import java.util.Map;
-import java.util.UUID;
+import java.util.*;
```
lilgreenbird, line 28: don't :)

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCallableStatement.java
```diff
@@ -487,7 +543,8 @@ public int getInt(int index) throws SQLServerException {
     public int getInt(String parameterName) throws SQLServerException {
         loggerExternal.entering(getClassNameLogging(), "getInt", parameterName);
         checkClosed();
-        Integer value = (Integer) getValue(findColumn(parameterName), JDBCType.INTEGER);
+        Integer value = (Integer) getValue(
+                findColumn(parameterName, CallableStatementGetterSetterMethod.isGetterMethod), JDBCType.INTEGER);
```
lilgreenbird, line 547: why are we adding this everywhere when this param isn't even used in the method?

src/main/java/com/microsoft/sqlserver/jdbc/dtv.java
```diff
@@ -861,10 +863,28 @@ private void sendTemporal(DTV dtv, JavaType javaType, Object value) throws SQLSe
                                             timestampNormalizedCalendar(calendar, javaType, conn.baseYear()),
                                             subSecondNanos, (valueLength), isOutParam, statement);
                                 }
-                            } else
-                                tdsWriter.writeRPCDateTime2(name,
-                                        timestampNormalizedCalendar(calendar, javaType, conn.baseYear()),
-                                        subSecondNanos, TDS.MAX_FRACTIONAL_SECONDS_SCALE, isOutParam);
+                            } else {
+                                if (jdbcType == JDBCType.SMALLDATETIME) {
```
lilgreenbird, line 867: this whole block seems very awkward, we're in a switch for these types and then we're checking the types again multiple times

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCallableStatement.java
```diff
@@ -1296,29 +1381,59 @@ public java.sql.Array getArray(int parameterIndex) throws SQLException {
...
+     *
      * @param columnName
      *        the name
      * @throws SQLServerException
      *         when an error occurs
      * @return the index
      */
-    private int findColumn(String columnName) throws SQLServerException {
-        if (null == parameterNames) {
+    private int findColumn(String columnName, CallableStatementGetterSetterMethod method) throws SQLServerException {
```
lilgreenbird, line 1408: method param not used

src/main/java/com/microsoft/sqlserver/jdbc/dtv.java
lilgreenbird: also, I know this wasn't added in this PR, but the check for isKatmaiOrLater is obselete as everything supported is > Katmai now so this should be removed since we will never be testing the else case

src/test/java/com/microsoft/sqlserver/jdbc/DTC.java
```diff
@@ -0,0 +1,113 @@
+package com.microsoft.sqlserver.jdbc;
```
lilgreenbird, line 1: file is badly named, all junit test files should be blahTest.java

lilgreenbird: also, why is this test in here instead of in the callablestatement folder?

lilgreenbird: also missing header

src/test/java/com/microsoft/sqlserver/jdbc/TestResource.java
```diff
@@ -76,6 +77,7 @@ protected Object[][] getContents() {
             {"R_supportUnwrapping", "{0} supports unwrapping."},
             {"R_cantAccessSnapshot", "Cant access the TRANSACTION_SNAPSHOT "},
             {"R_newConnectionShouldBeValid", "Newly created connection should be valid"},
+            {"R_outputParamCastError", "class java.lang.Integer cannot be cast to class [B (java.lang.Integer and [B are in module java.base of loader 'bootstrap')"},
```
lilgreenbird, line 80: why is this here? this is not a driver error?

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/DateAndTimeTypeTest.java
```diff
@@ -7,14 +7,7 @@
...
 
-import java.sql.Connection;
-import java.sql.Date;
-import java.sql.PreparedStatement;
-import java.sql.ResultSet;
-import java.sql.SQLException;
-import java.sql.Statement;
-import java.sql.Time;
-import java.sql.Timestamp;
+import java.sql.*;
```
lilgreenbird, line 10: don't use wild card

src/main/java/com/microsoft/sqlserver/jdbc/dtv.java
tkyc: I divided the cases into each switch case. I'm guessing in the past all 3 date/time types used the same `tdsWriter` call and it just got messy over time.

pom.xml
lilgreenbird, line 53: this is not necessary in this PR anymore, right?

tkyc: Oops, yes

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -1155,11 +1195,40 @@ private boolean doPrepExec(TDSWriter tdsWriter, Parameter[] params, boolean hasN
...
+    }
+
+    /**
+     * Checks if the parameter is a TVP type.
+     *
+     * @param params
+     * @return
+     * @throws SQLServerException
+     */
+    private boolean isTVPType(Parameter[] params) throws SQLServerException {
```
Jeffery-Wasty, line 1222: SonarQube marks line 1225 as critical, citing that a static field should not be changed in a non-static method. This method can be and should be marked static. 

Nitpick: remove exception that is never thrown.
```suggestion
    private static boolean isTVPType(Parameter[] params) {
```

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerStatement.java
```diff
@@ -112,10 +118,13 @@ final boolean wasExecuted() {
         return null != tdsReader;
     }
 
+    /** Return parameter for stored procedure calls */
+    Parameter returnParam;
```
Jeffery-Wasty, line 122: SonarQube has a critical error - Fields in a Serializable class must themselves be either Serializable or transient. Is this possible?
```suggestion
    transient Parameter returnParam;
```

src/test/java/com/microsoft/sqlserver/jdbc/TestResource.java
```diff
@@ -43,7 +43,8 @@ protected Object[][] getContents() {
             {"R_lengthTruncated", " The inserted length is truncated or not correct!"},
             {"R_timeValueTruncated", " The time value is truncated or not correct!"},
             {"R_invalidErrorMessage", "Invalid Error Message: "},
-            {"R_kerberosNativeGSSFailure", "No valid credentials provided (Mechanism level: Failed to find any Kerberos tgt)"},
+            {"R_kerberosNativeGSSFailure",
+                    "No valid credentials provided (Mechanism level: Failed to find any Kerberos tgt)"},
```
Jeffery-Wasty, line 47: Not part of this PR, but while you're here please change.
```suggestion
                    "No valid credentials provided (Mechanism level: Failed to find any Kerberos target)"},
```

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/DateAndTimeTypeTest.java
```diff
@@ -295,17 +299,105 @@ public void testSendTimestampAsDatetimeoffset() throws Exception {
         assertEquals(expected, actual);
     }
 
+    @Test
+    public void testCstmtRegisterDatetimeOutParameterDateTimeParameterTypeDatetime() throws Exception {
+        String expected = "3160-08-17 19:09:06.937";
+        String actual = null;
+        String query = "{CALL " + AbstractSQLGenerator.escapeIdentifier(datetimeSproc) + "(?)}";
+
+        // The following should be the T-SQL executed. Since datetimeParameterType=dateTime, the registererd param type
```
Jeffery-Wasty, line 308:
```suggestion
        // The following should be the T-SQL executed. Since datetimeParameterType=dateTime, the registered parameter type
```

src/test/java/com/microsoft/sqlserver/jdbc/callablestatement/CallableStatementTest.java
```diff
@@ -29,11 +31,6 @@
 import org.junit.platform.runner.JUnitPlatform;
 import org.junit.runner.RunWith;
 
-import com.microsoft.sqlserver.jdbc.RandomUtil;
-import com.microsoft.sqlserver.jdbc.SQLServerCallableStatement;
-import com.microsoft.sqlserver.jdbc.SQLServerDataSource;
-import com.microsoft.sqlserver.jdbc.TestResource;
-import com.microsoft.sqlserver.jdbc.TestUtils;
```
Jeffery-Wasty, line 36: Restore proper imports (remove line 25).

src/test/java/com/microsoft/sqlserver/jdbc/callablestatement/CallableStatementTest.java
```diff
@@ -95,6 +104,21 @@ public static void setupTest() throws Exception {
...
             createGetObjectOffsetDateTimeProcedure(stmt);
+            createProcedureZeroParams();
+            createOutOfOrderSproc();
+            createByParamNameSproc();
+            createUserDefinedFunction();
+        }
+    }
+
+    @Test
+    public void testCallableStatementClosedConnection() throws Exception {
```
Jeffery-Wasty, line 115: No exceptions can be thrown.
```suggestion
    public void testCallableStatementClosedConnection() {
```

src/test/java/com/microsoft/sqlserver/jdbc/callablestatement/CallableStatementTest.java
```diff
@@ -342,6 +365,120 @@ public void inputParamsTest() throws SQLException {
...
+
+        try (CallableStatement cs = connection.prepareCall(call)) {
+            cs.registerOutParameter(1, Types.INTEGER);
+            cs.execute();
+            assertEquals(1, cs.getInt(1));
+        }
+    }
+
+    @Test
+    public void testSprocCastingError() throws SQLException {
```
Jeffery-Wasty, line 380:
```suggestion
    public void testSprocCastingError() {
```

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCallableStatement.java
```diff
@@ -73,12 +73,25 @@ public class SQLServerCallableStatement extends SQLServerPreparedStatement imple
...
+    private boolean isReturnValueAccessed = false;
+
     /** map */
     private Map<String, Integer> map = new ConcurrentHashMap<>();
 
     /** atomic integer */
     AtomicInteger ai = new AtomicInteger(0);
 
+    /**
+     * Enum to check if the callablestatement is a setter or getter method. This enum is used in method findColumn where
```
Jeffery-Wasty, line 86:
```suggestion
     * Enum to check if the callable statement is a setter or getter method. This enum is used in method findColumn where
```

Jeffery-Wasty (dismissed):

Jeffery-Wasty (approved):

src/main/java/com/microsoft/sqlserver/jdbc/ISQLServerDataSource.java
```diff
@@ -609,6 +609,22 @@ public interface ISQLServerDataSource extends javax.sql.CommonDataSource {
      */
     boolean getUseDefaultGSSCredential();
 
+    /**
+     * Sets whether or not sp_sproc_columns will be used for parameter name lookup.
+     *
+     * @param useFlexibleCallableStatements
+     *        When set to true, sp_sproc_columns is not used for parameter name lookup
+     *        in callable statements.
```
David-Engel, line 617: My suggestion:
```suggestion
     *        When set to false, sp_sproc_columns is not used for parameter name lookup
     *        in callable statements. This eliminates a round trip to the server but imposes limitations
     *        on how parameters are set. When set to false, applications must either reference
     *        parameters by name or by index, not both. Parameters must also be set in the same
     *        order as the stored procedure definition.
```

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCallableStatement.java
```diff
@@ -73,12 +73,25 @@ public class SQLServerCallableStatement extends SQLServerPreparedStatement imple
...
+     * we get the parameter/column by name
+     *
+     */
+    enum CallableStatementGetterSetterMethod {
+        IS_SETTER_METHOD,
+       IS_GETTER_METHOD
```
David-Engel, line 92:
```suggestion
        IS_GETTER_METHOD
```

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -7813,6 +7829,26 @@ public void setAccessTokenCallbackClass(String accessTokenCallbackClass) {
...
+     */
+    public boolean getUseFlexibleCallableStatements() {
+        return this.useFlexibleCallableStatements;
+    }
+
+    /**
+     * Sets whether or not sp_sproc_columns will be used for parameter name lookup.
+     *
+     * @param useFlexibleCallableStatements
+     *        When set to true, sp_sproc_columns is not used for parameter name lookup
```
David-Engel, line 7845: Duplicate whatever we decide on in ISQLServerConnection.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -98,6 +98,8 @@ public class SQLServerPreparedStatement extends SQLServerStatement implements IS
     /** Set to true if the statement is a stored procedure call that expects a return value */
     final boolean bReturnValueSyntax;
 
+
+
```
David-Engel, line 102:
```suggestion
```

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerStatement.java
```diff
@@ -56,6 +56,7 @@ public class SQLServerStatement implements ISQLServerStatement {
     /**
      * Always update serialVersionUID when prompted.
      */
+
```
David-Engel, line 59:
```suggestion
```

David-Engel (commented):
Please add tests that exercise CallableStatement with useFlexibleCallableStatements set to false and validate success and failure cases. We should detect if they are mixing named/index and throw an error. Not sure if we can detect out of order params.

Jeffery-Wasty (dismissed):

Jeffery-Wasty (approved):

PR #2195, opened 2023-08-16
Escape schemaPattern for getSchemas

Jeffery-Wasty:
`schemaPattern` was the only pattern in `SQLServerDatabaseMetadata` that was not being escaped, it should be.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
@@ -1518,7 +1518,7 @@ public final class SQLServerDatabaseMetaData implements java.sql.DatabaseMetaDat
         if (loggerExternal.isLoggable(Level.FINER) && Util.isActivityTraceOn()) {
             loggerExternal.finer(toString() + ACTIVITY_ID + ActivityCorrelator.getCurrent().toString());
         }
-        return getSchemasInternal(catalog, schemaPattern);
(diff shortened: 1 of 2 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
```diff
@@ -1518,7 +1518,8 @@ public java.sql.ResultSet getSchemas(String catalog, String schemaPattern) throw
         if (loggerExternal.isLoggable(Level.FINER) && Util.isActivityTraceOn()) {
             loggerExternal.finer(toString() + ACTIVITY_ID + ActivityCorrelator.getCurrent().toString());
         }
-        return getSchemasInternal(catalog, schemaPattern);
+        String schema = escapeIDName(schemaPattern);
+        return getSchemasInternal(catalog, schema);
```
Jeffery-Wasty, line 1522: Alternatively, we can just pass in `schemaPattern` and escape inside of `getSchemasInternal`.

Jeffery-Wasty:
I'll add a test.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
Jeffery-Wasty: Added

lilgreenbird (approved):

tkyc (approved):

PR #2209, opened 2023-08-31
Reverted the IBM check to vendor check AND added an additional vm name check to remove false positives

Jeffery-Wasty:
The current `isIBM` check causes issues with OSGi environments. Instead, we're reverting to the previous check with `vendor` name, and adding an additional check with `vm` name to remove false positives, e.g. Semaru builds.

original issue #2139


MrEasy:
easier than https://github.com/microsoft/mssql-jdbc/pull/2208 👍🏻

lilgreenbird (approved):

David-Engel (approved):

PR #2215, opened 2023-09-11
Fix tests for testing encrypt options

lilgreenbird:


src/test/java/com/microsoft/sqlserver/jdbc/fips/FipsTest.java
```diff
@@ -86,7 +86,7 @@ public void fipsPropertyTest() throws Exception {
         props.remove(Constants.ENCRYPT);
         props.remove(Constants.TRUST_SERVER_CERTIFICATE);
 
-        try (Connection con = PrepUtil.getConnection(connectionString, props)) {
+        try (Connection con = PrepUtil.getConnection(connectionString + "encrypt=false", props)) {
```
Jeffery-Wasty, line 89:
```suggestion
        try (Connection con = PrepUtil.getConnection(connectionString + ";encrypt=false", props)) {
```

lilgreenbird: no, not trying to change connectionString, just AEConnectionString

Jeffery-Wasty: Not needed.

Jeffery-Wasty (approved):

src/test/java/com/microsoft/sqlserver/jdbc/fips/FipsTest.java
lilgreenbird: hmm this shouldn't be here it's being ignored, should be removed

Jeffery-Wasty (approved):

lilgreenbird:

Issue #2224, opened 2023-09-28
SQLServerBulkCopy destination column metadata is caching but is not reusing the results, so metadata calls are repeated


tkyc:
When performing a bulk insert, the driver does a destination column metadata query. However, this metadata is not used despite being saved.

The metadata query being ran is the following:

select collation_name, encryption_type from sys.columns where object_id=OBJECT_ID('table') order by column_id ASC

PR #2231, opened 2023-10-06
Fixed bulkcopy metadata query to make use of cached data if available

tkyc:
Prior to this PR, the column metadata query call was made every single time despite data being cached. We now only make the call if no data is cached.

Fixes https://github.com/microsoft/mssql-jdbc/issues/2224.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
@@ -1768,10 +1768,12 @@ public class SQLServerBulkCopy implements java.lang.AutoCloseable, java.io.Seria
                 // Unable to retrieve metadata for destination
                 throw new SQLServerException(SQLServerException.getErrString("R_unableRetrieveColMeta"), e);
             } finally {
-                if (null != rs)
+                if (null != rs) {
                     rs.close();
-                if (null != stmt)
+                }
+                if (null != stmt) {
                     stmt.close();
+                }
             }
         }
     }
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
@@ -1724,53 +1724,55 @@ public class SQLServerBulkCopy implements java.lang.AutoCloseable, java.io.Seria
         SQLServerStatement stmt = null;
         String metaDataQuery = null;
 
-        try {
-            if (null != destinationTableMetadata) {
-                rs = (SQLServerResultSet) destinationTableMetadata;
-            } else {
-                stmt = (SQLServerStatement) connection.createStatement(ResultSet.TYPE_FORWARD_ONLY,
-                        ResultSet.CONCUR_READ_ONLY, connection.getHoldability(), stmtColumnEncriptionSetting);
+        if (null == destColumnMetadata || destColumnMetadata.isEmpty()) {
+            try {
+                if (null != destinationTableMetadata) {
+                    rs = (SQLServerResultSet) destinationTableMetadata;
+                } else {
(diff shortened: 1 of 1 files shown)


David-Engel (dismissed):

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
```diff
@@ -1724,53 +1724,55 @@ private void getDestinationMetadata() throws SQLServerException {
...
+                    rs = (SQLServerResultSet) destinationTableMetadata;
+                } else {
+                    stmt = (SQLServerStatement) connection.createStatement(ResultSet.TYPE_FORWARD_ONLY,
+                            ResultSet.CONCUR_READ_ONLY, connection.getHoldability(), stmtColumnEncriptionSetting);
 
-                // Get destination metadata
-                rs = stmt.executeQueryInternal(
-                        "sp_executesql N'SET FMTONLY ON SELECT * FROM " + escapedDestinationTableName + " '");
-            }
+                    // Get destination metadata
```
lilgreenbird, line 1735: are these tabs here?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
```diff
@@ -1724,53 +1724,55 @@ private void getDestinationMetadata() throws SQLServerException {
...
-                                rsMoreMetaData.getString("collation_name"), bulkCopyEncryptionType));
-                    } else {
-                        destColumnMetadata.put(i, new BulkColumnMetaData(rs.getColumn(i)));
                     }
                 }
+            } catch (SQLException e) {
+                // Unable to retrieve metadata for destination
+                throw new SQLServerException(SQLServerException.getErrString("R_unableRetrieveColMeta"), e);
+            } finally {
+                if (null != rs)
```
Jeffery-Wasty, line 1771: Braces around if statements please.

lilgreenbird (approved):

lilgreenbird (approved):

mmimica:
When is metadata actually being reused?

From what I see, `SQLServerBulkCopy#getDestinationMetadata` is only ever called once from `SQLServerPreparedStatement#execute[Large]Batch`, where a new `SQLServerBulkCopy` object is created for each call.

I think the point of SQLServerBulkCopy#destColumnMetadata is not re-use, but merely having the state around to avoid passing it around as a method argument.

tkyc:
> When is metadata actually being reused?

The wording on the issue can probably be corrected/improved on my end. Originally, the problem scenario occurs when a user repeatedly calls `SQLServerBulkCopy#writeToServer(ISQLServerBulkData sourceData)` (for the same `SQLServerBulkCopy` object). Doing so, the metadata query call would be repeated as well and the `destColumnMetadata` would be repopulated each time which impacted performance. After this PR fix, the subsequent bulk inserts wouldn't make the metadata query call and would "reuse" the existing `destColumnMetadata` that was set from a prior `writeToServer` call. Feel free to Let me know of better phrasing or wording to convey the problem/fix.

> From what I see, SQLServerBulkCopy#getDestinationMetadata is only ever called once from SQLServerPreparedStatement#execute[Large]Batch, where a new SQLServerBulkCopy object is created for each call.

Is this on the main branch of the driver project? On my end I see, `SQLServerBulkCopy#getDestinationMetadata` being called from `SQLServerBulkCopy#writeToServer()`, which is then called by `SQLServerBulkCopy#writeToServer(ISQLServerBulkData sourceData)`. Let me know if I'm incorrect or missing any details here. Thanks.

mmimica:
Are you saying users use `SQLServerBulkCopy` class directly? In that case it makes sense, yes.

tkyc:
> Are you saying users use SQLServerBulkCopy class directly?

That's right.

PR #2212, opened 2023-09-06
Reintroduce previously removed exit condition in login loop + fix retry count

Jeffery-Wasty:
With the introduction of Idle Connection Resiliency in 10.2, we would continually retry on transient errors, until loginTimeout was reached. This was due to the omission of an exit condition. The condition has been added back in.

As well, we were retrying one more time than we should have been, this has been adjusted.

PR #2234, opened 2023-10-10
Fix inner login loop repeating until loginTimeout is reached

Jeffery-Wasty:
Same as #2212, except, where that PR had to be reverted b/c of changes to retry count, retry count is not touched here. Only changes to the inner login loop are present here. This ensures compatibility with current tests.


Jeffery-Wasty:
This will re resolved in the retry rewrite.

Issue #1299, opened 2020-04-01
[FEATURE REQUEST] Message handler - To intercept messages and potentially downgrade any SQLException to SQLWarning


goranschwarz:
Message handler - To intercept messages and potentially downgrade any SQLException to SQLWarning

This can for example be used to:
* logg all messages into a DB-access-errorlog
* simulate clients like 'sqlcmd' or 'ssms'... where a SQL Batch isn't aborted on first error, instead it continues to "read out" the TDS Stream

I'm not sure where the best entry point for adding this functionality, but possibly the TDS Parser, onError() then call the installed message handler (for the Connection)...
* if message handler returns SQLException then add it to the SQLExceptions list
* if message handler returns SQLWarning then add it to the SQLWarnings list

Below is an example of Sybase jConnect implementation of this:
```
// Create a new message handler which will be used for jConnect
SybMessageHandler newMsgHandler = new SybMessageHandler()
{
	@Override
	public SQLException messageHandler(SQLException sqle)
	{
		// Increment Statistics
		if (sqle instanceof SQLWarning)
			incSqlWarningCount();
		else
			incSqlExceptionCount();

		// Add it to the progress dialog
		progress.addMessage(sqle);

		// If we want to STOP if we get any errors...
		// Then we should return the origin Exception
		// SQLException will abort current SQL Batch, while SQLWarnings will continue to execute
		if (_abortOnDbMessages)
			return sqle;

		// Downgrade ALL messages to SQLWarnings, so executions wont be interuppted.
		return AseConnectionUtils.sqlExceptionToWarning(sqle);
	}
};

if (conn instanceof SybConnection)
{
	// get current message handler, so we can restore it later on
	curMsgHandler = ((SybConnection)conn).getSybMessageHandler();

	// Install the new message handler
	((SybConnection)conn).setSybMessageHandler(newMsgHandler);
}
```

FYI: Below is the interface used in jConnect... if something similar is needed...
```
public interface SybMessageHandler 
{
	SQLException messageHandler(SQLException ex);
}
```

and normally a SybSQLException also implemets a interface called `EedInfo`, which basically holds the same methods as https://github.com/microsoft/mssql-jdbc/blob/d4362039923a77caa080152d75d18fc8d1c7ff88/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerError.java
```
public interface EedInfo {
	int getState();
	int getSeverity();
	String getServerName();
	String getProcedureName();
	int getLineNumber();
[...]

Issue #2115, opened 2023-04-28
Exception chaining doesn't work in JDBC driver


akley:
## Driver Version
11.2.3.jre8, but it doesn't work with any version, wether JAVA or JAVA 11

## SQL Server Version
Microsoft SQL Server 2019 (RTM-CU20) (KB5024276) - 15.0.4312.2 (X64)   Apr  1 2023 12:10:46   Copyright (C) 2019 Microsoft Corporation  Standard Edition (64-bit) on Windows Server 2022 Standard 10.0 <X64> (Build 20348: ) (Hypervisor) 

## Problem description

- plain procedure call via Statement.exeute()
- the procedure P1 calls another procedure P2
- the inner procedure P2 raise an error with '_raiserror('P2 - inner exception text', 16, 1) with nowait_' and return 1
- the outer procedure P1 evaluate the return value and raise an error again '_raiserror('P1 - P2 raised an error', 16, 1) with nowait_'
- in the catch block of the jdbc call only the first exception which was transported to the client ("P2 - inner exception text") will be shown. When I'll try _e.getNextException()_ the returned object is NULL. Why?

It seems like the _getNextException()_ and _setNextException()_ methods are not used in _SQLServerException_ that extend _SQLException_ .

When I inspect the raised exception, the property 'next' of SQLException is NULL

All other calls are working fine!

Btw: in  SQL Management Studion, both exceptions are returned to the client, as I excpected.

[p1.txt](https://github.com/microsoft/mssql-jdbc/files/11353834/p1.txt)
[p2.txt](https://github.com/microsoft/mssql-jdbc/files/11353835/p2.txt)

[Test_Chained_Exceptions.java.txt](https://github.com/microsoft/mssql-jdbc/files/11354020/Test_Chained_Exceptions.java.txt)

It would help me a lot if you could give me an answer here. Thanks very much!
Best Regards
Alex

lilgreenbird:
hi @akley 

Thanks for bringing this to our attention, you are correct looks like this is currently not supported I have added this as an feature enhancement to be considered when we do planning for the next semester. 

I do notice our docs erroneously includes these unsupported APIs I will look into getting this updated, thanks.

lilgreenbird:
docs update merged

akley:
getNextExcption()  has been removed from the api doc, but the example still shows: "se = se.getNextException();"
Please remove this also from documentations.

Cheers
Alex

Jeffery-Wasty:
[...]

PR #2251, opened 2023-11-13
Server Message Handler and SQLException Chaining

goranschwarz:
This is a solution to issue https://github.com/microsoft/mssql-jdbc/issues/1299   

## ServerMessageHandler
You can use the message handler for the following
* "message feedback"
   *  Display Server messages from a long running SQL Statement... Like `RAISERROR ('Progress message...', 0, 1) WITH NOWAIT`   
   * Or Status messages from a running backup...
* "Universal" error logging
  * Your error-message handler can contain the logic for handling all error logging
* "Universal" error handling
  * Error-handling logic can be placed in your error-message handler, instead of being repeated throughout your application.
* Remapping of error-message severity (based on application requirements)
   * Your error-message handler can contain logic for recognizing specific error messages, and downgrading or upgrading their severity based on application considerations rather than the severity rating of the server.   
   * For example, during a cleanup operation that deletes old "stuff", you might want to downgrade the severity of "bad things".    
      However, you may want to upgrade the severity in other circumstances.
   * If you write a "SQL Query Tool" (like SSMS) you might not want the Errors to terminate current SQL Batch (throw SQLException), but instead "map" the errors as a "warning"...

##  SQLException Chaining
I also found out that Exceptions do not contains any "extra messages" that is produced by the SQL Server.   
For example when creating a stored proc with several errors in it, you will only see the first problem...
So this just adds any "extra" Messages to the first `SQLWarning` by `setNextException`


goranschwarz:

goranschwarz:

tkyc:
Thanks for the contribution. In case review gets a bit delayed, just to let you know PR is on our radar.

goranschwarz:
Thanks for the update, it would be nice to have it in release 12.5.0, if possible ;)

Jeffery-Wasty:
Hi @goranschwarz,

With our 12.5.0 release planned for today, that doesn't look likely. The next release to include these changes will likely be our 12.6.0 GA release, slated for January 31, 2024.

Jeffery-Wasty:
Adding on to my previous comment. This PR will NOT be included for the January 31 GA release. We'll look into including it in one of the following stable releases after 12.6.0.

Jeffery-Wasty:
Hi @goranschwarz,

This PR introduces a major change, that we consider to be a breaking change. It's also fairly large. These factors necessitate a approach to testing and review, different than what we would use with smaller PRs. With release planned for January 31, this gives us 7 weeks until a code freeze, plenty of time, but the team is currently engaged with other, high profile items. Its unclear when we would be able to start testing and reviewing, and we wouldn't want to rush such a large enhancement.  We'll try to get to this as soon as we can, we don't want a community-contributed PR to sit for an extended amount of time, but timing-wise, it doesn't look like this can make our 12.6.0 GA release.

goranschwarz:
OK; That sounds fare!

Let me know if I can help in any way...

lilgreenbird:
as mentioned above, this isn't going to make the GA release. I've put this into the backlog for now we will look at this again after the Jan release.

tkyc:
Looks good in our internal pipelines. But, it looks like exception chaining isn't working for the scenario described in #2115. From debugging, it looks like the TDS stream is returning only one error message and so only one error message is set.

goranschwarz:
I wanted to change as little as possible!  
The code for `Server Message Handler` is just "hooking into" the current code, **without** changing the *flow* of how things are processed (TDS Streaming wise)

@tkyc Regarding `SQLException chaining`
* The code in the PR will only work for commands that returns several Errors in **one** SQL Command   
   (and nested stored procedure calls is **not** "one" command if you look at the TDS Stream, and how we are currently processing it.)

To solve #2115 (several error messages in nested procedures), or multiple TDS_ERR in the TDS Stream  
We need to do **a lot** more changes (below is possibly why)
* SQLExceptions are thrown **early**, when reading the TDS Stream
* and on `stmnt.close()` we read the **remaining TDS Stream**, which contains those Error Messages (but at that time we have already thrown the SQLException)
So we need to *read out* more of the TDS Stream before SQLExceptions are throws...   
And to do that I think **a lot** of changes and **tests** are needed

What I'm trying to accomplish with the "Server Message Handler", is to be **able** to **downgrade** a `SQLException` to a `SQLWarning`   
(hence trying to be more like SSMS, meaning: continue to read the Data/Messages without aborting (by an SQLException) until the SQL Batch is complete)

### I did some test of the "nested" procedure errors (with some debug printing in the MS JDBC Driver)
This to trace what TDS Stream Tokens that was received (hence where did the Error messages go...)   
   
Simple `stmnt.executeUpdate(sql)` code
Note: I also tested with `stmnt.executeQuery(sql)` and `stmnt.execute(sql)`   
And they shows the "same" output as below   

Test code (I can post the full if you want to):
```
			System.out.println("");
			System.out.println("#####################################################");
			System.out.println("### executeUpdate('" + sql + "')");
			System.out.println("#####################################################");
			Statement stmnt = conn.createStatement();
			try //(Statement stmnt = conn.createStatement())
			{
				stmnt.executeUpdate(sql);

				SQLWarning ex = stmnt.getWarnings();
				while(ex != null)
				{
					System.out.println(">>> WARNING: Msg=" + ex.getErrorCode() + ", xxx=" + ex.getSQLState() + ", Text=|" + ex.getMessage() + "|.");
					ex = ex.getNextWarning();
				}
			}
			catch (SQLException ex)
			{
				while(ex != null)
				{
					System.out.println(">>> EXCEPTION: Msg=" + ex.getErrorCode() + ", xxx=" + ex.getSQLState() + ", Text=|" + ex.getMessage() + "|.");
					ex = ex.getNextException();
				}

				ex = stmnt.getWarnings();
				while(ex != null)
				{
					System.out.println(">>> WARNING: Msg=" + ex.getErrorCode() + ", xxx=" + ex.getSQLState() + ", Text=|" + ex.getMessage() + "|.");
					ex = ex.getNextException();
				}
			}
			finally {
				System.out.println("### stmnt.close()");
				stmnt.close();
			}
```
And it's output
```
#####################################################
### executeUpdate('exec p1 @doSelect = 0')
#####################################################
  -- TDS PARSE(): tdsTokenType=255, getTokenName=TDS_DONEINPROC (0xFF)
     - AT: NextResult.onDone(): doneToken.isError() = false
  -- TDS PARSE(): tdsTokenType=255, getTokenName=TDS_DONEINPROC (0xFF)
     - AT: NextResult.onDone(): doneToken.isError() = false
  -- TDS PARSE(): tdsTokenType=170, getTokenName=TDS_ERR (0xAA)
     - AT: TDSTokenHandler.addDatabaseError(): logContext='getNextResult', databaseError=P2 - inner EXCEPTION text. sev=16
  -- TDS PARSE(): tdsTokenType=255, getTokenName=TDS_DONEINPROC (0xFF)
     - AT: NextResult.onDone(): doneToken.isError() = true <<<<<<<<<<<<<<<<<<<<<<<<<<<<<< parsing will be set to FALSE in the TDS-PARSE() loop.
>>> EXCEPTION: Msg=50000, xxx=S0001, Text=|P2 - inner EXCEPTION text. sev=16|.
### stmnt.close()
  -- TDS PARSE(): tdsTokenType=171, getTokenName=TDS_MSG (0xAB)
  -- TDS PARSE(): tdsTokenType=255, getTokenName=TDS_DONEINPROC (0xFF)
     - AT: NextResult.onDone(): doneToken.isError() = false
  -- TDS PARSE(): tdsTokenType=255, getTokenName=TDS_DONEINPROC (0xFF)
     - AT: NextResult.onDone(): doneToken.isError() = false
  -- TDS PARSE(): tdsTokenType=255, getTokenName=TDS_DONEINPROC (0xFF)
     - AT: NextResult.onDone(): doneToken.isError() = false
  -- TDS PARSE(): tdsTokenType=255, getTokenName=TDS_DONEINPROC (0xFF)
     - AT: NextResult.onDone(): doneToken.isError() = false
  -- TDS PARSE(): tdsTokenType=255, getTokenName=TDS_DONEINPROC (0xFF)
     - AT: NextResult.onDone(): doneToken.isError() = false
  -- TDS PARSE(): tdsTokenType=170, getTokenName=TDS_ERR (0xAA)
     - AT: TDSTokenHandler.addDatabaseError(): logContext='getNextResult', databaseError=P1 - P2 raised an EXCEPTION. sev=16
  -- TDS PARSE(): tdsTokenType=255, getTokenName=TDS_DONEINPROC (0xFF)
     - AT: NextResult.onDone(): doneToken.isError() = true <<<<<<<<<<<<<<<<<<<<<<<<<<<<<< parsing will be set to FALSE in the TDS-PARSE() loop.
  -- TDS PARSE(): tdsTokenType=171, getTokenName=TDS_MSG (0xAB)
  -- TDS PARSE(): tdsTokenType=255, getTokenName=TDS_DONEINPROC (0xFF)
     - AT: NextResult.onDone(): doneToken.isError() = false
  -- TDS PARSE(): tdsTokenType=121, getTokenName=TDS_RET_STAT (0x79)
  -- TDS PARSE(): tdsTokenType=254, getTokenName=TDS_DONEPROC (0xFE)
  -- TDS PARSE(): tdsTokenType=-1, getTokenName=unknown token (0xFFFFFFFF)
```


### Extra info trying to "fix" that we Read more TDS Tokens before Throwing SQLException
Yes I did some test, which initially looked good but I "scrapped" them pretty quickly
The change was to comment out the below line at: SQLServerStatement.java, approx row: 1587
```
                // If the current command (whatever it was) produced an error then stop parsing and propagate it up.
                // In this case, the command is likely to be a RAISERROR, but it could be anything.
System.out.println("     - AT: NextResult.onDone(): doneToken.isError() = " + doneToken.isError() + (doneToken.isError() ? " <<<<<<<<<<<<<<<<<<<<<<<<<<<<<< parsing will be set to FALSE in the TDS-PARSE() loop." : ""));
//                if (doneToken.isError())
//                    return false;

                // In all other cases, keep parsing
                return true;
```
Then it worked...
*BUT* When mixing a couple of `SELECT 'some text'` in the procedures it didn't work anymore...   

### For completeness here are the Stored procedure code
```
drop procedure p2
go
drop procedure p1
go
create procedure p2 (@doSelect int = 0)
as
begin
	if (@doSelect > 0)
		select 'SELECT at -- P2 - inner execution'

	raiserror('P2 - inner EXCEPTION text. sev=16', 16, 1) with nowait
	raiserror('P2 - inner WARNING text. sev=10', 10, 1) with nowait
--	return 0
	return 1
end
go
create procedure p1 (@doSelect int = 0)
as
begin
	declare @rc int

	if (@doSelect > 0)
		select 'SELECT at -- P1 - inner execution -- BEFORE'

	exec @rc=p2 @doSelect

	if (@doSelect > 0)
		select 'SELECT at -- P1 - inner execution -- AFTER'

	if (@rc != 0)
	begin
		raiserror('P1 - P2 raised an EXCEPTION. sev=16', 16, 1) with nowait
		raiserror('P1 - P2 warning an WARNING. sev=10', 10, 1) with nowait
	end
end
go
```

### For even more completeness here are a "path" of what I LOCALLY changed to get the DEBUG Prints
```
diff --git a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerStatement.java b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerStatement.java
index b14f40e..ad0a17d 100644
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerStatement.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerStatement.java
@@ -1583,6 +1583,7 @@
                 }
                 // If the current command (whatever it was) produced an error then stop parsing and propagate it up.
                 // In this case, the command is likely to be a RAISERROR, but it could be anything.
+System.out.println("     - AT: NextResult.onDone(): doneToken.isError() = " + doneToken.isError() + (doneToken.isError() ? " <<<<<<<<<<<<<<<<<<<<<<<<<<<<<< parsing will be set to FALSE in the TDS-PARSE() loop." : ""));
                 if (doneToken.isError())
                     return false;
 
diff --git a/src/main/java/com/microsoft/sqlserver/jdbc/tdsparser.java b/src/main/java/com/microsoft/sqlserver/jdbc/tdsparser.java
index 1a79a43..4a9a03b 100644
--- a/src/main/java/com/microsoft/sqlserver/jdbc/tdsparser.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/tdsparser.java
@@ -64,6 +64,7 @@
         boolean isFeatureExtAck = false;
         while (parsing) {
             int tdsTokenType = tdsReader.peekTokenType();
+System.out.println("  -- TDS PARSE(): tdsTokenType="+tdsTokenType+", getTokenName="+com.microsoft.sqlserver.jdbc.TDS.getTokenName(tdsTokenType));
             if (isLogging) {
                 logger.finest(tdsReader.toString() + ": " + tdsTokenHandler.logContext + ": Processing "
                         + ((-1 == tdsTokenType) ? "EOF" : TDS.getTokenName(tdsTokenType)));
@@ -142,6 +143,7 @@
                     throwUnexpectedTokenException(tdsReader, tdsTokenHandler.logContext);
                     break;
             }
+//System.out.println("  " + (parsing ? "  " : "<<") + "            : parsing == " + parsing + (parsing ? "" : " <<<<<<<<<<<<<<< BREAKING WHILE LOOP <<<<<<<<<<<<<<<"));
         }
 
         // if TDS_FEATURE_EXTENSION_ACK is not received verify if TDS_FEATURE_EXT_AE was sent
@@ -198,6 +200,8 @@
     }
 
     public void addDatabaseError(SQLServerError databaseError) {
+System.out.println("     - AT: TDSTokenHandler.addDatabaseError(): logContext='"+this.logContext+"', databaseError="+databaseError.getErrorMessage());
+//new Exception("databaseError="+databaseError.getErrorMessage()).printStackTrace();
         if (this.databaseError == null) {
             this.databaseError = databaseError;
         } else {
@@ -247,6 +251,7 @@
     boolean onDone(TDSReader tdsReader) throws SQLServerException {
+System.out.println("     - AT tdsParser.onDone(): doneToken.isFinal()="+doneToken.isFinal());
         if (doneToken.isFinal()) {
             // Response is completely processed hence decrement unprocessed response count.
             tdsReader.getConnection().getSessionRecovery().decrementUnprocessedResponseCount();
```


hmmm... This become longer than initially intended :)   
But I hope it made sense, and some clarity!

tkyc:
@goranschwarz 

Thanks for the effort on the detailed response. That clarifies everything. Even though I've been on this project a while there's still things I do not know :smile:. I appreciate the help. Agreed, forget about the exception chaining case I mentioned. 

So far I don't have any other concerns. I'll urge my other team members to start looking at the PR as well.

src/test/java/com/microsoft/sqlserver/jdbc/msghandler/MsgHandlerTest.java
```diff
@@ -0,0 +1,497 @@
...
+
+import com.microsoft.sqlserver.jdbc.ISQLServerMessage;
+import com.microsoft.sqlserver.jdbc.ISQLServerMessageHandler;
+import com.microsoft.sqlserver.jdbc.SQLServerConnection;
+import com.microsoft.sqlserver.jdbc.TestResource;
+import com.microsoft.sqlserver.testframework.AbstractTest;
+
+
+@RunWith(JUnitPlatform.class)
+public class MsgHandlerTest extends AbstractTest {
```
lilgreenbird, line 35: Also change folder name to messageHandler
```suggestion
public class MessageHandlerTest extends AbstractTest {

```

src/test/java/com/microsoft/sqlserver/jdbc/msghandler/MsgHandlerTest.java
```diff
@@ -0,0 +1,497 @@
+/*
+ * Microsoft JDBC Driver for SQL Server Copyright(c) Microsoft Corporation All rights reserved. This program is made
+ * available under the terms of the MIT License. See the LICENSE file in the project root for more information.
+ */
+package com.microsoft.sqlserver.jdbc.msghandler;
```
lilgreenbird, line 5:
```suggestion
package com.microsoft.sqlserver.jdbc.messageHandler;

```

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerInfoMessage.java
```diff
@@ -0,0 +1,124 @@
...
+     * Upgrade a Info message into a Error message
+     * <p>
+     * This simply create a SQLServerError from this SQLServerInfoMessage. 
+     * 
+     * @param newErrorSeverity  - If you want to change the ErrorSeverity (-1: leave unchanged)
+     * @param newErrorNumber    - If you want to change the ErrorNumber   (-1: leave unchanged)
+     * 
+     * @return
+     */
+    public ISQLServerMessage toSQLServerError(int newErrorSeverity, int newErrorNumber)
```
lilgreenbird, line 106: this is only used internally can be private

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -4744,6 +4744,25 @@ void addWarning(String warningString) {
...
+            } else {
+                sqlWarnings.setNextWarning(sqlWarning);
+            }
+        } finally {
+            warningSynchronization.unlock();
+        }
+    }
+
+    // Any changes to SQLWarnings should be synchronized.
+    void addWarning(ISQLServerMessage sqlServerMessage) {
```
lilgreenbird, line 4762: since we now have 2 of these, add comment on how they differ and how they are used

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerInfoMessage.java
```diff
@@ -0,0 +1,124 @@
+/*
+ * Microsoft JDBC Driver for SQL Server Copyright(c) Microsoft Corporation All rights reserved. This program is made
+ * available under the terms of the MIT License. See the LICENSE file in the project root for more information.
+ */
+
+package com.microsoft.sqlserver.jdbc;
+
+import java.sql.SQLException;
+
+public final class SQLServerInfoMessage extends StreamPacket implements ISQLServerMessage {
```
lilgreenbird, line 10: javadoc for class description

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerInfoMessage.java
```diff
@@ -0,0 +1,124 @@
...
+        return toSQLServerError(newErrorSeverity, -1);
+    }
+
+    /**
+     * Upgrade a Info message into a Error message
+     * <p>
+     * This simply create a SQLServerError from this SQLServerInfoMessage. 
+     * 
+     * @param newErrorSeverity  - If you want to change the ErrorSeverity (-1: leave unchanged)
+     * @param newErrorNumber    - If you want to change the ErrorNumber   (-1: leave unchanged)
```
lilgreenbird, line 102: using -1 seem a bit of a hack, can we just create diff overloaded helpers to handle diff scenarios instead

src/main/java/com/microsoft/sqlserver/jdbc/ISQLServerMessageHandler.java
```diff
@@ -0,0 +1,78 @@
+/*
+ * Microsoft JDBC Driver for SQL Server Copyright(c) Microsoft Corporation All rights reserved. This program is made
+ * available under the terms of the MIT License. See the LICENSE file in the project root for more information.
+ */
+package com.microsoft.sqlserver.jdbc;
+
+public interface ISQLServerMessageHandler
```
lilgreenbird, line 7: javadoc class description

src/main/java/com/microsoft/sqlserver/jdbc/ISQLServerMessageHandler.java
```diff
@@ -0,0 +1,78 @@
...
+     * Example code:
+     * <pre>
+     *  public ISQLServerMessage messageHandler(ISQLServerMessage srvErrorOrWarning)
+     *  {
+     *      ISQLServerMessage retObj = srvErrorOrWarning;
+     *
+     *      if (srvErrorOrWarning.isErrorMessage()) {
+     *
+     *          // Downgrade: 2601 -- Cannot insert duplicate key row...
+     *          if (2601 == srvErrorOrWarning.getErrorNumber()) {
```
lilgreenbird, line 41: instead of hardcoding this here, define the list of these special numbers above

src/main/java/com/microsoft/sqlserver/jdbc/ISQLServerMessageHandler.java
```diff
@@ -0,0 +1,78 @@
...
+     *       Your error-message handler can contain logic for recognizing specific error messages, and downgrading or upgrading 
+     *       their severity based on application considerations rather than the severity rating of the server. 
+     *       For example, during a cleanup operation that deletes old rows, you might want to downgrade the severity of a 
+     *       message that a row does not exist. However, you may want to upgrade the severity in other circumstances.
+     *   </li>
+     * </ul>
+     * 
+     * Example code:
+     * <pre>
+     *  public ISQLServerMessage messageHandler(ISQLServerMessage srvErrorOrWarning)
```
lilgreenbird, line 34: nitpick: since this is a public API, can you please use a full spelling i.e. serverErrorOrWarning

src/main/java/com/microsoft/sqlserver/jdbc/ISQLServerMessageHandler.java
```diff
@@ -0,0 +1,78 @@
...
+     *       This results in the warning being added to the warning-message chain.
+     *   </li>
+     *   <li><b>SQLServerError</b> object<br>
+     *       If the originating message is a SQL warning (SQLServerInfoMessage object), messageHandler can evaluate 
+     *       the SQL warning as urgent and create and return a SQL exception (SQLServerError object)
+     *       to be thrown once control is returned to the JDBC Driver.
+     *   </li>
+     * </ul>
+     */
+    ISQLServerMessage messageHandler(ISQLServerMessage srvErrorOrWarning);
```
lilgreenbird, line 77: nitpick: since this is a public API, can you please use a full spelling i.e. serverErrorOrWarning

lilgreenbird (changes requested):
Please make sure to use the [formatter ](https://github.com/microsoft/mssql-jdbc/blob/main/mssql-jdbc_formatter.xml) on all files. I'm seeing a lot of  spacing diffs when I pulled this branch.

Also, pls resolve conflicts

src/main/java/com/microsoft/sqlserver/jdbc/ISQLServerMessageHandler.java
lilgreenbird, line 77: nitpick: since this is a public API, can you please use a full spelling instead
```suggestion
    ISQLServerMessage messageHandler(ISQLServerMessage serverErrorOrWarning);

```

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -8263,6 +8282,34 @@ public void setIPAddressPreference(String iPAddressPreference) {
...
+
+
+    /** Message handler */
+    private transient ISQLServerMessageHandler serverMessageHandler;
+    
+    /**
+     * Set current message handler
+     * 
+     * @param messageHandler
+     * @return The previously installed message handler (null if none)
```
lilgreenbird, line 8295: why does the setter need to return anything? there is already a getter

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -8263,6 +8282,34 @@ public void setIPAddressPreference(String iPAddressPreference) {
...
+     */
+    @Override
+    public ISQLServerMessageHandler setServerMessageHandler(ISQLServerMessageHandler messageHandler)
+    {
+    	ISQLServerMessageHandler installedMessageHandler = this.serverMessageHandler;
+    	this.serverMessageHandler = messageHandler;
```
lilgreenbird, line 8301: this overwrites any existing, maybe should check and log a message?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerInfoMessage.java
goranschwarz: It should stay public, since the idea is to use them in the MessageHandler, which is located in client code

src/test/java/com/microsoft/sqlserver/jdbc/msghandler/MsgHandlerTest.java
goranschwarz: done

src/test/java/com/microsoft/sqlserver/jdbc/msghandler/MsgHandlerTest.java
goranschwarz: done

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
goranschwarz: done

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerInfoMessage.java
goranschwarz: done

goranschwarz: I also added javadoc class description for SQLServerWarning

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerInfoMessage.java
goranschwarz: Not 100% sure what you say here!
But I have:
* `toSQLServerError()`    
   to be used if **no** changes to the 'errorNumber' and 'errorSeverity'
* `toSQLServerError(int newErrorSeverity)`    
   to be used if **only** 'errorNumber' should be changed
* `toSQLServerError(int newErrorSeverity, int newErrorNumber)`     
   to be used if you want to change **both** the 'errorNumber' and 'errorSeverity'

The first two methods just calls: `toSQLServerError(int newErrorSeverity, int newErrorNumber)`   
So we don't have to duplicate the code...


Give me an example of how I should do it...

src/main/java/com/microsoft/sqlserver/jdbc/ISQLServerMessageHandler.java
goranschwarz: done

src/main/java/com/microsoft/sqlserver/jdbc/ISQLServerMessageHandler.java
goranschwarz: done

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
goranschwarz: This is just convenience...   
Meaning: not have to call getter if you want to save the previous installed handler...   
Which may be used if you want to restore previous installed handler at a later point!

For example the java Properties.getProperty("someName") does the same thing!
https://docs.oracle.com/javase/8/docs/api/java/util/Properties.html#setProperty-java.lang.String-java.lang.String-

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
goranschwarz: Yes a setter usually overwrites previous values...
If you don't want to overwrite it, use the getter and check if you have one...

If you want: I can generate a log message (let me know the LogLevel)   
But from my perspective, I would leave it as is

goranschwarz:
@lilgreenbird 
Just to be sure: I guess you have seen my answers to your "review comments"?
But you are busy with other stuff...

Jeffery-Wasty (approved):

lilgreenbird (approved):

tkyc (approved):

PR #2272, opened 2023-12-11
Added cleanup task to start of every execute.

Jeffery-Wasty:
Under certain conditions, prepared statement cache can end up being filled quicker than expected, while being cleaned less often than expected. This adds additional cache clean triggers making the cleaning much more frequent.


ecki:
Do you really want to force=true the cleanup, doesn’t that add roundtrips?

Jeffery-Wasty:
For sure this can be improved. I have only opened this PR so the user can see our proposed changes and try testing them out, and see if that resolves the issue. For an actual fix, we will look into a cleaner and more performance-friendly solution.

ecki:
> can see our proposed changes and try testing them out, and see if that resolves the issue.

ah yes, I understand, makes sense.

it will check if it fixes the handles accumulated in my reproducer test, but of course not sure if that is the exact problem of OP

Jeffery-Wasty:
No we're not, but from the information we have, this is the best guess.

David-Engel:
For the production fix, I'd set force to false. (Assuming this is the fix.) We don't need to be super aggressive on the cleanup. It's just that if an application is doing millions of executions of a PreparedStatement and some of them occasionally result in an invalidated prepared statement handle, we should clean up periodically to avoid continuously increasing memory usage. Current cleanup only happens on close of the PreparedStatement and connection close.

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/PreparedStatementTest.java
```diff
@@ -400,9 +398,7 @@ public void testStatementPoolingEviction() throws SQLException {
                 }
 
                 // If we use it, now discard queue should be "full".
-                if (0 == testNo)
-                    assertSame(discardedStatementCount, con.getDiscardedServerPreparedStatementCount());
-                else
+                if (0 != testNo)
```
lilgreenbird, line 401: the test case for testNo==0 was removed we're not testing this anymore?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -2199,6 +2201,7 @@ public int[] executeBatch() throws SQLServerException, BatchUpdateException, SQL
             loggerExternal.finer(toString() + ACTIVITY_ID + ActivityCorrelator.getCurrent().toString());
         }
         checkClosed();
+        connection.unprepareUnreferencedPreparedStatementHandles(false);
```
lilgreenbird, line 2204: there are more occurrences of executeStatement in the file should we be doing this for every one of them?

lilgreenbird (approved):

barryw-mssql (approved):

PR #2273, opened 2023-12-12
Update retries to not retry when connectRetryCount=0

lilgreenbird:


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -1859,91 +1879,7 @@ Connection connect(Properties propsIn, SQLServerPooledConnection pooledConnectio
             loginTimeoutSeconds *= 10;
```
tkyc, line 1879: `loginTimeoutSeconds` doesn't look like it's used. We should probably remove it.

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -2460,13 +2463,32 @@ Socket findSocket(String hostName, int portNumber, int timeoutInMilliSeconds, bo
...
+
+                if (inetAddrs != null && inetAddrs.length > IP_ADDRESS_LIMIT) {
+                    MessageFormat form = new MessageFormat(
+                            SQLServerException.getErrString("R_ipAddressLimitWithMultiSubnetFailover"));
+                    Object[] msgArgs = {Integer.toString(IP_ADDRESS_LIMIT)};
+                    String errorStr = form.format(msgArgs);
+                    // we do not want any retry to happen here. So, terminate the connection
+                    // as the config is unsupported.
+                    conn.terminate(SQLServerException.DRIVER_ERROR_UNSUPPORTED_CONFIG, errorStr);
+                }
```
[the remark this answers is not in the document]
lilgreenbird: this code isn't new it was there before

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -3347,22 +3336,13 @@ private void login(String primary, String primaryInstanceName, int primaryPortNu
...
-                        || SQLServerException.DRIVER_ERROR_SSL_FAILED == driverErrorCode // SSL failure
-                        || SQLServerException.DRIVER_ERROR_INTERMITTENT_TLS_FAILED == driverErrorCode // TLS1.2 failure
-                        || SQLServerException.DRIVER_ERROR_UNSUPPORTED_CONFIG == driverErrorCode // unsupported config
-                                                                                                 // (eg Sphinx, invalid
-                                                                                                 // packetsize, etc)
-                        || (SQLServerException.ERROR_SOCKET_TIMEOUT == driverErrorCode // socket timeout
-                                && (!isDBMirroring || attemptNumber > 0)) // If mirroring, only close after failover has been tried (attempt >= 1)
-                        || timerHasExpired(timerExpire)
-                // for non-dbmirroring cases, do not retry after tcp socket connection succeeds
+                if (shouldFailConnection(errorCode, driverErrorCode, isDBMirroring, useTnir, attemptNumber,
```
tkyc, line 3342: If a transient error were to happen here, it looks like it would immediately terminate the connection. Prior, the logic was that if a transient error occurred, the driver would attempt another connection.

If a redirection failure were to happen, and a transient error was thrown, the driver wouldn't go back to the GW to attempt another connection.

PR #2284, opened 2024-01-02
Execute cstmt directly - Additional testing and changes

tkyc:
These changes make it so that when `useFlexibleCallableStatements=false`, setting/registering of parameters is stricter eg. either use all index or use named parameters (no mixing of index/named parameters). The reason for this is because the exceptions/cases to using index vs. named parameters are too nuanced.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCallableStatement.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCallableStatement.java
@@ -2550,8 +2550,6 @@ public class SQLServerCallableStatement extends SQLServerPreparedStatement imple
             loggerExternal.entering(getClassNameLogging(), "registerOutParameter",
                     new Object[] {index, sqlType, typeName});
 
-        checkClosed();
-
         registerOutParameterByName(index, sqlType);
 
         loggerExternal.exiting(getClassNameLogging(), "registerOutParameter");
@@ -2569,8 +2567,6 @@ public class SQLServerCallableStatement extends SQLServerPreparedStatement imple
         if (loggerExternal.isLoggable(java.util.logging.Level.FINER))
             loggerExternal.entering(getClassNameLogging(), "registerOutParameter",
                     new Object[] {index, sqlType, scale});
-
-        checkClosed();
 
         registerOutParameterByName(index, sqlType);
         inOutParam[index - 1].setOutScale(scale);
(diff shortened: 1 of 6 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerStatement.java
```diff
@@ -1136,6 +1136,14 @@ void checkClosed() throws SQLServerException {
         }
     }
 
+    void setByIndex() throws SQLServerException {
+        isSetByIndex = true;
+        if (!connection.getUseFlexibleCallableStatements() && isSetByName && isSetByIndex) {
+            SQLServerException.makeFromDriverError(connection, this,
+                    SQLServerException.getErrString("R_noNamedAndIndexedParameters"), null, false);
```
lilgreenbird, line 1143: update "R_noNamedAndIndexedParameters" to be more direct and less wordy

lilgreenbird (approved):

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCallableStatement.java
```diff
@@ -2495,8 +2543,17 @@ public void registerOutParameter(String parameterName, int sqlType, String typeN
...
-        registerOutParameter(findColumn(parameterName, CallableStatementGetterSetterMethod.IS_SETTER_METHOD), sqlType,
-                typeName);
+
+        int index = findColumn(parameterName, CallableStatementGetterSetterMethod.IS_SETTER_METHOD);
+
+        if (loggerExternal.isLoggable(java.util.logging.Level.FINER))
+            loggerExternal.entering(getClassNameLogging(), "registerOutParameter",
+                    new Object[] {index, sqlType, typeName});
+
+        checkClosed();
```
Jeffery-Wasty, line 2553: How likely is it that the connection is open before findColumn() and closed after? +2 more instances of the same below.

Jeffery-Wasty (approved):

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCallableStatement.java
```diff
@@ -2518,8 +2581,17 @@ public void registerOutParameter(String parameterName, int sqlType, int precisio
             loggerExternal.entering(getClassNameLogging(), "registerOutParameter",
                     new Object[] {parameterName, sqlType, scale});
         checkClosed();
-        registerOutParameter(findColumn(parameterName, CallableStatementGetterSetterMethod.IS_SETTER_METHOD), sqlType,
-                precision, scale);
+
+        int index = findColumn(parameterName, CallableStatementGetterSetterMethod.IS_SETTER_METHOD);
+
+        if (loggerExternal.isLoggable(java.util.logging.Level.FINER))
```
lilgreenbird, line 2587: we're doing this twice there's already a log entering above

lilgreenbird (approved):

Jeffery-Wasty (approved):

Issue #2271, opened 2023-12-04
Bulk Copy API does not perform Timezone conversion for DATETIME2(3) with PreparedStatement.setTimestamp


PriyadharshiniP:
## Driver version
mssql-jdbc version 12.4.1.jre11

## SQL Server version
Microsoft SQL Server 2019 (RTM-CU18) (KB5017593) - 15.0.4261.1 (X64)   Sep 12 2022 15:07:06   Copyright (C) 2019 Microsoft Corporation  Developer Edition (64-bit) on Windows 10 Pro 10.0 <X64> (Build 19045: ) 

## Client Operating System
Windows

## JAVA/JVM version
Java 17

## Problem description

**DATETIME2(3)** type is supported by the Bulk Copy API.  DATETIME2 is not there in the unsupported datatypes list as per: https://learn.microsoft.com/en-us/sql/connect/jdbc/using-bulk-copy-with-the-jdbc-driver?view=sql-server-ver16

Nevertheless, it was discovered that there appears to be NO timezone conversion with Bulk copy API for DATETIME2(3) column type.
The following is specified in the PreparedStatement.setTimestamp contract:  "Sets the designated parameter to the given {@code java.sql.Timestamp} value,using the given {@code Calendar} object.  The driver usesthe {@code Calendar} object to construct an SQL {@code TIMESTAMP} value, which the driver then sends to the database.  With a {@code Calendar} object, the driver can calculate the timestamptaking into account a custom timezone.  If no{@code Calendar} object is specified, the driver uses the default timezone, which is that of the virtual machine running the application." 
Bulk copy API seems to ignore this calendar settings and inserts the timestamp value in the default timezone.
Please see the following stand alone Java code that inserts the first row WITHOUT bulk API and the second row WITH bulk API.

```
package com.test;

import java.sql.Statement;
import java.sql.Timestamp;
import java.text.SimpleDateFormat;
import java.util.ArrayList;
import java.util.Calendar;
import java.util.Date;
import java.util.List;
import java.util.TimeZone;
import java.sql.Connection;
import java.sql.DriverManager;
import java.sql.PreparedStatement;

public class TestBulkAPI {
	public static void main(String[] args) throws Exception {
		String tableNameBulkCopyAPI = "testTable";
		List<Timestamp> loT = new ArrayList<>();
		Calendar gmtCal = Calendar.getInstance(TimeZone.getTimeZone("GMT"));
[...]

PR #2291, opened 2024-01-10
Apply timezone to Timestamp when using bulkcopy for batch insert

tkyc:
Correctly applies the timezone to Timestamps when inserted using batch insert with bulkcopy.

Fixes https://github.com/microsoft/mssql-jdbc/issues/2271


PriyadharshiniP:
Hi @tkyc 

I appreciate you making the changes.

I observed that the fix had the following problems:
1. Could you kindly have a look at the applyTimezoneTo method, which appears to use 
`Date date = new Date(**cal**.getTimeInMillis());` rather than `Date date = new Date(**ts.getTime()**); }?
If you have a time value other than today's time in your unit test, **BatchExecutionTest** for **_testValidTimezoneForTimestampBatchInsertWithBulkCopy_**, it will begin to break. The unit test would have seemed to be successful because the timestamp and calendar will often have the same current time value. However, this unit test would begin to fail if you had a different timestamp value.
For example, the unit test failed when I modified the timestamp milliseconds from `long ms = Timestamp.valueOf(todayString).getTime();} to `long ms = 1433338533461L;`. that leads to the failure.

image

2. In the event that the batch insert for bulk copy API falls back to the original batch insert, this fix does not seem to function. An additional conversion occurs when the fallback to no bulk copy API is used, leading to a double conversion for GMT.
To replicate this issue, Use the existing `BulkCopyResultSetCursorTest` for `testMultiplePreparedStatementAndResultSet ` unit test method alone. Connection string should have  "useBulkCopyForBatchInsert=true;sendTemporalDataTypesAsStringForBulkCopy=false;". 
Because bulk copy does not allow IDENTITY columns, using bulk copy would fail and result in inaccurate database values with fallback.
image

tkyc:
@PriyadharshiniP Thanks for the feedback. I'll do more investigating and I'll look into correcting the implementation.

> May I ask if the changes need to be done at SQLServerBulkCopy ?

That was my initial thought also, but when I looked over SQLServerBulkCopy it doesn't look like I'm able to apply the timezone in the TDS request for bulkcopy.

For example, I sort of expected code like this in SQLServerBulkCopy, which is part of the normal and regular batch insert code path. The code below takes into consideration the the timezone by sending the timezone info in the TDS request. So the timezone is applied during insertion by the server:

```
                                    tdsWriter.writeEncryptedRPCDateTime2(name,
                                            timestampNormalizedCalendar(calendar, javaType, conn.baseYear()),
                                            subSecondNanos, (valueLength), isOutParam, statement);
```

But I'm guessing because we're doing a bulkcopy, it has to send the timestamp value as is. And so, either the user or driver needs to apply the timezone conversion beforehand.

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/BatchExecutionTest.java
```diff
@@ -54,6 +60,8 @@ public class BatchExecutionTest extends AbstractTest {
     private static String ctstable3;
     private static String ctstable4;
     private static String ctstable3Procedure1;
+    private static String timestampTable = AbstractSQLGenerator
```
lilgreenbird, line 63: this table should be added to dropTable() for cleanup

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/BatchExecutionTest.java
```diff
@@ -99,6 +107,55 @@ public void testBatchUpdateCountTrueOnFirstPstmtSpPrepare() throws Exception {
...
+
+        try (Connection con = DriverManager.getConnection(connectionString); Statement stmt = con.createStatement();
+                PreparedStatement pstmt = con.prepareStatement("INSERT INTO " + timestampTable + " VALUES(?)")) {
+
+            String dropSql = "IF EXISTS (SELECT * FROM dbo.sysobjects WHERE id = object_id(N'"
+                    + TestUtils.escapeSingleQuotes(timestampTable)
+                    + "') and OBJECTPROPERTY(id, N'IsUserTable') = 1) DROP TABLE " + timestampTable;
+            stmt.execute(dropSql);
+
+            String createSql = "CREATE TABLE" + timestampTable + " (c1 DATETIME2(3))";
```
lilgreenbird, line 123: this table should be added to dropTable() or add finally to cleanup

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/BatchExecutionTest.java
```diff
@@ -99,6 +107,55 @@ public void testBatchUpdateCountTrueOnFirstPstmtSpPrepare() throws Exception {
...
 
+    @Test
+    public void testValidTimezoneForTimestampBatchInsertWithBulkCopy() throws Exception {
+        Calendar gmtCal = Calendar.getInstance(TimeZone.getTimeZone("GMT"));
+        long ms = 1578743412000L;
+
+        try (Connection con = DriverManager.getConnection(connectionString); Statement stmt = con.createStatement();
+                PreparedStatement pstmt = con.prepareStatement("INSERT INTO " + timestampTable + " VALUES(?)")) {
+
+            String dropSql = "IF EXISTS (SELECT * FROM dbo.sysobjects WHERE id = object_id(N'"
```
lilgreenbird, line 118: dropTableIfExists

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/BatchExecutionTest.java
```diff
@@ -13,11 +13,17 @@
...
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
+import java.sql.Timestamp;
+import java.text.SimpleDateFormat;
 import java.util.Arrays;
+import java.util.Calendar;
+import java.util.Date;
+import java.util.TimeZone;
```
lilgreenbird, line 26: pls cleanup unused

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/BatchExecutionTest.java
```diff
@@ -99,6 +107,55 @@ public void testBatchUpdateCountTrueOnFirstPstmtSpPrepare() throws Exception {
...
+
+            Timestamp ts0; // Timestamp batch inserted without bulkcopy
+            Timestamp ts1; // Timestamp batch inserted with bulkcopy
+
+            rs.next();
+            ts0 = rs.getTimestamp(1);
+            rs.next();
+            ts1 = rs.getTimestamp(1);
+
+            assertEquals(ts0, ts1);
```
lilgreenbird, line 155: maybe add a bit more debug info if it fails? I think assertEquals will just tell u false it doesn't equal but not the values

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/BatchExecutionTest.java
```diff
@@ -99,6 +107,55 @@ public void testBatchUpdateCountTrueOnFirstPstmtSpPrepare() throws Exception {
...
+
+            Timestamp timestamp = new Timestamp(ms);
+
+            pstmt.setTimestamp(1, timestamp, gmtCal);
+            pstmt.addBatch();
+            pstmt.executeBatch();
+        }
+
+        try (Connection con = DriverManager.getConnection(connectionString); Statement stmt = con.createStatement()) {
+            ResultSet rs = stmt.executeQuery("SELECT * FROM " + timestampTable);
```
lilgreenbird, line 145: ResultSet is leaked

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -2938,6 +2947,44 @@ private boolean checkSQLLength(int length) {
         return true;
     }
 
+    private String getDateFormatPattern(int nanos) {
```
lilgreenbird, line 2950: add comments on what this does

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -2938,6 +2947,44 @@ private boolean checkSQLLength(int length) {
...
+            if (i == 0) {
+                pattern.append(".S");
+            }
+            pattern.append("S");
+        }
+
+        return pattern.toString();
+    }
+
+    private java.sql.Timestamp applyTimezoneTo(java.sql.Timestamp ts, Calendar cal) {
```
lilgreenbird, line 2977: add comment for method

PriyadharshiniP:
> 1. **BatchExecutionTest**

Thanks for fixing the issue.
1.  Additionally, it appears that not all dates—particularly those with daylight savings time—cannot be converted successfully using the applyTimezoneTo method . In your BatchExecutionTest unit test case , if you change the ms to ` long ms = 1696127400000L;` that equates to :

```
GMT: Sunday, October 1, 2023 2:30:00 AM
My local time zone: Sunday, October 1, 2023 1:30:00 PM [GMT+11:00]
```

This time "GMT: Sunday, October 1, 2023 2:30:00 AM" doesn't exist in my local time zone. The non bulk version correctly inserts the date as `2023-10-01 02:30:00.000` but the bulk version incorrectly inserts as `2023-10-01 03:30:00.000` and the BatchExecutionTest  will fail if we use  ` ts0 = rs.getTimestamp(1, gmtCal);. and ts1 = rs.getTimestamp(1, gmtCal);`
I think this is because of using Timestamp.valueOf or any timestamp related methods that will use only local time for conversions.

image

2. As the standard batch insert without bulk copy seems to work in all cases, I started to look re-use the tdsWriter.writeEncryptedRPCDateTime2 in SQLServerBulkCopy itself, however while attempting to fix it, I encounter certain problems.
To store the parameters needed for the writeEncryptedRPCDateTime2 method, I made a new class. Next, update the SQLServerBulkBatchInsertRecord class with those values. Fill in the datetime2 field in SQLServerBulkCopy in accordance with the screenshots below :
image
image
However, the following error appears:  `java.sql.BatchUpdateException: While reading current row from host, a premature end-of-message was encountered--an incoming data stream was interrupted when the server expected to see more data. The host program may have terminated. Ensure that you are using a supported client application programming interface (API).` 
Kindly inform me if you can assist with this.

Could you see if we can accomplish something along these lines?

tkyc:
@PriyadharshiniP Thanks for collaborating.

> As the standard batch insert without bulk copy seems to work in all cases, I started to look re-use the tdsWriter.writeEncryptedRPCDateTime2 in SQLServerBulkCopy

Yeah, I struggled with getting that working as well when I initially looked at the code. In `SQLServerBulkCopy`, it seems the only thing written to the TDS request is column metadata info and data type info and it doesn't look like any input values are written eg. the timestamp in our case for batch insert with bulkcopy. When I traced the batch insert bulkcopy call path:

```
writeToServer -> sendBulkLoadBCP -> ... -> doInsertBulk -> sendBulkCopyCommand
```

In the above call path, I expected to end up somewhere where the timestamp value is written to the TDS request (something/somewhere where the logic `writeEncryptedRPCDateTime2 ` can be used) but I don't. I'll do some more debugging, maybe I missed some things initially.

> Additionally, it appears that not all dates—particularly those with daylight savings time—cannot be converted successfully using the applyTimezoneTo method

Thanks for the help on testing. I'll see what I can do here...

tkyc:
@PriyadharshiniP What's your exact local time zone? I've switched my PC's time zone setting to various different time zones that doesn't have day light savings and wasn't able to repro the issue. I double checked the local time zone used in my JDK and it matches my PC time zone.

PriyadharshiniP:
My local timezone is Australia/Sydney

PriyadharshiniP:
> @PriyadharshiniP What's your exact local time zone? I've switched my PC's time zone setting to various different time zones that doesn't have day light savings and wasn't able to repro the issue with `1696127400000L`. I double checked the local time zone used in my JDK and it matches the PC time zone set when I was experimenting.
> 
> Maybe I need to do something like below explicitly to repro your issue?
> 
> ```
> TimeZone tz = TimeZone.getTimeZone("your-time-zone");
> TimeZone.setDefault(tz);
> ```

I guess you could do something like this in your unit test  :`System.setProperty("user.timezone","Australia/Sydney");`

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/BatchExecutionTest.java
tkyc: `assertEquals` will output an expected vs. actual message, so no need for an actual failure message.

tkyc:
@PriyadharshiniP I've just pushed a new fix implementation. I did some testing of it and it seem to have addressed all your concerns so far. Let me know if there's any additional problems. Thanks again for the collab.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
```diff
@@ -2480,10 +2480,13 @@ else if (4 >= bulkScale)
...
-                                String timeStampValue = colValue.toString();
-                                tdsWriter.writeTime(java.sql.Timestamp.valueOf(timeStampValue), bulkScale);
+                                Timestamp ts = java.sql.Timestamp.valueOf(colValue.toString());
+                                tdsWriter.writeTime(ts, bulkScale, cal);
                                 // Send only the date part
-                                tdsWriter.writeDate(timeStampValue.substring(0, timeStampValue.lastIndexOf(' ')));
+                                tdsWriter.writeDate(ts.getTime(), cal);
+                                //tdsWriter.writeRPCDateTime2(name,
+                                //        timestampNormalizedCalendar(calendar, javaType, conn.baseYear()),
+                                //        subSecondNanos, TDS.MAX_FRACTIONAL_SECONDS_SCALE, isOutParam);
```
Jeffery-Wasty, line 2489: Remove commented out code

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -3836,6 +3850,22 @@ void writeTime(java.sql.Timestamp value, int scale) throws SQLServerException {
         writeScaledTemporal(calendar, subSecondNanos, scale, SSType.TIME);
     }
 
+    void writeTime(java.sql.Timestamp value, int scale, Calendar cal) throws SQLServerException {
+        GregorianCalendar calendar = initializeCalender(TimeZone.getDefault());
+        long utcMillis; // Value to which the calendar is to be set (in milliseconds 1/1/1970 00:00:00 GMT)
+        int subSecondNanos;
+        utcMillis = value.getTime();
+        subSecondNanos = value.getNanos();
```
Jeffery-Wasty, line 3858: Declarations and assignments can be combined.
```suggestion
        long utcMillis = value.getTime(); // Value to which the calendar is to be set (in milliseconds 1/1/1970 00:00:00 GMT)
        int subSecondNanos = value.getNanos();
```

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/BatchExecutionTest.java
```diff
@@ -54,6 +60,10 @@ public class BatchExecutionTest extends AbstractTest {
     private static String ctstable3;
     private static String ctstable4;
     private static String ctstable3Procedure1;
+    private static String timestampTable1 = AbstractSQLGenerator
+            .escapeIdentifier(RandomUtil.getIdentifier("timestamptable1"));
+    private static String timestampTable2 = AbstractSQLGenerator
+            .escapeIdentifier(RandomUtil.getIdentifier("timestampTableBatchInsert2"));
```
lilgreenbird, line 66: stupid nitpick, usually we use same identifier as the name.

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/BatchExecutionTest.java
```diff
@@ -99,6 +109,186 @@ public void testBatchUpdateCountTrueOnFirstPstmtSpPrepare() throws Exception {
...
+            Timestamp timestamp = new Timestamp(ms);
+
+            pstmt.setTimestamp(1, timestamp, gmtCal);
+            pstmt.addBatch();
+            pstmt.executeBatch();
+        }
+
+        // Compare Timestamp values inserted, should be the same
+        try (Connection con = DriverManager.getConnection(connectionString); Statement stmt = con.createStatement()) {
+            try (ResultSet rs = stmt.executeQuery("SELECT * FROM " + timestampTable1)) {
```
lilgreenbird, line 146: why not combine to same try?

PriyadharshiniP:
> @PriyadharshiniP I've just pushed a new fix implementation. I did some testing of it and it seem to have addressed all your concerns so far. Let me know if there's any additional problems. Thanks again for the collab.

Hi @tkyc 
Thanks for fixing it so quickly . 
I checked all dates this year in my local time zone, and it seems that there will still be problems with the bulk copy inserting the date in my local environment around spring forward dates. The time stamp millisecond values are 1712415600000L and 1743935400000L, to be more precise.
Noted the use of `java.sql.Timestamp.valueOf(colValue.toString())` in SQLServerBulkCopy, this seems to be the cause of this issue as it seems to alter the timestamp milliseconds value. Using the `(Timestamp)colValue` seems to work for all dates. Could you please check and fix this as well?

image
image

tkyc:
@PriyadharshiniP Oops, my mistake. Calling valueOf will cause a conversion. Fixed.

Jeffery-Wasty (dismissed):

lilgreenbird (approved):

Jeffery-Wasty (approved):

PriyadharshiniP:
> @PriyadharshiniP Oops, my mistake. Calling valueOf will cause a conversion. Fixed.

Thanks @tkyc :
It seems to be fine for my tests.
Please let me  know when the release would be available ?

Regards

tkyc:
@PriyadharshiniP this will make it in the release that's on the 31st of this month.

Issue #2288, opened 2024-01-05
When connecting with encrypt=true method, the Connection.abort(Executor) method does not work correctly


FischerRo2:
## Driver version
Provide the JDBC driver version (e.g. 10.2.0).
12.4.1
## SQL Server version
Microsoft SQL Server 2017 (RTM-CU31-GDR) (KB5021126) - 14.0.3460.9 (X64)   Jan 25 2023 08:42:43   Copyright (C) 2017 Microsoft Corporation  Developer Edition (64-bit) on Windows Server 2012 R2 Datacenter 6.3 <X64> (Build 9600: ) (Hypervisor) 
## Client Operating System
Windows 10

## JAVA/JVM version
11.0.13+8

## Problem description
if using Connections with encrypt=true it ist not possible to break  an running Call () to the Database with the jdbc Connection Method 
This is the stack trace of the hanging connection, which cannot be aborted using the abort(executor) method from the other thread
![grafik](https://github.com/microsoft/mssql-jdbc/assets/139137701/ae19730c-92f3-402f-b058-01053c6e0e52)
In my test I using die command "waitfor DELAY '00:10:00' to simulate a bad long running SQL. 

The Problem is that with enabled ssl encryption the method "final void close()" in the class TDSChannel is calling disableSSL():
```
        if (null != sslSocket)
            disableSSL();
```
and this Method is using the same java lock witch is using in the method ProxyInputStream.readInternal(...).
## Expected behavior
The corresponding socket is closed and the hanging thread aborts with a SQLException

## Actual behavior
The on DB call hanging Thread can not abort.

barryw-mssql:
Thanks for the ticket.

This is an interesting issue. I am looking into the problem and will try a replicate the issue for further investigation.

lilgreenbird:
hi @FischerRo2 

Can you please share any repro code that would reproduce the issue? What is the connection string you are using?

Also, just to clarify, are you only seeing this issue when `encrypt=true`? i.e. does it work with `encrypt=false`?

FischerRo2:
hi @lilgreenbird 
yes i only see the issue with encrypt=true or without the property, respectively.

With encrypt=false, correctly the db connection socket is closed and the running thread waiting for the database connection is interrupted with an exception. 
Simple Java test case: 
 [AbortTest.txt](https://github.com/microsoft/mssql-jdbc/files/13867094/AbortTest.txt)
```java
package jdbc;

[...]

Issue #2265, opened 2023-11-27
DeadLock in JDBC driver close() vs object finalizer


karel-haruda-abra:
## Driver version
12.4.1.jre11

## SQL Server version
Windows Server 2019 Standard (10.0)

## Client Operating System
Windows Server 2019 Standard (10.0)

## JAVA/JVM version
JAVA 11

## Problem description
A deadlock occurs in the JDBC driver

## Error message/stack trace
I have only call stacks of threads which are in deadlock

**Thread 1** 
ID16 ABRABI - Database connection checker	WAITING	java.util.concurrent.locks.ReentrantLock$NonfairSync@444316f1	Finalizer	ID3	0d 00:20:59,953
	
java.base@11.0.20.1/jdk.internal.misc.Unsafe.park(Native Method)
java.base@11.0.20.1/java.util.concurrent.locks.LockSupport.park(LockSupport.java:194)
java.base@11.0.20.1/java.util.concurrent.locks.AbstractQueuedSynchronizer.parkAndCheckInterrupt(AbstractQueuedSynchronizer.java:885)
java.base@11.0.20.1/java.util.concurrent.locks.AbstractQueuedSynchronizer.acquireQueued(AbstractQueuedSynchronizer.java:917)
java.base@11.0.20.1/java.util.concurrent.locks.AbstractQueuedSynchronizer.acquire(AbstractQueuedSynchronizer.java:1240)
java.base@11.0.20.1/java.util.concurrent.locks.ReentrantLock.lock(ReentrantLock.java:267)
java.base@11.0.20.1/sun.security.ssl.SSLSocketImpl.waitForClose(SSLSocketImpl.java:1817)
java.base@11.0.20.1/sun.security.ssl.SSLSocketImpl.closeSocket(SSLSocketImpl.java:1795)
java.base@11.0.20.1/sun.security.ssl.SSLSocketImpl.shutdown(SSLSocketImpl.java:1750)
java.base@11.0.20.1/sun.security.ssl.SSLSocketImpl.bruteForceCloseInput(SSLSocketImpl.java:800)
java.base@11.0.20.1/sun.security.ssl.SSLSocketImpl.duplexCloseInput(SSLSocketImpl.java:780)
java.base@11.0.20.1/sun.security.ssl.SSLSocketImpl.close(SSLSocketImpl.java:590)
com.microsoft.sqlserver.jdbc.TDSChannel.disableSSL(IOBuffer.java:817)
[...]

PR #2295, opened 2024-01-11
Fix lock issues around disableSSL and IOBuffer

Jeffery-Wasty:
When migrating from `synchronized` keyword to locks, any method that used `synchronized(this)` was then locked using the same lock declared at the top of TDSChannel. This is not correct as most of the `synchronzed(this)` methods were also in an inner class, but disableSSL was not. Therefore, disableSSL should have a separate lock. This resolves the deadlock issues found in #2265 and #2288.


src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -672,7 +672,7 @@ final boolean isLoggingPackets() {
     int numMsgsSent = 0;
     int numMsgsRcvd = 0;
 
-    private final transient Lock lock = new ReentrantLock();
+    private final transient Lock TDSChannelLock = new ReentrantLock();
```
David-Engel, line 675: Minor/Optional suggestion:
Aren't local variables supposed to start with a lowercase letter?
```suggestion
    private final transient Lock tdsChannelLock = new ReentrantLock();
```

David-Engel (dismissed):
I scanned the previous loom PR again and this looks like the complete fix to me. The only other suggestion I might make is to rename the `lock` variable introduced in the original PR in IOBuffer's TDSReader class to `tdsReaderLock` to make its scope clear for future developers.

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
Jeffery-Wasty: I thought since TDS was an acronym, not a word, we preserve the case for clarity. But looking over again, this is done nowhere else, so yes, it should be lowercase.

lilgreenbird (approved):

David-Engel (approved):

barryw-mssql (approved):

tkyc (approved):

Issue #2323, opened 2024-02-07
mssql-jdbc:12.6.0.jre11 - 'sp_getapplock' expects parameter '@LockMode'


driverpt:
## Driver version
Provide the JDBC driver version (e.g. 12.6.0.jre11).

## SQL Server version
2017

## Client Operating System
Docker

## JAVA/JVM version
17

## Table schema
N/A

## Problem description
We're currently using Spring Boot w/ Flyway Migration (You can use TestContainers to debug the issue). We recently upgraded to 12.6.0-jre11 Driver and this caused Flyway to break.

The Query that Flyway uses seems to be fine, but probably something changed in the Driver Library that broke it.

Please check the link to the source code of the query.

## Expected behavior
Flyway to execute

## Actual behavior
Check Stack Trace

## Error message/stack trace
```
Caused by: com.microsoft.sqlserver.jdbc.SQLServerException: Procedure or function 'sp_getapplock' expects parameter '@LockMode', which was not supplied.
at com.microsoft.sqlserver.jdbc.SQLServerException.makeFromDatabaseError(SQLServerException.java:261)
at com.microsoft.sqlserver.jdbc.SQLServerStatement.getNextResult(SQLServerStatement.java:1752)
at com.microsoft.sqlserver.jdbc.SQLServerPreparedStatement.doExecutePreparedStatement(SQLServerPreparedStatement.java:657)
at com.microsoft.sqlserver.jdbc.SQLServerPreparedStatement$PrepStmtExecCmd.doExecute(SQLServerPreparedStatement.java:576)
at com.microsoft.sqlserver.jdbc.TDSCommand.execute(IOBuffer.java:7739)
at com.microsoft.sqlserver.jdbc.SQLServerConnection.executeCommand(SQLServerConnection.java:4384)
at com.microsoft.sqlserver.jdbc.SQLServerStatement.executeCommand(SQLServerStatement.java:293)
at com.microsoft.sqlserver.jdbc.SQLServerStatement.executeStatement(SQLServerStatement.java:263)
at com.microsoft.sqlserver.jdbc.SQLServerPreparedStatement.execute(SQLServerPreparedStatement.java:553)
at com.zaxxer.hikari.pool.ProxyPreparedStatement.execute(ProxyPreparedStatement.java:44)
at com.zaxxer.hikari.pool.HikariProxyPreparedStatement.execute(HikariProxyPreparedStatement.java)
[...]

PR #2325, opened 2024-02-08
Re-added support for stored procedure 'exec' escape syntax in CallableStatements

tkyc:
Prior to 12.6.0, all cstmts calls were wrapped int T-SQL sp_executesql, sp_prepexec etc.. calls before being sent off to the server. Doing so means the raw/unprocessed user's sql was passed into these T-SQL calls as parameter values and so the driver was able to execute "EXEC" calls against the server eg. `EXEC sp_columns_100 ?,?,?,?....`.

For example:
```
exec sp_executesql N'EXEC sp_getapplock @Resource=@P0, @LockTimeout='3600000', @LockMode='Exclusive', @LockOwner='Session',N'@P0 nvarchar(4000)',N'testParamValue'
```

This PR enables this prior behaviour again. 

https://github.com/microsoft/mssql-jdbc/issues/2323

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
@@ -29,6 +29,7 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Vector;
 import java.util.logging.Level;
+import java.util.regex.Pattern;
 
 import com.microsoft.sqlserver.jdbc.SQLServerConnection.CityHash128Key;
 import com.microsoft.sqlserver.jdbc.SQLServerConnection.PreparedStatementHandle;
@@ -70,7 +71,9 @@ public class SQLServerPreparedStatement extends SQLServerStatement implements IS
     /** Processed SQL statement text, may not be same as what user initially passed. */
     final String userSQL;
 
-    private boolean isExecCommand;
+    private boolean isExecEscapeSyntax;
+
+    private boolean isCallEscapeSyntax;
 
     /** Parameter positions in processed SQL statement text. */
     final int[] userSQLParamPositions;
@@ -255,7 +258,8 @@ public class SQLServerPreparedStatement extends SQLServerStatement implements IS
         procedureName = parsedSQL.procedureName;
         bReturnValueSyntax = parsedSQL.bReturnValueSyntax;
         userSQL = parsedSQL.processedSQL;
-        isExecCommand = isExecCommand(sql);
+        isExecEscapeSyntax = isExecEscapeSyntax(sql);
+        isCallEscapeSyntax = isCallEscapeSyntax(sql);
         userSQLParamPositions = parsedSQL.parameterPositions;
         initParams(userSQLParamPositions.length);
         useBulkCopyForBatchInsert = conn.getUseBulkCopyForBatchInsert();
@@ -1217,9 +1221,11 @@ public class SQLServerPreparedStatement extends SQLServerStatement implements IS
         // In order to execute sprocs directly the following must be true:
         // 1. There must be a sproc name
         // 2. There must be parameters
-        // 3. Parameters must not be a TVP typ
-        // Note: isExecCommand check is a workaround to allow prior behaviour of doing EXEC calls
-        return (null != procedureName && paramCount != 0 && !isTVPType(params) && !isExecCommand);
+        // 3. Parameters must not be a TVP type
+        // 4. Compliant CALL escape syntax
+        // If isExecEscapeSyntax is true, EXEC escape syntax is used then use prior behaviour to
+        // execute the procedure
+        return (null != procedureName && paramCount != 0 && !isTVPType(params) && isCallEscapeSyntax && !isExecEscapeSyntax);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -1230,6 +1234,11 @@ private boolean isTVPType(Parameter[] params) throws SQLServerException {
         return false;
     }
 
+    private boolean isExecCommand() {
+        String command = userSQLUnprocessed.split(" ")[0].toLowerCase();
+        return command.equals("exec") || command.equals("execute");
```
lilgreenbird, line 1239: just use equalsIgnoreCase?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -253,6 +256,7 @@ private boolean resetPrepStmtHandle(boolean discardCurrentCacheItem) {
         procedureName = parsedSQL.procedureName;
         bReturnValueSyntax = parsedSQL.bReturnValueSyntax;
         userSQL = parsedSQL.processedSQL;
+        userSQLUnprocessed = sql;
```
lilgreenbird, line 259: if we do the check here then we won't need to keep a copy of the sql command

tkyc: I was thinking for the future where we want to do it the "jTDS" way, we'd need to track the unprocessed sql for the SQL Batch TDS message. But I'll switch to boolean for now.

src/test/java/com/microsoft/sqlserver/jdbc/callablestatement/CallableStatementTest.java
```diff
@@ -1067,9 +1067,83 @@ public void testRegisteringOutputByIndexandAcquiringOutputParamByName() throws S
...
+        }
+    }
+
+    @Test
+    public void testExecSystemStoredProcedureNoIndexedParametersResultSet() throws SQLException {
+        String call = "execute sp_sproc_columns_100 sp_getapplock, @ODBCVer=3, @fUsePattern=0";
+
+        try (CallableStatement cstmt = connection.prepareCall(call)) {
+
+            try (ResultSet rs = cstmt.executeQuery()) {
```
lilgreenbird, line 1101: nitpick, can combine the 2 trys

src/test/java/com/microsoft/sqlserver/jdbc/callablestatement/CallableStatementTest.java
```diff
@@ -1067,9 +1067,83 @@ public void testRegisteringOutputByIndexandAcquiringOutputParamByName() throws S
...
+
+        int paramIndex = 0;
+
+        for (String sproc : sprocs) {
+            try (CallableStatement cstmt = connection.prepareCall(sproc)) {
+                cstmt.setObject(1, params[paramIndex]);
+                cstmt.execute();
+                paramIndex++;
+            } catch (Exception e) {
+                fail("Failed executing '" + sproc + "' with indexed parameter '" + params[paramIndex]);
```
lilgreenbird, line 1139: print the exception msg for easier debugging if this fails

src/test/java/com/microsoft/sqlserver/jdbc/callablestatement/CallableStatementTest.java
```diff
@@ -1067,9 +1067,83 @@ public void testRegisteringOutputByIndexandAcquiringOutputParamByName() throws S
         }
     }
 
+    @Test
+    public void testExecuteSystemStoredProcedureNamedParametersAndIndexedParameterNoResultset() throws SQLException {
+        String call = "EXEC sp_getapplock @Resource=?, @LockTimeout='60', @LockMode='Exclusive', @LockOwner='Session'";
```
lilgreenbird, line 1072: can we choose a shorter timeout value? in case of problems then this won't have to wait too long

src/test/java/com/microsoft/sqlserver/jdbc/callablestatement/CallableStatementTest.java
tkyc: It will already print.

lilgreenbird (approved):

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -1230,6 +1239,11 @@ private boolean isTVPType(Parameter[] params) throws SQLServerException {
         return false;
     }
 
+    private boolean isExecCommand(String sql) {
+        String command = sql.split(" ")[0];
```
David-Engel, line 1243: Not sure if this will be comprehensive. What about tabs instead of spaces? Leading spaces? Carriage returns?

I would also consider turning this around the create a check to ensure the statement adheres to the format in the JDBC spec or at least one we know will work in an RPC call.

tkyc: > What about tabs instead of spaces? Leading spaces? Carriage returns?

I was thinking of failing these cases. So, the query string must start as "EXEC ...".

> I would also consider turning this around the create a check to ensure the statement adheres to the format in the JDBC spec or at least one we know will work in an RPC call.

I'll add that change in.

tkyc: Right nvm, since spaces, tabs and carriage returns worked in the past, I'll correct it so that anything that worked in the past will work the same now.

lilgreenbird (commented):
change the title to something more descriptive instead of just regression fix

David-Engel (approved):

lilgreenbird (approved):

Jeffery-Wasty (approved):

PR #2326, opened 2024-02-09
Fix to ensure metadata returned follows JDBC data type specs

barryw-mssql:
ResultSetMetadata of the ResultSet from DatabaseMetadata.getColumns() is Incorrect for Some Columns (1751)

Fix ensures the correct data types for each metadata column by mapping the data type from the SQL Server to the type specified in the JDBC spec


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
```diff
@@ -721,6 +730,48 @@ public java.sql.ResultSet getColumns(String catalog, String schema, String table
...
+                     getTypesDWColumns.put(23, NVARCHAR);       // SS_UDT_CATALOG_NAME
+                     getTypesDWColumns.put(24, NVARCHAR);       // SS_UDT_SCHEMA_NAME
+                     getTypesDWColumns.put(25, NVARCHAR);       // SS_UDT_ASSEMBLY_TYPE_NAME
+                     getTypesDWColumns.put(26, NVARCHAR);       // SS_XML_SCHEMACOLLECTION_CATALOG_NAME
+                     getTypesDWColumns.put(27, NVARCHAR);       // SS_XML_SCHEMACOLLECTION_SCHEMA_NAME
+                     getTypesDWColumns.put(28, NVARCHAR);       // SS_XML_SCHEMACOLLECTION_NAME
+                }
+                
+                // Ensure there is a data type for every metadata column
+                assert(getColumnsDWColumns.size() == getTypesDWColumns.size());
```
lilgreenbird, line 774: don't assert, handle the error properly

src/test/java/com/microsoft/sqlserver/jdbc/databasemetadata/DatabaseMetaDataTest.java
```diff
@@ -887,6 +920,57 @@ public void testGetImportedKeysDW() throws SQLException {
         }
     }
 
+    @Test
+    public void testValidateColumnMetadata() throws SQLException {
```
lilgreenbird, line 924: I don't think this test actually test the fix, it just tests that all the columns are there and in the order expected, which would never fail because that's the order we put them in. What the issue that the classes weren't cast to the correct types so they're not assignable. Take a look at the repro in the origin issue it shows what the problem is and what we should be checking for

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
```diff
@@ -257,12 +257,18 @@ private void checkClosed() throws SQLServerException {
     private static final String IS_GENERATEDCOLUMN = "IS_GENERATEDCOLUMN";
     private static final String IS_AUTOINCREMENT = "IS_AUTOINCREMENT";
     private static final String ACTIVITY_ID = " ActivityId: ";
+    
+    private static final String NVARCHAR = "NVARCHAR";
+    private static final String VARCHAR = "VARCHAR";
+    private static final String INTEGER = "INT";
+    private static final String SMALLINT = "SMALLINT";
```
lilgreenbird, line 264: instead of adding these extra string defns, why not just use the actual sql types?

src/test/java/com/microsoft/sqlserver/jdbc/databasemetadata/DatabaseMetaDataTest.java
```diff
@@ -887,6 +920,84 @@ public void testGetImportedKeysDW() throws SQLException {
...
+                                TestResource.getResource("R_classLoaderNotFoundForColumnType"));
+                    Object[] msgArgs = {columnClassName, columnLabel};
+                    fail(form.format(msgArgs));
+                }
+                
+                Class<?> expectedClass = getColumnMetaDataClass.get(columnLabel);
+                                
+                assert(expectedClass != null);
+    
+                if (!expectedClass.isAssignableFrom(columnClass)) {
```
Jeffery-Wasty, line 989: Remove assertion
```suggestion
                if (null != expectedClass && !expectedClass.isAssignableFrom(columnClass)) {
```

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
```diff
@@ -25,6 +25,7 @@
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.logging.Level;
+import com.microsoft.sqlserver.jdbc.JDBCType;
```
Jeffery-Wasty, line 28: Unused.
```suggestion
```

src/test/java/com/microsoft/sqlserver/jdbc/databasemetadata/DatabaseMetaDataTest.java
```diff
@@ -887,6 +920,84 @@ public void testGetImportedKeysDW() throws SQLException {
...
+                {
+                    MessageFormat form = new MessageFormat(
+                                TestResource.getResource("R_classLoaderNotFoundForColumnType"));
+                    Object[] msgArgs = {columnClassName, columnLabel};
+                    fail(form.format(msgArgs));
+                }
+                
+                Class<?> expectedClass = getColumnMetaDataClass.get(columnLabel);
+                                
+                assert(expectedClass != null);
```
lilgreenbird, line 987: don't assert

lilgreenbird, line 987: add null check instead of just assert

src/test/java/com/microsoft/sqlserver/jdbc/databasemetadata/DatabaseMetaDataTest.java
```diff
@@ -887,6 +920,84 @@ public void testGetImportedKeysDW() throws SQLException {
...
+                {
+                    columnClass = Class.forName(columnClassName);
+                }
+                catch (ClassNotFoundException ex)
+                {
+                    MessageFormat form = new MessageFormat(
+                                TestResource.getResource("R_classLoaderNotFoundForColumnType"));
+                    Object[] msgArgs = {columnClassName, columnLabel};
+                    fail(form.format(msgArgs));
+                }
```
lilgreenbird, line 983: I think we just need to assertTrue that the expectedClass.getName().equals(columnClassName) as per repro in the issue

src/test/java/com/microsoft/sqlserver/jdbc/databasemetadata/DatabaseMetaDataTest.java
lilgreenbird, line 989: same here, just assertTrue that expectedCalss.isAssignableFrom(columnClass)

lilgreenbird (changes requested):
pls run formatter

lilgreenbird (changes requested):
there should be a test case to cover the error msg R_colCountNotMatchColTypeCount path

src/test/java/com/microsoft/sqlserver/jdbc/databasemetadata/DatabaseMetaDataTest.java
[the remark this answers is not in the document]
lilgreenbird: just run formatter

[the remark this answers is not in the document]
Jeffery-Wasty: ^ this

src/test/java/com/microsoft/sqlserver/jdbc/databasemetadata/DatabaseMetaDataTest.java
```diff
@@ -887,6 +920,67 @@ public void testGetImportedKeysDW() throws SQLException {
...
+            // Ensure that there is an expected class for every column in the metadata result set
+            assertEquals(metadata.getColumnCount(), getColumnMetaDataClass.size());
+
+            for (int i = 1; i < metadata.getColumnCount(); i++) {
+                String columnLabel = metadata.getColumnLabel(i);
+                String columnClassName = metadata.getColumnClassName(i);
+                Class<?> expectedClass = getColumnMetaDataClass.get(columnLabel);
+
+                // Ensure the metadata column is in the metadata column class map
+                assertNotNull(expectedClass);
```
Jeffery-Wasty, line 976: This was mentioned in a previous comment above, but ensure that `expectedClass` is not null, don't assert.

Jeffery-Wasty:
```suggestion
                if (null != expectedClass) {
```

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
lilgreenbird: ^ this is not addressed. This import is not necessary

src/test/java/com/microsoft/sqlserver/jdbc/databasemetadata/DatabaseMetaDataTest.java
lilgreenbird: umm, should this fail? is null a failure scenario? if we did the null check then it would just skip

src/test/java/com/microsoft/sqlserver/jdbc/databasemetadata/DatabaseMetaDataTest.java
```diff
@@ -63,6 +63,39 @@ public class DatabaseMetaDataTest extends AbstractTest {
     private static final String functionName = RandomUtil.getIdentifier("DBMetadataFunction");
     private static Map<Integer, String> getColumnsDWColumns = null;
     private static Map<Integer, String> getImportedKeysDWColumns = null;
+    private static Map<String, Class<?>> getColumnMetaDataClass = null;
```
lilgreenbird, line 66: this is only ever used in the 1 new test, doesn't need to be in the class scope

src/test/java/com/microsoft/sqlserver/jdbc/databasemetadata/DatabaseMetaDataTest.java
lilgreenbird: did we accidentally remove the check for isAssignableFrom that was in previous commit?

src/test/java/com/microsoft/sqlserver/jdbc/TestResource.java
```diff
@@ -212,5 +212,8 @@ protected Object[][] getContents() {
             {"R_failedFedauth", "Failed to acquire fedauth token: "},
             {"R_noLoginModulesConfiguredForJdbcDriver",
                     "javax.security.auth.login.LoginException (No LoginModules configured for SQLJDBCDriver)"},
-            {"R_unexpectedThreadCount", "Thread count is higher than expected."}};
+            {"R_unexpectedThreadCount", "Thread count is higher than expected."},
+            {"R_classLoaderNotFoundForColumnType", "Class Loader for type {0} not found for column {1}"},
+            {"R_classNotAssignable", "Class {0} is not assignable from class {1} for column {2}"},
+            {"R_expectedClassDoesNotMatchActualClass", "Expected column class {0} does not match actual column class {1} for column {2}."}};
```
lilgreenbird, line 218: these are not used

Jeffery-Wasty: Specifically `R_classLoaderNotFoundForColumnType` and `R_classNotAssignable`.

src/test/java/com/microsoft/sqlserver/jdbc/databasemetadata/DatabaseMetaDataTest.java
```diff
@@ -887,6 +920,72 @@ public void testGetImportedKeysDW() throws SQLException {
         }
     }
 
+    @Test
+    // Validates the metadata data types defined by JDBC spec
+    // https://docs.oracle.com/javase/8/docs/api/java/sql/DatabaseMetaData.html#getColumns-java.lang.String-java.lang.String-java.lang.String-java.lang.String-
+    public void testValidateColumnMetadata() throws SQLException {
```
Jeffery-Wasty, line 926: Style. Please use JavaDoc comments and @Test should be after comment/before header.

src/test/java/com/microsoft/sqlserver/jdbc/databasemetadata/DatabaseMetaDataTest.java
```diff
@@ -887,6 +920,72 @@ public void testGetImportedKeysDW() throws SQLException {
...
+            for (int i = 1; i < metadata.getColumnCount(); i++) {
+                String columnLabel = metadata.getColumnLabel(i);
+                String columnClassName = metadata.getColumnClassName(i);
+                Class<?> expectedClass = getColumnMetaDataClass.get(columnLabel);
+
+                // Ensure the metadata column is in the metadata column class map
+                assertNotNull(expectedClass);
+
+                // Ensure the actual and expected column metadata types match
+                if (columnClassName != expectedClass.getName()) {
```
Jeffery-Wasty, line 979: Please use .equals() instead of == or !=
```suggestion
                if (!columnClassName.equals(expectedClass.getName())) {
```

barryw-mssql: Completed

src/test/java/com/microsoft/sqlserver/jdbc/databasemetadata/DatabaseMetaDataTest.java
barryw-mssql: Changed the logic for checking the returned column data type matches the expected data type to simplify the logic. The new logic removes this Class Loader test. There is a new tests that actually compares the actual and expected class name for each column but does not use an assert. Instead a R_expectedClassDoesNotMatchActualClass message is used which includes the actual and expected data types as well as the column. This was done because if just an assert was used it would not show which column was generating the issue in the metadata result set.

src/test/java/com/microsoft/sqlserver/jdbc/databasemetadata/DatabaseMetaDataTest.java
barryw-mssql: This assert has been removed and replaced with a null check and will use AssertionUtils.fail() with a descriptive error message

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
barryw-mssql: Import has been removed

src/test/java/com/microsoft/sqlserver/jdbc/databasemetadata/DatabaseMetaDataTest.java
barryw-mssql: Logic has been modified to check the JDBC class types in the metadata result set, instead of checking the SQL Server data types for the result set. The logic was greatly simplified by using in Integer class instead of a generic Object class as the expected class for the BUFFER_LENGTH column (i.e. which is not actually used in the JDBC spec).

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
barryw-mssql: Done

src/test/java/com/microsoft/sqlserver/jdbc/databasemetadata/DatabaseMetaDataTest.java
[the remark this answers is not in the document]
barryw-mssql: Done

src/test/java/com/microsoft/sqlserver/jdbc/databasemetadata/DatabaseMetaDataTest.java
barryw-mssql: Logic is no longer used so this request is moot

src/test/java/com/microsoft/sqlserver/jdbc/databasemetadata/DatabaseMetaDataTest.java
```diff
@@ -887,6 +920,84 @@ public void testGetImportedKeysDW() throws SQLException {
...
+    
+                if (!expectedClass.isAssignableFrom(columnClass)) {
+                    MessageFormat form = new MessageFormat(
+                                TestResource.getResource("R_classNotAssignable"));
+                    Object[] msgArgs = {expectedClass.getName(), columnClassName, columnLabel};
+                    fail(form.format(msgArgs));
+                }
+            }
+        } catch (SQLException e) {
+            fail(e.getMessage());
```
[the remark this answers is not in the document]
barryw-mssql: Removed the catch as most other tests in the file do not catch

src/test/java/com/microsoft/sqlserver/jdbc/databasemetadata/DatabaseMetaDataTest.java
```diff
@@ -887,6 +920,84 @@ public void testGetImportedKeysDW() throws SQLException {
         }
     }
 
+    @Test
+    // Validates the metadata data types defined by JDBC spec
+    // https://docs.oracle.com/javase/8/docs/api/java/sql/DatabaseMetaData.html#getColumns-java.lang.String-java.lang.String-java.lang.String-java.lang.String-
+    public void testValidateColumnMetadata() throws SQLException {
```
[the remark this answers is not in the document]
barryw-mssql: Removed the method catch so the throws SQLException statement is now useful for the method

src/test/java/com/microsoft/sqlserver/jdbc/databasemetadata/DatabaseMetaDataTest.java
barryw-mssql: The assignable class check is no longer used.

src/test/java/com/microsoft/sqlserver/jdbc/databasemetadata/DatabaseMetaDataTest.java
barryw-mssql: The expected class can be null if the the column label/name in the expected class lookup table does not match the actual column label from the result set (i.e. the lookup by column name will fail). This shouldn't happen but it is possible if there was a future code change or perhaps the returned metadata columns change.

Yes the IsAssignableFrom() is no longer used as I changed the BUFFER_LENGTH metadata class from Object to Integer. It was because of this generic Object class that we needed to use the IsAssignableFrom() check. Is The reason why it was originally specified as a generic Object was that the JDBC spec does not use the BUFFER_LENGTH column and so I reflected that as a generic Object being the expected data type. But in the code I have always ensured that BUFFER_LENGTH was an Integer. Therefore to simplify the logic and remove the IsAssignableFrom() check, I switched the expected class for the BUFFER_LENGTH column from Object to Integer.

src/test/java/com/microsoft/sqlserver/jdbc/databasemetadata/DatabaseMetaDataTest.java
barryw-mssql: Done

src/test/java/com/microsoft/sqlserver/jdbc/TestResource.java
barryw-mssql: Removed

src/test/java/com/microsoft/sqlserver/jdbc/databasemetadata/DatabaseMetaDataTest.java
barryw-mssql: Done

src/test/java/com/microsoft/sqlserver/jdbc/databasemetadata/DatabaseMetaDataTest.java
barryw-mssql: Done

tkyc (approved):

Jeffery-Wasty (approved):

David-Engel (approved):

PR #2330, opened 2024-02-16
Use Re-entrant locks instead of synchronized blocks to make driver loom friendly

mthakur77:
Loom project: https://openjdk.java.net/projects/loom/

Using synchronized keyword pins the virtual thread on the platform thread degrading the performance. Using ReentrantLock should prevent it from happening.

Sample code:
```
void foo() {
  synchronized(this) {
     ... // performs IO or something blocking
  }
}
```
With code like:
```
// loom friendly ...

private final Lock lock = new ReentrantLock();

void foo() {
  lock.lock();
  try {
     ... // performs IO or something blocking
  } finally {
     lock.unlock();
  }
}
```

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
@@ -1440,8 +1440,13 @@ final class TDSChannel implements Serializable {
         }
 
         @Override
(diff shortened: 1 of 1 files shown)


lilgreenbird:
@mthakur77 thanks for bringing this to our attention, this was incorrectly added before I've removed this in PR #2337 so this should no longer be necessary. I've also checked there are no more **synchronized**  keywords in the driver after this.

PR #2339, opened 2024-02-25
Allow constructing a `microsoft.sql.DateTimeOffset` instance from a `java.time.OffsetDateTime` value

arashi01:
Operates similarly to existing constructor however avoids additional expense of creating Timestamp instances, and increased timezone related error potential, by directly converting java.time.OffsetDateTime instances.

Reverse of existing `DateTimeOffset.getOffsetDateTime()` method.

Rounds precision to closest 100ns matching existing semantics of `Timestamp` based constructor.


arashi01:

arashi01:
@lilgreenbird Caught a bug caused by missing upwards rounding of second values if nanos are within 50ns of the next second. Corrected it and added a unit test. Behaviour now fits with existing semantics.

lilgreenbird:
@arashi01 really sorry for this but could you please open a new PR for this? your PR came in while we were doing some updates to the pipelines and for some reason it's stuck in an old version.

arashi01:
@lilgreenbird No worries! Done.

PR #2340, opened 2024-02-26
Allow constructing a `microsoft.sql.DateTimeOffset` instance from a `java.time.OffsetDateTime` value

arashi01:
Operates similarly to existing constructor however avoids additional expense of creating Timestamp instances and increased timezone related error potential, by directly converting java.time.OffsetDateTime instances.

Reverse of existing `DateTimeOffset.getOffsetDateTime()` method.

Rounds precision to closest 100ns matching existing semantics of `Timestamp` based constructor. Values where nanos are within 50ns of the next second are rounded upwards to the next second; as per current semantics.

Supercedes #2339

### DIFF
--- a/src/main/java/microsoft/sql/DateTimeOffset.java
+++ b/src/main/java/microsoft/sql/DateTimeOffset.java
@@ -73,6 +73,22 @@ public final class DateTimeOffset implements java.io.Serializable, java.lang.Com
     }
 
     /**
+     * Constructs a DateTimeOffset from an existing java.time.OffsetDateTime
+     *
+     * @param offsetDateTime
+     *        A java.time.OffsetDateTime value
+     * @apiNote DateTimeOffset represents values to 100 nanosecond precision. If the java.time.OffsetDateTime instance
+     * represents a value that is more precise,  the value is rounded to the nearest multiple of 100 nanoseconds. Values
+     * within 50 nanoseconds of the next second are rounded up to the next second.
+     */
+    private DateTimeOffset(java.time.OffsetDateTime offsetDateTime) {
+        int hundredNanos = ((offsetDateTime.getNano() + 50) / 100);
+        this.utcMillis = (offsetDateTime.toEpochSecond() * 1000) + (hundredNanos / HUNDRED_NANOS_PER_SECOND * 1000);
+        this.nanos = 100 * (hundredNanos % HUNDRED_NANOS_PER_SECOND);
+        this.minutesOffset = offsetDateTime.getOffset().getTotalSeconds() / 60;
+    }
+
+    /**
      * Converts a java.sql.Timestamp value with an integer offset to the equivalent DateTimeOffset value
      * 
      * @param timestamp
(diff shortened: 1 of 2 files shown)


tkyc:
FYI, I did a force push to fix the pipeline issue.

arashi01:
Also just noticed the second unit test (`valueOfOffsetDateTimeRounding()`) can cause a false test failure if the test is run at any time with a 59 seconds value, since it uses OffsetDateTime.now() for it's seed data.

Will fix it now. Rather than give the test value a wholly static value I think it's probably better to just fix the "seconds" value, similar to the "nanos" value.

arashi01:
Fixed the test. Fixed the test value to `OffsetDateTime.now().withSecond(58)`.

arashi01:
@tkyc Just a heads up. The current HEAD doesn't include the updated test. See https://github.com/microsoft/mssql-jdbc/compare/ae407c4ff5be6062fc63536c90edb532ffb45a0b..859c32db2a2d9822595d9d8b73849ae2962270b7

tkyc:
@arashi01 

My mistake, could you push again? I had to do another force push which overwrote your last commit.

lilgreenbird:
the reason the force-push was done by @tkyc was to get rid of the previous commits which was tied to the old versions of the pipelines for some reason and we wanted to remove CI-mssql-jdbc pipeline from the PR validation check. But now it seems with your latest force-push this came right back :( 
Well I guess we can just ignore the failures on this particular pipeline.

src/test/java/microsoft/sql/DateTimeOffsetTest.java
```diff
@@ -0,0 +1,54 @@
+/*
+ * Microsoft JDBC Driver for SQL Server Copyright(c) Microsoft Corporation All rights reserved. This program is made
```
lilgreenbird, line 2: driver tests are in the src/test/java/com/microsoft/sqlserver/jdbc folder.

also, this 1 test doesn't really need a separate file this can just be added to [DataTypeTests.java](https://github.com/microsoft/mssql-jdbc/blob/main/src/test/java/com/microsoft/sqlserver/jdbc/datatypes/DataTypesTest.java)

src/test/java/microsoft/sql/DateTimeOffsetTest.java
```diff
@@ -0,0 +1,54 @@
+/*
+ * Microsoft JDBC Driver for SQL Server Copyright(c) Microsoft Corporation All rights reserved. This program is made
+ * available under the terms of the MIT License. See the LICENSE file in the project root for more information.
+ */
+
+package microsoft.sql;
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.DisplayName;
```
lilgreenbird, line 9: nice to have, but we don't do this in the other tests so to keep things consistent, pls remove

lilgreenbird (changes requested):
pls use [formatter ](https://github.com/microsoft/mssql-jdbc/blob/main/mssql-jdbc_formatter.xml)to format all files

arashi01:
> pls use [formatter ](https://github.com/microsoft/mssql-jdbc/blob/main/mssql-jdbc_formatter.xml)to format all files

Done. Although used Intellij's Eclipse code style import support to use the mssql-jdbc_formatter.xml. If theres still an issue I'll install Eclipse and rerun.

lilgreenbird (approved):

Jeffery-Wasty (approved):

tkyc (approved):

barryw-mssql (commented):
Looks fine

Issue #2354, opened 2024-03-19
loginTimouts overwrites socketTimeout for SELECTs


jesudornenkrone:
## Driver version
12.6.0 (before 12.4.2)

## SQL Server version
Microsoft SQL Server 2016 (SP3-GDR)

## Client Operating System
Windows 11 23H2 and Windows Server 2012 R2

## JAVA/JVM version
Java 21 and Java 17

## Problem description
Using mssql-jdbc 12.4.2.jre11 everything was working as expected, but since we updated to 12.6.0.jre11 we get "java.net.SocketTimeoutException" with some long running SELECTs.

It is fairly easy to reproduce, by setting the socketTimeout to 90s and loginTimeout to 10s (i.e.: jdbc:sqlserver://192...;socketTimeout=90000;loginTimeout=10). Then we start a SELECT which takes the MSSQL Server more than 10s to return the first result.

Caused by: java.net.SocketTimeoutException: Read timed out
                at java.base/sun.nio.ch.NioSocketImpl.timedRead(NioSocketImpl.java:283)
                at java.base/sun.nio.ch.NioSocketImpl.implRead(NioSocketImpl.java:309)
                at java.base/sun.nio.ch.NioSocketImpl.read(NioSocketImpl.java:350)
                at java.base/sun.nio.ch.NioSocketImpl$1.read(NioSocketImpl.java:803)
                at java.base/java.net.Socket$SocketInputStream.read(Socket.java:976)
                at com.microsoft.sqlserver.jdbc.TDSChannel$ProxyInputStream.readInternal(IOBuffer.java:1206)
                at com.microsoft.sqlserver.jdbc.TDSChannel$ProxyInputStream.read(IOBuffer.java:1192)
                at com.microsoft.sqlserver.jdbc.TDSChannel$ProxyInputStream.readInternal(IOBuffer.java:1206)
                at com.microsoft.sqlserver.jdbc.TDSChannel$ProxyInputStream.read(IOBuffer.java:1192)

Could this be related to this change in version 12.6.0 "Changed socketTimeout to ensure it's always less than or equal to loginTimeout https://github.com/microsoft/mssql-jdbc/pull/2280"?

We normally use the default 30s loginTimeout, because we want the application to fail fast, if DB is not available. But we have some slow SELECTs fo which we require a much higher socketTimeout.

Thank you.

tkyc:
[...]

PR #2355, opened 2024-03-19
socketTimeout should reset to original value after a successful connection open

tkyc:
https://github.com/microsoft/mssql-jdbc/issues/2354

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
@@ -630,7 +630,7 @@ final class TDSChannel implements Serializable {
     }
 
     // Socket for raw TCP/IP communications with SQL Server
-    private transient Socket tcpSocket;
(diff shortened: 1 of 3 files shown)


barryw-mssql (approved):

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -630,7 +630,7 @@ final TDSReader getReader(TDSCommand command) {
     }
 
     // Socket for raw TCP/IP communications with SQL Server
-    private transient Socket tcpSocket;
+    transient Socket tcpSocket;
```
lilgreenbird, line 633: I think maybe better to add a method here so we don't have to change this

Jeffery-Wasty (dismissed):

Jeffery-Wasty (approved):

barryw-mssql (approved):

lilgreenbird (approved):

PR #2360, opened 2024-03-21
Enhancing Test Case Granularity for `testGetLocalDateTimeTypes`

Codegass:
### Description
As a newcomer to the mssql-jdbc community, I've been exploring the test suite and noticed an opportunity for minor improvement in the [`testGetLocalDateTimeTypes`](https://github.com/microsoft/mssql-jdbc/blob/9a8849b60e74d1d01508fa61f944d9e635cab700/src/test/java/com/microsoft/sqlserver/jdbc/datatypes/DataTypesTest.java#L1877-L1915) test case. Currently, this test case covers two distinct scenarios related to handling DATETIMEOFFSET conversions: one with the default behavior and another where the offset is ignored. I propose splitting this into two separate test cases to enhance test granularity and clarity.

Splitting the test allows each case to focus on a unique behavior, making it easier to pinpoint issues when one scenario fails without impacting the other. This approach not only aids in more precise problem identification across different scenarios but also reduces individual test execution time, which can be beneficial when debugging and resolving bugs.

### Key Points
* **Split testGetLocalDateTimeTypes into Two Separate Tests**: Create distinct test cases for the default conversion behavior and the behavior when ignoring time zone offsets.
* **Improve Test Clarity and Focus**: Each test will focus on a specific conversion scenario, enhancing the ability to diagnose and address issues in targeted areas of the codebase.
* **Reduce Single Test Execution Time**: By isolating the scenarios, each test's runtime is minimized, making the process of running tests more efficient during bug fixes and development.

I hope these changes will make our testing suite more robust and user-friendly. I'm open to discussion and feedback on this proposal, as I'm eager to learn from the community and contribute positively.

### DIFF
--- a/src/test/java/com/microsoft/sqlserver/jdbc/datatypes/DataTypesTest.java
+++ b/src/test/java/com/microsoft/sqlserver/jdbc/datatypes/DataTypesTest.java
@@ -1875,8 +1875,11 @@ public class DataTypesTest extends AbstractTest {
 
     @Test
     public void testGetLocalDateTimeTypesWithDefaultOffsetDateTimeConversion() throws Exception {
-        OffsetDateTime value = prepareOffsetDateTime();
-        LocalDateTime valueWithOffsetConversion = convertOffsetDateTimeToLocalDateTime(value);
+        OffsetDateTime value = OffsetDateTime.now().truncatedTo(ChronoUnit.MILLIS); // Linux has more precision than SQL Server
+        int offsetSeconds = value.getOffset().getTotalSeconds();
+        offsetSeconds += offsetSeconds < 0 ? 3600 : -3600;
+        value = value.withOffsetSameLocal(ZoneOffset.ofTotalSeconds(offsetSeconds));
+        LocalDateTime valueWithOffsetConversion = value.atZoneSameInstant(ZoneId.systemDefault()).toLocalDateTime();
 
         try (SQLServerConnection conn = PrepUtil.getConnection(connectionString)) {
             try (PreparedStatement stmt = conn.prepareStatement("SELECT ?")) {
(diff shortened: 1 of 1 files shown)


Codegass:

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/DataTypesTest.java
```diff
@@ -1946,6 +1944,17 @@ public void testDateTimeOffsetValueOfOffsetDateTime() throws Exception {
...
+        OffsetDateTime value = OffsetDateTime.now().truncatedTo(ChronoUnit.MILLIS);
+        int offsetSeconds = value.getOffset().getTotalSeconds();
+        offsetSeconds += offsetSeconds < 0 ? 3600 : -3600;
+        return value.withOffsetSameLocal(ZoneOffset.ofTotalSeconds(offsetSeconds));
+    }
+
+    private LocalDateTime convertOffsetDateTimeToLocalDateTime(OffsetDateTime value) {
+        return value.atZoneSameInstant(ZoneId.systemDefault()).toLocalDateTime();
+    }
+
```
lilgreenbird, line 1957: I'm not opposed to splitting the tests but refactoring into these methods seem unnecessary since they are only called once, why not just include them in the tests themselves?

Codegass: Thanks for the feedback! The decision to refactor was driven by enhancing test readability and making future maintenance easier. This approach will improve the clarity of our test cases and reduces debugging rerun time when debugging for specific scenario. When tests are more granular and focused, identifying and addressing issues becomes faster, leading to more efficient debugging cycles.

Even though the methods are called only once now, this structure allows us to easily expand or modify tests with minimal changes. The reduction in debugging rerun time and the enhancement in clarity are pivotal, as they ensure that our test suite remains robust, facilitating quicker modifications.

Jeffery-Wasty (dismissed):

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/DataTypesTest.java
lilgreenbird: I don't think it makes it any easier to read. If this changes in the future it can be changed then for now there is no need to do unnecessary refactoring and extra method calls

Codegass: Thanks for suggestion! Your perspective on the overall development picture is very helpful. I agree that maintainability should be balanced with immediate clarity. Let's keep the tests straightforward as you've advised.

Jeffery-Wasty (approved):

Jeffery-Wasty:
Hi @Codegass,

We're closing this because of reasons explained above by @lilgreenbird. Additionally, separating tests can lead to an increased test time, and this is not something we want. 

Please feel free to comment below if you have any additional questions or concerns about your pull request.

PR #2324, opened 2024-02-08
Add support for TDSType.GUID (#1582)

vxel:
This PR intends to provide support for sending UUID parameters as `TDSType.GUID (0x24)` on the wire and (hopefully) implements feature request #1582. 
It currently requires the Java parameter be given to the driver as `String` with `JDBCType.GUID`, i.e. :
```
st.setObject(index, myUUID.toString(), microsoft.sql.Types.GUID);
```

Issue #1582, opened 2021-05-18
[FEATURE REQUEST] Allow sending UUID in binary to database


knutwannheden:
## Is your feature request related to a problem? If so, please give a short summary of the problem and how the feature would resolve it
I am trying to execute SQL queries with `UUID` objects in Java mapping to `uniqueidentifier` values in the database. This works, but when passing a `UUID` value as an input to a query using `PreparedStatement#setObject()` the value ends up being transferred as a string over the wire. This is inefficient both computationally (`UUID#toString()`) and also transfers more data over the network (36 bytes vs. 16 bytes).

## Describe the preferred solution
When reading `uniqueidentifier` values from a `ResultSet` I can call `ResultSet#getObject(int, Class)` with the second parameter as `UUID.class`. This will give back the correct UUID value as expected without first converting the byte array to a String. Essentially I would want to have the same possibility when calling `PreparedStatement#setObject()` with a `UUID` object as input.

## Describe alternatives you've considered
The alternative (workaround) I now use to do this is to manually convert the `UUID` value to a `byte[]` value (essentially using the same logic as in `Util#asGuidByteArray(UUID)`) and then give that `byte[]` object to `PreparedStatement#setObject()`.

## Additional context
Looking at the code I found this relevant block: https://github.com/microsoft/mssql-jdbc/blob/1f550f13021b7110d073388725e2d79168614c3c/src/main/java/com/microsoft/sqlserver/jdbc/dtv.java#L1597-L1606

Apparently it looks like what I want is possible, but only when `cryptoMeta` is set. I don't understand how encryption is related to this.

v-zhangw:
Hi Knutwannheden,

This looks like a possible improvement to add in the future. It looks like this separation in flow was deliberately put in place, so there is risk of breaking something if we are not careful. Regardless it has been backlogged.

vxel:
Looking at the code, it seems that there is currently no way to pass a UUID as a `TDSType.GUID` (0x24) on the wire.
When you pass the parameter as `byte[]`, the TDS type seems to be `TDSType.BIGVARBINARY` (0x45).
When you pass it as `String`, the TDS type is `TDSType.NVARCHAR` (0xE7).

I may contribute a PR for this feature.

knutwannheden:
[...]

PR #2365, opened 2024-03-25
Revert "Add support for TDSType.GUID (#1582)"

tkyc:
Reverts microsoft/mssql-jdbc#2324


Jeffery-Wasty (approved):

tkyc:
@vxel 

After more testing with Always Encrypted. It looks like it's erroring out with:

```
The incoming tabular data stream (TDS) remote procedure call (RPC) protocol stream is incorrect. Parameter 5 (""): Data type 0x00 is unknown.
```

The test that failed was `testOutputProcedureCharInorder`. Sorry that this wasn't more extensively tested before the merge. We'll need to look more into this.

lilgreenbird (approved):

PR #2381, opened 2024-04-04
Add Assertion to the `SharedTimerTest`

### LINKED ISSUE #2357: Add Assertion to the `SharedTimer` Test

When I am walking through the unit test of the mssql-jdbc, I noticed that currently the [`SharedTimerTest`](https://github.com/microsoft/mssql-jdbc/blob/9a8849b60e74d1d01508fa61f944d9e635cab700/src/test/java/com/microsoft/sqlserver/jdbc/SharedTimerTest.java#L9-L27) does not contain any assertion statement to verify the behavior and state of the `SharedTimer` after execution of its `getTimer()` method in a multithreaded environment. The test ensures that references are added and removed, but does not assert the expected state of the SharedTimer (e.g., it is stopped and cleaned up properly after all references are removed). This could potentially mask issues with how `SharedTimer` manages its lifecycle and resources.

Current Implementation:
```java
@Test
    void getTimer() throws InterruptedException, ExecutionException, TimeoutException {
        final int iterations = 500;

        ExecutorService executor = Executors.newFixedThreadPool(2);
        try {
            ArrayList<CompletableFuture<?>> futures = new ArrayList<>(iterations);
            for (int i = 0; i < iterations; i++) {
                futures.add(CompletableFuture.runAsync(() -> SharedTimer.getTimer().removeRef(), executor));
            }

            CompletableFuture.allOf(futures.toArray(new CompletableFuture[0])).get(2, TimeUnit.MINUTES);
        } finally {
            executor.shutdown();
        }
    }
```

I'd like to propose a minor improvement by adding a simple assertion at the end of the getTimer test method to verify that the SharedTimer has indeed been stopped and its resources have been cleaned up after all references to it have been removed. This could be accomplished by asserting that SharedTimer.isRunning() returns false, indicating that the internal ScheduledThreadPoolExecutor has been shut down and the SharedTimer instance has been cleaned up. Here is the suggested code snippet for this improvement:
```java
    @Test
    void getTimer() throws InterruptedException, ExecutionException, TimeoutException {
        final int iterations = 500;
        ExecutorService executor = Executors.newFixedThreadPool(2);

        try {
            ArrayList<CompletableFuture<?>> futures = new ArrayList<>(iterations);
            for (int i = 0; i < iterations; i++) {
                futures.add(CompletableFuture.runAsync(() -> SharedTimer.getTimer().removeRef(), executor));
            }

            CompletableFuture.allOf(futures.toArray(new CompletableFuture[0])).get(2, TimeUnit.MINUTES);
        } finally {
            executor.shutdown();
            if (!executor.awaitTermination(800, TimeUnit.MILLISECONDS)) {
                executor.shutdownNow();
            }
        }

        assertFalse(SharedTimer.isRunning(), "SharedTimer should be stopped after all references are removed.");
    }
```

An alternative would be to incorporate more detailed logging within the SharedTimer class regarding its state transitions (e.g., when references are added or removed, and when it starts or stops). While this would provide more insight during test execution, it would not replace the need for explicit assertions to verify the correct behavior programmatically.

This suggestion does not directly relate to any specific JDBC Specifications.

## Reference Implementation

----
If this suggestion is helpful, I am more than happy to submit a Pull Request to implement the refactoring.

Codegass:
Resolve #2357 

When I am walking through the unit test of the mssql-jdbc, I noticed that currently the [`SharedTimerTest`](https://github.com/microsoft/mssql-jdbc/blob/9a8849b60e74d1d01508fa61f944d9e635cab700/src/test/java/com/microsoft/sqlserver/jdbc/SharedTimerTest.java#L9-L27) does not contain any assertion statement to verify the behavior and state of the `SharedTimer` after execution of its `getTimer()` method in a multithreaded environment. The test ensures that references are added and removed, but does not assert the expected state of the SharedTimer (e.g., it is stopped and cleaned up properly after all references are removed). This could potentially mask issues with how `SharedTimer` manages its lifecycle and resources.

Current Implementation:
```java
@Test
    void getTimer() throws InterruptedException, ExecutionException, TimeoutException {
        final int iterations = 500;

        ExecutorService executor = Executors.newFixedThreadPool(2);
        try {
            ArrayList<CompletableFuture<?>> futures = new ArrayList<>(iterations);
            for (int i = 0; i < iterations; i++) {
                futures.add(CompletableFuture.runAsync(() -> SharedTimer.getTimer().removeRef(), executor));
            }

            CompletableFuture.allOf(futures.toArray(new CompletableFuture[0])).get(2, TimeUnit.MINUTES);
        } finally {
            executor.shutdown();
        }
    }
```

I'd like to propose a minor improvement by adding a simple assertion at the end of the getTimer test method to verify that the SharedTimer has indeed been stopped and its resources have been cleaned up after all references to it have been removed. This could be accomplished by asserting that SharedTimer.isRunning() returns false, indicating that the internal ScheduledThreadPoolExecutor has been shut down and the SharedTimer instance has been cleaned up. Here is the suggested code snippet for this improvement:
```java
    @Test
    void getTimer() throws InterruptedException, ExecutionException, TimeoutException {
        final int iterations = 500;
        ExecutorService executor = Executors.newFixedThreadPool(2);

        try {
            ArrayList<CompletableFuture<?>> futures = new ArrayList<>(iterations);
            for (int i = 0; i < iterations; i++) {
                futures.add(CompletableFuture.runAsync(() -> SharedTimer.getTimer().removeRef(), executor));
            }

            CompletableFuture.allOf(futures.toArray(new CompletableFuture[0])).get(2, TimeUnit.MINUTES);
        } finally {
            executor.shutdown();
            if (!executor.awaitTermination(800, TimeUnit.MILLISECONDS)) {
                executor.shutdownNow();
            }
        }

        assertFalse(SharedTimer.isRunning(), "SharedTimer should be stopped after all references are removed.");
    }
```

An alternative would be to incorporate more detailed logging within the SharedTimer class regarding its state transitions (e.g., when references are added or removed, and when it starts or stops). While this would provide more insight during test execution, it would not replace the need for explicit assertions to verify the correct behavior programmatically.

This suggestion does not directly relate to any specific JDBC Specifications.


src/test/java/com/microsoft/sqlserver/jdbc/SharedTimerTest.java
```diff
@@ -5,14 +5,16 @@
 import java.util.ArrayList;
 import java.util.concurrent.*;
 
+import static org.junit.jupiter.api.Assertions.assertFalse;
+
```
Jeffery-Wasty, line 9: Not part of your changes, but since we have the opportunity, please remove the * import as that's our preferred style. Also reorder as seen below. Applying our formatter will also give you this suggestion.

```
import static org.junit.jupiter.api.Assertions.assertFalse;

import java.util.ArrayList;
import java.util.concurrent.CompletableFuture;
import java.util.concurrent.ExecutionException;
import java.util.concurrent.ExecutorService;
import java.util.concurrent.Executors;
import java.util.concurrent.TimeUnit;
import java.util.concurrent.TimeoutException;

import org.junit.jupiter.api.Test;
```

src/test/java/com/microsoft/sqlserver/jdbc/SharedTimerTest.java
```diff
@@ -22,6 +24,11 @@ void getTimer() throws InterruptedException, ExecutionException, TimeoutExceptio
             CompletableFuture.allOf(futures.toArray(new CompletableFuture[0])).get(2, TimeUnit.MINUTES);
         } finally {
             executor.shutdown();
+            if (!executor.awaitTermination(800, TimeUnit.MILLISECONDS)) {
+                executor.shutdownNow();
```
lilgreenbird, line 28: this seems really short. We tests in multiple environments including Azure DB which may require longer so this looks like it would cause intermittent failures. A quick search in other tests show that we wait 5 seconds there. We would also need to run full tests in our test lab to confirm it's ok in this case but pretty sure 800 ms would be too short

lilgreenbird (changes requested):
please update as per review comments

src/test/java/com/microsoft/sqlserver/jdbc/SharedTimerTest.java
Codegass: Hi @lilgreenbird, since I don't have the experiment env for the Azure DB, I have only test it on my local machine. As you suggested, I changed it to 5000ms, I will leave a comment in the test to mention that this may needs some confirmation with the set lab.

src/test/java/com/microsoft/sqlserver/jdbc/SharedTimerTest.java
```diff
@@ -22,6 +24,11 @@ void getTimer() throws InterruptedException, ExecutionException, TimeoutExceptio
...
+            }
         }
+
+        assertFalse(SharedTimer.isRunning(), "SharedTimer should be stopped after all references are removed.");
```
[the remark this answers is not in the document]
Codegass: Updated

src/test/java/com/microsoft/sqlserver/jdbc/SharedTimerTest.java
Jeffery-Wasty: Thanks for the change, we'll run these changes and see if the timeout value works for our test lab.

Jeffery-Wasty (approved):

tkyc (approved):

lilgreenbird (approved):

barryw-mssql (approved):

PR #2396, opened 2024-04-23
Configurable Retry Logic I - Statement Retry

Jeffery-Wasty:
ConfigurableRetryLogic
--------------------------

Rules can be defined either in the connection string or the `mssql-jdbc.properties` file. In both cases, the rules follow this format: `retryExec={rule1;rule2;rule3}`.

During connection initialization, the static instance of `ConfigurableRetryLogic` is fetched (`getInstance()`). Since the instance is null, a reread is triggered (`reread()`). 

Within the reread, the `lastTimeRead` is set up, this is to ensure that in the future, the CRL instance is not updated all the time, only when the interval has been reached (default 30 seconds).

Next the rules are setup (`setUpRules()`). If there are rules to read from the connection string, they are used to populate the internal ruleset, if not, then rules are read from the `mssql-jdbc.properties` file. That is, the connection string takes priority over the properties file. After the rules are read, they are parsed into `ConfigRetryRule` objects in `createRules()`.

In `createRules()`, the list of String rules are used to create a HashMap of `ConfigRetryRule` objects

ConfigRetryRule
-----------------

* The rules are parsed in `parse()` removing extra elements that could remain after reading from connection string or file. As well the rules is split based on a `:`.

* Valid rules can have the following formats:
    * {error(s):numberOfRetries}----------------------------------------------------i.e. {2714:1}
    * {error(s):numberOfRetries,initialRetryTime}----------------------------------i.e. {2714:1,3}
    * {error(s):numberOfRetries,intialRetryTime(operand)}------------------------i.e. {2714,1,3+}
    * {error(s):numberOfRetries,intitialRetryTime(operand)retryChange}---------i.e. {2714,1,3+2}
    * {error(s):numberOfRetries:query}---------------------------------------------i.e. {2714:1:CREATE}
    * {error(s):numberOfRetries,intialRetryTime:query}----------------------------i.e. {2714,1,3:CREATE}
    * {error(s):numberOfRetries,intialRetryTime(operand):query}-----------------i.e. {2714,1,3+:CREATE}
    * {error(s):numberOfRetries,intialRetryTime(operand)retryChange:query}----i.e. {2714,1,3+2:CREATE}

* The elements from each rule are fetched using `addElement`. If a rule is <2 or >3 parts, an error is returned. Otherwise, the first part is set as the `retryError`, and the second part (the timings) are parsed further. The first part of the timings is always the `retryCount`. We check if its an integer and greater than 0, if so it is set as `retryCount`. If there are two parts to the timing, then the second part can be the `initalRetryTime` (initial time to wait between retries), the `initialRetryTime` + `operand` (the change we make to the retry time each retry, * or +), or the `initialRetryTime` + `operand` + `retryChange` (the numeric change that is applied to each wait time).
* The second part of timings is parsed, if it contains a *, this is the multiplicative change. We store this in `operand`. The value proceeding the operand is the `initialRetryTime`, and and value following the operand is the `retryChange`. If it doesn't exist, its set to the default.
* The same applies to +. If there is neither operand, then we know we only have the `initialRetryTime` and so we set `initialRetryTime` from the value left in the timings.
* If the rule is 3 parts, then the optional query part has been included (apply this rule to only specific queries). This part of the rules is then set under `retryQueries`.
* Wait time is calculated from the elements we parsed (`calcWaitTimes`) and stored in an ArrayList `waitTimes`.
* With this the rule has been turned into a `ConfigRetryRule`. Another check has to be done to ensure that there is only one `retryError` per rule. If multiple errors were passed in (separated by a comma), then we need to create new rules from the rule, each containing only one `retryError`. Finally all of the rules are put into the stmtRules HashMap.

SQLServerStatement
----------
* Execution of the statement is tried. If there is an error, previously the error was thrown.
* Now, we set a continue flag (`cont`) to false, and attempt to determine whether we can retry. We attempt to fetch the `ConfigRetryRule` that corresponds to the sqlServerError. If there is none, behavior continues as previously defined.
* If there is a match, we check if there are retries remaining. Statement retries are tracked via `retryAttempt` and compared to the `ConfigRetryRule`'s `retryCount`.
* If there are retries remaining, we need to ensure this error also matches in regards to the query. The last query was stored prior to execution (see SQLServerStatement 860 and SQLServerPreparedStatement 583). If there is no query defined for the rule, or the query matches, we proceed.
* We then have to wait until we can retry again. The wait times were calculated previously and can easily be fetched using the `retryAttempt`.
* After waiting we increment `retryAttempt`, and set the `cont` flag to true, allowing us to repeat and try executing again.
* We exit out of this retry in the following ways:
    * On successful execution
    * If we do not find a perfect match in the list of Conifgurable Retry Logic errors (`retryError` and `query`).
    * If we run out of retries


src/main/java/com/microsoft/sqlserver/jdbc/ConfigurableRetryLogic.java
```diff
@@ -0,0 +1,389 @@
...
+
+    /**
+     * Fetches the static instance of ConfigurableRetryLogic, instantiating it if it hasn't already been. Each time the instance
+     * is fetched, we check if a re-read is needed, and do so if properties should be re-read.
+     *
+     * @return The static instance of ConfigurableRetryLogic
+     * @throws SQLServerException
+     *         an exception
+     */
+    public static synchronized ConfigurableRetryLogic getInstance() throws SQLServerException {
```
lilgreenbird, line 44: use locks for loom

src/main/java/com/microsoft/sqlserver/jdbc/ConfigurableRetryLogic.java
```diff
@@ -0,0 +1,343 @@
...
+            if (driverInstance == null) {
+                driverInstance = new ConfigurableRetryLogic();
+            } else {
+                reread();
+            }
+            return driverInstance;
+        } finally {
+            CRL_LOCK.unlock();
+        }
+    }
```
tkyc, line 62: We should do double checked locking here. Something like below:

```suggestion
    public static ConfigurableRetryLogic getInstance() throws SQLServerException {
        if (driverInstance == null) {
            CRL_LOCK.lock();

            try {
                if (driverInstance == null) {
                    driverInstance = new ConfigurableRetryLogic();
                } else {
                    reread();
                }
            } finally {
                CRL_LOCK.unlock();
            }

        } else {
            reread();
        }

        return driverInstance;
    }
```

src/main/java/com/microsoft/sqlserver/jdbc/ConfigurableRetryLogic.java
```diff
@@ -0,0 +1,343 @@
...
+                while ((readLine = buffer.readLine()) != null) {
+                    if (readLine.startsWith("retryExec")) {
+                        String value = readLine.split("=")[1];
+                        Collections.addAll(list, value.split(";"));
+                    }
+                }
+            }
+        } catch (IOException e) {
+            if (CONFIGURABLE_RETRY_LOGGER.isLoggable(java.util.logging.Level.FINEST)) {
+                CONFIGURABLE_RETRY_LOGGER.finest("No properties file exists or file is badly formatted.");
```
tkyc, line 163: I've seen this in other places in the project also. Maybe we should add these to the Resource file so these strings can be localized. Just a thought, not suggesting it to be done in this PR.

tkyc, line 163: Do we want to check if the file exists? eg. If file exists, parse it, and if it's formatted badly error out completely. Otherwise, if file doesn't exist continue on silently.

Right now, if we just look for the IOException we can't tell if it's a bad file.

src/main/java/com/microsoft/sqlserver/jdbc/ConfigurableRetryLogic.java
```diff
@@ -0,0 +1,343 @@
...
+import java.util.Date;
+import java.util.HashMap;
+import java.util.LinkedList;
+import java.util.Map;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+
+
+public class ConfigurableRetryLogic {
+    private final static int INTERVAL_BETWEEN_READS = 30000;
```
barryw-mssql, line 25: Consider including Units (e.g. milliseconds) in the constant variable name. This makes it easier to understand the use of the variable later (e.g. in calculations that may have different units)

src/test/java/com/microsoft/sqlserver/jdbc/ConfigurableRetryLogicTest.java
lilgreenbird, line 1: I think this is big enough feature to warrant a sep folder. We will have tests more tests for configurable connection later too

src/main/java/com/microsoft/sqlserver/jdbc/ConfigurableRetryLogic.java
```diff
@@ -0,0 +1,343 @@
...
+        return parsed.split(":");
+    }
+
+    private void parameterIsNumeric(String value) throws SQLServerException {
+        if (!StringUtils.isNumeric(value)) {
+            String[] arr = value.split(",");
+            for (String error : arr) {
+                if (!StringUtils.isNumeric(error)) {
+                    MessageFormat form = new MessageFormat(SQLServerException.getErrString("R_InvalidParameterFormat"));
+                    throw new SQLServerException(null, form.format(new Object[] {}), null, 0, true);
```
lilgreenbird, line 227: we should add what the error is to the error string

src/main/java/com/microsoft/sqlserver/jdbc/ConfigurableRetryLogic.java
```diff
@@ -0,0 +1,343 @@
...
+            parameterIsNumeric(rule[0]);
+            retryError = rule[0];
+            String[] timings = rule[1].split(",");
+            parameterIsNumeric(timings[0]);
+
+            if (Integer.parseInt(timings[0]) > 0) {
+                retryCount = Integer.parseInt(timings[0]);
+            } else {
+                MessageFormat form = new MessageFormat(SQLServerException.getErrString("R_InvalidParameterFormat"));
+                throw new SQLServerException(null, form.format(new Object[] {}), null, 0, true);
```
lilgreenbird, line 244: see above

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResource.java
```diff
@@ -547,6 +548,9 @@ protected Object[][] getContents() {
         {"R_InvalidSqlQuery", "Invalid SQL Query: {0}"},
         {"R_InvalidScale", "Scale of input value is larger than the maximum allowed by SQL Server."},
         {"R_colCountNotMatchColTypeCount", "Number of provided columns {0} does not match the column data types definition {1}."},
+        {"R_InvalidParameterFormat", "One or more supplied parameters are not correct."},
+        {"R_InvalidRuleFormat", "Wrong number of parameters supplied to rule."},
+        {"R_InvalidRetryInterval", "Current retry interval is longer than queryTimeout."},
```
lilgreenbird, line 553: in all of these include the invalid error into the error string

src/main/java/com/microsoft/sqlserver/jdbc/ConfigurableRetryLogic.java
```diff
@@ -0,0 +1,343 @@
...
+            for (String error : arr) {
+                if (!StringUtils.isNumeric(error)) {
+                    MessageFormat form = new MessageFormat(SQLServerException.getErrString("R_InvalidParameterFormat"));
+                    throw new SQLServerException(null, form.format(new Object[] {}), null, 0, true);
+                }
+            }
+        }
+    }
+
+    private void addElements(String[] rule) throws SQLServerException {
```
lilgreenbird, line 233: please add comment header to describe parsing logic here

src/main/java/com/microsoft/sqlserver/jdbc/ISQLServerDataSource.java
```diff
@@ -1350,4 +1350,18 @@ public interface ISQLServerDataSource extends javax.sql.CommonDataSource {
      * @return calcBigDecimalPrecision boolean value
      */
     boolean getCalcBigDecimalPrecision();
+
+    /**
+     * Returns value of 'retryExec' from Connection String.
+     *
+     * @param retryExec
```
lilgreenbird, line 1357: this needs a description or will get javadoc warning

src/main/java/com/microsoft/sqlserver/jdbc/ISQLServerDataSource.java
```diff
@@ -1350,4 +1350,18 @@ public interface ISQLServerDataSource extends javax.sql.CommonDataSource {
...
+     */
+    void setRetryExec(String retryExec);
+
+    /**
+     * Sets the value for 'retryExec' property
+     *
+     * @return retryExec String value
```
lilgreenbird, line 1364: nit pick: we usually put description in sep line

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -2006,6 +2016,7 @@ Connection connect(Properties propsIn, SQLServerPooledConnection pooledConnectio
                         throw e;
                     } else {
                         // only retry if transient error
+
```
lilgreenbird, line 2019: nitpick: extra blank line

src/main/java/com/microsoft/sqlserver/jdbc/ConfigurableRetryLogic.java
```diff
@@ -0,0 +1,343 @@
...
+import java.util.Collections;
+import java.util.Date;
+import java.util.HashMap;
+import java.util.LinkedList;
+import java.util.Map;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+
+
+public class ConfigurableRetryLogic {
```
lilgreenbird, line 24: need javadoc description

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDriver.java
```diff
@@ -610,7 +610,8 @@ enum SQLServerDriverStringProperty {
     ENCRYPT("encrypt", EncryptOption.TRUE.toString()),
     SERVER_CERTIFICATE("serverCertificate", ""),
     DATETIME_DATATYPE("datetimeParameterType", DatetimeType.DATETIME2.toString()),
-    ACCESS_TOKEN_CALLBACK_CLASS("accessTokenCallbackClass", "");
+    ACCESS_TOKEN_CALLBACK_CLASS("accessTokenCallbackClass", ""),
+    RETRY_EXEC("retryExec", "");
```
lilgreenbird, line 614: did David ok this name? seems a bit too vague to me I would think something more descriptive would be better as all our other properties are more descriptive

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerStatement.java
```diff
@@ -258,22 +259,68 @@ final void executeStatement(TDSCommand newStmtCmd) throws SQLServerException, SQ
...
+                executeCommand(newStmtCmd);
+            } catch (SQLServerException e) {
+                SQLServerError sqlServerError = e.getSQLServerError();
+                ConfigRetryRule rule = null;
+
+                if (null != sqlServerError) {
+                    rule = ConfigurableRetryLogic.getInstance().searchRuleSet(e.getSQLServerError().getErrorNumber());
+                }
+
+                if (null != rule && retryAttempt < rule.getRetryCount()) {
```
lilgreenbird, line 278: add comment description pls

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerStatement.java
```diff
@@ -258,22 +259,68 @@ final void executeStatement(TDSCommand newStmtCmd) throws SQLServerException, SQ
...
+                    boolean meetsQueryMatch = true;
```
lilgreenbird, line 279: not sure about this variable name - what does "meetsQueryMatch" mean?

src/main/java/com/microsoft/sqlserver/jdbc/ConfigurableRetryLogic.java
```diff
@@ -0,0 +1,343 @@
...
+        try {
+            String className = new Object() {}.getClass().getEnclosingClass().getName();
+            String location = Class.forName(className).getProtectionDomain().getCodeSource().getLocation().getPath();
+            location = location.substring(0, location.length() - 16);
+            URI uri = new URI(location + "/");
+            return uri.getPath();
+        } catch (Exception e) {
+            if (CONFIGURABLE_RETRY_LOGGER.isLoggable(java.util.logging.Level.FINEST)) {
+                CONFIGURABLE_RETRY_LOGGER.finest("Unable to get current class path for properties file reading.");
+            }
```
lilgreenbird, line 140: we're eating the exception?

src/main/java/com/microsoft/sqlserver/jdbc/ConfigurableRetryLogic.java
lilgreenbird: why are we eating the exception and not throwing?

lilgreenbird (dismissed):
needs comment blocks everywhere, especially class headers and all public

src/main/java/com/microsoft/sqlserver/jdbc/ConfigurableRetryLogic.java
```diff
@@ -0,0 +1,343 @@
...
+            } else {
+                reread();
+            }
+            return driverInstance;
+        } finally {
+            CRL_LOCK.unlock();
+        }
+    }
+
+    private static void reread() throws SQLServerException {
```
lilgreenbird, line 64: a bit of a nitpick but I don't really like the name. There's nothing here that RE reads anything it just reads. The fact that you are calling it to re-read is just how you are using this method

src/main/java/com/microsoft/sqlserver/jdbc/ConfigurableRetryLogic.java
```diff
@@ -0,0 +1,343 @@
...
+            if (entry.getKey() == ruleToSearch) {
+                return entry.getValue();
+            }
+        }
+        return null;
+    }
+}
+
+
+class ConfigRetryRule {
```
lilgreenbird, line 181: 1 class per file pls

src/main/java/com/microsoft/sqlserver/jdbc/ConfigurableRetryLogic.java
```diff
@@ -0,0 +1,343 @@
...
+        createRules(temp);
+    }
+
+    private static void createRules(LinkedList<String> listOfRules) throws SQLServerException {
+        stmtRules = new HashMap<>();
+
+        for (String potentialRule : listOfRules) {
+            ConfigRetryRule rule = new ConfigRetryRule(potentialRule);
+
+            if (rule.getError().contains(",")) {
```
lilgreenbird, line 117: define constants for all these special chars

src/main/java/com/microsoft/sqlserver/jdbc/ConfigurableRetryLogic.java
```diff
@@ -0,0 +1,343 @@
...
+        String filePath = getCurrentClassPath();
+        LinkedList<String> list = new LinkedList<>();
+
+        try {
+            File f = new File(filePath + DEFAULT_PROPS_FILE);
+            timeLastModified = f.lastModified();
+            try (BufferedReader buffer = new BufferedReader(new FileReader(f))) {
+                String readLine;
+                while ((readLine = buffer.readLine()) != null) {
+                    if (readLine.startsWith("retryExec")) {
```
lilgreenbird, line 155: define constant

src/main/java/com/microsoft/sqlserver/jdbc/ConfigRetryRule.java
```diff
@@ -0,0 +1,279 @@
...
+import java.text.MessageFormat;
+import java.util.ArrayList;
+
+
+/**
+ * The ConfigRetryRule object is what is used by the ConfigurableRetryLogic class to handle statement retries. Each
+ * ConfigRetryRule object allows for one rule.
+ *
+ */
+public class ConfigRetryRule {
```
lilgreenbird, line 17: The naming is inconsistent with ConfigurableRetryLogic class. Should this be ConfigurableRetryRule?

src/main/java/com/microsoft/sqlserver/jdbc/ConfigRetryRule.java
```diff
@@ -0,0 +1,279 @@
...
+    private String retryQueries = "";
+    private final String PLUS_SIGN = "+";
+    private final String MULTIPLICATION_SIGN = "*";
+    private final String COMMA = ",";
+    private final String ZERO = "0";
+
+    private ArrayList<Integer> waitTimes = new ArrayList<>();
+
+    /**
+     * Default constructor
```
lilgreenbird, line 32: this is not a default constructor

src/main/java/com/microsoft/sqlserver/jdbc/ConfigRetryRule.java
lilgreenbird: why does this need to be public?

src/main/java/com/microsoft/sqlserver/jdbc/ConfigRetryRule.java
```diff
@@ -0,0 +1,279 @@
...
+     * Allows constructing a ConfigRetryRule object from another ConfigRetryRule object. Used when the first object has
+     * multiple errors provided. We pass in the multi-error object and create 1 new object for each error in the initial
+     * object.
+     *
+     * @param rule
+     *        The rule used to construct the ConfigRetryRule object
+     * @param base
+     *        The ConfigRetryRule object to base the new objects off of
+     */
+    public ConfigRetryRule(String rule, ConfigRetryRule base) {
```
lilgreenbird, line 54: the namings are confusing. This copies a config retry rule, and also takes another rule? both are rules

src/main/java/com/microsoft/sqlserver/jdbc/ConfigRetryRule.java
```diff
@@ -0,0 +1,279 @@
...
+        copyFromExisting(base);
+        this.retryError = rule;
+    }
+
+    private void copyFromExisting(ConfigRetryRule base) {
```
lilgreenbird, line 59: this is just a copy, specifically a shallow copy. How caller uses this shouldn't be part of the name

lilgreenbird: the variable name base is also irrelevant, it's just a rule

src/main/java/com/microsoft/sqlserver/jdbc/ConfigRetryRule.java
```diff
@@ -0,0 +1,279 @@
...
+        this.retryError = base.getError();
+        this.operand = base.getOperand();
+        this.initialRetryTime = base.getInitialRetryTime();
+        this.retryChange = base.getRetryChange();
+        this.retryCount = base.getRetryCount();
+        this.retryQueries = base.getRetryQueries();
+        this.waitTimes = base.getWaitTimes();
+    }
+
+    private String[] parse(String rule) {
```
lilgreenbird, line 69: add comment describing what this is for

src/main/java/com/microsoft/sqlserver/jdbc/ConfigRetryRule.java
```diff
@@ -0,0 +1,279 @@
...
+    /**
+     * Checks if the value passed in is numeric. In the case where the value contains a comma, the value must be a
+     * multi-error value, e.g. 2714,2716. This must be separated, and each error checked separately.
+     *
+     * @param value
+     *        The value to be checked
+     * @throws SQLServerException
+     *         if a non-numeric value is passed in
+     */
+    private void checkParameterIsNumeric(String value) throws SQLServerException {
```
lilgreenbird, line 90: would there be other checks? maybe better to just call this checkParameter

src/main/java/com/microsoft/sqlserver/jdbc/ConfigRetryRule.java
```diff
@@ -0,0 +1,279 @@
...
+                retryQueries = (rule[2].equals(ZERO) ? "" : rule[2].toLowerCase());
+            }
+        } else {
+            MessageFormat form = new MessageFormat(SQLServerException.getErrString("R_InvalidRuleFormat"));
+            Object[] msgArgs = {rule.length};
+            throw new SQLServerException(null, form.format(msgArgs), null, 0, true);
+        }
+    }
+
+    private void calculateWaitTimes() {
```
lilgreenbird, line 194: comment

src/main/java/com/microsoft/sqlserver/jdbc/ConfigRetryRule.java
```diff
@@ -0,0 +1,279 @@
...
+        this.operand = base.getOperand();
+        this.initialRetryTime = base.getInitialRetryTime();
+        this.retryChange = base.getRetryChange();
+        this.retryCount = base.getRetryCount();
+        this.retryQueries = base.getRetryQueries();
+        this.waitTimes = base.getWaitTimes();
+    }
+
+    private String[] parse(String rule) {
+        if (rule.endsWith(":")) {
```
lilgreenbird, line 70: why isn't the colon defined too?

src/main/java/com/microsoft/sqlserver/jdbc/ConfigRetryRule.java
```diff
@@ -0,0 +1,279 @@
...
+            rule = rule + ZERO; // Add a zero to make below parsing easier
+        }
+
+        rule = rule.replace("{", "");
```
lilgreenbird, line 74: and braces?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerStatement.java
```diff
@@ -258,22 +259,66 @@ final void executeStatement(TDSCommand newStmtCmd) throws SQLServerException, SQ
...
+            try {
+                // (Re)execute this Statement with the new command
+                executeCommand(newStmtCmd);
+            } catch (SQLServerException e) {
+                SQLServerError sqlServerError = e.getSQLServerError();
+                ConfigRetryRule rule = null;
+
+                if (null != sqlServerError) {
+                    rule = ConfigurableRetryLogic.getInstance().searchRuleSet(e.getSQLServerError().getErrorNumber());
+                }
```
lilgreenbird, line 276: can we just getInstance once

src/test/java/com/microsoft/sqlserver/jdbc/configurableretry/ConfigurableRetryLogicTest.java
```diff
@@ -0,0 +1,453 @@
...
+    private static final String tableName = AbstractSQLGenerator
+            .escapeIdentifier(RandomUtil.getIdentifier("crlTestTable"));
+
+    @BeforeAll
+    public static void setupTests() throws Exception {
+        setConnection();
+    }
+
+    /**
+     * Test that the SQLServerConnection methods getRetryExec and setRetryExec are working.
```
lilgreenbird, line 48: test that * "works" is not very descriptive, specify what the test is testing. Multiple occurrences

src/test/java/com/microsoft/sqlserver/jdbc/configurableretry/ConfigurableRetryLogicTest.java
```diff
@@ -0,0 +1,453 @@
...
+     *         if an exception occurs
+     */
+    @Test
+    public void testRetryExecConnectionStringOption() throws Exception {
+        try (SQLServerConnection conn = (SQLServerConnection) DriverManager.getConnection(connectionString);
+                Statement s = conn.createStatement()) {
+            String test = conn.getRetryExec();
+            assertTrue(test.isEmpty());
+            conn.setRetryExec("{2714:3,2*2:CREATE;2715:1,3}");
+            PreparedStatement ps = conn.prepareStatement("create table " + tableName + " (c1 int null);");
```
lilgreenbird, line 60: leak, use try-with-resource or close in finally. Multiple occurrences

src/main/java/com/microsoft/sqlserver/jdbc/ConfigRetryRule.java
```diff
@@ -0,0 +1,279 @@
...
+        return initialRetryTime;
+    }
+
+    /**
+     * Returns the retry change (timing change to apply to wait times) for this ConfigRetryRule object.
+     *
+     * @return
+     *         The retry change
+     */
+    public int getRetryChange() {
```
lilgreenbird, line 246: this is not used anywhere outside of this class, why does it need to be public? as a matter of fact, why does this need to exist at all if it's just within this class

src/main/java/com/microsoft/sqlserver/jdbc/ConfigurableRetryLogic.java
```diff
@@ -0,0 +1,218 @@
...
+ * Allows configurable statement retry through the use of the 'retryExec' connection property. Each rule read in is
+ * converted to ConfigRetryRule objects, which are stored and referenced during statement retry.
+ *
+ */
+public class ConfigurableRetryLogic {
+    private final static int INTERVAL_BETWEEN_READS_IN_MS = 30000;
+    private final static String DEFAULT_PROPS_FILE = "mssql-jdbc.properties";
+    private static final java.util.logging.Logger CONFIGURABLE_RETRY_LOGGER = java.util.logging.Logger
+            .getLogger("com.microsoft.sqlserver.jdbc.ConfigurableRetryLogic");
+    private static ConfigurableRetryLogic driverInstance = null;
```
lilgreenbird, line 35: confusing name, why is a configurable retry logic object named driverInstance

src/main/java/com/microsoft/sqlserver/jdbc/ConfigurableRetryLogic.java
```diff
@@ -0,0 +1,218 @@
...
+     * If it has been INTERVAL_BETWEEN_READS_IN_MS (30 secs) since last read, see if we last did a file read, if so
+     * only reread if the file has been modified. If no file read, set up rules using the prev. connection string rules.
+     *
+     * @throws SQLServerException
+     *         when an exception occurs
+     */
+    private static void refreshRuleSet() throws SQLServerException {
+        long currentTime = new Date().getTime();
+        if ((currentTime - timeLastRead) >= INTERVAL_BETWEEN_READS_IN_MS) {
+            // If it has been 30 secs, reread
```
lilgreenbird, line 89: the value may change, do not mention hardcoded value.
Just something like "if it has been > an interval wait time"

src/main/java/com/microsoft/sqlserver/jdbc/ConfigRetryRule.java
Jeffery-Wasty: Private, but added.

src/main/java/com/microsoft/sqlserver/jdbc/ConfigRetryRule.java
Jeffery-Wasty: Private, but added.

src/main/java/com/microsoft/sqlserver/jdbc/ConfigRetryRule.java
Jeffery-Wasty: Only used in this method. Shouldn't define a constant for what is essentially a local variable.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerStatement.java
Jeffery-Wasty: sure

src/main/java/com/microsoft/sqlserver/jdbc/ConfigurableRetryLogic.java
```diff
@@ -0,0 +1,217 @@
...
+ * Allows configurable statement retry through the use of the 'retryExec' connection property. Each rule read in is
+ * converted to ConfigRetryRule objects, which are stored and referenced during statement retry.
+ *
+ */
+public class ConfigurableRetryLogic {
+    private final static int INTERVAL_BETWEEN_READS_IN_MS = 30000;
+    private final static String DEFAULT_PROPS_FILE = "mssql-jdbc.properties";
+    private static final java.util.logging.Logger CONFIGURABLE_RETRY_LOGGER = java.util.logging.Logger
+            .getLogger("com.microsoft.sqlserver.jdbc.ConfigurableRetryLogic");
+    private static ConfigurableRetryLogic singleInstance = null;
```
barryw-mssql, line 35: Consider making singeInstance singleton variable volatile. It looks like you are using the Double Checked Locking Idiom to perform the synchronization (i.e. check null, lock, and recheck null). However this does not work in Java. Please check this blog describing the issue in both C++ and Java:
https://www.cs.umd.edu/~pugh/java/memoryModel/DoubleCheckedLocking.html

Here is a 2nd blog describing the issue and why volatile fixes the problem:
https://stackoverflow.com/questions/11639746/what-is-the-point-of-making-the-singleton-instance-volatile-while-using-double-l

Jeffery-Wasty: Good comment, but the `singleInstance` is actually thread-safe because its a static initializer, so I can just remove the locks altogether.

tkyc (approved):

src/main/java/com/microsoft/sqlserver/jdbc/ConfigurableRetryRule.java
```diff
@@ -0,0 +1,259 @@
...
+     * @throws SQLServerException
+     *         if a non-numeric value is passed in
+     */
+    private void checkParameter(String value) throws SQLServerException {
+        if (!StringUtils.isNumeric(value)) {
+            String[] arr = value.split(COMMA);
+            for (String error : arr) {
+                if (!StringUtils.isNumeric(error)) {
+                    MessageFormat form = new MessageFormat(SQLServerException.getErrString("R_InvalidParameterFormat"));
+                    Object[] msgArgs = {error, "Not a positive number"};
```
lilgreenbird, line 102: hardcoded error string, this should be in resource file.
Also, we already have a R_invalidParameterNumber can we re-use that?

src/main/java/com/microsoft/sqlserver/jdbc/ConfigurableRetryRule.java
```diff
@@ -0,0 +1,259 @@
...
+
+    /**
+     * Construct a ConfigurableRetryRule object from a String rule.
+     *
+     * @param rule
+     *        The rule used to construct the ConfigRetryRule object
+     * @throws SQLServerException
+     *         If there is a problem parsing the rule
+     */
+    public ConfigurableRetryRule(String rule) throws SQLServerException {
```
lilgreenbird, line 39: this doesn't need to be public

src/main/java/com/microsoft/sqlserver/jdbc/ConfigurableRetryRule.java
```diff
@@ -0,0 +1,259 @@
...
+     * Allows constructing a ConfigRetryRule object from another ConfigRetryRule object. Used when the first object has
+     * multiple errors provided. We pass in the multi-error object and create 1 new object for each error in the initial
+     * object.
+     *
+     * @param newRule
+     *        The rule used to construct the ConfigRetryRule object
+     * @param baseRule
+     *        The ConfigRetryRule object to base the new objects off of
+     */
+    public ConfigurableRetryRule(String newRule, ConfigurableRetryRule baseRule) {
```
lilgreenbird, line 54: doesn't need to be public

src/main/java/com/microsoft/sqlserver/jdbc/ConfigurableRetryLogic.java
```diff
@@ -0,0 +1,206 @@
...
+            timeLastModified = f.lastModified();
+        } catch (FileNotFoundException e) {
+            // If the file is not found either A) We're not using CRL OR B) the path is wrong. Do not error out, instead
+            // log a message.
+            if (CONFIGURABLE_RETRY_LOGGER.isLoggable(java.util.logging.Level.FINER)) {
+                CONFIGURABLE_RETRY_LOGGER.finest("No properties file exists or the file path is incorrect.");
+            }
+        } catch (IOException e) {
+            MessageFormat form = new MessageFormat(SQLServerException.getErrString("R_errorReadingStream"));
+            Object[] msgArgs = {e.toString()};
```
lilgreenbird, line 191: this seems to be the wrong error msg used, also would be good to add the file path to the error msg

src/main/java/com/microsoft/sqlserver/jdbc/ConfigurableRetryLogic.java
```diff
@@ -0,0 +1,206 @@
...
+                        Collections.addAll(list, value.split(SEMI_COLON));
+                    }
+                }
+            }
+            timeLastModified = f.lastModified();
+        } catch (FileNotFoundException e) {
+            // If the file is not found either A) We're not using CRL OR B) the path is wrong. Do not error out, instead
+            // log a message.
+            if (CONFIGURABLE_RETRY_LOGGER.isLoggable(java.util.logging.Level.FINER)) {
+                CONFIGURABLE_RETRY_LOGGER.finest("No properties file exists or the file path is incorrect.");
```
lilgreenbird, line 187: just use something like "file not found" and add path to log file

src/main/java/com/microsoft/sqlserver/jdbc/ConfigurableRetryLogic.java
```diff
@@ -0,0 +1,206 @@
...
+            }
+        } catch (IOException e) {
+            MessageFormat form = new MessageFormat(SQLServerException.getErrString("R_errorReadingStream"));
+            Object[] msgArgs = {e.toString()};
+            throw new SQLServerException(form.format(msgArgs), null, 0, e);
+        }
```
lilgreenbird, line 193: add test as per above

src/main/java/com/microsoft/sqlserver/jdbc/ConfigurableRetryRule.java
Jeffery-Wasty: I didn't see that before. Replaced.

src/main/java/com/microsoft/sqlserver/jdbc/ConfigurableRetryLogic.java
Jeffery-Wasty: Not possible.

src/main/java/com/microsoft/sqlserver/jdbc/ConfigurableRetryLogic.java
```diff
@@ -0,0 +1,206 @@
...
+
+    /**
+     * Fetches the static instance of ConfigurableRetryLogic, instantiating it if it hasn't already been.
+     * Each time the instance is fetched, we check if a re-read is needed, and do so if properties should be re-read.
+     *
+     * @return The static instance of ConfigurableRetryLogic
+     * @throws SQLServerException
+     *         an exception
+     */
+    public static ConfigurableRetryLogic getInstance() throws SQLServerException {
```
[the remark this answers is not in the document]
Jeffery-Wasty: We cannot use synchronized in order to maintain compatibility with loom. https://github.com/microsoft/mssql-jdbc/issues/1154

src/main/java/com/microsoft/sqlserver/jdbc/ConfigurableRetryLogic.java
```diff
@@ -0,0 +1,206 @@
...
+import java.util.LinkedList;
+import java.util.Map;
+
+
+/**
+ * Allows configurable statement retry through the use of the 'retryExec' connection property. Each rule read in is
+ * converted to ConfigRetryRule objects, which are stored and referenced during statement retry.
+ *
+ */
+public class ConfigurableRetryLogic {
```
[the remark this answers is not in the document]
Jeffery-Wasty: This would not be consistent with already named classes. It's a good idea, but will need to be saved for some future point where we're doing mass code rewrite.

src/main/java/com/microsoft/sqlserver/jdbc/ConfigurableRetryLogic.java
```diff
@@ -0,0 +1,206 @@
...
+        } catch (IOException e) {
+            MessageFormat form = new MessageFormat(SQLServerException.getErrString("R_errorReadingStream"));
+            Object[] msgArgs = {e.toString()};
+            throw new SQLServerException(form.format(msgArgs), null, 0, e);
+        }
+        return list;
+    }
+
+    ConfigurableRetryRule searchRuleSet(int ruleToSearchFor) throws SQLServerException {
+        refreshRuleSet();
```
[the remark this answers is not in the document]
Jeffery-Wasty: I don't see any issues with `refreshRuleSet` and threading mainly because of the 30 sec requirement. We read not only when the file is updated, but if 30 secs has passed. So therefore, a user should not expect changes in the props file to be immediately reflected. So if a thread reads one set of props, and another thread another, i) the "wrong" thread will be updated in the next run through, and ii) this means one extra statement retry, which in comparison to the 30 sec wait is inconsequential. That is, for that thread, the change in props was reflected 30 secs and some milliseconds after file change, as opposed to 30 secs.

For the members, I see nothing that would cause issues. As most, an out of sync `timeLastModified` would lead to an extra read, which is fine.

src/main/java/com/microsoft/sqlserver/jdbc/ConfigurableRetryLogic.java
Jeffery-Wasty: An IO Exception is thrown if the above BufferedReader was interrupted during read. This appears to be the correct error for this case, and is used in our driver in other cases where we have an potential interrupted reader. What should be used instead?

Jeffery-Wasty: My mistake, fixed.

src/main/java/com/microsoft/sqlserver/jdbc/ConfigurableRetryLogic.java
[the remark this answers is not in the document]
barryw-mssql: when I said "private static class members do not need to be protected" in my previous comment, I should have said "private static class members do not need to be thread-safe"

src/main/java/com/microsoft/sqlserver/jdbc/ConfigurableRetryLogic.java
barryw-mssql: It is statically initialized to NULL so there is no ConfigurableRetryLogic object instance. We are using lazy construction in getInstance() to create the object at dynamically at run time.

barryw-mssql: by load time I mean when the java class is loaded into memory and all the appropriate static methods have been invoked but before a client thread is given access to the classes in the JAR. In this case it would be either a thread from the calling application or a thread that we created at run time. This is why I am saying that we are using lazy construction. It means that the ConfigurableRetryLogic instance is not constructed by Java which is thread safe but by a client thread or a thread spawned by a client thread at run time. Just because the getInstance() method is static, it does not mean it is implicitly thread safe. It only means that there is one instance and that multiple threads can and will access that static method

src/main/java/com/microsoft/sqlserver/jdbc/ConfigurableRetryLogic.java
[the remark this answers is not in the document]
barryw-mssql: I'm not sure if I agree with that assessment. This is all new code and so we can write it to adhere to any naming conventions we want, even if that means we have changed the conventions to a new standard. I am not asking to retroactively change code in the main branch to adhere to a new naming standard, just to do it for the new code being added

lilgreenbird (approved):

src/main/java/com/microsoft/sqlserver/jdbc/ConfigurableRetryLogic.java
```diff
@@ -56,12 +59,22 @@ private ConfigurableRetryLogic() throws SQLServerException {
      *         an exception
      */
     public static ConfigurableRetryLogic getInstance() throws SQLServerException {
-        // No need for lock; static initializer singleInstance is thread-safe
+        // No need for lock; static initializer singleInstance is thread-safe]
```
barryw-mssql: Please alter or remove comment as it is no longer correct

src/main/java/com/microsoft/sqlserver/jdbc/ConfigurableRetryLogic.java
```diff
@@ -0,0 +1,219 @@
...
+ *
+ */
+public class ConfigurableRetryLogic {
+    private final static int INTERVAL_BETWEEN_READS_IN_MS = 30000;
+    private final static String DEFAULT_PROPS_FILE = "mssql-jdbc.properties";
+    private static final Lock CRL_LOCK = new ReentrantLock();
+    private static final java.util.logging.Logger CONFIGURABLE_RETRY_LOGGER = java.util.logging.Logger
+            .getLogger("com.microsoft.sqlserver.jdbc.ConfigurableRetryLogic");
+    private static ConfigurableRetryLogic singleInstance;
+    private static long timeLastModified;
```
barryw-mssql, line 37: As we are not serializing the calls to the statis ConfigurableRetryLogic static methods (i.e. except for the getInstance() method, all the static class objects are inherently not thread safe. Most Java statements are not atomic. That means that internally even a Java assignment operation internally within the JVM results in several statements that are executed. Because of this it is possible in a multithreaded scenario for there to be a switch in the middle of a Java assignment statement (e.g. hypothetically the Java assignment statement results in 4 separate statements in the JVM) and switches to another thread also performing that same assignment statement resulting in corrupted data for one or possibly both threads.

To address this scenario, Java has added atomic class types for manipulating simple objects in a thread safe manner (i.e. the serialization is implicitly performed in the atomic class and does not need to be explicitly performed in our own code). Please convert these static objects shared between thread to use AtomicInteger, AtomicLong, and AtomicReference. This will ensure that a single Java statement manipulating these will be atomic and cannot be impacted by a switch between threads.

Form more details please refer to: https://naveen-metta.medium.com/atomic-operations-in-java-mastering-thread-safety-and-concurrency-7c3360ec0bc5

Jeffery-Wasty: Fair point, done.

barryw-mssql (approved):

barryw-mssql (approved):

tkyc (approved):

lilgreenbird (approved):

PR #2417, opened 2024-05-15
Profiles for new LTS JDK

tkyc:


build.gradle
```diff
@@ -46,6 +46,19 @@ if (!hasProperty('buildProfile') || (hasProperty('buildProfile') && buildProfile
 	targetCompatibility = 22
 }
 
+if (!hasProperty('buildProfile') || (hasProperty('buildProfile') && buildProfile == "jre21")) {
```
Jeffery-Wasty, line 49:
```suggestion
if (hasProperty('buildProfile') && buildProfile == "jre21") {
```

We want default to be 22.

Jeffery-Wasty (approved):

build.gradle
Jeffery-Wasty: Never mind, if there is no build profile it will build with 22, no need to remove.

barryw-mssql, line 49: Please correct me if I am wrong but isn't (!hasProperty('buildProfile') || (hasProperty('buildProfile') always true?

barryw-mssql: Sorry ... missed the order of brackets. Never mind

barryw-mssql (approved):

Issue #2424, opened 2024-05-22
Implementation bug in `SqlAuthenticationToken` constructor accepting unix epoch


konjac:
## Driver version
12.0

## SQL Server version
Any

## Client Operating System
Any

## JAVA/JVM version
Any

## Table schema
Any

## Problem description
The implementation of  `public SqlAuthenticationToken(String accessToken, long expiresOn)` is different from its document. The doc says `expiresOn` is in seconds while `java.util.Date` is accepting timestamp in milliseconds.

## Expected behavior
Convert the input `expiresOn` to milliseconds when creating `java.util.Date`

## Actual behavior
Output of what you actually see.

## Error message/stack trace
Complete error message and stack trace.

## JDBC trace logs
Provide the JDBC driver trace logs.  Instructions can be found here: https://docs.microsoft.com/sql/connect/jdbc/tracing-driver-operation

Jeffery-Wasty:
Thank you for the PR, we'll take a look and get back to you with our decision.

PR #2425, opened 2024-05-22
Fix SqlAuthenticationToken constructor accepting unix epoch

konjac:
This is to address #2424


konjac:
@barryw-mssql @tkyc Could you help to review this fix?

barryw-mssql (approved):

Jeffery-Wasty:
@konjac Thanks for the PR, we'll take a look at it and get back to you with our thoughts.

Jeffery-Wasty:
It looks like this is an issue of the documentation being wrong, not the implementation. When creating the `SQLAuthenticationToken` we are indeed passing in millisecond values, but the Javadoc should be changed to reflect that.

tkyc (approved):

Jeffery-Wasty (changes requested):
Hold off on merging for a bit

tkyc (dismissed):

Jeffery-Wasty (approved):

barryw-mssql (approved):

PR #2468, opened 2024-07-03
Add logging info to token caching

lilgreenbird:


Jeffery-Wasty (approved):

tkyc (approved):

David-Engel (commented):
Just a note: I would have done FINER logging here. This isn't a hot path and it seems silly to require FINEST logging (with all the packet logging that comes with) to get this info for troubleshooting.

PR #2482, opened 2024-07-24
Prepared statement objects should be reusable

tkyc:
Discovered that when attempting to reuse the same pstmt object, the driver would error out with `R_multipleQueriesNotAllowed` and consume that error.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
@@ -2681,13 +2681,6 @@ public class SQLServerPreparedStatement extends SQLServerStatement implements IS
 
     private void checkAdditionalQuery() {
         while (checkAndRemoveCommentsAndSpace(true)) {}
-
-        // At this point, if localUserSQL is not empty (after removing all whitespaces, semicolons and comments), we
-        // have a
-        // new query. reject this.
-        if (localUserSQL.length() > 0) {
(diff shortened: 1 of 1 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -2681,13 +2681,6 @@ private void checkValidColumns(TypeInfo ti) throws SQLServerException {
 
     private void checkAdditionalQuery() {
         while (checkAndRemoveCommentsAndSpace(true)) {}
```
Jeffery-Wasty, line 2683: At this point the method is just:

```
private void checkAdditionalQuery() {
        while (checkAndRemoveCommentsAndSpace(true)) {}
    }
```

May as well replace the two `checkAdditionalQuery()` method calls at 2317 and 2512, with `while (checkAndRemoveCommentsAndSpace(true)) {};`

tkyc: No, I'll be duping the code.

Jeffery-Wasty (dismissed):

lilgreenbird (changes requested):
add test?

tkyc:
This error isn't explicitly thrown, so I can't really test by not looking for it. In the old code, this error was consumed and the driver would keep running.

Jeffery-Wasty (approved):

lilgreenbird (approved):

PR #2484, opened 2024-08-06
Added a missing unlock in `getDestinationMetadata()`

Jeffery-Wasty:


barryw-mssql (approved):

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
```diff
@@ -1762,6 +1762,8 @@ private void getDestinationMetadata() throws SQLServerException {
                     } finally {
                         DESTINATION_COL_METADATA_LOCK.unlock();
                     }
+                } else {
+                    DESTINATION_COL_METADATA_LOCK.unlock();
```
David-Engel, line 1766: Do we need to lock at all if the cache get is successful?

David-Engel: Never mind, I didn't read up higher in the code where we did the first get.

David-Engel, line 1766: A better pattern would be to move the start of the try block to the line below the lock(). That way you have one lock and one unlock and you are sure to always unlock in the finally after taking the lock.

David-Engel:
Also, you should add a test to cover the scenario that the reporting issue manifests.

barryw-mssql (approved):

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/BatchExecutionTest.java
```diff
@@ -206,6 +211,82 @@ public void testSqlServerBulkCopyCachingConnectionLevel() throws Exception {
...
+        long ms = 1578743412000L;
+        long timeOut = 90000; // 90 seconds
+        int NUMBER_SIMULTANEOUS_INSERTS = 40;
+
+        try (SQLServerConnection con = (SQLServerConnection) DriverManager.getConnection(connectionString
+                + ";useBulkCopyForBatchInsert=true;cacheBulkCopyMetadata=true;sendTemporalDataTypesAsStringForBulkCopy=false;");
+                Statement stmt = con.createStatement()) {
+
+            // Needs to be on a JDK version greater than 8
+            assumeTrue(TestUtils.getJVMVersion() > 8);
```
lilgreenbird, line 226: This is not necessary.
As discussed pls move this line in the other test to beginning of the test and add comment

Jeffery-Wasty: I've moved it above, but it is necessary as we need reflection to clear the cache in case of deadlock.

barryw-mssql (approved):

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/BatchExecutionTest.java
```diff
@@ -206,6 +211,82 @@ public void testSqlServerBulkCopyCachingConnectionLevel() throws Exception {
...
 
+    @Test
+    public void testSqlServerBulkCopyCachingConnectionLevelMultiThreaded() throws Exception {
+        Calendar gmtCal = Calendar.getInstance(TimeZone.getTimeZone("GMT"));
+        long ms = 1578743412000L;
+        long timeOut = 30000; // 30 seconds
+        int NUMBER_SIMULTANEOUS_INSERTS = 5;
+
+        // Needs to be on a JDK version greater than 8
+        assumeTrue(TestUtils.getJVMVersion() > 8);
```
lilgreenbird, line 222: nitpick but, we don't need to define all those vars if we're going to skip.

lilgreenbird (approved):

lilgreenbird (dismissed):
it's failing code coverage?

lilgreenbird (approved):

barryw-mssql (approved):

PR #2449, opened 2024-06-11
Fixed timestamp string conversion error for cstmt

tkyc:
Fixes https://github.com/microsoft/mssql-jdbc/issues/2436.

PR #2410, opened 2024-05-08
Execute stored procedures directly for RPC calls

tkyc:
RPC calls were falling back to old logic of being wrapped in sp_executesql calls to execute. With the PR, moving forward, RPC calls are also directly executed.

PR #2349, opened 2024-03-18
Fix calling procedures with output parameters by their four-part syntax

tkyc:
https://github.com/microsoft/mssql-jdbc/issues/2344

PR #2488, opened 2024-08-14
Revert Execute Stored Procedure directly feature

tkyc:
Revert of the following:

Fixed regression with specifying arg names in call syntax (#2480)
0e97689f50f30c70dcfec89bcc89537ea622b8b5

Fixed parentheses parsing for stored procedures and functions names (#2467)
8f53c026bc62ee3852bd0ea9a513288aad9155fc

Fixed CallableStatement default value regression (#2452)
93070514af439954021f90d69cd44b20cf4625e0

Fixed timestamp string conversion error for cstmt (#2449)
975407889ef90522d5414897072c970eaaf0117a

Execute stored procedures directly for RPC calls (#2410)
df5bfa6b80ba419042498b343aed697b7807b810

Fix calling procedures with output parameters by their four-part syntax (#2349)
aa466376a1ee28ed069a8753487ace65b2855f2c

Re-added support for stored procedure 'exec' escape syntax in CallableStatements (#2325)
ba88da8e97d10b2ffda325599699a37525981ffd

Execute cstmt directly - Additional testing and changes (#2284)
92cfe0d2162985f33aa2d0723cdf99ca27ff361f

Execute Stored Procedures Directly (#2154)
11680a61e38ec7b0075cd60376156b8824ee93de


pom.xml
```diff
@@ -43,7 +43,7 @@
 			xAzureSQLDW - - - - For tests not compatible with Azure Data Warehouse - 
 			xAzureSQLMI - - - - For tests not compatible with Azure SQL Managed Instance 
 			NTLM  - - - - - - - For tests using NTLM Authentication mode (excluded by default)
-			kerberos  - - - - - For tests using Kerberos authentication (excluded by default)
+			kerberos -  - - - - For tests using Kerberos authentication (excluded by default)
```
Jeffery-Wasty, line 46: You can manually leave this out of the revert since its correct.

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -4882,9 +4881,10 @@ void writeRPCStringUnicode(String sName, String sValue, boolean bOut, SQLCollati
          * Use PLP encoding if either OUT params were specified or if the user query exceeds
          * DataTypes.SHORT_VARTYPE_MAX_BYTES
          */
-        if ((nValueLen > DataTypes.SHORT_VARTYPE_MAX_BYTES || bOut) && !isNonPLP) {
+        if (nValueLen > DataTypes.SHORT_VARTYPE_MAX_BYTES || bOut) {
             writeRPCNameValType(sName, bOut, TDSType.NVARCHAR);
 
+            // Handle Yukon v*max type header here.
```
Jeffery-Wasty, line 4887: As well with all these Yukon comments, can leave them out too.

src/main/java/com/microsoft/sqlserver/jdbc/Parameter.java
```diff
@@ -445,14 +397,10 @@ boolean isValueGotten() {
...
-        if (null == getterDTV) {
+        if (null == getterDTV)
             getterDTV = new DTV();
-        }
-
-        if (null != tdsReader) {
-            deriveTypeInfo(tdsReader);
-        }
 
+        deriveTypeInfo(tdsReader);
```
lilgreenbird, line 403: why remove the null check?

tkyc: TdsReader should not be null. Before the revert, it could potentially be null because we may or may not have the value type info depending on how the parameters were set. eg. either set by name or by index for the cstmt.

lilgreenbird (approved):

Jeffery-Wasty (dismissed):

Jeffery-Wasty (approved):

lilgreenbird (approved):

b0fh042:
will this ever be brought back at some point? the way the old 'wrapper' works makes it impossible to return (n)varchar(max) from a stored procedure, which is very limiting... - thanks.

PR #2513, opened 2024-09-18
Fix retries when `connectRetryCount` is set to 1 or greater

Jeffery-Wasty:
Fixes the following issues:

* We attempt one **too many** connection retries. 
    * This is resolved by the change in SQLServerConnection::1997
* We wait between the initial connection attempt, and the first retry attempt, when we shouldn't. 
    * This is resolved by SQLServerConnection::2030 - 2031

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
@@ -2026,6 +2026,7 @@ public class SQLServerConnection implements ISQLServerConnection, java.io.Serial
                                     + sqlServerError.getErrorNumber() + ". Wait for connectRetryInterval("
                                     + connectRetryInterval + ")s before retry.");
                         }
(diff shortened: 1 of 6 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -132,7 +132,7 @@ public class SQLServerConnection implements ISQLServerConnection, java.io.Serial
     /**
      * Back off interval in ms for retries
      */
-    static final int BACKOFF_INTERVAL = 100;
+    static final int BACKOFF_INTERVAL = 500;
```
lilgreenbird, line 135: this shouldn't exist we should be using the connectRetryInterval from connection string

barryw-mssql (approved):

tkyc (approved):

lilgreenbird (approved):

Issue #2530, opened 2024-10-22
[QUESTION] Active Directory (Authentication=ActiveDirectoryServicePrincipalCertificate). Cannot parse the PVK, PVK file does not contain the correct header..does it not support encryped keys?


muskaan62:
## Question
**ActiveDirectoryServicePrincipalCertificate Authentication with client certifiicate private key and private key associated with password is not working.**

```
String jdbcUrl = "jdbc:sqlserver://******;" +
                "database=*******;" +
                "encrypt=true;"+
                "trustServerCertificate=false;" +
                "authentication=ActiveDirectoryServicePrincipalCertificate;"+
                "user=*****;"+
                "clientCertificate= ******;"+
                "clientKey=******;"+
                "clientKeyPassword=*****";
```

I am trying to connect with azuresql db with the above url using (clientcert,private key,password) for authentication
Note(clientcert with private key is working)
client cert with password also working.
only the above combination is giving the below error..I have verified the private key as well it contains valid header

`-----BEGIN ENCRYPTED PRIVATE KEY-----
`

```
Exception in thread "main" com.microsoft.sqlserver.jdbc.SQLServerException: Failed to authenticate the user xxxxxxxx in Active Directory (Authentication=ActiveDirectoryServicePrincipalCertificate). Cannot parse the PVK, PVK file does not contain the correct header.
	at com.microsoft.sqlserver.jdbc.SQLServerMSAL4JUtils.getCorrectedException(SQLServerMSAL4JUtils.java:488)
	at com.microsoft.sqlserver.jdbc.SQLServerMSAL4JUtils.getSqlFedAuthTokenPrincipalCertificate(SQLServerMSAL4JUtils.java:319)
	at com.microsoft.sqlserver.jdbc.SQLServerConnection.getFedAuthToken(SQLServerConnection.java:6089)
	at com.microsoft.sqlserver.jdbc.SQLServerConnection.onFedAuthInfo(SQLServerConnection.java:6012)
	at com.microsoft.sqlserver.jdbc.SQLServerConnection.processFedAuthInfo(SQLServerConnection.java:5846)
	at com.microsoft.sqlserver.jdbc.TDSTokenHandler.onFedAuthInfo(tdsparser.java:346)
	at com.microsoft.sqlserver.jdbc.TDSParser.parse(tdsparser.java:130)
	at com.microsoft.sqlserver.jdbc.TDSParser.parse(tdsparser.java:42)
[...]

PR #2532, opened 2024-11-01
fix issue-2530 encrypted private keys is not working issue

muskaan62:
client certificate without password (the test you provide for this scenario) [working]
client certificate with password [working]
client certificate with private key [working]
**client certificate with private key and key password (i.e means encrypted key) [bug]**
The above 3 ways are working fine but fourth is not ..
The logic for decrypt the private key is restricted to pvk though providing pvk key also its not working types and RSA private keys when decrypting also failing so added a fix for this. related to [https://github.com/microsoft/mssql-jdbc/issues/2530](url)

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCertificateUtils.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCertificateUtils.java
@@ -57,6 +57,11 @@ import org.bouncycastle.openssl.PEMKeyPair;
 import org.bouncycastle.openssl.PEMParser;
 import org.bouncycastle.openssl.jcajce.JcaPEMKeyConverter;
 import org.bouncycastle.openssl.jcajce.JcePEMDecryptorProviderBuilder;
+import org.bouncycastle.openssl.jcajce.JceOpenSSLPKCS8DecryptorProviderBuilder;
+import org.bouncycastle.operator.InputDecryptorProvider;
+import org.bouncycastle.operator.OperatorCreationException;
+import org.bouncycastle.pkcs.PKCS8EncryptedPrivateKeyInfo;
+import org.bouncycastle.pkcs.PKCSException;
 
 
 final class SQLServerCertificateUtils {
@@ -318,6 +323,8 @@ final class SQLServerCertificateUtils {
     private static final String CLIENT_KEY = "client-key";
     // PKCS#1 format
     private static final String PEM_RSA_PRIVATE_START = "-----BEGIN RSA PRIVATE KEY-----";
+    private static final String PEM_RSA_ENCRYPTED_PRIVATE_START =
+            "-----BEGIN ENCRYPTED PRIVATE KEY-----";
     // PVK format
     private static final long PVK_MAGIC = 0xB0B5F11EL;
     private static final byte[] RSA2_MAGIC = {82, 83, 65, 50};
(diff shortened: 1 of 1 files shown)


muskaan62:
@lilgreenbird FYI

lilgreenbird:
thanks the team will take a look

muskaan62:
Thanks @lilgreenbird can we expect this fix in 12.8.1 version or it will be available with new version of driver?

tkyc (approved):

lilgreenbird:
@muskaan62 please see above you will need to agree to CLA before the PR can be considered and tests run on it

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCertificateUtils.java
```diff
@@ -57,6 +57,11 @@
 import org.bouncycastle.openssl.PEMParser;
 import org.bouncycastle.openssl.jcajce.JcaPEMKeyConverter;
 import org.bouncycastle.openssl.jcajce.JcePEMDecryptorProviderBuilder;
+import org.bouncycastle.openssl.jcajce.JceOpenSSLPKCS8DecryptorProviderBuilder;
+import org.bouncycastle.operator.InputDecryptorProvider;
+import org.bouncycastle.operator.OperatorCreationException;
+import org.bouncycastle.pkcs.PKCS8EncryptedPrivateKeyInfo;
+import org.bouncycastle.pkcs.PKCSException;
```
David-Engel, line 64: You'll need to fix the issue without using Bouncy Castle. Try to use java.security instead. We want to limit Bouncy Castle use to Java 8 and drop it when Java 8 support ends.

muskaan62:
@microsoft-github-policy-service agree company="Gaian Solutions"

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCertificateUtils.java
muskaan62: @David-Engel Thanks..The existing code is already using bouncy castle for private keys, i've just added one more if condition for encrypted keys.

David-Engel: I understand that. But the existing code is only used within the context of the Always Encrypted with secure enclaves feature with Java 8 only, (https://github.com/microsoft/mssql-jdbc/blob/main/README.md#dependencies)

We can't extend the use of Bouncy Castle for other contexts/features without approval from our security team, as they didn't like the use of an external encryption library at all. We only got approval because it would be dropped when support for Java 8 ends.

David-Engel: To clarify, I meant: If you have to limit this fix to Java 11+ only **_without using Bouncy Castle_**, I'm fine with that. We can't extend our usage of Bouncy Castle to Java 11.

muskaan62: Thanks @David-Engel 
If i Understood correctly, then you are saying to fix this issue for java 17+ and the issue will still exist for java 8 and java 11??
We are using Java 11 in our applications and require a fix that works specifically for this version. I attempted to address the issue using the Java Security API, but encountered errors related to the provider. Upon investigation, it appears these errors stem from a known bug in JDK 11 ([JDK-8231581](https://bugs.openjdk.org/browse/JDK-8231581)) that remains unresolved.
we need a solution for Java 11. Most suggestions I found online recommend using Bouncy Castle. If you have any references or approaches to resolve this for Java 11, I would be glad to explore them. Thank you for your help!

lilgreenbird: Bouncy Castle has security issues it is not recommended to be used and is not approved by Microsoft. We are looking into a solution that does not require Bouncy Castle.

Syed-SnapLogic: @lilgreenbird , thanks for the reply.

Can you pls help me more details around the security issues being present in the bouncy castle library?  That will help me to see if I also need to have BC in my app or not.  The more details around it would be appreciated.

j2-z:
I'm also encountering the same issue regarding to the encrypted private key when using `authentication=ActiveDirectoryServicePrincipalCertificate` with mssql-jdbc 12.8.1.jre8.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCertificateUtils.java
Syed-SnapLogic: Any response here @lilgreenbird or @David-Engel ?

David-Engel: > Thanks @David-Engel If i Understood correctly, then you are saying to fix this issue for java 17+ and the issue will still exist for java 8 and java 11?? We are using Java 11 in our applications and require a fix that works specifically for this version. I attempted to address the issue using the Java Security API, but encountered errors related to the provider. Upon investigation, it appears these errors stem from a known bug in JDK 11 ([JDK-8231581](https://bugs.openjdk.org/browse/JDK-8231581)) that remains unresolved. we need a solution for Java 11. Most suggestions I found online recommend using Bouncy Castle. If you have any references or approaches to resolve this for Java 11, I would be glad to explore them. Thank you for your help!

@muskaan62 I'm just saying we can't expand the scenarios under which we use BC. If this means the fix is limited, so be it.

muskaan62:
The PR is still open, for now i have the workaround from my end by using keystore to put the certificate key and when getting the key back from keystore iam getting unencrypted key.

PR #2539, opened 2024-11-14
Deprecation of Security Manager fix

tkyc:
Kerberos auth depends on calling `Subject.getSubject()`, which is deprecated since JDK 17. Fall back to acquiring the subject through the replacement API when the deprecation exception is thrown.


Jeffery-Wasty (approved):

src/main/java/com/microsoft/sqlserver/jdbc/KerbAuthentication.java
```diff
@@ -94,8 +96,22 @@ private void initAuthInit() throws SQLServerException {
...
                 try {
-                    java.security.AccessControlContext context = java.security.AccessController.getContext();
-                    currentSubject = Subject.getSubject(context);
+
+                    try {
+                        java.security.AccessControlContext context = java.security.AccessController.getContext();
+                        currentSubject = Subject.getSubject(context);
+
+                    } catch (UnsupportedOperationException ue) {
+                        if (authLogger.isLoggable(Level.FINE)) {
```
machavan, line 105: Would it be needed ( or good to have) to also check the exact exception message as part of the UnsupportedOperationException ?

tkyc: https://bugs.java.com/bugdatabase/view_bug?bug_id=8328643
>When the Security Manager is removed in a future release, the `Subject.getSubject(AccessControlContext)` method will be degraded further to throw `UnsupportedOperationException` unconditionally

Should be ok, since when an UnsupportedOperationException is thrown, it means security manager deprecation.

lilgreenbird (approved):

PR #2555, opened 2024-12-04
Incident#567732673 - Add provision to set SQLServerBulkCopy options in PreparedStatement

machavan:
Description:

Add provision to set SQLServerBulkCopyOptions in PreparedStatement via 
-new connection string options
-a new method setBulkCopyOptions

Testing:
- All tests pass
- Added a new test BatchExecutionWithBCOptionsTest
- TBD: Currently new tests have been added only for CheckConstraint option. Need to add tests for all the other newly addded options.


muskan124947:
Added tests for all the options in BatchExecutionWithBCOptionsTest

tkyc (approved):

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopyOptions.java
```diff
@@ -103,6 +103,22 @@ public SQLServerBulkCopyOptions() {
         useInternalTransaction = false;
         allowEncryptedValueModifications = false;
     }
+    
+    /**
+     * Constructs a SQLServerBulkCopySettings class using defaults from given connection
+     */
+    SQLServerBulkCopyOptions(SQLServerConnection conn) {
```
divang, line 110: Is there a particular reason for using a package-level modifier? The default constructor is public.

machavan: Yes, this one is only for internal consumption. 

For applications, the way to set SQLServerBulkCopyOptions  is to set on a given PreparedStatement using SQLServerPreparedStatement::setBulkCopyOptions API.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResource.java
```diff
@@ -466,6 +466,15 @@ protected Object[][] getContents() {
...
+        {"R_bulkCopyOptionDefaultsTimeoutPropertyDescription", "The default timeout for bulk copy operations"},
+        {"R_bulkCopyOptionDefaultsCheckConstraintsPropertyDescription", "Whether to check constraints during bulk copy operations"},
+        {"R_bulkCopyOptionDefaultsFireTriggersPropertyDescription", "Whether to fire triggers during bulk copy operations"},
+        {"R_bulkCopyOptionDefaultsKeepIdentityPropertyDescription", "Whether to keep identity values during bulk copy operations"},
+        {"R_bulkCopyOptionDefaultsKeepNullsPropertyDescription", "Whether to keep null values during bulk copy operations"},
+        {"R_bulkCopyOptionDefaultsTableLockPropertyDescription", "Whether to use table lock during bulk copy operations"},
+        {"R_bulkCopyOptionDefaultsUseInternalTransactionPropertyDescription", "Whether to use internal transaction during bulk copy operations"},
+        {"R_bulkCopyOptionDefaultsAllowEncryptedValueModificationsPropertyDescription", "Whether to allow encrypted value modifications during bulk copy operations"},        
         {"R_UnknownDataClsTokenNumber", "Unknown token for Data Classification."}, // From Server
```
divang, line 478: can we add '.' in all message contents.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDataSource.java
```diff
@@ -1025,6 +1025,122 @@ public boolean getUseBulkCopyForBatchInsert() {
                 SQLServerDriverBooleanProperty.USE_BULK_COPY_FOR_BATCH_INSERT.getDefaultValue());
     }
 
+    @Override
+    public void setBulkCopyOptionDefaultsBatchSize(int bulkCopyOptionDefaultsBatchSize) {
```
divang, line 1029: Could the method name be setBulkCopyOptionBatchSize instead of setBulkCopyOptionDefaultsBatchSize?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDriver.java
```diff
@@ -644,7 +644,9 @@ enum SQLServerDriverIntProperty {
     STATEMENT_POOLING_CACHE_SIZE("statementPoolingCacheSize", SQLServerConnection.DEFAULT_STATEMENT_POOLING_CACHE_SIZE),
     CANCEL_QUERY_TIMEOUT("cancelQueryTimeout", -1),
     CONNECT_RETRY_COUNT("connectRetryCount", 1, 0, 255),
-    CONNECT_RETRY_INTERVAL("connectRetryInterval", 10, 1, 60);
+    CONNECT_RETRY_INTERVAL("connectRetryInterval", 10, 1, 60),
+    BULK_COPY_OPTION_DEFAULTS_BATCH_SIZE("bulkCopyOptionDefaultsBatchSize", 0),
```
divang, line 648: Could the method name be BULK_COPY_OPTION_BATCH_SIZE/BULK_COPY_BATCH_SIZE instead of BULK_COPY_OPTION_DEFAULTS_BATCH_SIZE? (If it is fine then we need to use same pattern in all methods and constants)

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResource.java
machavan: done

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDataSource.java
machavan: The Default in the name indicate that we are setting connection level defaults for these options which can be overriden on statement basis

src/test/java/com/microsoft/sqlserver/jdbc/preparedStatement/BatchExecutionWithBCOptionsTest.java
```diff
@@ -0,0 +1,913 @@
...
+                }
+            }
+        } catch (SQLException e) {
+            fail(TestResource.getResource("R_unexpectedException") + e.getMessage());
+        }
+    }
+
+    /**
+     * Test with useBulkCopyBatchInsert=true and
+     * bulkCopyOptionDefaultsTableLock=true
```
divang, line 427: Could you also add the false case scenario for bulkCopyOptionDefaultsTableLock=false?

src/test/java/com/microsoft/sqlserver/jdbc/preparedStatement/BatchExecutionWithBCOptionsTest.java
```diff
@@ -0,0 +1,913 @@
...
+                }
+            }
+        } catch (SQLException e) {
+            fail(TestResource.getResource("R_unexpectedException") + e.getMessage());
+        }
+    }
+
+    /**
+     * Test with useBulkCopyBatchInsert=true and
+     * bulkCopyOptionDefaultsKeepIdentity=true
```
divang, line 210: Could we also add a scenario for bulkCopyOptionDefaultsKeepIdentity=false?

src/test/java/com/microsoft/sqlserver/jdbc/preparedStatement/BatchExecutionWithBCOptionsTest.java
```diff
@@ -0,0 +1,913 @@
...
+                }
+            }
+        } catch (SQLException e) {
+            fail(TestResource.getResource("R_unexpectedException") + e.getMessage());
+        }
+    }
+
+    /**
+     * Test with useBulkCopyBatchInsert=true and
+     * bulkCopyOptionDefaultsUseInternalTransaction=true
```
divang, line 796: Could we also add scenario for bulkCopyOptionDefaultsUseInternalTransaction=false?

src/test/java/com/microsoft/sqlserver/jdbc/preparedStatement/BatchExecutionWithBCOptionsTest.java
```diff
@@ -0,0 +1,913 @@
...
+        } catch (SQLException e) {
+            if (!e.getMessage().contains("UseInternalTransaction option cannot be set to TRUE")) {
+                fail(TestResource.getResource("R_unexpectedException") + e.getMessage());
+            }
+        }
+    }
+
+    /**
+     * Test with useBulkCopyBatchInsert=true and
+     * bulkCopyOptionDefaultsAllowEncryptedValueModifications=true
```
divang, line 841: Could we also add scenario for bulkCopyOptionDefaultsAllowEncryptedValueModifications=false?

src/test/java/com/microsoft/sqlserver/jdbc/preparedStatement/BatchExecutionWithBCOptionsTest.java
muskan124947: As discussed, added a scenario where options are set to true and insert fails

src/test/java/com/microsoft/sqlserver/jdbc/preparedStatement/BatchExecutionWithBCOptionsTest.java
muskan124947: To test failure scenario, added a test case where bulkCopyOptionDefaultsTableLock=true and insert fails

src/test/java/com/microsoft/sqlserver/jdbc/preparedStatement/BatchExecutionWithBCOptionsTest.java
```diff
@@ -0,0 +1,913 @@
...
+    }
+
+    /**
+     * Test with useBulkCopyBatchInsert=true and
+     * bulkCopyOptionDefaultsFireTriggers=true where insert fails
+     *
+     * @throws SQLException
+     */
+    @Test
+    public void testBulkCopyOptionDefaultsFireTriggersFailure() throws Exception {
```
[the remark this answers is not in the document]
muskan124947: It will work same as default way
For failure scenario, added test case where bulkCopyOptionDefaultsFireTriggers=true where insert fails

src/test/java/com/microsoft/sqlserver/jdbc/preparedStatement/BatchExecutionWithBCOptionsTest.java
muskan124947: Added test case to ensure that the existing functionality remains intact and is not affected.

src/test/java/com/microsoft/sqlserver/jdbc/preparedStatement/BatchExecutionWithBCOptionsTest.java
muskan124947: test case to ensure that the existing functionality remains intact and is not affected.

divang (dismissed):

divang (dismissed):

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResource.java
```diff
@@ -465,7 +465,16 @@ protected Object[][] getContents() {
...
+        {"R_bulkCopyOptionDefaultsTimeoutPropertyDescription", "The default timeout for bulk copy operations."},
+        {"R_bulkCopyOptionDefaultsCheckConstraintsPropertyDescription", "Determines whether to check constraints during bulk copy operations."},
+        {"R_bulkCopyOptionDefaultsFireTriggersPropertyDescription", "Determines whether to fire triggers during bulk copy operations."},
+        {"R_bulkCopyOptionDefaultsKeepIdentityPropertyDescription", "Determines whether to keep identity values during bulk copy operations."},
+        {"R_bulkCopyOptionDefaultsKeepNullsPropertyDescription", "Determines whether to keep null values during bulk copy operations."},
+        {"R_bulkCopyOptionDefaultsTableLockPropertyDescription", "Determines whether to use table lock during bulk copy operations."},
+        {"R_bulkCopyOptionDefaultsUseInternalTransactionPropertyDescription", "Determines whether to use internal transaction during bulk copy operations."},
+        {"R_bulkCopyOptionDefaultsAllowEncryptedValueModificationsPropertyDescription", "Determines whether to allow encrypted value modifications during bulk copy operations."},
```
David-Engel, line 477: The descriptions here make me want to ensure we don't confuse users on what these settings apply to. As it is, it sounds to me like they would apply to the defaults when using the bulk copy APIs directly. E.g. Instantiating a new SQLServerBulkCopyOptions object outside of the context of a connection or not even passing options to a SQLServerBulkCopy object. See the many bulk copy API examples in the docs: https://learn.microsoft.com/en-us/sql/connect/jdbc/using-bulk-copy-with-the-jdbc-driver?view=sql-server-ver16#performing-a-dedicated-bulk-copy-operation-in-a-transaction for what I'm talking about where these new settings **_don't_** apply

I want to make sure it is clear that these options only apply within the scope of the useBulkCopyForBatchInsert option for PreparedStatements. (And without making the option name too long.)

I'm thinking along the lines of: bulkCopyForBatchInsertXXX
- bulkCopyForBatchInsertBatchSize
- bulkCopyForBatchInsertTimeout
- bulkCopyForBatchInsertFireTriggers
- etc...

Then these descriptions would become:

- The default batch size for bulk copy operations created from batch insert operations.
- The default timeout for bulk copy operations created from batch insert operations.
- Determines whether to check constraints during bulk copy operations created from batch insert operations.
- etc...

@lilgreenbird @tkyc - Let me know if these make more sense to you.

machavan: My understanding was the `default` in the name conveyed that these are connection defaults while one could use this connection for individual bulk copy operations with different option values.

lilgreenbird: yes agree this makes more sense, the "default" word is unnecessary and seems a bit misleading

David-Engel: > My understanding was the `default` in the name conveyed that these are connection defaults while one could use this connection for individual bulk copy operations with different option values.

Functionally, what does that look like? Meaning, given a set of code like the examples in the docs, what would the behavior change be?

The original request for this feature was to be able to control the bulk copy options used within the context of useBulkCopyForBatchInsert. Given that SQLServerBulkCopyOptions live outside of the context of a connection, I'm not sure it makes sense to try to apply the new options to contexts outside of useBulkCopyForBatchInsert. It might just cause confusion.

machavan: Agree, will make the changes.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -2224,7 +2224,7 @@ public int[] executeBatch() throws SQLServerException, BatchUpdateException, SQL
 
                         if (null == bcOperation) {
                             bcOperation = new SQLServerBulkCopy(connection);
-                            SQLServerBulkCopyOptions option = new SQLServerBulkCopyOptions();
+                            SQLServerBulkCopyOptions option = new SQLServerBulkCopyOptions(connection);
                             option.setBulkCopyTimeout(queryTimeout);
```
machavan, line 2228: I think since we have the timeout already configurable and coming from connection we can remove this

`option.setBulkCopyTimeout(queryTimeout)` from here?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -855,6 +855,224 @@ public void setUseBulkCopyForBatchInsert(boolean useBulkCopyForBatchInsert) {
...
+     *        the bulkCopyOptionDefaultsBatchSize value to set.
+     */
+    public void setBulkCopyOptionDefaultsBatchSize(int bulkCopyOptionDefaultsBatchSize) {
+        this.bulkCopyOptionDefaultsBatchSize = bulkCopyOptionDefaultsBatchSize;
+    }
+
+    /** 
+     * The default timeout for bulk copy operations.
+     */
+    private int bulkCopyOptionDefaultsTimeout = 0;
```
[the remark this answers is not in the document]
machavan: done

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
David-Engel: Oh, I missed this detail. We should definitely respect queryTimeout in this context. In fact, given queryTimeout, We don't need bulkCopyForBatchInsertTimeout as they should always be the same.

This prompted me to think more on each new option. We can also leave out bulkCopyForBatchInsertUseInternalTransaction. It doesn't make sense in the context of a PreparedStatement, which can its own transaction logic around it.

muskan124947:
ADO pipeline -> https://sqlclientdrivers.visualstudio.com/mssql-jdbc/_build/results?buildId=105965&view=results

tkyc (approved):

Jeffery-Wasty (approved):

David-Engel (approved):

muskan124947:
ADO pipeline is green post merging this changeset
https://sqlclientdrivers.visualstudio.com/mssql-jdbc/_build/results?buildId=106929&view=results

PR #2558, opened 2024-12-06
JSON datatype support

divang:
### Description

This pull request introduces support for the JSON datatype in the Microsoft JDBC Driver for SQL Server. With these changes, applications using the driver can now read from and write to SQL Server columns of type JSON, enabling seamless integration with semi-structured data.

1. Insertion of JSON Data into Table
```
@Test
public void insertJsonData() throws SQLException {
    String insertSql = "INSERT INTO " + AbstractSQLGenerator.escapeIdentifier(tableName) + " (data) VALUES (?)";
    String json = "{\"name\":\"John\",\"skills\":[\"Java\",\"SQL\"]}";
    try (PreparedStatement pstmt = connection.prepareStatement(insertSql)) {
        pstmt.setString(1, json);
        pstmt.executeUpdate();
    }
}
```
2. Select JSON Data from Table
```
@Test
public void getJsonData() throws SQLException {
    String query = "SELECT data FROM " + AbstractSQLGenerator.escapeIdentifier(tableName);
    try (PreparedStatement stmt = connection.prepareStatement(query);
         ResultSet rs = stmt.executeQuery()) {
        while (rs.next()) {
            String json = rs.getString("data");
            System.out.println("JSON: " + json);
        }
    }
}
```
3. Stored Procedure Call with JSON IN/OUT
```
@Test
public void testJsonStoredProcedureInputOutput() throws SQLException {
    createProcedure();
    try (CallableStatement cstmt = connection.prepareCall(call)) {
        String inputJson = "{\"key\":\"value\"}";
        cstmt.setString(1, inputJson);
        cstmt.registerOutParameter(2, microsoft.sql.Types.JSON);

        cstmt.execute();
        String outputJson = cstmt.getString(2);
        System.out.println("Output JSON: " + outputJson);
    }
}
```
4. Bulk Copy from CSV to Table with JSON Column
```
@Test
public void testBulkCopyJsonFromCSV() throws Exception {
    String dstTable = RandomUtil.getIdentifier("BulkJsonTest");
    String fileName = filePath + jsonInputFile;
         SQLServerBulkCSVFileRecord fileRecord = new SQLServerBulkCSVFileRecord(fileName, encoding, ",", false)) {
        stmt.executeUpdate("CREATE TABLE " + dstTable + " (c1 BIT, c2 nchar(50), c3 JSON)");
        fileRecord.addColumnMetadata(1, "c1", java.sql.Types.BIT, 0, 0);
        fileRecord.addColumnMetadata(2, "c2", java.sql.Types.NCHAR, 10, 0);
        fileRecord.addColumnMetadata(3, "c3", microsoft.sql.Types.JSON, 0, 0);
    }
}
```

#### Key Changes

- **Type System and Metadata**
  - Added `JSON` to the type system (enums such as `TDSType`, `SSType`, `JDBCType`, and their categories).
  - Updated metadata handling to detect, map, and propagate JSON types correctly.

- **Connection and Feature Negotiation**
  - Implemented feature negotiation for JSON support in `SQLServerConnection`, including versioning and server capability detection.
  - Extended the TDS protocol handling and connection handshake to advertise and recognize JSON support.

- **Bulk Copy and Data Table Support**
  - Modified `SQLServerBulkCopy`, `SQLServerBulkCSVFileRecord`, and `SQLServerDataTable` to handle JSON columns in bulk operations.
  - Ensured proper streaming, encoding, and null handling for JSON columns during bulk inserts and exports.

- **Parameter and Value Handling**
  - Updated parameter setting, reading, and writing logic to process JSON values seamlessly.
  - Ensured correct behavior for both application-supplied and database-returned JSON data.

- **Test and Build Integration**
  - Added `JSONTest` exclusion tags in `build.gradle` and `pom.xml` for targeted testing.
  - Enabled grouping and exclusion of JSON-specific tests for various build profiles.

#### Motivation

The addition of JSON datatype support allows Java applications to leverage SQL Server’s native JSON features directly through the JDBC driver. This enhancement addresses modern data requirements by enabling efficient storage, retrieval, and manipulation of semi-structured JSON data within relational workflows.

#### Testing

- Comprehensive unit and integration tests were added for JSON columns.
- Tests ensure correct data round-trip, null handling, and edge-case coverage.

---

For details on all code changes, see the [full diff](https://github.com/microsoft/mssql-jdbc/pull/2558/files).


src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/RegressionTest.java
```diff
@@ -238,6 +238,30 @@ public void testXmlQuery() throws SQLException {
...
+    /**
+     * Tests Json query
+     * 
+     * @throws SQLException
+     */
+    @Test
+    public void testJsonQuery() throws SQLException {
+        try (Connection connection = getConnection(); Statement stmt = connection.createStatement()) {
+            tableName = RandomUtil.getIdentifier("try_SQLJSON_Table");
+            TestUtils.dropTableIfExists(AbstractSQLGenerator.escapeIdentifier(tableName), stmt);
```
machavan, line 250: We can consider enhancing this test to insert a couple of records, update the first record with a where predicate on JSON field itself, delete the second record (with a predicate), read the records back and validate that its 1 record with expected value of the json column.

src/test/java/com/microsoft/sqlserver/jdbc/callablestatement/CallableStatementTest.java
```diff
@@ -40,577 +40,572 @@
...
-			.escapeIdentifier(RandomUtil.getIdentifier("manyParam_Procedure"));
-	private static String currentTimeProc = AbstractSQLGenerator
-			.escapeIdentifier(RandomUtil.getIdentifier("currentTime_Procedure"));
-	private static String manyParamUserDefinedType = AbstractSQLGenerator
-			.escapeIdentifier(RandomUtil.getIdentifier("manyParam_definedType"));
-	private static String zeroParamSproc = AbstractSQLGenerator
-			.escapeIdentifier(RandomUtil.getIdentifier("zeroParamSproc"));
-	private static String tableNameJSON = "TestJSONTable";
+    private static String tableNameGUID = AbstractSQLGenerator
+            .escapeIdentifier(RandomUtil.getIdentifier("uniqueidentifier_Table"));
```
divang, line 51: Please maintain the current code formatting.

src/main/java/microsoft/sql/Types.java
```diff
@@ -74,4 +74,10 @@ private Types() {
      * Microsoft SQL type GEOGRAPHY.
      */
     public static final int GEOGRAPHY = -158;
+    
+    /**
+     * The constant in the Java programming language, sometimes referred to as a type code, that identifies the
+     * Microsoft SQL type JSON.
+     */
+    public static final int JSON = -159;
```
lilgreenbird, line 82: just wondering, where does this come from? I looked into this before but couldnt find any reference

src/main/java/com/microsoft/sqlserver/jdbc/dtv.java
```diff
@@ -294,7 +295,10 @@ final class SendByRPCOp extends DTVExecuteOp {
         void execute(DTV dtv, String strValue) throws SQLServerException {
             if (dtv.getJdbcType() == JDBCType.GUID) {
                 tdsWriter.writeRPCUUID(name, UUID.fromString(strValue), isOutParam);
-            } else {
+            } else if (dtv.getJdbcType() == JDBCType.JSON) {
+                tdsWriter.writeRPCJson(name, strValue, isOutParam);
+            }
```
lilgreenbird, line 300: coding standard - pls run formatter

src/main/java/com/microsoft/sqlserver/jdbc/dtv.java
```diff
@@ -115,6 +115,7 @@ abstract class DTVExecuteOp {
     abstract void execute(DTV dtv, TVP tvpValue) throws SQLServerException;
 
     abstract void execute(DTV dtv, SqlVariant sqlVariantValue) throws SQLServerException;
+
```
lilgreenbird, line 118: nitpick - this isn't necessary

lilgreenbird (changes requested):
need to add tests for datatype conversions. See PR for sql_variant for example

src/main/java/microsoft/sql/Types.java
divang: This class (ie. microsoft.sql.Types.java) defines constants that are used to identify SQL types specific to Microsoft SQL Server, which are not defined in java.sql.Types. The value -159 is simply the next sequential number in our microsoft.sql.Types.java.
 
public class MicrosoftSqlTypes {

    /**
     * The constant in the Java programming language, sometimes referred to as a type code, that identifies the
     * Microsoft SQL type GEOGRAPHY.
     */
    public static final int GEOGRAPHY = **-158;**

    /**
     * The constant in the Java programming language, sometimes referred to as a type code, that identifies the
     * Microsoft SQL type JSON.
     */
    public static final int JSON = **-159;**
}

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyISQLServerBulkRecordTest.java
```diff
@@ -150,6 +150,37 @@ public void testBulkCopyDateTimePrecision() throws SQLException {
         }
     }
 
+    @Test
+    public void testBulkCopyJSON() throws SQLException {
```
tkyc, line 154: Before this merges, just a heads up and reminder that both internal and public pipelines will need the new SQL Server version that supports JSON.

David-Engel (approved):

src/main/java/com/microsoft/sqlserver/jdbc/DataTypes.java
```diff
@@ -970,7 +985,7 @@ boolean isBinary() {
      * @return true if the JDBC type is textual
      */
     private final static EnumSet<Category> textualCategories = EnumSet.of(Category.CHARACTER, Category.LONG_CHARACTER,
-            Category.CLOB, Category.NCHARACTER, Category.LONG_NCHARACTER, Category.NCLOB);
+    Category.CLOB, Category.NCHARACTER, Category.LONG_NCHARACTER, Category.NCLOB, Category.JSON); //FIXME: JSON is textual?
```
saurabh500, line 988: you have a //fixme comment. Was this resolved? how does textualCategories impact APIs ?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerParameterMetaData.java
```diff
@@ -166,6 +166,8 @@ private void parseQueryMeta(ResultSet rsQueryMeta) throws SQLServerException {
                             qm.precision = 8;
                         } else if (SSType.XML == ssType) {
                             qm.precision = SQLServerDatabaseMetaData.MAXLOBSIZE / 2;
+                        }  else if (SSType.JSON == ssType) {
+                            qm.precision = SQLServerDatabaseMetaData.MAXLOBSIZE / 2;
```
saurabh500, line 170: What is the meaning of `precision` in JSON's context ?

David-Engel (dismissed):

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerParameterMetaData.java
divang: Actually, we followed the XML datatype code path, so added it, but it is not required. Tested the code.

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyISQLServerBulkRecordTest.java
divang: As discussed with the .Net driver team too, we need to tag all JSON unit test cases and skip them in our pipelines until SQL Server with JSON support is available. If required, we can manually run those JSON JUnit test cases.

src/test/java/com/microsoft/sqlserver/jdbc/tvp/TVPTypesTest.java
```diff
@@ -149,6 +149,36 @@ public void testXML() throws SQLException {
...
+                "INSERT INTO " + AbstractSQLGenerator.escapeIdentifier(tableName) + " select * from ? ;")) {
+            pstmt.setStructured(1, tvpName, tvp);
+
+            pstmt.execute();
+
+            try (Connection con = getConnection(); Statement stmt = con.createStatement();
+                    ResultSet rs = stmt.executeQuery(
+                            "select c1 from " + AbstractSQLGenerator.escapeIdentifier(tableName) + " ORDER BY rowId")) {
+                while (rs.next())
+                    assertEquals(rs.getString(1), value);
```
machavan, line 177: have a test that does getString("<columnName>") instead of index

src/test/java/com/microsoft/sqlserver/jdbc/tvp/TVPTypesTest.java
```diff
@@ -349,6 +379,38 @@ public void testTVPXMLStoredProcedure() throws SQLException {
...
+
+        value = "{\"severity\":\"TRACE\",\"duration\":200,\"date\":\"2024-12-17T15:45:56\"}";
+
+        tvp = new SQLServerDataTable();
+        tvp.addColumnMetadata("c1",  microsoft.sql.Types.JSON);
+        tvp.addRow(value);
+
+        final String sql = "{call " + AbstractSQLGenerator.escapeIdentifier(procedureName) + "(?)}";
+
+        try (SQLServerCallableStatement callableStmt = (SQLServerCallableStatement) connection.prepareCall(sql)) {
```
machavan, line 401: do we support a TVP as out parameter, if yes we can have a test that selects a TVP that contains a row containing JSON fields.

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/RegressionTest.java
```diff
@@ -238,6 +238,78 @@ public void testXmlQuery() throws SQLException {
...
+                pstmt.setObject(3, "{\"key22\":\"value22\"}");
+                pstmt.executeUpdate();
+            }
+
+            sql = "insert into " + AbstractSQLGenerator.escapeIdentifier(tableName) + " values (?, ?,?)";
+                        try (SQLServerPreparedStatement pstmt = (SQLServerPreparedStatement) connection.prepareStatement(sql)) {
+                pstmt.setInt(1, pkRow3);
+                pstmt.setObject(2, "{\"key31\":\"value31\"}");
+                pstmt.setObject(3, "{\"key32\":\"value32\"}");
+                pstmt.executeUpdate();
```
machavan, line 278: have a mix of executeUpdate, addBatch +executeBatch, execute in various tests

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
```diff
@@ -997,6 +999,7 @@ private void writeTypeInfo(TDSWriter tdsWriter, int srcJdbcType, int srcScale, i
 
             case java.sql.Types.LONGVARCHAR:
             case java.sql.Types.VARCHAR: // 0xA7
+            case microsoft.sql.Types.JSON:
```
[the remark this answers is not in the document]
divang: Yes, in case of JSON type, unicodeConversionRequired() method returns false, so code flow does not fall into that.

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyISQLServerBulkRecordTest.java
```diff
@@ -150,6 +150,209 @@ public void testBulkCopyDateTimePrecision() throws SQLException {
...
+
+        try (Connection conn = DriverManager.getConnection(connectionString);) {
+            try (Statement dstStmt = conn.createStatement();
+                    SQLServerBulkCopy bulkCopy = new SQLServerBulkCopy(conn)) {
+
+                dstStmt.executeUpdate(
+                        "CREATE TABLE " + dstTable + " (testCol JSON);");
+
+                bulkCopy.setDestinationTableName(dstTable);
+                String data = "{\"key\":\"value\"}";
```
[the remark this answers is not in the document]
muskan124947: This scenario is already covered by testBulkCopyMultipleJsonRows, testBulkCopyMultipleJsonRowsAndColumns
Also we have a test in which json has multiple key:value pair inside it

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyISQLServerBulkRecordTest.java
```diff
@@ -150,6 +150,209 @@ public void testBulkCopyDateTimePrecision() throws SQLException {
...
+
+        try (Connection conn = DriverManager.getConnection(connectionString);) {
+            try (Statement dstStmt = conn.createStatement();
+                    SQLServerBulkCopy bulkCopy = new SQLServerBulkCopy(conn)) {
+
+                dstStmt.executeUpdate(
+                        "CREATE TABLE " + dstTable + " (testCol JSON);");
+
+                bulkCopy.setDestinationTableName(dstTable);
+                String data1 = "{\"key1\":\"value1\"}";
```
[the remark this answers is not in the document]
muskan124947: done

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyISQLServerBulkRecordTest.java
```diff
@@ -150,6 +150,209 @@ public void testBulkCopyDateTimePrecision() throws SQLException {
...
+        try (Connection conn = DriverManager.getConnection(connectionString);) {
+            try (Statement dstStmt = conn.createStatement();
+                    SQLServerBulkCopy bulkCopy = new SQLServerBulkCopy(conn)) {
+
+                dstStmt.executeUpdate(
+                        "CREATE TABLE " + dstTable + " (testCol JSON);");
+
+                bulkCopy.setDestinationTableName(dstTable);
+                String data1 = "{\"key1\":{\"nestedKey1\":\"nestedValue1\"}}";
+                String data2 = "{\"key2\":{\"nestedKey2\":\"nestedValue2\"}}";
```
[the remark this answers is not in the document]
muskan124947: done

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyCSVTest.java
```diff
@@ -93,13 +96,13 @@ public static void setUpConnection() throws Exception {
     @Test
     @DisplayName("Test SQLServerBulkCSVFileRecord")
     public void testCSV() {
-        String fileName = filePath + inputFile;
+        String fileName = filePath + jsonInputFile;
```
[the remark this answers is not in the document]
Ananya2: As discussed, we have added a new CSV with an additional JSON column without modifying the existing test cases.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCSVFileRecord.java
```diff
@@ -601,11 +605,19 @@ public void setEscapeColumnDelimitersCSV(boolean escapeDelimiters) {
         this.escapeDelimiters = escapeDelimiters;
     }
 
+    private static boolean isJson(String token) {
+        return token.startsWith("{") && token.endsWith("}");
```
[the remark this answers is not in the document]
Ananya2: Extra spaces are not an issue, as the parseString() method correctly handles delimiters and braces while preserving JSON structure. Also, we have added test cases to validate nested JSON and multiple key-value pairs within JSON.
https://github.com/microsoft/mssql-jdbc/pull/2558/commits/7e8b1affd217eb5e6dc51a40526ca5db0fb4098f

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/RegressionTest.java
divang: Added mix of those method calls.

src/test/java/com/microsoft/sqlserver/jdbc/databasemetadata/DatabaseMetaDataTest.java
```diff
@@ -1034,6 +1034,26 @@ public void shouldEscapeSchemaName() throws SQLException {
...
+                    + " (c1 JSON null);";
+            stmt.execute(sql);
+
+            String query = "SELECT * FROM " + AbstractSQLGenerator.escapeIdentifier(jsonTableName);
+            try (Statement statement = connection.createStatement();
+                    ResultSet resultSet = statement.executeQuery(query)) {
+
+                ResultSetMetaData metaData = resultSet.getMetaData();
+                String columnType = metaData.getColumnTypeName(1);
+                assertTrue("JSON".equalsIgnoreCase(columnType));
```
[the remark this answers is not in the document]
divang: Added more validation for metadata related.

src/test/java/com/microsoft/sqlserver/jdbc/tvp/TVPTypesTest.java
Ananya2: SQL Server does not support using TVPs as output parameters. TVPs are READONLY, and they cannot be modified or assigned inside the stored procedure.
https://learn.microsoft.com/en-us/dotnet/framework/data/adonet/sql/table-valued-parameters

src/test/java/com/microsoft/sqlserver/jdbc/tvp/TVPTypesTest.java
Ananya2: done

src/test/java/com/microsoft/sqlserver/jdbc/JSONFunctionTest.java
```diff
@@ -0,0 +1,971 @@
...
+import org.junit.jupiter.api.Tag;
+import org.junit.platform.runner.JUnitPlatform;
+import org.junit.runner.RunWith;
+
+import com.microsoft.sqlserver.testframework.AbstractSQLGenerator;
+import com.microsoft.sqlserver.testframework.AbstractTest;
+import com.microsoft.sqlserver.testframework.Constants;
+
+@RunWith(JUnitPlatform.class)
+@DisplayName("Test Json Functions")
```
muskan124947, line 30: 1. location of this test class
2. Tag xAzureSQLDW ?
3. OPENJSON() not supported by native JSON datatype
To verify (https://learn.microsoft.com/en-us/sql/t-sql/functions/openjson-transact-sql?view=sql-server-ver16)

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCSVFileRecord.java
```diff
@@ -639,24 +651,34 @@ private static String[] escapeQuotesRFC4180(String[] tokens) throws SQLServerExc
...
     private static String[] parseString(String buffer, String delimiter) {
         ArrayList<String> tokens = new ArrayList<>();
-        int position = 0;
+        StringBuilder currentToken = new StringBuilder();
         boolean quoted = false;
+        int braceCount = 0; // track nested JSON
 
         for (int i = 0; i < buffer.length(); i++) {
-            if (buffer.charAt(i) == doubleQuoteChar) {
+            char c = buffer.charAt(i);
```
David-Engel, line 659: This is a hot path when parsing data for bulk load. Effort was specifically made to avoid allocations here to maintain perf. Can you avoid using the StringBuilder and instead just reference substrings within the buffer?

Also, what if we are just parsing a string field that has braces in it? Like `Testing random } chars`. Or rows that have a field with braces but not double quotes?
`test,test,testing } testing,test`
`ResourceString1,Error on row {0},Description`
`ResourceString2,{0} does not exist,Description`

How should we detect that a field in CSV format is JSON? Thinking about it, I think we should expect JSON data to be escaped as string data within CSV, as the CSV spec doesn't have a provision for representing JSON. Example:
```
id,name,details
1,John Doe,"{""age"": 30, ""city"": ""New York""}"
2,Jane Smith,"{""age"": 25, ""city"": ""Los Angeles""}"
3,Bob Johnson,"{""age"": 40, ""city"": ""Chicago""}"
```

Note: You'll need to modify the new test CSV data to be CSV compliant, too.

src/test/java/com/microsoft/sqlserver/testframework/Constants.java
```diff
@@ -29,6 +29,7 @@ private Constants() {}
      * reqExternalSetup  - For tests requiring external setup
      * clientCertAuth  - - For tests requiring client certificate authentication setup
      * Fedauth - - - - - - For Fedauth tests
+     * JSONTest -  - - - - For tests requiring JSON setup
```
lilgreenbird, line 32: why do we need to add a separate tag for this? this should always run along with the other datatype tests

src/test/java/com/microsoft/sqlserver/testframework/AbstractTest.java
```diff
@@ -158,8 +158,7 @@ public static void setup() throws Exception {
             // no config file used
         }
 
-        connectionString = getConfiguredPropertyOrEnv(Constants.MSSQL_JDBC_TEST_CONNECTION_PROPERTIES);
-
+        connectionString = getConfiguredPropertyOrEnv(Constants.MSSQL_JDBC_TEST_CONNECTION_PROPERTIES);
```
lilgreenbird, line 161: why remove the blank line?

src/test/java/com/microsoft/sqlserver/testframework/Constants.java
David-Engel: @lilgreenbird We would need to exclude these tests against test endpoints that don't support JSON. (I'm kind of surprised we don't have one for Data Classification, for example. But it looks like we've used SQL Server version for that, which isn't quite proper.)

lilgreenbird: we need a new tag for new SQL Server version that supports JSON. (SQL Server 2023 or ?)

lilgreenbird (changes requested):
the tests need to take into consideration the SQL server version. JSON support is only avail on SQL 2023 (or ?) so this means we need to exclude this for all that version and below. Need new tag for this new SQL Server version

src/test/java/com/microsoft/sqlserver/testframework/Constants.java
lilgreenbird: yeah, why not? data classification was in 2019 only

src/main/java/com/microsoft/sqlserver/jdbc/dtv.java
```diff
@@ -1459,6 +1461,7 @@ final void executeOp(DTVExecuteOp op) throws SQLServerException {
                 case NVARCHAR:
                 case LONGNVARCHAR:
                 case NCLOB:
+                case JSON:
```
lilgreenbird, line 1464: I think we're missing a case in the else (line #1599). This makes me think we're also missing a test case that will go to that path.  Check SQL_VARIANT to see where this is tested.

lilgreenbird: I think we also need a case for JSON in readSqlVariant, and, a test for that

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
```diff
@@ -2176,6 +2181,7 @@ else if (null != sourceCryptoMeta) {
                 case microsoft.sql.Types.DATETIMEOFFSET:
                 case microsoft.sql.Types.DATETIME:
                 case microsoft.sql.Types.SMALLDATETIME:
+                case microsoft.sql.Types.JSON:
```
muskan124947: This is not required. This conversion is for temporal data types.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -6852,6 +6885,23 @@ private void onFeatureExtAck(byte featureId, byte[] data) throws SQLServerExcept
                 serverSupportsVector = true;
                 break;
             }
+            
+            case TDS.TDS_FEATURE_EXT_JSONSUPPORT: {
+                if (connectionlogger.isLoggable(Level.FINER)) {
```
David-Engel:
```suggestion
                if (connectionlogger.isLoggable(Level.FINE)) {
```

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResource.java
```diff
@@ -314,6 +314,7 @@ protected Object[][] getContents() {
         {"R_AE_NotSupportedByServer", "SQL Server in use does not support column encryption."},
         {"R_InvalidAEVersionNumber", "Received invalid version number \"{0}\" for Always Encrypted."}, // From server
         {"R_InvalidVectorVersionNumber", "Received invalid version number \"{0}\" for vector feature negotiation."},
+        {"R_InvalidJSONVersionNumber", "Received invalid version number \"{0}\" for JSON."},
```
David-Engel:
```suggestion
        {"R_InvalidJSONVersionNumber", "Received invalid version number \"{0}\" for JSON feature negotiation."},
```

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResource.java
```diff
@@ -488,6 +489,7 @@ protected Object[][] getContents() {
         {"R_unknownUTF8SupportValue", "Unknown value for UTF8 support."},
         {"R_unknownAzureSQLDNSCachingValue", "Unknown value for Azure SQL DNS Caching."},
         {"R_unknownVectorSupportValue", "Unexpected version value received for vector support feature negotiation."},
+        {"R_unknownJSONSupportValue", "Unknown value for JSON support."},
```
David-Engel:
```suggestion
        {"R_unknownJSONSupportValue", "Unexpected version value received for JSON support feature negotiation."},
```

David-Engel (commented):
Partial review (21/31)

src/test/java/com/microsoft/sqlserver/jdbc/callablestatement/CallableStatementTest.java
```diff
@@ -75,6 +75,9 @@ public class CallableStatementTest extends AbstractTest {
             .escapeIdentifier(RandomUtil.getIdentifier("manyParam_definedType"));
     private static String zeroParamSproc = AbstractSQLGenerator
             .escapeIdentifier(RandomUtil.getIdentifier("zeroParamSproc"));
+    private static String tableNameJSON = "TestJSONTable";
```
David-Engel: Table name should be randomized to avoid concurrent test interference.

src/test/java/com/microsoft/sqlserver/jdbc/callablestatement/CallableStatementTest.java
```diff
@@ -75,6 +75,9 @@ public class CallableStatementTest extends AbstractTest {
...
+	private static String procedureNameJSON = AbstractSQLGenerator
+			.escapeIdentifier(RandomUtil.getIdentifier("TestJSONProcedure"));
```
David-Engel: NIT: Spacing
```suggestion
    private static String procedureNameJSON = AbstractSQLGenerator
            .escapeIdentifier(RandomUtil.getIdentifier("TestJSONProcedure"));
```

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/JSONFunctionTest.java
```diff
@@ -0,0 +1,1862 @@
...
+
+    /**
+     * Test ISJSON function with valid and invalid JSON value.
+     * ISJSON -> Tests whether a string contains valid JSON.
+     */
+    @Test
+    @Tag(Constants.JSONTest)
+    public void testISJSONWithVariousInputs() throws SQLException {
+        String dstTable = TestUtils
+                .escapeSingleQuotes(AbstractSQLGenerator.escapeIdentifier("dstTableNew"));
```
David-Engel: Randomize table name

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/JSONFunctionTest.java
```diff
@@ -0,0 +1,1862 @@
...
+                        "ISJSON(testCol) AS isJsonValid " +
+                        "FROM " + dstTable;
+                try (ResultSet rs = dstStmt.executeQuery(select)) {
+                    assertTrue(rs.next());
+                    assertEquals(validJson, rs.getString("testCol"));
+                    assertEquals(1, rs.getInt("isJsonValid"));
+                }
+
+            } catch (Exception e) {
+                fail(e.getMessage());
```
David-Engel: This catch (and all the others like it) are unnecessary. The uncaught Exception will provide better information in the test failure results.

David-Engel (commented):
Another partial review 23/31

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/JSONFunctionTest.java
```diff
@@ -0,0 +1,1828 @@
...
+    @Tag(Constants.JSONTest)
+    public void testJSONArrayAggWithThreeElements() throws SQLException {
+        String select = "SELECT JSON_ARRAYAGG(c1) AS jsonArrayAgg FROM (VALUES ('c'), ('b'), ('a')) AS t(c1)";
+        try (Connection conn = DriverManager.getConnection(connectionString);
+                Statement stmt = conn.createStatement();
+                ResultSet rs = stmt.executeQuery(select)) {
+            assertTrue(rs.next());
+            assertEquals("[\"c\",\"b\",\"a\"]", rs.getString("jsonArrayAgg"));
+        } catch (Exception e) {
+            fail(e.getMessage());
```
David-Engel: The unhandled exception will provide better context to the test failure versus catching and just surfacing the message.
```suggestion
```

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/JSONFunctionTest.java
```diff
@@ -0,0 +1,1828 @@
...
+    @Tag(Constants.JSONTest)
+    public void testJSONArrayAggWithOrderedElements() throws SQLException {
+        String select = "SELECT JSON_ARRAYAGG(c1 ORDER BY c1) AS jsonArrayAgg FROM (VALUES ('c'), ('b'), ('a')) AS t(c1)";
+        try (Connection conn = DriverManager.getConnection(connectionString);
+                Statement stmt = conn.createStatement();
+                ResultSet rs = stmt.executeQuery(select)) {
+            assertTrue(rs.next());
+            assertEquals("[\"a\",\"b\",\"c\"]", rs.getString("jsonArrayAgg"));
+        } catch (Exception e) {
+            fail(e.getMessage());
```
David-Engel:
```suggestion
```

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/JSONFunctionTest.java
```diff
@@ -0,0 +1,1828 @@
...
+                "SET @info = JSON_MODIFY(JSON_MODIFY(JSON_MODIFY(@info, '$.name', 'Mike'), '$.surname', 'Smith'), 'append $.skills', 'Azure'); "
+                +
+                "SELECT @info AS info;";
+        try (Connection conn = DriverManager.getConnection(connectionString);
+                Statement stmt = conn.createStatement();
+                ResultSet rs = stmt.executeQuery(update)) {
+            assertTrue(rs.next());
+            assertEquals(expectedJson, rs.getString("info"));
+        } catch (Exception e) {
+            fail(e.getMessage());
```
David-Engel:
```suggestion
```

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/JSONFunctionTest.java
```diff
@@ -0,0 +1,1828 @@
...
+                "SET @product = JSON_MODIFY(JSON_MODIFY(@product, '$.Price', CAST(JSON_VALUE(@product, '$.price') AS NUMERIC(4, 2))), '$.price', NULL); "
+                +
+                "SELECT @product AS product;";
+        try (Connection conn = DriverManager.getConnection(connectionString);
+                Statement stmt = conn.createStatement();
+                ResultSet rs = stmt.executeQuery(update)) {
+            assertTrue(rs.next());
+            assertEquals(expectedJson, rs.getString("product"));
+        } catch (Exception e) {
+            fail(e.getMessage());
```
David-Engel:
```suggestion
```

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/JSONFunctionTest.java
```diff
@@ -0,0 +1,1828 @@
...
+                "SET @stats = JSON_MODIFY(@stats, '$.click_count', CAST(JSON_VALUE(@stats, '$.click_count') AS INT) + 1); "
+                +
+                "SELECT @stats AS stats;";
+        try (Connection conn = DriverManager.getConnection(connectionString);
+                Statement stmt = conn.createStatement();
+                ResultSet rs = stmt.executeQuery(update)) {
+            assertTrue(rs.next());
+            assertEquals(expectedJson, rs.getString("stats"));
+        } catch (Exception e) {
+            fail(e.getMessage());
```
David-Engel:
```suggestion
```

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/JSONFunctionTest.java
```diff
@@ -0,0 +1,1828 @@
...
+    @Tag(Constants.JSONTest)
+    public void testJSONObjectEmpty() throws SQLException {
+        String select = "SELECT JSON_OBJECT() AS jsonObject";
+        try (Connection conn = DriverManager.getConnection(connectionString);
+                Statement stmt = conn.createStatement();
+                ResultSet rs = stmt.executeQuery(select)) {
+            assertTrue(rs.next());
+            assertEquals("{}", rs.getString("jsonObject"));
+        } catch (Exception e) {
+            fail(e.getMessage());
```
David-Engel:
```suggestion
```

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/JSONFunctionTest.java
```diff
@@ -0,0 +1,1828 @@
...
+    @Tag(Constants.JSONTest)
+    public void testJSONObjectWithMultipleKeys() throws SQLException {
+        String select = "SELECT JSON_OBJECT('name':'value', 'type':NULL ABSENT ON NULL) AS jsonObject";
+        try (Connection conn = DriverManager.getConnection(connectionString);
+                Statement stmt = conn.createStatement();
+                ResultSet rs = stmt.executeQuery(select)) {
+            assertTrue(rs.next());
+            assertEquals("{\"name\":\"value\"}", rs.getString("jsonObject"));
+        } catch (Exception e) {
+            fail(e.getMessage());
```
David-Engel:
```suggestion
```

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/JSONFunctionTest.java
```diff
@@ -0,0 +1,1828 @@
...
+    @Tag(Constants.JSONTest)
+    public void testJSONObjectWithJsonArray() throws SQLException {
+        String select = "SELECT JSON_OBJECT('name':'value', 'type':JSON_ARRAY(1, 2)) AS jsonObject";
+        try (Connection conn = DriverManager.getConnection(connectionString);
+                Statement stmt = conn.createStatement();
+                ResultSet rs = stmt.executeQuery(select)) {
+            assertTrue(rs.next());
+            assertEquals("{\"name\":\"value\",\"type\":[1,2]}", rs.getString("jsonObject"));
+        } catch (Exception e) {
+            fail(e.getMessage());
```
David-Engel:
```suggestion
```

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/JSONFunctionTest.java
```diff
@@ -0,0 +1,1828 @@
...
+    @Tag(Constants.JSONTest)
+    public void testJSONObjectWithNestedJsonObject() throws SQLException {
+        String select = "SELECT JSON_OBJECT('name':'value', 'type':JSON_OBJECT('type_id':1, 'name':'a')) AS jsonObject";
+        try (Connection conn = DriverManager.getConnection(connectionString);
+                Statement stmt = conn.createStatement();
+                ResultSet rs = stmt.executeQuery(select)) {
+            assertTrue(rs.next());
+            assertEquals("{\"name\":\"value\",\"type\":{\"type_id\":1,\"name\":\"a\"}}", rs.getString("jsonObject"));
+        } catch (Exception e) {
+            fail(e.getMessage());
```
David-Engel:
```suggestion
```

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/JSONFunctionTest.java
```diff
@@ -0,0 +1,1828 @@
...
+                ResultSet rs = stmt.executeQuery(select)) {
+            while (rs.next()) {
+                int sessionId = rs.getInt("session_id");
+                String info = rs.getString("info");
+                assertTrue(info.contains("\"security_id\":\""));
+                assertTrue(info.contains("\"login\":\""));
+                assertTrue(info.contains("\"status\":\""));
+            }
+        } catch (Exception e) {
+            fail(e.getMessage());
```
David-Engel:
```suggestion
```

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/JSONFunctionTest.java
```diff
@@ -0,0 +1,1828 @@
...
+    @Tag(Constants.JSONTest)
+    public void testJSONObjectAggWithThreeProperties() throws SQLException {
+        String select = "SELECT JSON_OBJECTAGG(c1:c2) AS jsonObjectAgg FROM (VALUES('key1', 'c'), ('key2', 'b'), ('key3','a')) AS t(c1, c2)";
+        try (Connection conn = DriverManager.getConnection(connectionString);
+                Statement stmt = conn.createStatement();
+                ResultSet rs = stmt.executeQuery(select)) {
+            assertTrue(rs.next());
+            assertEquals("{\"key1\":\"c\",\"key2\":\"b\",\"key3\":\"a\"}", rs.getString("jsonObjectAgg"));
+        } catch (Exception e) {
+            fail(e.getMessage());
```
David-Engel:
```suggestion
```

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/JSONFunctionTest.java
```diff
@@ -0,0 +1,1828 @@
...
+        String jsonInfo = "{\"info\":{\"address\":[{\"town\":\"Paris\"},{\"town\":\"London\"}]}}";
+        String select = "DECLARE @jsonInfo AS JSON = N'" + jsonInfo + "'; " +
+                "SELECT JSON_PATH_EXISTS(@jsonInfo, '$.info.address') AS pathExists";
+        try (Connection conn = DriverManager.getConnection(connectionString);
+                Statement stmt = conn.createStatement();
+                ResultSet rs = stmt.executeQuery(select)) {
+            assertTrue(rs.next());
+            assertEquals(1, rs.getInt("pathExists"));
+        } catch (Exception e) {
+            fail(e.getMessage());
```
David-Engel:
```suggestion
```

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/JSONFunctionTest.java
```diff
@@ -0,0 +1,1828 @@
...
+        String jsonInfo = "{\"info\":{\"address\":[{\"town\":\"Paris\"},{\"town\":\"London\"}]}}";
+        String select = "DECLARE @jsonInfo AS JSON = N'" + jsonInfo + "'; " +
+                "SELECT JSON_PATH_EXISTS(@jsonInfo, '$.info.addresses') AS pathExists";
+        try (Connection conn = DriverManager.getConnection(connectionString);
+                Statement stmt = conn.createStatement();
+                ResultSet rs = stmt.executeQuery(select)) {
+            assertTrue(rs.next());
+            assertEquals(0, rs.getInt("pathExists"));
+        } catch (Exception e) {
+            fail(e.getMessage());
```
David-Engel:
```suggestion
```

src/test/java/com/microsoft/sqlserver/jdbc/resultset/ResultSetTest.java
```diff
@@ -709,6 +719,43 @@ public void testResultSetClientCursorInitializerSqlErrorState() {
...
+                    rs.next();
+                    assertEquals(123, rs.getShort("c1"));
+                    assertEquals(123, rs.getInt("c1"));
+                    assertEquals(123f, rs.getFloat("c1"));
+                    assertEquals(123L, rs.getLong("c1"));
+                    assertEquals(123d, rs.getDouble("c1"));
+                    assertEquals(new BigDecimal(123), rs.getBigDecimal("c1"));
+                }
+            } catch (Exception e) {
+                fail(e.getMessage());
```
David-Engel:
```suggestion
```

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyISQLServerBulkRecordTest.java
```diff
@@ -186,6 +186,299 @@ public void testBulkCopyVector() throws SQLException {
                     assertNotNull(resultVector, "Retrieved vector is null.");
                     assertEquals(3, resultVector.getDimensionCount(), "Dimension count mismatch.");
                     assertArrayEquals(vectorData, resultVector.getData(), "Vector data mismatch.");
+                } catch (Exception e) {
+                    fail(e.getMessage());
```
David-Engel:
```suggestion
```

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyISQLServerBulkRecordTest.java
```diff
@@ -186,6 +186,299 @@ public void testBulkCopyVector() throws SQLException {
...
+                String select = "SELECT * FROM " + dstTable;
+                try (ResultSet rs = dstStmt.executeQuery(select)) {
+                    assertTrue(rs.next());
+                    assertEquals(data1, rs.getString(1));
+                    assertTrue(rs.next());
+                    assertEquals(data2, rs.getString(1));
+                }
+
+            } catch (Exception e) {
+                fail(e.getMessage());
```
David-Engel:
```suggestion
```

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyISQLServerBulkRecordTest.java
```diff
@@ -186,6 +186,299 @@ public void testBulkCopyVector() throws SQLException {
...
+                    assertTrue(rs.next());
+                    assertEquals(data1, rs.getString(1));
+                    assertTrue(rs.next());
+                    assertEquals(data2, rs.getString(1));
+                    assertTrue(rs.next());
+                    assertEquals(data3, rs.getString(1));
+                }
+
+            } catch (Exception e) {
+                fail(e.getMessage());
```
David-Engel:
```suggestion
```

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyISQLServerBulkRecordTest.java
```diff
@@ -186,6 +186,299 @@ public void testBulkCopyVector() throws SQLException {
...
+                    assertTrue(rs.next());
+                    assertTrue(data1.equals(rs.getObject(1)));
+                    assertTrue(rs.next());
+                    assertTrue(data2.equals(rs.getObject(1)));
+                    assertTrue(rs.next());
+                    assertTrue(data3.equals(rs.getObject(1)));
+                }
+
+            } catch (Exception e) {
+                fail(e.getMessage());
```
David-Engel:
```suggestion
```

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyISQLServerBulkRecordTest.java
```diff
@@ -186,6 +186,299 @@ public void testBulkCopyVector() throws SQLException {
...
+                    assertTrue(rs.next());
+                    assertTrue(data1Col1.equals(rs.getObject(1)));
+                    assertTrue(data1Col2.equals(rs.getObject(2)));
+                    assertTrue(rs.next());
+                    assertTrue(data2Col1.equals(rs.getObject(1)));
+                    assertTrue(data2Col2.equals(rs.getObject(2)));
+                }
+
+            } catch (Exception e) {
+                fail(e.getMessage());
```
David-Engel:
```suggestion
```

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyISQLServerBulkRecordTest.java
```diff
@@ -186,6 +186,299 @@ public void testBulkCopyVector() throws SQLException {
...
+                    assertTrue(rs.next());
+                    assertTrue(data1.equals(rs.getObject(1)));
+                    assertTrue(rs.next());
+                    assertTrue(data2.equals(rs.getObject(1)));
+                    assertTrue(rs.next());
+                    assertEquals(data3, rs.getString(1));
+                }
+
+            } catch (Exception e) {
+                fail(e.getMessage());
```
David-Engel:
```suggestion
```

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyISQLServerBulkRecordTest.java
```diff
@@ -186,6 +186,299 @@ public void testBulkCopyVector() throws SQLException {
...
+                bulkCopy.writeToServer(new BulkRecordJSON(data));
+
+                String select = "SELECT * FROM " + dstTable;
+                try (ResultSet rs = dstStmt.executeQuery(select)) {
+                    assertTrue(rs.next());
+                    assertTrue(data.equals(rs.getObject(1)));
+                }
+
+            } catch (Exception e) {
+                fail(e.getMessage());
```
David-Engel:
```suggestion
```

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyISQLServerBulkRecordTest.java
```diff
@@ -198,6 +491,56 @@ public void testBulkCopyVector() throws SQLException {
...
+                try (ResultSet rs = dstStmt.executeQuery(select)) {
+                    while (rs.next()) {
+                        count2++;
+                    }
+                }
+
+                assertEquals(count1, count2);
+
+            } catch (Exception e) {
+                fail(e.getMessage());
```
David-Engel:
```suggestion
```

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/JSONFunctionTest.java
```diff
@@ -0,0 +1,1828 @@
...
+                    while ((charsRead = clobReader.read(buffer)) != -1) {
+                        writer.write(buffer, 0, charsRead);
+                    }
+                }
+            }
+            
+            assertTrue(filesAreEqual(Path.of(JSON_FILE_PATH), tempFile));
+        
+        } catch (Exception e) {
+            fail("Test failed due to: " + e.getMessage());
```
David-Engel:
```suggestion
```

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/JSONFunctionTest.java
```diff
@@ -0,0 +1,1828 @@
...
```
David-Engel:
```suggestion
```

src/test/java/com/microsoft/sqlserver/jdbc/resultset/ResultSetTest.java
```diff
@@ -94,13 +95,14 @@ public void cleanUp() throws Exception {
      */
     @Test
     @Tag(Constants.xAzureSQLDW)
+    @Tag(Constants.JSONTest)
```
David-Engel: This concerns me a bit. I assume we are generally excluding JSON tests initially, until we stand up some new servers/images that support it in test pipelines. This effectively turns this test off for all the non-JSON stuff in it, too, until that happens. It would be better to split JSON into its own test here.

src/test/java/com/microsoft/sqlserver/testframework/AbstractTest.java
```diff
@@ -163,8 +163,7 @@ public static void setup() throws Exception {
...
```
David-Engel: This seems like an unneeded change to this file. I'd revert it so this file is unchanged in this PR.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -1687,8 +1687,12 @@ public final void setObject(int n, Object obj, int jdbcType) throws SQLServerExc
             precision = vector.getDimensionCount();
             scale = (int) VectorUtils.getScaleByte(vector.getVectorDimensionType());
         }
-
-        setObject(setterGetParam(n), obj, JavaType.of(obj), JDBCType.of(jdbcType), scale, precision, false, n, tvpName);
+        
+        if (microsoft.sql.Types.JSON == jdbcType) {
```
machavan, line 1691: Why did we need this special handling, could we avoid it ?

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyISQLServerBulkRecordTest.java
```diff
@@ -186,10 +186,334 @@ public void testBulkCopyVector() throws SQLException {
...
+            }
+        }
+    }
+
+    /**
+     * Test bulk copy with empty JSON document
+     */
+    @Test
+    @Tag(Constants.JSONTest)
+    public void testBulkCopyWithEmptyJsonDocument() throws SQLException {
```
machavan, line 237: Do we have tests for Bulk COPY API ( and with and without sendStringParameterAsUnicode set to tru/false ) ? If not let us add.

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyISQLServerBulkRecordTest.java
```diff
@@ -186,10 +186,334 @@ public void testBulkCopyVector() throws SQLException {
...
+            }
+        }
+    }
+
+    /**
+     * Test bulk copy with nested JSON documents.
+     */
+    @Test
+    @Tag(Constants.JSONTest)
+    public void testBulkCopyNestedJsonRows() throws SQLException {
```
machavan, line 461: A table to table bulk copy from source to destination table with source table storing vector data in JSON column while destination table having VECTOR data type will be interesting to have.

src/test/java/com/microsoft/sqlserver/jdbc/resultset/ResultSetTest.java
```diff
@@ -709,6 +710,73 @@ public void testResultSetClientCursorInitializerSqlErrorState() {
...
+    public void testCastOnJSON() throws SQLException {
+        String dstTable = TestUtils
+                .escapeSingleQuotes(AbstractSQLGenerator.escapeIdentifier(RandomUtil.getIdentifier("dstTable")));
+
+        String jsonData = "{\"key\":\"123\"}";
+
+        try (Connection conn = DriverManager.getConnection(connectionString)) {
+            try (Statement stmt = conn.createStatement()) {
+                stmt.executeUpdate("CREATE TABLE " + dstTable + " (jsonData JSON)");
+                stmt.executeUpdate("INSERT INTO " + dstTable + " VALUES (CAST('" + jsonData + "' AS JSON))");
```
machavan, line 727: Lets have some test also use the `execute` API

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
Ananya2: This ensures that nulls and objects are sent as NVARCHAR, as expected by SQL Server for JSON parameters. Without this, the driver sends them using an incorrect JDBC type (such as JSON), resulting in an invalid TDS stream.

machavan: OK, this place would check this for every value of every type ? Could we explore to o this somewhere down in IOBuffer, where there is already null specific coding?

Ananya2: Discussed offline.

Issue #2534, opened 2024-11-09
Stored procedure call return wrong BigDecimal scale.


DmitriyGod:
## Driver version
latest

## SQL Server version
mcr.microsoft.com/mssql/server:2022-latest

## Client Operating System
any

## JAVA/JVM version
any

## Table schema
no

## Problem description
Stored procedure call return wrong BigDecimal scale.
## Expected behavior
Return type defined scale
## Actual behavior
Return wrong scale.
## Error message/stack trace
```Java
var stmt = connection.createStatement();
stmt.execute("""
    create procedure test_bigdecimal
        @big_decimal_type      decimal(15, 3)       ,
        @big_decimal_type_o    decimal(15, 3) output
       as begin
           set @big_decimal_type_o = @big_decimal_type;
       end;
    """);
var call = connection.prepareCall("""
    {call test_bigdecimal(100.241, ?)}""");
call.registerOutParameter(1, Types.DECIMAL);
call.execute();

var actual = call.getBigDecimal(1);

var expected = new BigDecimal("100.241");

Assertions.assertEquals(expected, actual);
```
This example shows that driver returns 4 scale instead of 3.
So here is example driver returns 4 scale instead of 5:
```Java
var stmt = connection.createStatement();
stmt.execute("""
    create procedure test_bigdecimal
        @big_decimal_type      decimal(15, 5)       ,
        @big_decimal_type_o    decimal(15, 5) output
       as begin
           set @big_decimal_type_o = @big_decimal_type;
       end;
    """);
var call = connection.prepareCall("""
    {call test_bigdecimal(100.24112, ?)}""");
call.registerOutParameter(1, Types.DECIMAL);
call.execute();

var actual = call.getBigDecimal(1);

var expected = new BigDecimal("100.24112");

Assertions.assertEquals(expected, actual);
```
I think there is some place in driver fixing scale.
## Any other details that can be helpful
-

## JDBC trace logs
(1)
Expected :100.241
Actual   :100.2410
(2)

Expected :100.24112
Actual   :100.2411

machavan:
While this is investigated further, does the below workaround help?

Changing 

call.registerOutParameter(1, Types.DECIMAL);

to 

call.registerOutParameter(1, Types.DOUBLE);`

Jeffery-Wasty:
Hi @DmitriyGod,

Please let us know if the above workaround works for you. We'll be moving this issue to our backlog, to be addressed at a later time.

[...]

PR #2559, opened 2024-12-06
Stored procedure call returns wrong BigDecimal scale. 

Ananya2:
https://github.com/microsoft/mssql-jdbc/issues/2534 - Set scale value dynamically for DECIMAL output parameters in registerOutParameter method.

This change retrieves the scale from the ParameterMetaData and applies it to the DECIMAL output parameter. This provides a more flexible approach for handling DECIMAL data types. 

As an alternative, users can directly call registerOutParameter(index, sqlType, scale) to specify the scale explicitly.


machavan (commented):
Let us also add a test that resembles the two customer reported scenarios

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCallableStatement.java
```diff
@@ -150,6 +151,16 @@ public void registerOutParameter(int index, int sqlType) throws SQLServerExcepti
...
                 param.setOutScale(7);
                 break;
+            case java.sql.Types.DECIMAL:
+                ParameterMetaData parameterMetaData = this.getParameterMetaData();
+                if (parameterMetaData != null) {
+                    try {
+                        // Fetch scale from metadata for DECIMAL type
+                        int scale = parameterMetaData.getScale(index);
+                        param.setOutScale(scale);
+                    } catch (SQLException e) {}
```
divang, line 161: Please add the logging for the exception to avoid silent failures.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCallableStatement.java
```diff
@@ -150,6 +151,18 @@ public void registerOutParameter(int index, int sqlType) throws SQLServerExcepti
...
                 break;
+            case java.sql.Types.DECIMAL:
+                ParameterMetaData parameterMetaData = this.getParameterMetaData();
+                if (parameterMetaData != null) {
+                    try {
+                        // Fetch scale from metadata for DECIMAL type
+                        int scale = parameterMetaData.getScale(index);
+                        param.setOutScale(scale);
+                    } catch (SQLException e) {
+                       loggerExternal.warning("Failed to fetch scale for DECIMAL type parameter at index " + index + ": " + e.getMessage());
```
divang: The SQLServerException needs to be explicitly thrown instead of being suppressed.

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/StatementTest.java
```diff
@@ -1195,15 +1196,66 @@ public void testJdbc41CallableStatementMethods() throws Exception {
...
+        @Test
+        public void testBigDecimalPrecision() throws SQLException {
+            try (Connection connection = getConnection()) {
+                String createProceduresSQL = "create procedure test_bigdecimal_3\n" +
+                "    @big_decimal_type      decimal(15, 3),\n" +
+                "    @big_decimal_type_o    decimal(15, 3) output\n" +
+                "as\n" +
+                "begin\n" +
+                "    set @big_decimal_type_o = @big_decimal_type;\n" +
+                "end;\n";
```
tkyc, line 1220: We should move the create and drop of the procedure to the `init()` and `terminate()`. If the test fails for some reason mid run, then the drop call doesn't get executed, and then next run the create fails since the proc already exists. And, the procedure name should be prefixed/suffixed with a unique identifier eg. using `AbstractSQLGenerator.escapeIdentifier`. The create proc might fail if the proc name isn't unique, if multiple runs of this test is running against the same server.

tkyc:
Looks like test `testJdbc41CallableStatementMethods` is failing.

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/StatementTest.java
lilgreenbird: also, pls use code formatter this doesn't look formatted

tkyc:
pinged you about an internal test fail

lilgreenbird (changes requested):
this failed tests in internal test lab

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/StatementTest.java
Ananya2: Updated the test case, please review.

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/BigDecimalPrecisionTest.java
```diff
@@ -0,0 +1,90 @@
...
+
+            try (Statement stmt = connection.createStatement()) {
+                stmt.execute(createProcedureSQL1);
+                stmt.execute(createProcedureSQL2);
+            }
+        }
+    }
+
+    @AfterEach
+    public void terminate() throws SQLException {
```
machavan, line 53: Let us use this approach for table creation and dropping in init and terminate methods as it provides idempotency on failures and repeated runs


        @BeforeEach
        public void init() throws Exception {
            try (Connection con = getConnection()) {
                con.setAutoCommit(false);
                try (Statement stmt = con.createStatement()) {
                    TestUtils.dropTableIfExists(tableName, stmt);
                    stmt.executeUpdate(
                            "CREATE TABLE " + tableName + " (col1 INT, col2 VARCHAR(" + TEST_STRING.length() + "))");
                }
                con.commit();
            }
        }



       @AfterEach
        public void terminate() throws Exception {
            try (Connection con = getConnection(); Statement stmt = con.createStatement()) {
                try {
                    TestUtils.dropTableIfExists(tableName, stmt);
                } catch (SQLException e) {
                    fail(TestResource.getResource("R_unexpectedException") + e.getMessage());
                }
            }
        }

Ramudaykumar:
There should be significant comments in code, so that a new person can understand the code.

Ananya2:
> There should be significant comments in code, so that a new person can understand the code.
Noted.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCallableStatement.java
```diff
@@ -150,6 +151,21 @@ public void registerOutParameter(int index, int sqlType) throws SQLServerExcepti
...
+                // The scale for the DECIMAL type is fetched from the ParameterMetaData.
+                // This provides flexibility to automatically apply the correct scale as per the database metadata.
+                ParameterMetaData parameterMetaData = this.getParameterMetaData();
+                if (parameterMetaData != null) {
+                    try {
+                        // Fetch scale from metadata for DECIMAL type
+                        int scale = parameterMetaData.getScale(index);
+                        param.setOutScale(scale);
+                    } catch (SQLException e) {
+                        loggerExternal.warning("Failed to fetch scale for DECIMAL type parameter at index " + index + ": " + e.getMessage());
```
lilgreenbird, line 165: should this fail? according to the spec this would only throw an error if a database access error occurs, seem quite a serious error

lilgreenbird (changes requested):
doesn't look like the formatter was run on BigDecimalPrecisionTest.java, pls make sure to run formatter on all files

lilgreenbird (commented):
there are already BigDecimal tests in StatementTest. Either just add the new tests here or, move all the BigDecimal test to this new file

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/BigDecimalPrecisionTest.java
```diff
@@ -0,0 +1,94 @@
...
+        try (Connection connection = getConnection()) {
+            String dropProcedureSQL = "DROP PROCEDURE IF EXISTS " + procName1 + ", " + procName2;
+            try (Statement stmt = connection.createStatement()) {
+                stmt.execute(dropProcedureSQL);
+            }
+            
+            String createProcedureSQL1 = "CREATE PROCEDURE " + procName1 + "\n" +
+                    "    @big_decimal_type      decimal(15, 3),\n" +
+                    "    @big_decimal_type_o    decimal(15, 3) OUTPUT\n" +
+                    "AS\n" +
```
lilgreenbird, line 38: add boundary cases for max scale 38. see BigDecimal tests in StatementTest.java

divang (approved):

tkyc (approved):

Ananya2:
Changes have already been requested as a part of https://github.com/microsoft/mssql-jdbc/pull/2582, hence closing the PR.

PR #2562, opened 2024-12-12
Introduced timeouts for MSAL calls.

machavan:
Description:

Calls to MSAL authentication library done from the Java driver code use future.get(), potentially leading to hangs in case the library does not receive timely response from the server. All these calls done as part of getFedAuthToken need to have timeouts imposed. The timeout period is calculated based on remaining time wrt login timeout, and passed to future.get(<timeout>, <timeunit>) calls with this fix.

Tests:
- Existing unit tests pass on local build
- All ADO tests passed
- Added new tests

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
@@ -6110,7 +6110,7 @@ public class SQLServerConnection implements ISQLServerConnection, java.io.Serial
         }
 
         while (true) {
-        	int millisecondsRemaining = timerRemaining(timerExpire);
+            int millisecondsRemaining = timerRemaining(timerExpire);
             if (authenticationString.equalsIgnoreCase(SqlAuthentication.ACTIVE_DIRECTORY_PASSWORD.toString())) {
                 fedAuthToken = SQLServerMSAL4JUtils.getSqlFedAuthToken(fedAuthInfo, user,
                         activeConnectionProperties.getProperty(SQLServerDriverStringProperty.PASSWORD.toString()),
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
@@ -6126,12 +6126,12 @@ public class SQLServerConnection implements ISQLServerConnection, java.io.Serial
 
                 if (null != managedIdentityClientId && !managedIdentityClientId.isEmpty()) {
                     fedAuthToken = SQLServerSecurityUtility.getManagedIdentityCredAuthToken(fedAuthInfo.spn,
-                            managedIdentityClientId);
+                            managedIdentityClientId, millisecondsRemaining);
                     break;
                 }
 
                 fedAuthToken = SQLServerSecurityUtility.getManagedIdentityCredAuthToken(fedAuthInfo.spn,
-                        activeConnectionProperties.getProperty(SQLServerDriverStringProperty.MSI_CLIENT_ID.toString()));
+                        activeConnectionProperties.getProperty(SQLServerDriverStringProperty.MSI_CLIENT_ID.toString()), millisecondsRemaining);
 
                 // Break out of the retry loop in successful case.
                 break;
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
@@ -6110,10 +6110,11 @@ public class SQLServerConnection implements ISQLServerConnection, java.io.Serial
         }
 
         while (true) {
+        	int millisecondsRemaining = timerRemaining(timerExpire);
             if (authenticationString.equalsIgnoreCase(SqlAuthentication.ACTIVE_DIRECTORY_PASSWORD.toString())) {
                 fedAuthToken = SQLServerMSAL4JUtils.getSqlFedAuthToken(fedAuthInfo, user,
                         activeConnectionProperties.getProperty(SQLServerDriverStringProperty.PASSWORD.toString()),
-                        authenticationString);
+                        authenticationString, millisecondsRemaining);
 
                 // Break out of the retry loop in successful case.
                 break;
@@ -6141,12 +6142,12 @@ public class SQLServerConnection implements ISQLServerConnection, java.io.Serial
                 if (aadPrincipalID != null && !aadPrincipalID.isEmpty() && aadPrincipalSecret != null
                         && !aadPrincipalSecret.isEmpty()) {
                     fedAuthToken = SQLServerMSAL4JUtils.getSqlFedAuthTokenPrincipal(fedAuthInfo, aadPrincipalID,
-                            aadPrincipalSecret, authenticationString);
+                            aadPrincipalSecret, authenticationString, millisecondsRemaining);
                 } else {
                     fedAuthToken = SQLServerMSAL4JUtils.getSqlFedAuthTokenPrincipal(fedAuthInfo,
                             activeConnectionProperties.getProperty(SQLServerDriverStringProperty.USER.toString()),
                             activeConnectionProperties.getProperty(SQLServerDriverStringProperty.PASSWORD.toString()),
-                            authenticationString);
+                            authenticationString, millisecondsRemaining);
                 }
 
                 // Break out of the retry loop in successful case.
@@ -6159,7 +6160,7 @@ public class SQLServerConnection implements ISQLServerConnection, java.io.Serial
                         activeConnectionProperties.getProperty(SQLServerDriverStringProperty.USER.toString()),
                         servicePrincipalCertificate,
                         activeConnectionProperties.getProperty(SQLServerDriverStringProperty.PASSWORD.toString()),
-                        servicePrincipalCertificateKey, servicePrincipalCertificatePassword, authenticationString);
+                        servicePrincipalCertificateKey, servicePrincipalCertificatePassword, authenticationString, millisecondsRemaining);
 
                 // Break out of the retry loop in successful case.
                 break;
@@ -6194,7 +6195,7 @@ public class SQLServerConnection implements ISQLServerConnection, java.io.Serial
                             throw new SQLServerException(form.format(msgArgs), null);
                         }
 
-                        int millisecondsRemaining = timerRemaining(timerExpire);
+                        millisecondsRemaining = timerRemaining(timerExpire);
                         if (ActiveDirectoryAuthentication.GET_ACCESS_TOKEN_TRANSIENT_ERROR != errorCategory
                                 || timerHasExpired(timerExpire) || (fedauthSleepInterval >= millisecondsRemaining)) {
 
@@ -6240,7 +6241,7 @@ public class SQLServerConnection implements ISQLServerConnection, java.io.Serial
                         Object[] msgArgs = {SQLServerDriver.AUTH_DLL_NAME, authenticationString};
                         throw new SQLServerException(form.format(msgArgs), null, 0, null);
                     }
-                    fedAuthToken = SQLServerMSAL4JUtils.getSqlFedAuthTokenIntegrated(fedAuthInfo, authenticationString);
+                    fedAuthToken = SQLServerMSAL4JUtils.getSqlFedAuthTokenIntegrated(fedAuthInfo, authenticationString, millisecondsRemaining);
                 }
                 // Break out of the retry loop in successful case.
                 break;
@@ -6248,7 +6249,7 @@ public class SQLServerConnection implements ISQLServerConnection, java.io.Serial
                     .equalsIgnoreCase(SqlAuthentication.ACTIVE_DIRECTORY_INTERACTIVE.toString())) {
                 // interactive flow
                 fedAuthToken = SQLServerMSAL4JUtils.getSqlFedAuthTokenInteractive(fedAuthInfo, user,
-                        authenticationString);
+                        authenticationString, millisecondsRemaining);
 
                 // Break out of the retry loop in successful case.
                 break;
@@ -6258,12 +6259,12 @@ public class SQLServerConnection implements ISQLServerConnection, java.io.Serial
 
                 if (null != managedIdentityClientId && !managedIdentityClientId.isEmpty()) {
                     fedAuthToken = SQLServerSecurityUtility.getDefaultAzureCredAuthToken(fedAuthInfo.spn,
-                            managedIdentityClientId);
+                            managedIdentityClientId, millisecondsRemaining);
                     break;
                 }
 
                 fedAuthToken = SQLServerSecurityUtility.getDefaultAzureCredAuthToken(fedAuthInfo.spn,
-                        activeConnectionProperties.getProperty(SQLServerDriverStringProperty.MSI_CLIENT_ID.toString()));
+                        activeConnectionProperties.getProperty(SQLServerDriverStringProperty.MSI_CLIENT_ID.toString()), millisecondsRemaining);
 
                 break;
             }
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerMSAL4JUtils.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerMSAL4JUtils.java
@@ -66,6 +66,7 @@ class SQLServerMSAL4JUtils {
     static final String SLASH_DEFAULT = "/.default";
     static final String ACCESS_TOKEN_EXPIRE = "access token expires: ";
     static final long TOKEN_WAIT_DURATION_MS = 20000;
+    static final long TOKEN_LOCK_WAIT_DURATION_MS = 5000;
     private static final TokenCacheMap TOKEN_CACHE_MAP = new TokenCacheMap();
 
     private final static String LOGCONTEXT = "MSAL version "
@@ -88,9 +89,11 @@ class SQLServerMSAL4JUtils {
             logger.finest(LOGCONTEXT + authenticationString + ": get FedAuth token for user: " + user);
         }
 
-        lock.lock();
-
+        boolean lockAcquired = false;
         try {
+            //Just try to acquire the lock and if can't then proceed to attempt to get the token
+            lockAcquired = lock.tryLock(Math.min(millisecondsRemaining, TOKEN_LOCK_WAIT_DURATION_MS), TimeUnit.MILLISECONDS);
+
             String hashedSecret = getHashedSecret(new String[] {fedAuthInfo.stsurl, user, password});
             PersistentTokenCacheAccessAspect persistentTokenCacheAccessAspect = TOKEN_CACHE_MAP.getEntry(user,
                     hashedSecret);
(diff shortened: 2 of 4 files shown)


src/test/java/com/microsoft/sqlserver/jdbc/SQLServerConnectionTest.java
```diff
@@ -1321,4 +1323,34 @@ public void testServerNameField() throws SQLException {
...
+
+    @Test
+    public void testGetSqlFedAuthTokenFailure() throws SQLException {
+        try (Connection conn = getConnection()){
+        	SqlFedAuthInfo fedAuthInfo = ((SQLServerConnection) conn).new SqlFedAuthInfo();
+        	fedAuthInfo.spn = "https://database.windows.net/";
+        	fedAuthInfo.stsurl = "https://login.windows.net/xxx";
+        	SqlAuthenticationToken fedAuthToken = SQLServerMSAL4JUtils.getSqlFedAuthToken(fedAuthInfo, "xxx",
+                    "xxx",SqlAuthentication.ACTIVE_DIRECTORY_PASSWORD.toString(), 0);
+        	fail("Expected the test to throw SQLServerException");
```
lilgreenbird, line 1336: fail(TestResource.getResource("R_expectedExceptionNotThrown"));

src/test/java/com/microsoft/sqlserver/jdbc/SQLServerConnectionTest.java
```diff
@@ -1321,4 +1323,34 @@ public void testServerNameField() throws SQLException {
...
+        } catch (SQLServerException e) {
+        	//test pass
```
lilgreenbird, line 1338: check that the error message is the one we expect?

src/test/java/com/microsoft/sqlserver/jdbc/SQLServerConnectionTest.java
```diff
@@ -1321,4 +1323,34 @@ public void testServerNameField() throws SQLException {
...
+
+    @Test
+    public void testGetSqlFedAuthTokenPrincipalFailure() throws SQLException {
+        try (Connection conn = getConnection()){
+        	SqlFedAuthInfo fedAuthInfo = ((SQLServerConnection) conn).new SqlFedAuthInfo();
+        	fedAuthInfo.spn = "https://database.windows.net/";
+        	fedAuthInfo.stsurl = "https://login.windows.net/xxx";
+        	SqlAuthenticationToken fedAuthToken = SQLServerMSAL4JUtils.getSqlFedAuthTokenPrincipal(fedAuthInfo, "xxx",
+                    "xxx",SqlAuthentication.ACTIVE_DIRECTORY_SERVICE_PRINCIPAL.toString(), 0);
+        	fail("Expected the test to throw SQLServerException");
```
lilgreenbird, line 1351: fail(TestResource.getResource("R_expectedExceptionNotThrown"))

src/test/java/com/microsoft/sqlserver/jdbc/SQLServerConnectionTest.java
```diff
@@ -1321,4 +1323,34 @@ public void testServerNameField() throws SQLException {
...
+        } catch (SQLServerException e) {
+    	}
```
lilgreenbird, line 1353: check error msg

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerMSAL4JUtils.java
```diff
@@ -297,7 +302,7 @@ static SqlAuthenticationToken getSqlFedAuthTokenPrincipalCertificate(SqlFedAuthI
 
             final CompletableFuture<IAuthenticationResult> future = clientApplication
                     .acquireToken(ClientCredentialParameters.builder(scopes).build());
-            final IAuthenticationResult authenticationResult = future.get();
+            final IAuthenticationResult authenticationResult = future.get(millisecondsRemaining, TimeUnit.MILLISECONDS);
```
Gueorgi, line 306: Why there is no below the same condition as above:
`} catch (TimeoutException e) {
            throw new SQLServerException(SQLServerException.getErrString("R_connectionTimedOut"), e);
        }`

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -6110,10 +6110,11 @@ private SqlAuthenticationToken getFedAuthToken(SqlFedAuthInfo fedAuthInfo) throw
         }
 
         while (true) {
+            int millisecondsRemaining = timerRemaining(timerExpire);
```
Gueorgi, line 6113: If this is 0 or negative does it make sense to try the calls below but not throw timed out exception here (or from timerRemaining call)? Or just let the first calls throw the exception?

src/test/java/com/microsoft/sqlserver/jdbc/SQLServerConnectionTest.java
Gueorgi: Also needs one test with negative time maybe

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
JJ-WM: there's min value of 1ms, but the question is still valid.

src/test/java/com/microsoft/sqlserver/jdbc/SQLServerConnectionTest.java
machavan: Done

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
machavan: right, we are just letting the first call throw it

src/test/java/com/microsoft/sqlserver/jdbc/SQLServerConnectionTest.java
machavan: done

src/test/java/com/microsoft/sqlserver/jdbc/SQLServerConnectionTest.java
machavan: done

src/test/java/com/microsoft/sqlserver/jdbc/SQLServerConnectionTest.java
machavan: done

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerMSAL4JUtils.java
```diff
@@ -151,10 +159,12 @@ static SqlAuthenticationToken getSqlFedAuthTokenPrincipal(SqlFedAuthInfo fedAuth
...
         Set<String> scopes = new HashSet<>();
         scopes.add(scope);
-
-        lock.lock();
-
+        
+        boolean lockAcquired = false;
         try {
+        	//Just try to acquire the lock and if can't then proceed to attempt to get the token
+            lockAcquired = lock.tryLock(Math.min(millisecondsRemaining, TOKEN_LOCK_WAIT_DURATION_MS), TimeUnit.MILLISECONDS);
```
001mertiya, line 166: Where are we using `lockAcquired`? We should go ahead only if it returns true(lock acquired), right?

machavan: The idea is to attempt acquiring the lock with a timeout of 5 seconds. If the lock cannot be acquired within this time, proceed to attempt token acquisition regardless of whether the earlier thread(s) have released the lock or not

`lockAcquired` is used to check in the finally block if the lock should be released or not.

machavan: The purpose is to optimize the token acquisition process, the first caller succeeding does caching which is then leveraged by subsequent threads.  However, if the first thread takes considerable time, then we want the others to also go and try after waiting for a while.

If we were to let say 30 threads try in parallel, they would all miss the cache and hit the AAD auth endpoints to get their tokens at the same time, stressing the auth endpoint.

JJ-WM: Understood the intent, but would Semaphore be a more readable solution?

machavan: Agree. The code has been updated along with detailed comment describing the intent.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerMSAL4JUtils.java
```diff
@@ -77,19 +80,28 @@ private SQLServerMSAL4JUtils() {
...
+            String authenticationString, int millisecondsRemaining) throws SQLServerException {
         ExecutorService executorService = Executors.newSingleThreadExecutor();
 
         if (logger.isLoggable(Level.FINEST)) {
             logger.finest(LOGCONTEXT + authenticationString + ": get FedAuth token for user: " + user);
         }
 
-        lock.lock();
-
+        boolean semAcquired = false;
```
lilgreenbird, line 93: coding standard - boolean should be named isXXX ie isSemAcquired

machavan: Done

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerMSAL4JUtils.java
```diff
@@ -219,9 +248,18 @@ static SqlAuthenticationToken getSqlFedAuthTokenPrincipalCertificate(SqlFedAuthI
         Set<String> scopes = new HashSet<>();
         scopes.add(scope);
 
-        lock.lock();
-
+        boolean isSemAcquired = false;
         try {
+            //
+            //Just try to acquire the semaphore and if can't then proceed to attempt to get the token.
```
divang, line 254: Can we intergate mockito framework and simulate this scenario?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerMSAL4JUtils.java
```diff
@@ -116,7 +128,7 @@ static SqlAuthenticationToken getSqlFedAuthToken(SqlFedAuthInfo fedAuthInfo, Str
                     .builder(Collections.singleton(fedAuthInfo.spn + SLASH_DEFAULT), user, password.toCharArray())
                     .build());
 
-            final IAuthenticationResult authenticationResult = future.get();
+            final IAuthenticationResult authenticationResult = future.get(Math.min(millisecondsRemaining, TOKEN_WAIT_DURATION_MS), TimeUnit.MILLISECONDS);
```
saurabh500, line 131: Why is this min computation needed? In case of infinite connection timeout ?

saurabh500: Also in case of infinite timeout, what is the value of millisecondsRemaining? Is it 0 or negative?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerMSAL4JUtils.java
```diff
@@ -77,19 +80,28 @@ private SQLServerMSAL4JUtils() {
...
         try {
+            //
+            //Just try to acquire the semaphore and if can't then proceed to attempt to get the token.
+            //The purpose is to optimize the token acquisition process, the first caller succeeding does caching 
+            //which is then leveraged by subsequent threads. However, if the first thread takes considerable time, 
+            //then we want the others to also go and try after waiting for a while.
+            //If we were to let say 30 threads try in parallel, they would all miss the cache and hit the AAD auth endpoints 
+            //to get their tokens at the same time, stressing the auth endpoint.
+            //
+            isSemAcquired = sem.tryAcquire(Math.min(millisecondsRemaining, TOKEN_SEM_WAIT_DURATION_MS), TimeUnit.MILLISECONDS);
```
saurabh500, line 103: Based on the comment, it feels like there should be a `if` condition check to see if the semaphore actually was acquired or not. 
If not acquired, then it means that someone else is holding the semaphore and take actions ? 
From the current code, whether the semaphore was acquired or not, would lead to next code to execute no ? 
If the semaphore was not acquired, then we may have waited long enough, and that should likely lead to a timeout of connection attempt ?

saurabh500: This pattern is used at a lot of places in the code. 
So same comment is applicable at all usage instance.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerMSAL4JUtils.java
```diff
@@ -340,10 +382,19 @@ static SqlAuthenticationToken getSqlFedAuthTokenIntegrated(SqlFedAuthInfo fedAut
...
+
+        	final PublicClientApplication pca = PublicClientApplication
```
saurabh500, line 397: Nit: The indentation of this line seems off by one tab. Or is it Github UI messing this up for me ? :D

src/test/java/com/microsoft/sqlserver/jdbc/SQLServerConnectionTest.java
```diff
@@ -1321,4 +1323,51 @@ public void testServerNameField() throws SQLException {
...
+    }
+
+    @Test
+    public void testGetSqlFedAuthTokenFailureNoWaiting() throws SQLException {
+        try (Connection conn = getConnection()){
+        	SqlFedAuthInfo fedAuthInfo = ((SQLServerConnection) conn).new SqlFedAuthInfo();
+        	fedAuthInfo.spn = "https://database.windows.net/";
+        	fedAuthInfo.stsurl = "https://login.windows.net/xxx";
+        	SqlAuthenticationToken fedAuthToken = SQLServerMSAL4JUtils.getSqlFedAuthToken(fedAuthInfo, "xxx",
+                    "xxx",SqlAuthentication.ACTIVE_DIRECTORY_PASSWORD.toString(), 0);
```
saurabh500, line 1350: Out of curiosity, time remaining of 0, does it mean no more time left, or does it mean infinite?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerMSAL4JUtils.java
machavan: millisecondsRemaining is computed based on loginTimeout and TOKEN_WAIT_DURATION_MS is set to 20 seconds based on previous experience that anything over 15 seconds is not likely to return a token at all and we should bail out and retry in the outer connection retry loop.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerMSAL4JUtils.java
machavan: The idea is to block for a bit to allow the first request to get a token then use the cached token. But if a request hangs, we don’t want to block everyone for very long. It’s not going to break anything if the semaphore isn’t acquired. It’s just means two callers might overwrite the same token cache entry.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerMSAL4JUtils.java
machavan: Done

src/test/java/com/microsoft/sqlserver/jdbc/SQLServerConnectionTest.java
machavan: Right 0 means no time left here for the purpose of this test

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerMSAL4JUtils.java
machavan: We have covered test scenarios in unit tests by directly calling methods with specified millisecondsRemaining as arguments. 

But we will surely want to add Mockito based tests as well to try out few more scenarios.

saurabh500 (approved):

divang (approved):

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerMSAL4JUtils.java
```diff
@@ -77,19 +80,28 @@ private SQLServerMSAL4JUtils() {
         throw new UnsupportedOperationException(SQLServerException.getErrString("R_notSupported"));
     }
 
-    private static final Lock lock = new ReentrantLock();
+    private static final Semaphore sem = new Semaphore(1);
```
imasud00, line 83: >Semaphore [](http://example.com/codeflow?start=45&length=9)

If I have more than 1 connections tring with different auth mode should we restrict this semaphore to 1. Unless this class has its own instance for each connectionstring. ?

machavan: The lock is replaced by semaphore here only to improve the readability of the code as requested in some code review comments here, since it is not really serving a mutual exclusion but rather a kind of timed that helps all the incoming requests not to bombard AAD server at once and also not to wait for a long time for a previous requester to finish (if it is taking a long time).

But this is an interesting point to consider if there should be a separate semaphore per Auth mode.

David-Engel: We could technically partition the locking/semaphore to be by auth endpoint (fedAuthInfo.stsurl), but for the purposes of simplicity we've done it globally. The primary use case is connection pools that open a lot of connections to the same server (and auth point) at the same time. Different pools/threads opening many simultaneous connections to different servers (and auth points) isn't a use case we have prioritized here. Let me know if you disagree.

imasud00: actually we should partition becasue if one token is expired we need to wait all others connecting string as well .
 Is my understanding correct ?

David-Engel: A small potential wait in that scenario, yes.

Partitioning will add yet another lock, though. Obtain lock1 to add a new TokenCacheMap entry that contains a lock for the entry. Obtain the TokenCacheMap entry lock. Release lock1. Proceed with token acquisition. Release TokenCacheMap entry lock. Don't forget to refactor the cache expiry code to take into account the new lock...

imasud00: how does this work when we are working with Integrated Login where each user token is going to be different.

imasud00: Create an instance of this class per connections string could solve unmercenary coupling amongst connection string.

However I am still curious on how does this work if a webapp is using integrated login or integrated login, where each token will be unique to the user.

machavan: Created an ADO story to explore this further. 

Resolving this discussion here.

divang (dismissed):

David-Engel (approved):

Issue #2546, opened 2024-11-22
DatabaseMetaData getIndexInfo does not include COLUMNSTORE Indexes


goranschwarz:
## Driver version
12.8.1.0

## SQL Server version
Microsoft SQL Server 2022 (RTM-GDR) (KB5046861) - 16.0.1135.2 (X64) 
	Oct 18 2024 15:31:58 
	Copyright (C) 2022 Microsoft Corporation
	Developer Edition (64-bit) on Windows 10 Pro 10.0 <X64> (Build 22631: ) (Hypervisor)

## Client Operating System
Windows (but really any)

## JAVA/JVM version
java version "21.0.1" 2023-10-17 LTS

## Table schema
WideWorldImporters.Sales.InvoiceLines

## Problem description
DatabaseMetaData getIndexInfo does not include COLUMNSTORE Indexes
* See Screenshot from one of my tools: ![Image](https://github.com/user-attachments/assets/a2861d0c-e8b3-4287-9957-780acce0c06a)
* And a Screenshot from SSMS: ![Image](https://github.com/user-attachments/assets/8cdba856-d3c2-4750-b4e6-d7c1b7d1c84a)

## Expected behavior
That **all** indexes that re on the table are included in the resultset

## Actual behavior
Output of what you actually see.

## Error message/stack trace
None

## Any other details that can be helpful
Nope

## JDBC trace logs
None

machavan:
The driver calls sp_statisitcs procedure to get the index info and it does not include the COLUMNSTORE indexes 

```
e.g. 
exec sp_statistics 'tcol', 'dbo', 'tpch', '%', 'N', 'E'
```

We will check why it does not return this info and get back on this.

Meanwhile, as a workaround, a query similar to following on sys.indexes returning this info can be used:

```


SELECT 
    db_name() AS CatalogName,
    sch.name AS SchemaName,
    t.name AS TableName,
    i.name AS IndexName,
    i.type_desc AS IndexType,
    i.is_unique AS IsUnique,
    c.name AS ColumnName,
    ic.key_ordinal AS ColumnOrder
FROM 
    sys.indexes i
INNER JOIN 
    sys.index_columns ic ON i.object_id = ic.object_id AND i.index_id = ic.index_id
INNER JOIN 
    sys.columns c ON ic.object_id = c.object_id AND ic.column_id = c.column_id
INNER JOIN 
    sys.tables t ON i.object_id = t.object_id
INNER JOIN 
[...]

PR #2563, opened 2024-12-13
Updated getIndexInfo() to include Columnstore indexes by using custom query.

Ananya2:
Replaced the use of the sp_statistics stored procedure with a custom query to retrieve index information as the sp_statistics procedure did not return Columnstore indexes, so a query using sys.indexes was implemented as a workaround. 
This new query ensures that all index types (Clustered, NonClustered, Columnstore) are included in the result set. 
Github Issue: https://github.com/microsoft/mssql-jdbc/issues/2546


src/test/java/com/microsoft/sqlserver/jdbc/DatabaseMetadataTest.java
```diff
@@ -0,0 +1,103 @@
...
+                    hasColumnstoreIndex = true;
+                } else if (indexType.equals("CLUSTERED")) {
+                    hasClusteredIndex = true;
+                } else if (indexType.equals("NONCLUSTERED")) {
+                    hasNonClusteredIndex = true;
+                }
+            }
+
+            assertTrue(hasClusteredIndex, "CLUSTERED index found.");
+            assertTrue(hasNonClusteredIndex, "NONCLUSTERED index found.");
```
machavan, line 99: the messages  would be "non-clustered index not found" as the assert should raise them when hasClusteredIndex is false

src/test/java/com/microsoft/sqlserver/jdbc/DatabaseMetadataTest.java
```diff
@@ -0,0 +1,103 @@
...
+            assertTrue(hasColumnstoreIndex, "COLUMNSTORE index found.");
```
machavan, line 100: the messages  would be "columnstore index not found" as the assert should raise them when hasClusteredIndex is false

src/test/java/com/microsoft/sqlserver/jdbc/DatabaseMetadataTest.java
```diff
@@ -0,0 +1,103 @@
...
+                    "    sys.schemas sch ON t.schema_id = sch.schema_id " +
+                   
+                    "WHERE t.name = '" + tableName + "' " +
+                          "AND sch.name = '" + schema + "' " +
+                    "ORDER BY " +
+                    "    t.name, i.name, ic.key_ordinal;";
+            rs = stmt.executeQuery(query);
+            System.out.println("Testing query " + rs + " " + rs.next());
+            while (rs.next()) {
+                String indexType = rs.getString("IndexType");
```
machavan, line 85: Will be best to validate every single field coming out of getIndexInfo call, with the result of this query.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
```diff
@@ -1197,39 +1197,31 @@ private ResultSet executeSPFkeys(String[] procParams) throws SQLException {
...
+                "i.type_desc AS IndexType, " +
+                "i.is_unique AS IsUnique, " +
+                "c.name AS ColumnName, " +
+                "ic.key_ordinal AS ColumnOrder " +
+                "FROM sys.indexes i " +
+                "INNER JOIN sys.index_columns ic ON i.object_id = ic.object_id AND i.index_id = ic.index_id " +
+                "INNER JOIN sys.columns c ON ic.object_id = c.object_id AND ic.column_id = c.column_id " +
+                "INNER JOIN sys.tables t ON i.object_id = t.object_id " +
+                "INNER JOIN sys.schemas sch ON t.schema_id = sch.schema_id " +
+                "WHERE t.name = '" + table + "' " +
```
machavan, line 1221: Let us check for case sensitivity of table name and schema name. Compare this with existing behavior on what happens with sp_statistics

src/test/java/com/microsoft/sqlserver/jdbc/DatabaseMetadataTest.java
```diff
@@ -0,0 +1,88 @@
...
+    public static void setupTests() throws Exception {
+        setConnection();
+    }
+
+    @Test
+    public void testGetIndexInfo() throws SQLException, SQLServerException {
+        ResultSet rs = null;
+        try (Connection connection = getConnection(); Statement stmt = connection.createStatement()) {
+        	TestUtils.dropTableIfExists(tableName, stmt);
+        	String createTableSQL = "CREATE TABLE " + tableName + " (" +
```
machavan, line 33: Let us use below approach with init and terminate methods for table creation and doppping as it provides idempotency across repeated test runs

        @BeforeEach
        public void init() throws Exception {
            try (Connection con = getConnection()) {
                con.setAutoCommit(false);
                try (Statement stmt = con.createStatement()) {
                    TestUtils.dropTableIfExists(tableName, stmt);
                    stmt.executeUpdate(
                            "CREATE TABLE " + tableName + " (col1 INT, col2 VARCHAR(" + TEST_STRING.length() + "))");
                }
                con.commit();
            }
        }



       @AfterEach
        public void terminate() throws Exception {
            try (Connection con = getConnection(); Statement stmt = con.createStatement()) {
                try {
                    TestUtils.dropTableIfExists(tableName, stmt);
                } catch (SQLException e) {
                    fail(TestResource.getResource("R_unexpectedException") + e.getMessage());
                }
            }
        }

Ananya2:
Changes have already been requested as a part of https://github.com/microsoft/mssql-jdbc/pull/2566, hence closing the PR.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
Ananya2: I’ve validated that the previous getIndexInfo was also not case-sensitive, similar to this change, and I have added a test case to validate the scenario. I am closing this PR and have added the same changes in [PR #2566](https://github.com/microsoft/mssql-jdbc/pull/2566). Please review.

PR #2338, opened 2024-02-22
Finish support for RFC4180 for CSV bulk insert operations

funkyjive:
RFC4180 specifies support for fields that contain newlines as long as they are quoted.  See section 6 in the specification here: https://www.ietf.org/rfc/rfc4180.txt

This PR enhances the read operation when the option escapeColumnDelimitersCSV is set to true to allow newline characters to survive the CSV parse process and find their way into the record.  An existing unit test was enhanced to test the behavior.

PR #2575, opened 2025-01-06
Fixed issue with SQLServerBulkCopy from CSV with setEscapeColumnDelimerts set to true

machavan:
Description:

The method `SQLServerBulkCSVFileRecord::readLineEscapeDelimiters` introduced for supporting newline characters as part of field values ( https://github.com/microsoft/mssql-jdbc/pull/2338)  missed to handle the case of CSV file not having a newline character at the end of the last row in CSV file, resulting into skipping the last record when setEscapeColumnDelimerts is set to true.

Tests:
- All existing tests units pass
- Introduced additional tests to verify the reported scenario.


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCSVFileRecord.java
```diff
@@ -247,7 +247,9 @@ private String readLineEscapeDelimiters() throws SQLServerException {
                 throw new SQLServerException(SQLServerException.getErrString("R_InvalidCSVQuotes"), null, 0, null);
             }
             if (c == -1) { // keep semantics of readLine() by returning a null when there is no more data
-                return null;
+                if (sb.isEmpty()) {
```
divang, line 250: can we make it single line in `if` condition?
if (c == -1 && sb.isEmpty()) { // keep semantics of readLine() by returning a null when there is no more data
                    return null;
    }

machavan: Done

divang (dismissed):

tkyc (dismissed):

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyCSVTest.java
```diff
@@ -197,12 +198,70 @@ public void testEscapeColumnDelimitersCSV() throws Exception {
...
+                    assertEquals(expectedEscaped[i][0], rs.getString("c1"));
+                    assertEquals(expectedEscaped[i][1], rs.getString("c2"));
+                    assertEquals(expectedEscaped[i][2], rs.getString("c3"));
+                    assertEquals(expectedEscaped[i][3], rs.getString("c4"));
+                    i++;
+                }
+                assertEquals(i, 12, "Expected to load 12 records, but loaded " + i + " records");
+            }
+
+            TestUtils.dropTableIfExists(tableName, stmt);
```
lilgreenbird, line 261: put this in finally

codewatchzen (approved):

yatharthgovil:
Hi @machavan 
Thanks for resolving the issue, I want to know one more thing what method Microsoft recommend to load data to the target mssql table We have seen some gain with using SqlServerBulkCopy api and using ISqlServereBulkRecord against loading data using the same class using csv file.
Would really appreciate if you could answer this query.

machavan:
> Hi @machavan Thanks for resolving the issue, I want to know one more thing what method Microsoft recommend to load data to the target mssql table We have seen some gain with using SqlServerBulkCopy api and using ISqlServereBulkRecord against loading data using the same class using csv file. Would really appreciate if you could answer this query.

Hi @yatharthgovil 

Yes, bulk Copy offers a significant performance advantage. 

Please refer: 
Bulk Copy: https://learn.microsoft.com/en-us/sql/connect/jdbc/using-bulk-copy-with-the-jdbc-driver?view=sql-server-ver16

and 

Bulk Copy API for Batch Insert: https://learn.microsoft.com/en-us/sql/connect/jdbc/use-bulk-copy-api-batch-insert-operation?view=sql-server-ver16

lilgreenbird (changes requested):
see comment re test cleanup

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCSVFileRecord.java
```diff
@@ -246,7 +246,7 @@ private String readLineEscapeDelimiters() throws SQLServerException {
             if (c == -1 && quoteCount % 2 != 0) { // stream ended, but we are within quotes -- data problem
                 throw new SQLServerException(SQLServerException.getErrString("R_InvalidCSVQuotes"), null, 0, null);
             }
-            if (c == -1) { // keep semantics of readLine() by returning a null when there is no more data
+            if ((c == -1) && sb.toString().isEmpty()) { // keep semantics of readLine() by returning a null when there is no more data
```
mmimica, line 249:
```suggestion
            if ((c == -1) && sb.length() == 0) { // keep semantics of readLine() by returning a null when there is no more data
```
toString() is heavy

yatharthgovil:
Hi @machavan can you please let us know when this will be available for us to consume also whats the plan will this fix be available with the next version release? if so can you please let us know when is that planned also will it be a minor version upgrade or a major version upgrade?
We would like to know the ETA for when its avaialble for us to consume.

Jeffery-Wasty (dismissed):

yatharthgovil:
Hi @machavan can you please provide clarification for the concern I have mentioned in the previous comment or provide any ETA for the same?

machavan:
> Hi @machavan can you please let us know when this will be available for us to consume also whats the plan will this fix be available with the next version release? if so can you please let us know when is that planned also will it be a minor version upgrade or a major version upgrade? We would like to know the ETA for when its avaialble for us to consume.

Hi @yatharthgovil 

We will look to include this fix in the next release, will update you shortly on the timeline for the same.

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyCSVTest.java
machavan: done

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCSVFileRecord.java
machavan: done

David-Engel (approved):

tkyc (approved):

lilgreenbird (approved):

Jeffery-Wasty (approved):

machavan:
ADO tests passed for this PR.

machavan:
> Hi @machavan can you please let us know your plan for next version release as to when the release with this fix included will be released. We would like to know when is it planned ?

Hi @yatharthgovil , We are planning for a release around Feb end/March start

PR #2579, opened 2025-01-09
Fix for `mssql-jdbc.properties` location logic

Jeffery-Wasty:
The logic for finding the `mssql-jdbc.properties` file, used in Configurable Retry Logic was correct for test environments, and when building the driver manually, but not correct when using the driver as a jar or dependency. This fixes that issue by only removing the `target/classes` suffix if needed.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/ConfigurableRetryLogic.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/ConfigurableRetryLogic.java
@@ -284,7 +284,6 @@ public class ConfigurableRetryLogic {
     private static String getCurrentClassPath() throws SQLServerException {
         String location = "";
         String className = "";
-        String locationSuffix = "target/classes/";
 
         try {
             className = new Object() {}.getClass().getEnclosingClass().getName();
@@ -295,7 +294,7 @@ public class ConfigurableRetryLogic {
                 // We check if the Path we get from the CodeSource location is a directory. If so, we are running
                 // from class files and should remove a suffix (i.e. the props file is in a different location from the
                 // location returned)
-                location = location.substring(0, location.length() - locationSuffix.length());
+                location = location.substring(0, location.length() - ("target/classes/").length());
             }
 
             return new URI(location).getPath() + DEFAULT_PROPS_FILE; // TODO: Allow custom paths
--- a/src/main/java/com/microsoft/sqlserver/jdbc/ConfigurableRetryLogic.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/ConfigurableRetryLogic.java
@@ -14,7 +14,6 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.nio.file.Files;
 import java.nio.file.Paths;
-import java.security.CodeSource;
 import java.text.MessageFormat;
 import java.util.Collections;
 import java.util.Date;
@@ -288,20 +287,18 @@ public class ConfigurableRetryLogic {
         String locationSuffix = "target/classes/";
 
         try {
-            // Attempt to get the location and CodeSource for this class
             className = new Object() {}.getClass().getEnclosingClass().getName();
             location = Class.forName(className).getProtectionDomain().getCodeSource().getLocation().getPath();
-            CodeSource codeSource = ConfigurableRetryLogic.class.getProtectionDomain().getCodeSource();
             
-            if (null != codeSource && Files.isDirectory(Paths.get(codeSource.getLocation().toURI()))) {
+            if (Files.isDirectory(Paths.get(
+                    ConfigurableRetryLogic.class.getProtectionDomain().getCodeSource().getLocation().toURI()))) {
                 // We check if the Path we get from the CodeSource location is a directory. If so, we are running
                 // from class files and should remove a suffix (i.e. the props file is in a different location from the 
                 // location returned)
                 location = location.substring(0, location.length() - locationSuffix.length());
             }
-
-            URI uri = new URI(location);
-            return uri.getPath() + DEFAULT_PROPS_FILE; // For now, we only allow "mssql-jdbc.properties" as file name.
+            
+            return  new URI(location).getPath() + DEFAULT_PROPS_FILE; // TODO: Allow custom paths
         } catch (URISyntaxException e) {
             MessageFormat form = new MessageFormat(SQLServerException.getErrString("R_URLInvalid"));
             Object[] msgArgs = {location};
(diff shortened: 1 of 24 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/ConfigurableRetryLogic.java
```diff
@@ -283,16 +282,25 @@ private static void createConnectionRules(LinkedList<String> listOfRules) throws
     private static String getCurrentClassPath() throws SQLServerException {
         String location = "";
         String className = "";
+        String locationSuffix = "target/classes/";
```
tkyc, line 285: What about `C:\target\classes\mssql-jdbc.jar` + `C:\target\classes\mssql-jdbc.properties`?. Unlikely scenario I suppose.

src/main/java/com/microsoft/sqlserver/jdbc/ConfigurableRetryLogic.java
```diff
@@ -283,16 +282,25 @@ private static void createConnectionRules(LinkedList<String> listOfRules) throws
     private static String getCurrentClassPath() throws SQLServerException {
         String location = "";
         String className = "";
+        String locationSuffix = "target/classes/";
 
         try {
             className = new Object() {}.getClass().getEnclosingClass().getName();
             location = Class.forName(className).getProtectionDomain().getCodeSource().getLocation().getPath();
```
tkyc, line 289: If I do a `java -jar mssql-jdbc.jar` with the above code, the location always includes the JAR name eg. `/path/to/mssql-jdbc.jar`. But, when I run it in my IDE I get `/path/to/`. Without the JAR name in the path.

Do you need to account for the cases mentioned?

src/main/java/com/microsoft/sqlserver/jdbc/ConfigurableRetryLogic.java
```diff
@@ -283,16 +282,25 @@ private static void createConnectionRules(LinkedList<String> listOfRules) throws
...
             className = new Object() {}.getClass().getEnclosingClass().getName();
             location = Class.forName(className).getProtectionDomain().getCodeSource().getLocation().getPath();
-            location = location.substring(0, location.length() - 16);
-            URI uri = new URI(location + FORWARD_SLASH);
+            // When the driver is used as a jar / as a dependency, the above will be the correct "main" directory where
+            // the props file should be placed (as per documentation). When testing, or building the driver manually,
+            // the above will return with a suffix "target/classes/" which must be removed, as, with the first case, the
+            // properties file should be in the main directory.
+
+            if (location.endsWith(locationSuffix)) {
```
machavan, line 295: Do we need to handle this case of testing/building the driver? 

Would it be fair to expect developers to place the the mssql-jdbc.properties file under `<path>/tagret/classes`  themselves while tresting/debugging etc. ?

debedb: IDEs may normally put things there from resources subdirectory, so probably yes?

src/main/java/com/microsoft/sqlserver/jdbc/ConfigurableRetryLogic.java
Jeffery-Wasty: Unlikely scenario that I'm not sure how we could address. In that case, they would not be able to use CRL with the props file. They would have to use the connection string property instead.

src/main/java/com/microsoft/sqlserver/jdbc/ConfigurableRetryLogic.java
Jeffery-Wasty: `Class.forName(className).getProtectionDomain().getCodeSource().getLocation().getPath();` will return the `/path/to` example you showed above, i.e. where the jar is without including the jar. This is the scenario I tested before pushing this PR.

src/main/java/com/microsoft/sqlserver/jdbc/ConfigurableRetryLogic.java
Jeffery-Wasty: It's not that the properties file is in `target/classes`. It should be in the main directory, and the documentation says this (or it will when we update it for 12.10.0). `Class.forName(className).getProtectionDomain().getCodeSource().getLocation().getPath();` returns the correct main directory when the driver is ran as a dependency or as a jar, when its ran while testing (i.e. running the source code for tests), the above snippet instead returns `<main_directory>/target/classes` which must be "corrected" to the main directory when the props file should be.

We can avoid this whole issue by only expected the props file in the main directory OR next to the jar, however this would mean we wouldn't be able to run internal testing for CRL using a props file. Also, since we're thinking of perhaps using the props file in the future for other connection properties, this would only be more of a limitation in the future.

Finally, I think this will be resolved when we institute custom locations for the props file as another connection string property (in our backlog). Where we only have to deal with "unusual" paths if they are specified in the prop.

src/main/java/com/microsoft/sqlserver/jdbc/ConfigurableRetryLogic.java
David-Engel: You could test to see if you are running from a jar file or classes. You would only need to use the location suffix in the classes scenario.
CoPilot suggested code:
```
        String classPath = RuntimeCheck.class.getProtectionDomain()
                                           .getCodeSource()
                                           .getLocation()
                                           .toURI()
                                           .getPath();
        if (Files.isDirectory(Paths.get(classPath))) {
            System.out.println("Running from class files");
        } else {
            System.out.println("Running from a JAR file");
        }
```

src/main/java/com/microsoft/sqlserver/jdbc/ConfigurableRetryLogic.java
David-Engel: @lilgreenbird See my related comment: https://github.com/microsoft/mssql-jdbc/pull/2579#discussion_r1927581870

Right, the env variable was left out of scope since we wanted to keep the user experience simple and be able to find the properties file without an app restart. If it makes testing easier, I'm not opposed to having the env variable to override the location of the properties file. But we should still look in the jar location when the var is not set.

Jeffery-Wasty: Thanks, I've switched out the logic to better detect beween jar or class files.

machavan (approved):

machavan (approved):

machavan:
ADO test runs were good for this change.

Issue #2576, opened 2025-01-07
IBM Semeru Runtime Certified Edition for z/OS, Kerberos and mssql-jdbc don't work together


trzpiot:
## Driver version

12.4.2.jre11

## Client Operating System

z/OS

## JAVA/JVM version

IBM Semeru Runtime Certified Edition for z/OS 17.0.7.0 (https://www.ibm.com/docs/en/semeru-runtime-ce-z/17)

## Problem description

The following exception is thrown:

```java
javax.security.auth.login.LoginException (No LoginModule found for com.ibm.security.auth.module.Krb5LoginModule)
```

Same problem as with https://github.com/microsoft/mssql-jdbc/issues/2139.

The check in https://github.com/microsoft/mssql-jdbc/pull/2209 on `java.vm.name` causes still false-positives, because the IBM Semeru Runtime Certified Edition for z/OS still uses `IBM J9 VM [...]` as `java.vm.name`. It then accesses `com.ibm.security.auth.module.Krb5LoginModule` instead of `com.sun.security.auth.module.Krb5LoginModule`. Only the latter is available on the Semeru runtime.

machavan:
Hi @trzpiot 

We are looking into addressing this issue. 

As a temporary workaround, you can setup a JAAS config and set a correct login module, 
- using jaasConfigurationName connection property
- or programmatically overriding the JAAS config and use the useDefaultJaasConfig connection property

MrEasy:
fyi: Change introduced #2608

PR #2581, opened 2025-01-10
IBM Semeru Runtime Certified Edition for z/OS, Kerberos and mssql-jdbc don't work together #2576

muskan124947:
Github Issue: https://github.com/microsoft/mssql-jdbc/issues/2576


machavan (dismissed):
Please run 
- azp run
-ADO tests with this branch

muskan124947:
tested with ado pipeline
https://sqlclientdrivers.visualstudio.com/mssql-jdbc/_build/results?buildId=105806&view=results

src/main/java/com/microsoft/sqlserver/jdbc/JaasConfiguration.java
```diff
@@ -19,26 +19,32 @@ public class JaasConfiguration extends Configuration {
...
+                // See
+                // https://www.ibm.com/support/knowledgecenter/SSYKE2_8.0.0/com.ibm.java.security.component.80.doc/security-component/jgssDocs/jaas_login_user.html
+                final String ibmLoginModule = "com.ibm.security.auth.module.Krb5LoginModule";
+                return new AppConfigurationEntry[] {
+                        new AppConfigurationEntry(ibmLoginModule, AppConfigurationEntry.LoginModuleControlFlag.SUFFICIENT,
+                                confDetailsWithoutPassword),
+                        new AppConfigurationEntry(ibmLoginModule, AppConfigurationEntry.LoginModuleControlFlag.SUFFICIENT,
+                                confDetailsWithPassword)};
+            } catch (ClassNotFoundException ex) {
+                throw new SQLServerException(SQLServerException.getErrString("R_moduleNotFound"), null);
```
David-Engel, line 46: Need to define R_moduleNotFound. "Neither com.sun.security.auth.module.Krb5LoginModule nor com.ibm.security.auth.module.Krb5LoginModule was found."

Also, exceptions are relatively heavy for perf. If we've gone down the IBM path once, we know we will always have to use that module in subsequent calls. So, we should keep track of that in a static boolean so we can jump straight to loading com.ibm.security.auth.module.Krb5LoginModule on future cycles.

muskan124947: Updated

src/main/java/com/microsoft/sqlserver/jdbc/KerbAuthentication.java
```diff
@@ -45,7 +45,11 @@ final class KerbAuthentication extends SSPIAuthentication {
     static {
         // Overrides the default JAAS configuration loader.
         // This one will forward to the default one in all cases but the default configuration is empty.
-        Configuration.setConfiguration(new JaasConfiguration(Configuration.getConfiguration()));
+        try {
+            Configuration.setConfiguration(new JaasConfiguration(Configuration.getConfiguration()));
+        } catch (SQLServerException e) {
+            e.printStackTrace();
```
lilgreenbird, line 51: we don't usually do this in the driver, throw the appropriate exception and error message

src/main/java/com/microsoft/sqlserver/jdbc/JaasConfiguration.java
```diff
@@ -19,26 +19,39 @@ public class JaasConfiguration extends Configuration {
     private final Configuration delegate;
     private AppConfigurationEntry[] defaultValue;
 
-    private static AppConfigurationEntry[] generateDefaultConfiguration() {
-        if (Util.isIBM()) {
+    private static boolean useIbmModule = false;
+
+    private static AppConfigurationEntry[] generateDefaultConfiguration() throws SQLServerException {
+        try {
+            if (useIbmModule) {
```
lilgreenbird, line 26: we only need to do this if isIBM(). Would it be better to just add this in Util.isIBM()?

src/main/java/com/microsoft/sqlserver/jdbc/JaasConfiguration.java
```diff
@@ -47,8 +60,10 @@ private static AppConfigurationEntry[] generateDefaultConfiguration() {
      * 
      * @param delegate
      *        a possibly null delegate
+     * @throws SQLServerException
+     *         if neither Kerberos module is found: com.sun.security.auth.module.Krb5LoginModule or com.ibm.security.auth.module.Krb5LoginModule
      */
-    JaasConfiguration(Configuration delegate) {
+    JaasConfiguration(Configuration delegate) throws SQLServerException {
```
lilgreenbird, line 66: codecov error, need a test case for this

src/main/java/com/microsoft/sqlserver/jdbc/JaasConfiguration.java
machavan: @lilgreenbird  Do you mean just adding the try catch logic in Util.isIBM method and leaving everything else unchanged ?

So something like this ? 


```
    private static Boolean isIBM = null;
    static boolean isIBM() {
        if (isIBM != null) {
            return isIBM;
        }

        String vmName = System.getProperty("java.vm.name");
        if (SYSTEM_JRE.startsWith("IBM") && vmName.startsWith("IBM")) {
            isIBM = true;
            return isIBM;
        } 

        try {
            Class.forName("com.ibm.security.auth.module.Krb5LoginModule");
            isIBM = true;
        } catch (ClassNotFoundException ex) {
            isIBM = false;
        }
        return isIBM;
    }

```

lilgreenbird: yes that's what I meant :)

divang (approved):

PR #2582, opened 2025-01-13
Stored procedure call returns wrong BigDecimal scale and integrated Mockito Dependency into JDBC Driver Codebase.

Ananya2:
https://github.com/microsoft/mssql-jdbc/issues/2534 - Set scale value dynamically for DECIMAL output parameters in registerOutParameter method.

This change retrieves the scale from the ParameterMetaData and applies it to the DECIMAL output parameter. This provides a more flexible approach for handling DECIMAL data types.

As an alternative, users can directly call registerOutParameter(index, sqlType, scale) to specify the scale explicitly.

And integrated Mockito Dependency into JDBC Driver Codebase. https://sqlclientdrivers.visualstudio.com/mssql-jdbc/_workitems/edit/33295


machavan (approved):
Please run ADO tests with this branch

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/StatementTest.java
```diff
@@ -2692,4 +2691,110 @@ public void terminate() throws Exception {
...
+
+    	@AfterEach
+    	public void terminate() throws SQLException {
+    		try (Connection connection = getConnection()) {
+    			try (Statement stmt = connection.createStatement()) {
+    				String dropProcedureSQL = "DROP PROCEDURE IF EXISTS " + procName1 + ", " + procName2 + ", "
+    						+ procNameMaxScale + ", " + procNameMaxPrecision;
+    				stmt.execute(dropProcedureSQL);
+    			}
+    		}
```
David-Engel, line 2751: There's a utility function to do this: `TestUtils.dropProcedureIfExists`.

Although, now that we no longer need to support SQL Server 2014, the `TestUtils.dropXXXIfExists` methods could be enhanced to use the `IF EXISTS` syntax and even take an array of strings to drop multiple objects in one command. But this enhancement is optional.

tkyc (approved):

Jeffery-Wasty (approved):

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCallableStatement.java
```diff
@@ -150,6 +151,22 @@ public void registerOutParameter(int index, int sqlType) throws SQLServerExcepti
             case microsoft.sql.Types.DATETIMEOFFSET:
                 param.setOutScale(7);
                 break;
+            case java.sql.Types.DECIMAL:
+                // Dynamically handle the scale for DECIMAL output parameters.
+                // The scale for the DECIMAL type is fetched from the ParameterMetaData.
+                // This provides flexibility to automatically apply the correct scale as per the database metadata.
+                ParameterMetaData parameterMetaData = this.getParameterMetaData();
```
David-Engel, line 158: I'm pretty sure this will incur a round trip to the database to get the metadata. That would be a significant performance hit if so. You can validate by running the SQL Server Profiler, watching executions against the DB while exercising this code. We would not want to add a round trip to the server for this.

David-Engel (commented):
This issue sounds very familiar... I will try to look for some history here.

Ananya2:
We’ll be closing this MR as, after reviewing the proposed changes, we’ve decided not to proceed with merging it. The update introduces an additional round trip to the database to fetch metadata, which could negatively impact performance. Minimizing unnecessary database calls is a key design principle of the driver, and adding this overhead would go against that objective.

A separate PR has been created to include the Mockito-related changes: https://github.com/microsoft/mssql-jdbc/pull/2644

PR #2594, opened 2025-01-28
Changed the scope of BULK_COPY_OPERATION_CACHE to connection.

machavan:
Description:

The BULK_COPY_OPERATION_CACHE should be defined at connection scope and not global scope to avoid incorrect cache hits across connections and potential OOM issues. 

Testing:
- Updated existing unit tests for this functionality accordingly.
- Existing unit tests as part of build pass with the change.


mmimica:
More importantly, it solves a problem when a driver instance is used to connect to different SQLServers. Tables from different databases could end up in the same static cache.

David-Engel (approved):

lilgreenbird (approved):

Jeffery-Wasty (approved):

tkyc (approved):

PR #2600, opened 2025-02-11
Perf: Check for null when getting DTV values (JDBC spec compliance - getBinaryStream /getAsciiStream will return null when the value is null)

machavan:
This fix of checking for null value of a DTVImpl before calling DTVImpl::getValue has shown some perf gains in our internal benchmark.

This fix will change behavior of getBinaryStream /getAsciiStream in ResultSet to be compliant with JDBC specs:
a Java input stream that delivers the database column value as a stream of one-byte ASCII characters; if the value is SQL NULL, the value returned is null


David-Engel (approved):

lilgreenbird (dismissed):
add test?

machavan:
> add test?

Verified that existing tests already cover this code path. This code path is hit for any get call on a ResultSet that returns NULL. e.g. many tests in ResultSetTest cover this.

divang (approved):

Jeffery-Wasty (approved):

tkyc (approved):

lilgreenbird:
test added in PR #2617

Issue #1611, opened 2021-07-06
Not able to build on my local machine due to data type not found sql exception. 


Avkashhirpara:
I am getting the test failure due to the following exception. 

com.microsoft.sqlserver.jdbc.SQLServerException: Column, parameter, or variable @P0. : Cannot find data type TVP_jdbc_avkash.hirpara_'44c0c0ef-3ab3-40ca-
98cd-ba486e8f0c35.
  at com.microsoft.sqlserver.jdbc.tvp.TVPAllTypesTest.testTVPDataTable(TVPAllTypesTest.java:172)
[ERROR] testQueryDateTVP  Time elapsed: 0.245 s  <<< ERROR!
com.microsoft.sqlserver.jdbc.SQLServerException: Column, parameter, or variable @P0. : Cannot find data type dateTVP_jdbc_avkash.hirpara_'8dc0b177-03a3-4
ea4-bebd-fe5914306ff0.
        at com.microsoft.sqlserver.jdbc.datatypes.DateAndTimeTypeTest.testQueryDateTVP(DateAndTimeTypeTest.java:127)

[ERROR] testQueryTimeTVP  Time elapsed: 0.252 s  <<< ERROR!
com.microsoft.sqlserver.jdbc.SQLServerException: Column, parameter, or variable @P0. : Cannot find data type timeTVP_jdbc_avkash.hirpara_'07614d0f-aa59-4
617-bc92-10a56ba6713e.
        at com.microsoft.sqlserver.jdbc.datatypes.DateAndTimeTypeTest.testQueryTimeTVP(DateAndTimeTypeTest.java:175)

[ERROR] testQueryTimestampTVP  Time elapsed: 0.233 s  <<< ERROR!
com.microsoft.sqlserver.jdbc.SQLServerException: Column, parameter, or variable @P0. : Cannot find data type timestampTVP_jdbc_avkash.hirpara_'9f04b8d3-3
004-4222-b49d-e836563730eb.
        at com.microsoft.sqlserver.jdbc.datatypes.DateAndTimeTypeTest.testQueryTimestampTVP(DateAndTimeTypeTest.java:151)

I am following all the instruction as per the readme.md. 

Thanks, 
Avkash

lilgreenbird:
hi @Avkashhirpara,

Can you tell us how you are building the driver? it looks like the errors are from the tests are these the only failure you are seeing or you can not run any of the junit tests at all? What is the connection string you are using?

[...]

PR #2607, opened 2025-02-17
Fix for build failing (#1611)

jocstar:
Fix for tests failing during a build due to data type not found errors occurring when user name property contains a period (#1611)


src/test/java/com/microsoft/sqlserver/jdbc/RandomUtil.java
```diff
@@ -41,7 +41,7 @@ static public String getIdentifier(String prefix, int maxLength, boolean unique,
         sb.append(prefix);
         sb.append("_");
         sb.append("jdbc_");
-        sb.append(System.getProperty("user.name"));
+        sb.append(System.getProperty("user.name").replace(".", "_"));
```
lilgreenbird, line 44: there are other characters which are not valid characters if we are to strip 1 then should get rid of all. Alternatively could just get rid of this altogether

Issue #553, opened 2017-11-16
CallableStatement: using named parameters to specify only certain parameter values


gordthompson:
(ref: [this Stack Overflow question](https://stackoverflow.com/q/47312673/2144390))

Section 13.3.2 of the JDBC 4.2 specification says

> Named parameters can be used to specify only the values that have no default value.
>
> In CODE EXAMPLE 13-21, the procedure COMPLEX_PROC takes ten parameters, but only the first and fifth parameters, PARAM_1 and PARAM_5, are required.
>
> CallableStatement cstmt = con.prepareCall("{CALL COMPLEX_PROC(?, ?)}";
> cstmt.setString("PARAM_1", "Price");
> cstmt.setFloat("PARAM_5", 150.25);

I just tried this with 6.3.4.jre8-preview.jar and I couldn't get it to work. It seemed that I had to provide a parameter placeholder for every possible parameter, (`(?,?,?,?,?,?,?,?,?,?)` in the above example), and I didn't see a way to specify `DEFAULT` for a parameter that I would have otherwise simply omitted.

In other words, it seems like named parameter support for CallableStatement in mssql-jdbc is currently limited to allowing us to specify the parameters in any order, but not to "pick and choose" the parameters we actually supply.

Have I missed something?

ulvii:
Hi @gordthompson ,
Thank you for creating the issue, I can confirm that this is a deficiency in the implementation.  #547 is potentially related, we will address the issue once/when the PR is merged.

JaapD:
@cheenamalhotra  It looks like it is in progress for more than two years. Is there any progress?

peterbae:
Hi @JaapD, we haven't been able to work on this, the team will let you know when we have an update.

kliakos:
Any update on whether this will be addressed soon?

lilgreenbird:
hi @kliakos 
Sorry to report that this issue is still in our backlog it didn't make it to the top of the list in triage.

jocstar:
As part of a large update we included migrating from JTDS to the MS JDBC driver but hit this problem in testing. This is a big issue for us so I have been looking into the possibility of providing a patch for it. I do have a proof of concept test working that will probably be suitable for our internal use but have a question before looking to take it any further publicly.

1) According to the JDBC4.1 spec - 
```
It is not possible to combine setting parameters with ordinals and with names in the same statement. 
If ordinals and names are used for parameters in the same statement, an SQLException is thrown
```
[...]

PR #2616, opened 2025-02-26
Changed CallableStatement to allow you to not provide a named paramet…

jocstar:
…er for parameters with a default value (#553)


jocstar:

jocstar:
I have found a problem with my work where I hadn't accounted for the possibility that the user had already named the parameters in the SQL which leads to problems if they aren;t in the same order  e.g.
CALL TESTPROC(@NAME=?, @ID=?)
stmt.setInt("@ID", 123);   stmt.registerOutParameter("@NAME")

PR #2618, opened 2025-02-27
Blackrock: Added QUOTED_IDENTIFIER and CONCAT_NULL_YIELDS_NULL flag in connection

divang:
Moved the QUOTED_IDENTIFIER and CONCAT_NULL_YIELDS_NULL flag setting in initResettableValues() method to handle the resetting of QUOTED_IDENTIFIER and CONCAT_NULL_YIELDS_NULL flag for both - new and pooled connection.


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -1828,6 +1828,40 @@ final void initResettableValues() {
...
+                quotedIdentifierValue = SQLServerDriverStringProperty.QUOTED_IDENTIFIER.getDefaultValue();
+                activeConnectionProperties.setProperty(quotedIdentifierProperty, quotedIdentifierValue);
+            }
+
+            String quotedIdentifierOption = OnOffOption.valueOfString(quotedIdentifierValue).toString();
+            if (quotedIdentifierOption.compareToIgnoreCase(OnOffOption.OFF.toString()) == 0) {
+                connectionCommand("SET QUOTED_IDENTIFIER OFF", "quotedIdentifier");
+            } else if (quotedIdentifierOption.compareToIgnoreCase(OnOffOption.ON.toString()) == 0) {
+                connectionCommand("SET QUOTED_IDENTIFIER ON", "quotedIdentifier");
+            }
```
David-Engel, line 1846: (Verify) I believe this will always be the default value for a JDBC connection and we don't need to send the ON command. The command is a perf hit, so we don't want to take the hit, if we don't need to.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -1828,6 +1828,40 @@ final void initResettableValues() {
...
+                concatNullYieldsNullValue = SQLServerDriverStringProperty.CONCAT_NULL_YIELDS_NULL.getDefaultValue();
+                activeConnectionProperties.setProperty(concatNullYieldsNullProperty, concatNullYieldsNullValue);
+            }
+            String concatNullYieldsNullOption = OnOffOption.valueOfString(concatNullYieldsNullValue).toString();
+            if (concatNullYieldsNullOption.compareToIgnoreCase(OnOffOption.OFF.toString()) == 0) {
+                connectionCommand("SET CONCAT_NULL_YIELDS_NULL OFF", "concatNullYieldsNull");
+
+            } else if (concatNullYieldsNullOption.compareToIgnoreCase(OnOffOption.ON.toString()) == 0) {
+                connectionCommand("SET CONCAT_NULL_YIELDS_NULL ON", "concatNullYieldsNull");
+            }
```
David-Engel, line 1861: (Verify) Similarly, this is the default state and we don't need to send ON.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -3534,7 +3568,9 @@ else if (0 == requestedPacketSize)
             if (connectionlogger.isLoggable(Level.FINER)) {
                 connectionlogger.finer(toString() + " End of connect");
             }
-        } catch (SocketException e) {
+        } catch (
+
+        SocketException e) {
```
David-Engel, line 3573:
```suggestion
        } catch (SocketException e) {
```

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
divang: Yes, server side it is default.

SET CONCAT_NULL_YIELDS_NULL OFF and the CONCAT_NULL_YIELDS_NULL OFF database option are deprecated. Starting with SQL Server 2017 (14.x), CONCAT_NULL_YIELDS_NULL is always set to ON. Deprecated features shouldn't be used in new applications. 

Ref: https://learn.microsoft.com/en-us/sql/t-sql/statements/set-concat-null-yields-null-transact-sql?view=sql-server-ver16

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
divang: fixed.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
divang: Yes, it is ON by default at server side.
From documentation-
When SET QUOTED_IDENTIFIER is ON (default)...
Ref: https://learn.microsoft.com/en-us/sql/t-sql/statements/set-quoted-identifier-transact-sql?view=sql-server-ver16#remarks

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -1828,6 +1828,36 @@ final void initResettableValues() {
         sqlWarnings = null;
         sCatalog = originalCatalog;
         databaseMetaData = null;
+
+        try{
```
lilgreenbird, line 1832: can we get the values once? initResetableValues gets called multiple times

lilgreenbird (dismissed):
there are formatting issues, looks like the formatter wasn't run

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
divang: Can we have three state of those QUOTED_IDENTIFIER and CONCAT_NULL_YIELDS_NULL?

ON
When the customer wants to enable the flag, every initResettableValues() method will call SET <flag name>. This will make the JDBC driver more chatty. 
Example: jdbc:sqlserver://DIVANGSHARMAPC:1433;quotedIdentifier=ON;userName=divang25;

OFF
When the customer wants to disable the flag, every initResettableValues() method will call SET <flag name>. This will make the JDBC driver more chatty. 
Example: jdbc:sqlserver://DIVANGSHARMAPC:1433;quotedIdentifier=OFF;userName=divang25;

NOT_SET (JDBC driver side default setting)
When the customer goes with the default server setting, the initResettableValues() method will not call SET <flag name>. 
Example: jdbc:sqlserver://DIVANGSHARMAPC:1433;userName=divang25; (flag is not set)

David-Engel: > How do we deal with a scenario where the option was turned OFF explicitly by the application and then application closes (connection going back to pool) the connection and again gets a connection (from the pol), the expectation of the application would be that this option is ON as set on the connection string, however it would find it to be OFF ?

@machavan 
This ties into the comment I made internally about the fODBC flag in OptionFlags2 of LOGIN7. fODBC=1 sets ANSI_DEFAULTS=ON during initial connection and thus RESETCONNECTION, which in turn sets QUOTED_IDENTIFIER=ON. (I thought ANSI_DEFAULTS=ON also set CONCAT_NULL_YIELDS_NULL=ON, but the docs don't indicate that, so we should validate actual behavior.) This means, without sending any SET commands after RESETCONNECTION, we should always already be in a state of QUOTED_IDENTIFIER=ON (and maybe CONCAT_NULL_YIELDS_NULL=ON - need to verify).

src/test/java/com/microsoft/sqlserver/jdbc/SQLServerConnectionTest.java
```diff
@@ -462,6 +468,126 @@ public void testConnectionPoolGetTwice() throws SQLException {
...
+     * @throws SQLException
+     */
+    @Test
+    public void testConcatNullYieldsNull() throws SQLException {
+        SQLServerDataSource ds = new SQLServerDataSource();
+        ds.setURL(connectionString);
+        ds.setConcatNullYieldsNull("OFF");
+        int expectedResultForNewConnection = 0;
+        // Server default is CONCAT_NULL_YIELDS_NULL = ON  
+        int expectedResultForPooledConnection = 1;
```
David-Engel, line 482: When we set concat null yields null to OFF, the expected result == 0 for all connections, even pooled connections. If that's not happening, we need to figure out what's not working correctly in the implementation so it can be fixed.

src/test/java/com/microsoft/sqlserver/jdbc/SQLServerConnectionTest.java
```diff
@@ -462,6 +468,126 @@ public void testConnectionPoolGetTwice() throws SQLException {
...
+
+        try (Connection con = ds.getConnection(); Statement stmt = con.createStatement()) {
+            try (ResultSet rs = stmt.executeQuery(sqlSelect)) {
+                if (rs.next()) {
+                    assertEquals(expectedResultForNewConnection, rs.getInt(1));
+                } else {
+                    assertTrue(false, "Expected row of data was not found.");
+                }
+            }
+        }
```
David-Engel, line 494: I'd extract this out into a helper that takes a Connection object so you don't have to duplicate this code so much.

src/test/java/com/microsoft/sqlserver/jdbc/SQLServerConnectionTest.java
```diff
@@ -462,6 +468,126 @@ public void testConnectionPoolGetTwice() throws SQLException {
...
+         // Repeat getConnection to put the physical connection through a RESETCONNECTION
+         try (Connection con = pc.getConnection(); Statement statement = con.createStatement()) {
+             try (ResultSet rs = statement.executeQuery(sqlSelect)) {
+                 if (rs.next()) {
+                     assertEquals(expectedResultForPooledConnection, rs.getInt(1));
+                 } else {
+                     assertTrue(false, "Expected row of data was not found.");
+                 }
+             }
+         }
```
David-Engel, line 521: Nit: Indentation for this block needs to be fixed.

src/test/java/com/microsoft/sqlserver/jdbc/SQLServerConnectionTest.java
```diff
@@ -462,6 +468,126 @@ public void testConnectionPoolGetTwice() throws SQLException {
         }
     }
 
+    /**
+     * Test connection properties: CONCAT_NULL_YIELDS_NULL with SQLServerXADataSource for new connection and pooled connection
+     * @throws SQLException
+     */
+    @Test
+    public void testConcatNullYieldsNull() throws SQLException {
```
David-Engel, line 476: High level: Both of these new tests should also test "ON" for each setting.

divang:
ADO pipeline is successful.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -4786,6 +4823,7 @@ boolean executeReconnectCommand(TDSCommand newCommand) throws SQLServerException
              */
             boolean commandComplete = false;
             try {
+                newCommand.createCounter(null, activeConnectionProperties);
```
David-Engel, line 4826: What's this change for?

David-Engel (commented):
Generally looks good to me. Just a couple questions/comments.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
divang: It is added by Terry and as per his comment - _Deadloack and null counter TDSCommand fix_

src/test/java/com/microsoft/sqlserver/jdbc/SQLServerConnectionTest.java
```diff
@@ -462,6 +468,130 @@ public void testConnectionPoolGetTwice() throws SQLException {
...
+        pdsWithOff.setURL(connectionString);
+        pdsWithOff.setConcatNullYieldsNull("OFF");
+
+        PooledConnection pcWithOff = pdsWithOff.getPooledConnection();
+        try {
+            testSessionPropertyValueHelper(pcWithOff.getConnection(), sessionPropertyName, expectedResultFlagOff);
+            // Repeat getConnection to put the physical connection through a RESETCONNECTION
+            testSessionPropertyValueHelper(pcWithOff.getConnection(), sessionPropertyName, expectedResultFlagOff);            
+        } catch (Exception e) {
+            fail(TestResource.getResource("R_unexpectedErrorMessage") + e.getMessage());
```
David-Engel, line 498:
```suggestion
```

David-Engel (approved):

machavan (approved):

divang:
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> 
>     
>   
> 
> >

divang:
> @divang please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

@microsoft-github-policy-service agree company="Microsoft"

divang:
@microsoft-github-policy-service agree company="Microsoft"

PR #2622, opened 2025-02-28
Remove scheme from URI before fetching path for CRL path check

Jeffery-Wasty:
When attempting to get the Path for use in Configurable Retry Logic patch check, this could fail if the scheme was not `file:`. This is resolved by removing the scheme component from the URI before feeding it into `Paths.get()`. This is done by checking if there is a scheme, and if so, removing anything before the first forward slash, including that slash,


Jeffery-Wasty:
Passes tests with run id = 110147.

machavan (approved):

David-Engel (approved):

tkyc (approved):

lilgreenbird (approved):

PR #2626, opened 2025-03-04
Adding OS, architecture and runtime details to applicationName

muskan124947:
By enhancing telemetry for SQL drivers, we can achieve a more nuanced understanding of the operating environments in use. This capability enables us to make strategic decisions about testing and support, ultimately improving our efficiency and reducing costs. Adopting a structured approach to telemetry collection will ensure we gather the most relevant and useful data, facilitating continuous improvement in our processes. 

Format 
{Driver Name}|{OS Type}|{Process Architecture}|{OS Details}|{Runtime Details} 

Maximum lengths (in Unicode characters) for each field are: 
Driver Name - 16 
OS Type - 10 
Process Architecture - 10 
OS Details - 44 
Runtime Details - 44

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDataSource.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDataSource.java
@@ -170,7 +170,7 @@ public class SQLServerDataSource
     @Override
     public String getApplicationName() {
         return getStringProperty(connectionProps, SQLServerDriverStringProperty.APPLICATION_NAME.toString(),
-        SQLServerDriverStringProperty.APPLICATION_NAME.getDefaultValue());
+        SQLServerDriver.constructedAppName);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDataSource.java
```diff
@@ -170,7 +170,7 @@ public void setApplicationName(String applicationName) {
     @Override
     public String getApplicationName() {
         return getStringProperty(connectionProps, SQLServerDriverStringProperty.APPLICATION_NAME.toString(),
-        SQLServerDriverStringProperty.APPLICATION_NAME.getDefaultValue());
+        SQLServerDriver.constructedAppName);
```
machavan, line 173: As decided to make a change only for interfaceLibName, let us keep the application name unchanged everywhere.

PR #2634, opened 2025-03-18
Vector datatype support

muskan124947:
**Description**
With the rise of AI and machine learning, the ability to handle vector data is crucial for applications like semantic search, recommendation systems, and more. By incorporating vector support directly into Microsoft SQL Server and Azure SQL Database, we eliminate the need for separate vector databases, streamlining data architecture and improving performance. 
The vector data type is designed to store vector data optimized for operations such as similarity search and machine learning applications. 

Vector support in Microsoft SQL Server JDBC driver is enabled through a new feature extension (FE identifier 0x0E) with a version handshake to negotiate vector capabilities with the server.

Since the standard JDBC specification does not include native support for the VECTOR type, the driver provides a custom implementation using the `microsoft.sql.Vector` class.

The vector object includes number of elements, a type indicator, and the actual data, as defined below:

    public enum VectorDimensionType { 
        FLOAT32 // 32-bit (single precision) float 
    }
    private VectorDimensionType vectorType; 
    private int dimensionCount; 
    private Object[] data;

This enhancement introduces support for a number of VECTOR data type operations, including inserts, selections, stored procedures, bulk inserts with `useBulkCopyForBatchInsert`, table-to-table bulk copies, CSV-to-table bulk copies, and table-valued parameters (TVP).

1. Insertion of vector data into table
```
@Test 
void insertVectorData() throws SQLException { 
        String insertSql = "INSERT INTO " + AbstractSQLGenerator.escapeIdentifier(tableName) + " (v) VALUES (?, ?)"; 
        Object[] data= new Float[] { 0.45f, 7.9f, 63.0f }; 
        Vector vector = new Vector(3, VectorDimensionType.FLOAT32, data); 
        //Vector vector = new Vector(3, 4, data); 
       //Vector nullVector = new Vector(3,4,null);
        try (PreparedStatement pstmt = connection.prepareStatement(insertSql))         { 
            pstmt.setObject(1, vector, microsoft.sql.Types.VECTOR);
            pstmt.executeUpdate(); 
        } 
}
```

2. Select vector data from table
```
@Test
    public void getVectorData() throws SQLException {
        String query = "SELECT v FROM " + AbstractSQLGenerator.escapeIdentifier(tableName);
        try (PreparedStatement stmt = connection.prepareStatement(query)) {
            try (ResultSet rs = stmt.executeQuery()) {
                assertTrue(rs.next(), "No result found for inserted vector.");
                ResultSetMetaData meta = rs.getMetaData();
                int columnCount = meta.getColumnCount();

                while (rs.next()) {
                    for (int i = 1; i <= columnCount; i++) {
                        String columnName = meta.getColumnName(i);
                        int columnType = meta.getColumnType(i); // from java.sql.Types

                        Object value = null;
                        switch (columnType) {
                            case Types.VARCHAR:
                            case Types.NVARCHAR:
                                value = rs.getString(i);
                                break;
                            case microsoft.sql.Types.VECTOR:
                                value = rs.getObject(i, microsoft.sql.Vector.class);
                        }

                        System.out.println(columnName + " = " + value + " (type: " + columnType + ")");
                    }
                    System.out.println("---");
                }
            }
        }
    }
```
 
3. Stored Procedure Call
```
@Test 
public void testVectorStoredProcedureInputOutput() throws SQLException { 
        createProcedure(); 
        try (SQLServerCallableStatement cstmt = (SQLServerCallableStatement) connection.prepareCall(call)) { 
            Vector vector= new Vector(3, VectorDimensionType.FLOAT32, new Float[]{0.5f, 1.0f, 1.5f}); 
            cstmt.setObject(1, vector, microsoft.sql.Types.VECTOR); 
            cstmt.registerOutParameter(2, microsoft.sql.Types.VECTOR, 3, 4);
            cstmt.execute(); 
        } 
   } 
```
   
4. Bulk Copy from CSV to table
```
@Test
    public void testBulkCopyVectorFromCSV() throws SQLException { 
        String dstTable = RandomUtil.getIdentifier("dstTable"); 
        String fileName = filePath + vectorInputCsvFile; 
                SQLServerBulkCSVFileRecord fileRecord = new SQLServerBulkCSVFileRecord(fileName, null, ",", true)) { 
            stmt.executeUpdate( 
                    "CREATE TABLE " + dstTable + " (id INT, vectorCol VECTOR(3));"); 
            fileRecord.addColumnMetadata(1, "vectoxrCol", microsoft.sql.Types.VECTOR, 3, 4);
        } 
} 
```

**Backward Compatibility**
If an application hasn't been updated to handle the VECTOR data type, the driver provides backward compatibility by allowing vector data types to be read using backward compatible types. This is controlled using the **vectorTypeSupport** connection string property.
Currently, the supported values are "off" (server sends vector types as JSON string data) and "v1" (server sends vector types of FLOAT32 as vector data). Default is "v1". Future values may be added as needed to support additional vector types. ("v2" may indicate additional support for FLOAT16 and/or INT32 vectors, for example).

**Testing**
1. Added test scenarios for all operations involving the VECTOR data type. Tests are implemented across the following classes: VectorTest, BulkCopyCSVTest, BulkCopyISQLServerBulkRecordTest, DatabaseMetaDataTest and BatchExecutionWithBulkCopyTest.
2. Vector-specific tests are tagged using the **vectorTest** exclusion tag to allow controlled execution.
3. All tests have been verified locally using a SQL Server build with vector support.
4. Code coverage will increase once vector related test cases are enabled.
_Note: Currently, tests tagged with vectorTest are excluded from ADO runs and Github checks. These will be enabled post the official build rollout._

**Performance**
Conducted performance benchmarking for vector data insertion and retrieval across different data volumes— 100, 1K, 10K, 100K, and 1M records.
Comparative analysis was performed locally to evaluate performance across various scenarios.

### 📊 Performance Metrics  
*(dimensionCount = 1998; includes serialization and deserialization time)*

| Number of Records | Insert Operation (ms) | Insert with `useBulkCopyForBatchInsert` (ms) | Bulk Copy Operation (ms) |
|-------------------|------------------------|----------------------------------------------|---------------------------|
| 1                 | 11                     | 202                                          | 249                       |
| 100               | 215                    | 243                                          | 275                       |
| 1,000             | 991                    | 592                                          | 598                       |
| 10,000            | 6,089                  | 1,881                                        | 2,026                     |
| 100,000           | 52,126                 | 14,226                                       | 14,502                    |
| 1,000,000         | 492,570                | 143,871                                      | 153,397                   |


src/main/java/com/microsoft/sqlserver/jdbc/DataTypes.java
```diff
@@ -266,7 +269,10 @@ enum GetterConversion {
 
         SQL_VARIANT(SSType.Category.SQL_VARIANT, EnumSet.of(JDBCType.Category.CHARACTER, JDBCType.Category.SQL_VARIANT,
                 JDBCType.Category.NUMERIC, JDBCType.Category.DATE, JDBCType.Category.TIME, JDBCType.Category.BINARY,
-                JDBCType.Category.TIMESTAMP, JDBCType.Category.NCHARACTER, JDBCType.Category.GUID));
+                JDBCType.Category.TIMESTAMP, JDBCType.Category.NCHARACTER, JDBCType.Category.GUID)),
+
+        VECTOR(SSType.Category.VECTOR, EnumSet.of(JDBCType.Category.BINARY, JDBCType.Category.LONG_BINARY,
+                JDBCType.Category.CHARACTER, JDBCType.Category.LONG_CHARACTER, JDBCType.Category.BLOB, JDBCType.Category.VECTOR));
```
muskan124947, line 275: Need to clean getter and setter conversions in DataTypes class

src/main/java/com/microsoft/sqlserver/jdbc/Column.java
```diff
@@ -189,6 +191,12 @@ Object getValue(JDBCType jdbcType, InputStreamGetterArgs getterArgs, Calendar ca
             SQLServerStatement statement) throws SQLServerException {
         Object value = getterDTV.getValue(jdbcType, typeInfo.getScale(), getterArgs, cal, typeInfo, cryptoMetadata,
                 tdsReader, statement);
+        if (jdbcType == JDBCType.VECTOR && value == null) {
+            int precision = typeInfo.getPrecision(); 
+            VectorDimensionType scale = typeInfo.getScale() == 4 ? VectorDimensionType.F32 : VectorDimensionType.F16;
```
divang, line 196: Could we move all hard-coded values to constant variables with intuitive and descriptive names?

src/main/java/com/microsoft/sqlserver/jdbc/DataTypes.java
divang: Can we add test cases to validate all data types in this list, including their conversions?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCSVFileRecord.java
```diff
@@ -421,6 +421,40 @@ public Object[] getRowData() throws SQLServerException {
...
 
+                        case microsoft.sql.Types.VECTOR: {
+                            int dimensionCount = cm.precision;
+                        
+                            if (dimensionCount <= 0) {
+                                throw new SQLServerException("Invalid vector dimension count.", null, 0, null);
+                            }
+                        
+                            String vectorData = data[pair.getKey() - 1].trim();
+                            microsoft.sql.Vector.VectorDimensionType vectorDimensionType = microsoft.sql.Vector.getVectorDimensionType(cm.scale);
```
divang, line 432: Is it possible to import the microsoft.sql.Vector package?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCSVFileRecord.java
```diff
@@ -421,6 +421,40 @@ public Object[] getRowData() throws SQLServerException {
...
+                        
+                            if ("NULL".equalsIgnoreCase(vectorData) || "\"NULL\"".equalsIgnoreCase(vectorData)) {
```
divang, line 434: "\"NULL\" and "NULL" should be constant variable

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCSVFileRecord.java
```diff
@@ -421,6 +421,40 @@ public Object[] getRowData() throws SQLServerException {
...
                         }
 
+                        case microsoft.sql.Types.VECTOR: {
+                            int dimensionCount = cm.precision;
+                        
+                            if (dimensionCount <= 0) {
+                                throw new SQLServerException("Invalid vector dimension count.", null, 0, null);
+                            }
+                        
+                            String vectorData = data[pair.getKey() - 1].trim();
```
divang, line 431: pair.getKey() - 1 is repeated multiple times. Store it in a variable to improve readability and performance.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCSVFileRecord.java
```diff
@@ -421,6 +421,40 @@ public Object[] getRowData() throws SQLServerException {
...
+                            microsoft.sql.Vector.VectorDimensionType vectorDimensionType = microsoft.sql.Vector.getVectorDimensionType(cm.scale);
+                        
+                            if ("NULL".equalsIgnoreCase(vectorData) || "\"NULL\"".equalsIgnoreCase(vectorData)) {
+                                dataRow[pair.getKey() - 1] = new microsoft.sql.Vector(dimensionCount, vectorDimensionType, null);
+                            } else {
+                                // Remove enclosing brackets or braces if present
+                                if ((vectorData.startsWith("[") && vectorData.endsWith("]")) || 
+                                    (vectorData.startsWith("{") && vectorData.endsWith("}"))) {
```
divang, line 439: The logic for removing brackets/braces is duplicated. Consider extracting it into a helper method for reusability. Ex:
private String removeEnclosingBrackets(String data) {
    if ((data.startsWith("[") && data.endsWith("]")) || (data.startsWith("{") && data.endsWith("}"))) {
        return data.substring(1, data.length() - 1);
    }
    return data;
}

src/main/java/com/microsoft/sqlserver/jdbc/dtv.java
```diff
@@ -2992,6 +3007,29 @@ public void apply(TypeInfo typeInfo, TDSReader tdsReader) throws SQLServerExcept
...
+             * @throws SQLServerException
+             *         when an error occurs
+             */
+            public void apply(TypeInfo typeInfo, TDSReader tdsReader) throws SQLServerException {
+                typeInfo.ssLenType = SSLenType.USHORTLENTYPE;
+                typeInfo.maxLength = tdsReader.readUnsignedShort();
+                typeInfo.displaySize = typeInfo.maxLength;
+                typeInfo.ssType = SSType.VECTOR;
+                int scaleByte = tdsReader.readUnsignedByte(); // Read the dimension type (scale)
+                typeInfo.scale = (scaleByte == 0) ? 4 : 2;
```
divang, line 3029: Please use the constant that will be created as suggested in the comment above.

src/main/java/microsoft/sql/Vector.java
```diff
@@ -0,0 +1,176 @@
...
+public final class Vector implements java.io.Serializable {
+    public enum VectorDimensionType {
+        F16, // 16-bit (half precision) float
+        F32 // 32-bit (single precision) float
+    }
+
+    public VectorDimensionType vectorType;
+    public int dimensionCount;
+
+    public float[] data;
```
divang, line 20: Can those variables be made private?

src/main/java/microsoft/sql/Vector.java
```diff
@@ -0,0 +1,176 @@
...
+     */
+    public static microsoft.sql.Vector fromBytes(byte[] bytes) {
+        if (bytes == null) {
+            return null;
+        }
+        if (bytes.length < 8) {
+            throw new IllegalArgumentException("Byte array length must be at least 8 bytes.");
+        }
+        if (bytes.length % 4 != 0) {
+            throw new IllegalArgumentException("Byte array length must be a multiple of 4.");
```
divang, line 40: Can we add the below error message in SQLServerResource.java
{"R_InvalidVectorLength", "Byte array length must be at least \"{0}\" bytes."}

src/main/java/com/microsoft/sqlserver/jdbc/Parameter.java
```diff
@@ -623,6 +633,17 @@ private void setTypeDefinition(DTV dtv) {
                         param.typeDefinition = VARBINARY_8K;
                     break;
 
+                case VECTOR:
+                    param.typeDefinition = SSType.VECTOR.toString();
+                    int precision = 0;
+                    if (param.isOutput() && scale < param.getOutScale())
+                        precision = param.getValueLength();
+                    else if (dtv.getSetterValue() != null) {
```
machavan, line 641: This logic should be moved into vector in a method getTypeDefinition as it is a vector concern

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
```diff
@@ -1039,6 +1040,13 @@ private void writeTypeInfo(TDSWriter tdsWriter, int srcJdbcType, int srcScale, i
                 }
                 break;
 
+            case microsoft.sql.Types.VECTOR: // 0xF5
+                tdsWriter.writeByte(TDSType.VECTOR.byteValue());
+                tdsWriter.writeShort((short) (8 + srcScale * srcPrecision)); //length
```
machavan, line 1045: Use the static methods in vector class to encapsulate vector concerns for both length computation and returning scale byte.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
```diff
@@ -3552,6 +3583,15 @@ private byte[] normalizedValue(JDBCType destJdbcType, Object value, JDBCType src
                         throw new SQLServerException(this, form.format(msgArgs), null, 0, false);
                     }
                     return byteArrayValue;
+                case VECTOR:
+                    Vector vector = (Vector) value;
+                    byteValue = vector.toBytes();
+                    if (byteValue.length > vector.getActualLength()) {
+                        MessageFormat form = new MessageFormat(SQLServerException.getErrString("R_InvalidDataForAE"));
```
machavan, line 3590: Used error string seems wrong : R_INvalidDataForAE.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCallableStatement.java
```diff
@@ -389,6 +389,9 @@ public void registerOutParameter(int index, int sqlType, int scale) throws SQLSe
 
         registerOutParameter(index, sqlType);
         inOutParam[index - 1].setOutScale(scale);
+        if (microsoft.sql.Types.VECTOR == sqlType) {
+            inOutParam[index - 1].setValueLength(4); // default 32-bit (single precision) float
```
machavan, line 393: Add a new static method in vector getDefaultPrecision() and return 4 from there

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDataTable.java
```diff
@@ -318,6 +320,22 @@ private void internalAddrow(JDBCType jdbcType, Object val, Object[] rowValues,
                     internalAddrow(internalJDBCType, val, rowValues, pair);
                     break;
 
+                case VECTOR:
+                    nValueLen = ((Vector) val).getActualLength();
+                    int scaleByte = ((Vector) val).getScale();
+                    int scale = Vector.getbytesPerDimensionFromScale(scaleByte);
```
machavan, line 326: method name getbytesPerDimensionFromScale -> getBytesPerDimensionFromScale

src/main/java/com/microsoft/sqlserver/jdbc/dtv.java
```diff
@@ -2992,6 +3009,29 @@ public void apply(TypeInfo typeInfo, TDSReader tdsReader) throws SQLServerExcept
...
+             *         when an error occurs
+             */
+            public void apply(TypeInfo typeInfo, TDSReader tdsReader) throws SQLServerException {
+                typeInfo.ssLenType = SSLenType.USHORTLENTYPE;
+                typeInfo.maxLength = tdsReader.readUnsignedShort();
+                typeInfo.displaySize = typeInfo.maxLength;
+                typeInfo.ssType = SSType.VECTOR;
+                int scaleByte = tdsReader.readUnsignedByte(); // Read the dimension type (scale)
+                typeInfo.scale = Vector.getbytesPerDimensionFromScale(scaleByte);
+                typeInfo.precision = (typeInfo.maxLength - 8) / typeInfo.scale;
```
machavan, line 3032: Add a static method in Vector class getPrecision and move this logic  there

machavan (commented):
Posting first pass of review comments

src/main/java/com/microsoft/sqlserver/jdbc/Parameter.java
```diff
@@ -623,6 +633,17 @@ private void setTypeDefinition(DTV dtv) {
                         param.typeDefinition = VARBINARY_8K;
                     break;
 
+                case VECTOR:
+                    param.typeDefinition = SSType.VECTOR.toString();
```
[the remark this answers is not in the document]
muskan124947: Here toString() is different. It isn't the one defined in Vector class.

```
enum SSType {
    VECTOR(Category.VECTOR, "vector", JDBCType.VECTOR);
} // where name is "vector"

private SSType(Category category, String name, JDBCType jdbcType) {
        this.category = category;
        this.name = name;
        this.jdbcType = jdbcType;
    }
    
@Override
    public String toString() {
        return name;
    }

```

src/main/java/com/microsoft/sqlserver/jdbc/DataTypes.java
muskan124947: Removed explicit vector mapping since operations work correctly without it.

src/main/java/com/microsoft/sqlserver/jdbc/DataTypes.java
```diff
@@ -795,7 +806,13 @@ enum SetterConversion {
...
         GEOMETRY(JDBCType.Category.GEOMETRY, EnumSet.of(JDBCType.Category.GEOMETRY)),
 
-        GEOGRAPHY(JDBCType.Category.GEOGRAPHY, EnumSet.of(JDBCType.Category.GEOGRAPHY));
+        GEOGRAPHY(JDBCType.Category.GEOGRAPHY, EnumSet.of(JDBCType.Category.GEOGRAPHY)),
+
+        VECTOR(JDBCType.Category.VECTOR, EnumSet.of(JDBCType.Category.NUMERIC, JDBCType.Category.DATE,
+        JDBCType.Category.TIME, JDBCType.Category.TIMESTAMP, JDBCType.Category.CHARACTER,
+        JDBCType.Category.LONG_CHARACTER, JDBCType.Category.NCHARACTER, JDBCType.Category.LONG_NCHARACTER,
+        JDBCType.Category.BINARY, JDBCType.Category.LONG_BINARY, JDBCType.Category.BLOB, JDBCType.Category.GUID,
+        JDBCType.Category.SQL_VARIANT, JDBCType.Category.VECTOR));
```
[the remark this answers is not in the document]
muskan124947: Removed explicit vector mapping since operations work correctly without it.

src/main/java/com/microsoft/sqlserver/jdbc/DataTypes.java
```diff
@@ -895,7 +912,13 @@ enum UpdaterConversion {
...
                 SSType.Category.NCHARACTER, SSType.Category.LONG_NCHARACTER)),
 
-        SQL_VARIANT(JDBCType.Category.SQL_VARIANT, EnumSet.of(SSType.Category.SQL_VARIANT));
+        SQL_VARIANT(JDBCType.Category.SQL_VARIANT, EnumSet.of(SSType.Category.SQL_VARIANT)),
+
+        VECTOR(JDBCType.Category.VECTOR, EnumSet.of(SSType.Category.NUMERIC, SSType.Category.DATETIME,
+        SSType.Category.CHARACTER, SSType.Category.LONG_CHARACTER, SSType.Category.NCHARACTER,
+        SSType.Category.LONG_NCHARACTER, SSType.Category.XML, SSType.Category.BINARY,
+        SSType.Category.LONG_BINARY, SSType.Category.UDT, SSType.Category.TIMESTAMP, SSType.Category.GUID,
+        SSType.Category.SQL_VARIANT, SSType.Category.VECTOR));
```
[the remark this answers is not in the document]
muskan124947: Removed explicit vector mapping since operations work correctly without it.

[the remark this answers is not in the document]
muskan124947: Updated the vector mapping, also the existing test cases validates the mappings added 
If these are not correctly added, then the operations fails

src/main/java/com/microsoft/sqlserver/jdbc/DataTypes.java
muskan124947: Updated the vector mapping, also the existing test cases validates the mappings added 
If these are not correctly added, then the operations fails

David-Engel (commented):
Partial review (about 1/3 of files).

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -6745,6 +6825,24 @@ private void onFeatureExtAck(byte featureId, byte[] data) throws SQLServerExcept
                 sessionRecovery.setConnectionRecoveryPossible(true);
                 break;
             }
+
+            case TDS.TDS_FEATURE_EXT_VECTORSUPPORT: {
+                if (connectionlogger.isLoggable(Level.FINER)) {
```
David-Engel:
```suggestion
                if (connectionlogger.isLoggable(Level.FINE)) {
```

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -7935,7 +8037,7 @@ public <T> T unwrap(Class<T> iface) throws SQLException {
 
     /** original bulkCopyForBatchInsertAllowEncryptedValueModifications flag */
     private boolean originalBulkCopyForBatchInsertAllowEncryptedValueModifications;
-
+
```
David-Engel: Nit: Stray empty space.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnectionPoolProxy.java
```diff
@@ -892,4 +892,23 @@ public void setBulkCopyForBatchInsertAllowEncryptedValueModifications(boolean bu
...
+     */
+    public String getVectorTypeSupport() {
+        return wrappedConnection.getVectorTypeSupport();
+    }
+
+    /**
+     * Sets the vectorTypeSupport value.
+     *
+     * @param vectorTypeSupport
+     *                          the vectorTypeSupport value to set ("off" or "v1").
```
David-Engel: Nit: Spacing
```suggestion
     *        the vectorTypeSupport value to set ("off" or "v1").

```

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -5611,6 +5663,34 @@ int writeDNSCacheFeatureRequest(boolean write, /* if false just calculates the l
...
     }
 
+    /**
+     * Writes the Vector Support feature request to the physical state object,
+     * unless vectorTypeSupport is "off". The request includes the feature ID,
+     * feature data length, and version number.
+     * 
+     * @param write
+     *                  If true, writes the feature request to the physical state
+     *                  object.
```
David-Engel: Nit: Spacing
```suggestion
     *        If true, writes the feature request to the physical state object.
```

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDataTable.java
```diff
@@ -324,6 +326,22 @@ else if (val instanceof OffsetDateTime)
...
+                    if (scale > currentColumnMetadata.scale) {
+                        currentColumnMetadata.scale = scale;
+                        columnMetadata.put(pair.getKey(), currentColumnMetadata);
+                    }
+                    if (precision > currentColumnMetadata.precision) {
+                        currentColumnMetadata.precision = precision;
+                        columnMetadata.put(pair.getKey(), currentColumnMetadata);
+                    }
+                    rowValues[key] = ((Vector) val).toBytes();
+                    break;
```
David-Engel: You're casting val to Vector multiple times. This is inefficient. Consider assigning it to a local variable.

Also, you are potentially updating the columnMetadata map twice when once would be more efficient.

```suggestion
                case VECTOR:
                    Vector vectorVal = (Vector) val;
                    nValueLen = vectorVal.getVectorLength();
                    int scaleByte = vectorVal.getScaleByte();
                    int scale = Vector.getBytesPerDimensionFromScale(scaleByte);
                    precision = vectorVal.getDimensionCount();
                    if (scale > currentColumnMetadata.scale || precision > currentColumnMetadata.precision) {
                        if (scale > currentColumnMetadata.scale) {
                            currentColumnMetadata.scale = scale;
                        }
                        if (precision > currentColumnMetadata.precision) {
                            currentColumnMetadata.precision = precision;
                        }
                        columnMetadata.put(pair.getKey(), currentColumnMetadata);
                    }

                    rowValues[key] = vectorVal.toBytes();
                    break;
```

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -1691,7 +1691,13 @@ public final void setObject(int n, Object obj, int jdbcType) throws SQLServerExc
         if (microsoft.sql.Types.STRUCTURED == jdbcType) {
             tvpName = getTVPNameFromObject(n, obj);
         }
-        setObject(setterGetParam(n), obj, JavaType.of(obj), JDBCType.of(jdbcType), null, null, false, n, tvpName);
+        int precision = 0, scale =0;
```
David-Engel: Nit
```suggestion
        int precision = 0, scale = 0;
```

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -1691,7 +1691,13 @@ public final void setObject(int n, Object obj, int jdbcType) throws SQLServerExc
...
+        if (microsoft.sql.Types.VECTOR == jdbcType) {
+            precision = microsoft.sql.Vector.valueOf(obj).getDimensionCount();
+            scale = microsoft.sql.Vector.valueOf(obj).getScaleByte();
+        }
+
+        setObject(setterGetParam(n), obj, JavaType.of(obj), JDBCType.of(jdbcType), scale, precision, false, n, tvpName);
```
David-Engel, line 1691: Slight parameter value change. Will setObject behave differently for other types if we pass 0 instead of null for scale and precision?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResource.java
```diff
@@ -339,6 +340,11 @@ protected Object[][] getContents() {
         {"R_unsupportedStmtColEncSetting", "SQLServerStatementColumnEncryptionSetting cannot be null."},
         {"R_unsupportedConversionAE", "The conversion from {0} to {1} is unsupported for encrypted column."},
         {"R_InvalidDataForAE", "The given value of type {0} from the data source cannot be converted to type {1} of the specified target column {2}."},
+        {"R_vectorTypeSupportPropertyDescription", "Determines the vector feature extension support during connection initialization. Valid values are \"off\" (do not send FE for vector) and \"v1\" (send FE for vector v1). Default is \"v1\"."},
```
David-Engel: I don't see this used. Does it need to be removed?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResource.java
```diff
@@ -313,6 +313,7 @@ protected Object[][] getContents() {
         {"R_StreamingDataTypeAE", "Data of length greater than {0} is not supported in encrypted {1} column."},
         {"R_AE_NotSupportedByServer", "SQL Server in use does not support column encryption."},
         {"R_InvalidAEVersionNumber", "Received invalid version number \"{0}\" for Always Encrypted."}, // From server
+        {"R_InvalidVectorVersionNumber", "Received invalid version number \"{0}\" for Vector."},
```
David-Engel: Recommend:
```suggestion
        {"R_InvalidVectorVersionNumber", "Received invalid version number \"{0}\" for Vector feature negotiation."},
```

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResource.java
```diff
@@ -480,6 +486,7 @@ protected Object[][] getContents() {
         {"R_InvalidDataClsVersionNumber", "Invalid version number {0} for Data Classification."}, // From Server
         {"R_unknownUTF8SupportValue", "Unknown value for UTF8 support."},
         {"R_unknownAzureSQLDNSCachingValue", "Unknown value for Azure SQL DNS Caching."},
+        {"R_unknownVectorSupportValue", "Unknown value for Vector support."},
```
David-Engel: We haven't been very consistent with the format of this error message in the past... I think this would be better:
```suggestion
        {"R_unknownVectorSupportValue", "Unexpected version value received for Vector support feature negotiation."},
```

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResultSet.java
```diff
@@ -3062,7 +3077,7 @@ public SQLXML getSQLXML(String columnLabel) throws SQLException {
         loggerExternal.exiting(getClassNameLogging(), "getSQLXML", xml);
         return xml;
     }
-
+
```
David-Engel: Nit: Stray white space added

src/main/java/microsoft/sql/Vector.java
```diff
@@ -0,0 +1,327 @@
...
+
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+import java.util.Arrays;
+import java.util.ResourceBundle;
+import java.text.MessageFormat;
+
+public final class Vector implements java.io.Serializable {
+    public enum VectorDimensionType {
+        float16, // 16-bit (half precision) float
```
David-Engel: Do we even want to expose float16, yet? What are the implications if we expose it before the server supports it? What if we only add/expose it later? (Don't remove code yet, just want to have the conversation before making a decision.)

David-Engel (commented):
Partial review. 24/35 files.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResource.java
muskan124947: If we add any connection string property, then it requires to add its property description here in SQLServerResource class. Else build fails with below error
Exception java.util.MissingResourceException: Can't find resource for bundle com.microsoft.sqlserver.jdbc.SQLServerResource, key R_vectorTypeSupportPropertyDescription

src/main/java/microsoft/sql/Vector.java
David-Engel: As discussed, we'll leave float16 out until server support comes along. (Here and all places below.)

src/main/java/microsoft/sql/Vector.java
```diff
@@ -0,0 +1,327 @@
...
+     */
+    public Vector(int precision, int scale, Object[] data) {
+        this(precision, getVectorDimensionTypeFromScaleValue(scale), data);
+    }
+
+    /**
+     * Converts a byte array to a Vector object. The byte array must contain the following:
+     * 8 bytes for header and 4 bytes per float value.
+     */
+    public static microsoft.sql.Vector fromBytes(byte[] bytes) {
```
David-Engel: Do we need the package here?
```suggestion
    public static Vector fromBytes(byte[] bytes) {
```

src/main/java/microsoft/sql/Vector.java
```diff
@@ -0,0 +1,327 @@
...
+     * Converts the vector to a string representation.
+     * @return A string representation of the vector.
+     */
+    @Override
+    public String toString() {
+        return "VECTOR(" + vectorType + ", " + dimensionCount + ") : " +
+                (data != null ? Arrays.toString(data) : "null");
+    }
+
+    public static microsoft.sql.Vector valueOf(Object obj) {
```
David-Engel: Do we need the package name here?
```suggestion
    public static Vector valueOf(Object obj) {
```

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -1025,6 +1025,42 @@ public void setBulkCopyForBatchInsertAllowEncryptedValueModifications(
...
+     * Sets the value of the vectorTypeSupport connection property.
+     *
+     * @param vectorTypeSupport
+     * A string that indicates the vector type support during connection initialization.
+     * Valid values are "off" (vector types are returned as strings) and "v1" (vectors of type float32 are returned as vectors).  
+     * Default is "v1".
+     */
+    @Override
+    public void setVectorTypeSupport(String vectorTypeSupport) {
+        if (!"off".equalsIgnoreCase(vectorTypeSupport) && !"v1".equalsIgnoreCase(vectorTypeSupport)) {
```
divang: Please create a constant for "off" value

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -5611,6 +5663,32 @@ int writeDNSCacheFeatureRequest(boolean write, /* if false just calculates the l
...
+     * @param write
+     * If true, writes the feature request to the physical state object.
+     * @param tdsWriter
+     * @return
+     * The length of the feature request in bytes, or 0 if vectorTypeSupport is "off".
+     * @throws SQLServerException
+     */
+    int writeVectorSupportFeatureRequest(boolean write,
+            TDSWriter tdsWriter) throws SQLServerException {
+        if ("off".equalsIgnoreCase(vectorTypeSupport)) {
```
divang: Please use above created a constant in place of "off" as hard code value here.

src/main/java/microsoft/sql/Vector.java
```diff
@@ -0,0 +1,325 @@
...
+        VectorDimensionType vectorType = getVectorDimensionType(vectorTypeByte);
+        int bytesPerDimension = getBytesPerDimensionFromScale(vectorType);
+
+        if (bytes.length < 8) {
+            MessageFormat form = new MessageFormat(
+                ResourceBundle.getBundle("com.microsoft.sqlserver.jdbc.SQLServerResource")
+                    .getString("R_vectorByteArrayLength"));
+            throw new IllegalArgumentException(form.format(new Object[0]));
+        }
+                if ((bytes.length - 8) % bytesPerDimension != 0) {
```
divang: Please apply the formator.

src/test/java/com/microsoft/sqlserver/jdbc/databasemetadata/DatabaseMetaDataTest.java
```diff
@@ -1083,6 +1083,73 @@ public void testGetSchemasWithAndWithoutCatalog() throws SQLException {
...
+            String sql = "CREATE TABLE " + AbstractSQLGenerator.escapeIdentifier(vectorTableName)
+                    + " (c1 VECTOR(3) NULL);";
+            stmt.execute(sql);
+
+            // Query the table and retrieve metadata
+            String query = "SELECT * FROM " + AbstractSQLGenerator.escapeIdentifier(vectorTableName);
+            try (Statement statement = connection.createStatement();
+                    ResultSet resultSet = statement.executeQuery(query)) {
+
+                ResultSetMetaData metaData = resultSet.getMetaData();
```
machavan, line 1108: Let us also add test for Connection.getDatabaseMetadata (that internally queries sp_columns_100)

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyISQLServerBulkRecordTest.java
```diff
@@ -150,6 +157,480 @@ public void testBulkCopyDateTimePrecision() throws SQLException {
...
+
+        try (Connection connection = DriverManager.getConnection(connectionString);
+                Statement statement = connection.createStatement()) {
+
+            // Create the source table with a VECTOR column
+            statement.executeUpdate("CREATE TABLE " + vectorTable + " (vectorCol VECTOR(3))");
+
+            // Insert sample data into the source table
+            Object[] data = new Object[] { 1.0f, 2.0f, 3.0f };
+            microsoft.sql.Vector vector = new microsoft.sql.Vector(data.length, microsoft.sql.Vector.VectorDimensionType.float32, data);
```
David-Engel: Would this be more readable with an import? Actually, I see we import microsoft.sql.Vector now. Do we need the full name here? (I see the pattern repeated elsewhere in this file. Please address all, if needed.)

David-Engel:
My approval isn't needed to merge this PR, at this point. I think the surface looks good. None of the small issues I identified are blockers. Just make sure the rest of the team is good with it and that CI is green.

Thanks!

David

src/test/java/com/microsoft/sqlserver/jdbc/databasemetadata/DatabaseMetaDataTest.java
muskan124947: Currently, sp_columns_100 displays column type as varbinary for VECTOR column

David-Engel (dismissed):

src/main/java/com/microsoft/sqlserver/jdbc/DataTypes.java
```diff
@@ -895,7 +907,10 @@ enum UpdaterConversion {
                 SSType.Category.DATETIMEOFFSET, SSType.Category.CHARACTER, SSType.Category.LONG_CHARACTER,
                 SSType.Category.NCHARACTER, SSType.Category.LONG_NCHARACTER)),
 
-        SQL_VARIANT(JDBCType.Category.SQL_VARIANT, EnumSet.of(SSType.Category.SQL_VARIANT));
+        SQL_VARIANT(JDBCType.Category.SQL_VARIANT, EnumSet.of(SSType.Category.SQL_VARIANT)),
+
+        VECTOR(JDBCType.Category.VECTOR, EnumSet.of(SSType.Category.CHARACTER, SSType.Category.LONG_CHARACTER,
```
divang, line 916: In UpdaterConversion, we have added vector mapping only for CHARACTER, but not for LONG_CHARACTER, even though Vector is mapped to both CHARACTER and LONG_CHARACTER. Do we know the reason for this?
 ```
CHARACTER(JDBCType.Category.CHARACTER, EnumSet.of(SSType.Category.NUMERIC, SSType.Category.DATE,
 
                SSType.Category.TIME, SSType.Category.DATETIME, SSType.Category.DATETIME2,
 
                SSType.Category.DATETIMEOFFSET, SSType.Category.CHARACTER, SSType.Category.LONG_CHARACTER,
 
                SSType.Category.NCHARACTER, SSType.Category.LONG_NCHARACTER, SSType.Category.XML,
 
                SSType.Category.BINARY, SSType.Category.LONG_BINARY, SSType.Category.UDT, SSType.Category.GUID,
 
                SSType.Category.TIMESTAMP, SSType.Category.SQL_VARIANT, SSType.Category.VECTOR)),
 
 
        LONG_CHARACTER(JDBCType.Category.LONG_CHARACTER, EnumSet.of(SSType.Category.CHARACTER,
 
                SSType.Category.LONG_CHARACTER, SSType.Category.NCHARACTER, SSType.Category.LONG_NCHARACTER,
 
                SSType.Category.XML, SSType.Category.BINARY, SSType.Category.LONG_BINARY)),
 

 
VECTOR(JDBCType.Category.VECTOR, EnumSet.of(SSType.Category.CHARACTER, SSType.Category.LONG_CHARACTER,
 
                SSType.Category.VECTOR));
                ```

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -1025,6 +1028,42 @@ public void setBulkCopyForBatchInsertAllowEncryptedValueModifications(
...
+    /**
+     * Sets the value of the vectorTypeSupport connection property.
+     *
+     * @param vectorTypeSupport
+     * A string that indicates the vector type support during connection initialization.
+     * Valid values are "off" (vector types are returned as strings) and "v1" (vectors of type float32 are returned as vectors).  
+     * Default is "v1".
+     */
+    @Override
+    public void setVectorTypeSupport(String vectorTypeSupport) {
```
divang, line 1058: Switching to a switch-case statement for validating the vectorTypeSupport string makes the code more readable and easier to extend in the future (e.g., when adding new supported values). It also centralizes the validation logic, reducing the risk of errors if more options are introduced.
The null check and normalization of the input string are handled well.

```
@Override
public void setVectorTypeSupport(String vectorTypeSupport) {
    if (vectorTypeSupport == null) {
        MessageFormat form = new MessageFormat(SQLServerException.getErrString("R_invalidVectorTypeSupport"));
        Object[] msgArgs = { "null" };
        throw new IllegalArgumentException(form.format(msgArgs));
    }
    switch (vectorTypeSupport.trim().toLowerCase()) {
        case VECTOR_SUPPORT_OFF:
        case VECTOR_SUPPORT_V1:
            this.vectorTypeSupport = vectorTypeSupport.toLowerCase();
            break;
        default:
            MessageFormat form = new MessageFormat(SQLServerException.getErrString("R_invalidVectorTypeSupport"));
            Object[] msgArgs = { vectorTypeSupport };
            throw new IllegalArgumentException(form.format(msgArgs));
    }
}

File: src/main/java/com/microsoft/sqlserver/jdbc/Parameter.java [hunk 40]
```diff
@@ -623,6 +625,12 @@ private void setTypeDefinition(DTV dtv) {
                         param.typeDefinition = VARBINARY_8K;
                     break;
 
+                case VECTOR:
```
divang, line 628: dtv.getSetterValue() may return null if the parameter value was never set or was explicitly set to null. If VectorUtils.getTypeDefinition() does not handle a null Vector argument, this will result in an NPE.

```
Vector vectorValue = (Vector) dtv.getSetterValue();
if (vectorValue != null) {
    param.typeDefinition = VectorUtils.getTypeDefinition(
        vectorValue, scale, param.isOutput(), 
        param.getOutScale(), param.getValueLength());
} else {
    param.typeDefinition = null; // or some default
}

File: src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCSVFileRecord.java [hunk 41]
```diff
@@ -421,6 +423,39 @@ public Object[] getRowData() throws SQLServerException {
                             break;
                         }
 
+                        case microsoft.sql.Types.VECTOR: {
+                            int dimensionCount = cm.precision;
```
divang, line 428: Can we add a few sample Vector data examples as comments below the Vector switch case? For instance, how they would appear in a CSV file—such as ["10.1", "1.2"] or [3.4, 5.6]. This would help improve readability and understanding.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
```diff
@@ -2121,6 +2131,9 @@ private void writeNullToTdsWriter(TDSWriter tdsWriter, int srcJdbcType,
             case microsoft.sql.Types.SQL_VARIANT:
                 tdsWriter.writeInt((byte) 0x00);
                 return;
+            case microsoft.sql.Types.VECTOR:
+                tdsWriter.writeShort((short) -1);
```
divang: Can we move microsoft.sql.Types.VECTOR case to existing java.sql.Types.CHAR case.

Explanation:
-> tdsWriter.writeByte((byte) 0xFF); tdsWriter.writeByte((byte) 0xFF);
Writes two bytes, each with the value 0xFF (255 unsigned, -1 signed).
The output is: FF FF (two separate bytes).

-> tdsWriter.writeShort((short) -1);
Writes a single 16-bit short value, which is 0xFFFF in two’s complement.
The output is also: FF FF (as a single short, but written as two bytes in big-endian or little-endian order depending on the writer’s implementation).

->Key difference:
Both write the same two bytes (FF FF), but the first does it as two separate byte writes, the second as a single short write (which is two bytes).
If the writer is big-endian, the result is identical. If the writer is little-endian, the byte order may differ depending on implementation.
Semantically, the first is two bytes, the second is one short (two bytes).
In most network protocols (including TDS), both will result in the same byte sequence: FF FF.


```        switch (srcJdbcType) {
            case java.sql.Types.CHAR:
            case java.sql.Types.NCHAR:
            case java.sql.Types.VARCHAR:
            case java.sql.Types.NVARCHAR:
            case java.sql.Types.BINARY:
            case java.sql.Types.VARBINARY:
            case java.sql.Types.LONGVARCHAR:
            case java.sql.Types.LONGNVARCHAR:
            case java.sql.Types.LONGVARBINARY:
            microsoft.sql.Types.VECTOR:
                if (isStreaming) {
                    tdsWriter.writeLong(PLPInputStream.PLP_NULL);
                } else {
                    tdsWriter.writeByte((byte) 0xFF);
                    tdsWriter.writeByte((byte) 0xFF);
                }
                return;

File: src/main/java/microsoft/sql/Vector.java [hunk 43]
```diff
@@ -0,0 +1,114 @@
...
+import java.text.MessageFormat;
+import java.util.MissingResourceException;
+import java.util.ResourceBundle;
+import java.util.Arrays;
+
+public final class Vector implements java.io.Serializable {
+
+    public enum VectorDimensionType {
+        // float16, 16-bit (half precision) float
+        float32 // 32-bit (single precision) float
```
divang: By Java naming conventions, enum constants should be in all uppercase letters :

public enum VectorDimensionType {
    // FLOAT16, // 16-bit (half precision) float
    FLOAT32  // 32-bit (single precision) float
}

src/test/java/com/microsoft/sqlserver/jdbc/databasemetadata/DatabaseMetaDataTest.java
machavan: Yes and this is expected to be supported in near future.

src/test/java/com/microsoft/sqlserver/jdbc/databasemetadata/DatabaseMetaDataTest.java
```diff
@@ -1083,6 +1083,73 @@ public void testGetSchemasWithAndWithoutCatalog() throws SQLException {
...
+                int columnCount = metaData.getColumnCount();
+                assertEquals(1, columnCount, "Column count should be 1");
+
+                // Validate column name
```
machavan, line 1112: Let us also add a placeholder for now to cover a test on PreparedStatement::getParameterMetadata as that will also be returning a wrong type for vector as of now. We can revisit this test later

src/main/java/com/microsoft/sqlserver/jdbc/VectorUtils.java
```diff
@@ -0,0 +1,264 @@
...
+            throw vectorException("R_vectorByteArrayLength");
+        }
+        if ((bytes.length - getHeaderLength()) % bytesPerDimension != 0) {
+            throw vectorException("R_vectorByteArrayMultipleOfBytesPerDimension", bytesPerDimension);
+        }
+
+        int objectCount = (bytes.length - getHeaderLength()) / bytesPerDimension; // 8 bytes for header
+        Object[] objectArray;
+        switch (vectorType) {
+            case float32:
```
divang: Can it more concise:
```
Object[] objectArray;
switch (vectorType) {
    // case FLOAT16:
    //     objectArray = new Short[objectCount];
    //     break;
    case FLOAT32:
    default:
        objectArray = new Float[objectCount];
        break;
}

File: src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCallableStatement.java [hunk 46]
```diff
@@ -1606,9 +1623,17 @@ public void setObject(String parameterName, Object value, int sqlType) throws SQ
...
                     findColumn(parameterName), tvpName);
-        } else
-            setObject(setterGetParam(findColumn(parameterName)), value, JavaType.of(value), JDBCType.of(sqlType), null,
-                    null, false, findColumn(parameterName), tvpName);
+        } else {
+            Integer scale = null, precision = null;
+            if (microsoft.sql.Types.VECTOR == sqlType && value instanceof Vector) {
+                Vector vector = (Vector) value;
+                precision = vector.getDimensionCount();
+                scale = (int) VectorUtils.getScaleByte(vector.getVectorDimensionType());
```
divang, line 1631: Defensive null check:
```
Vector.VectorDimensionType dimType = vector.getVectorDimensionType();
    if (dimType != null) {
        scale = (int) VectorUtils.getScaleByte(dimType);
    }

File: src/main/java/com/microsoft/sqlserver/jdbc/DataTypes.java [hunk 47]
```diff
@@ -230,7 +233,7 @@ enum GetterConversion {
 
         CHARACTER(SSType.Category.CHARACTER, EnumSet.of(JDBCType.Category.NUMERIC, JDBCType.Category.DATE,
```
divang: Do we need JDBCType.Category.VECTOR in GetterConversion for the CHARACTER type, and vice versa?

src/main/java/com/microsoft/sqlserver/jdbc/DataTypes.java
```diff
@@ -266,7 +269,9 @@ enum GetterConversion {
 
         SQL_VARIANT(SSType.Category.SQL_VARIANT, EnumSet.of(JDBCType.Category.CHARACTER, JDBCType.Category.SQL_VARIANT,
                 JDBCType.Category.NUMERIC, JDBCType.Category.DATE, JDBCType.Category.TIME, JDBCType.Category.BINARY,
-                JDBCType.Category.TIMESTAMP, JDBCType.Category.NCHARACTER, JDBCType.Category.GUID));
+                JDBCType.Category.TIMESTAMP, JDBCType.Category.NCHARACTER, JDBCType.Category.GUID)),
+
+        VECTOR(SSType.Category.VECTOR, EnumSet.of(JDBCType.Category.CHARACTER, JDBCType.Category.VECTOR));
```
divang: (similar to above) Do we need JDBCType.Category.VECTOR in GetterConversion for the CHARACTER type, and vice versa?

src/main/java/com/microsoft/sqlserver/jdbc/DataTypes.java
```diff
@@ -733,7 +743,7 @@ enum SetterConversion {
                 JDBCType.Category.TIME, JDBCType.Category.TIMESTAMP, JDBCType.Category.DATETIMEOFFSET,
                 JDBCType.Category.CHARACTER, JDBCType.Category.LONG_CHARACTER, JDBCType.Category.NCHARACTER,
                 JDBCType.Category.LONG_NCHARACTER, JDBCType.Category.BINARY, JDBCType.Category.LONG_BINARY,
-                JDBCType.Category.GUID, JDBCType.Category.SQL_VARIANT)),
+                JDBCType.Category.GUID, JDBCType.Category.SQL_VARIANT, JDBCType.Category.VECTOR)),
```
divang: (similar to above) Do we need JDBCType.Category.VECTOR in GetterConversion for the CHARACTER type, and vice versa?

src/main/java/com/microsoft/sqlserver/jdbc/DataTypes.java
```diff
@@ -795,7 +805,9 @@ enum SetterConversion {
 
         GEOMETRY(JDBCType.Category.GEOMETRY, EnumSet.of(JDBCType.Category.GEOMETRY)),
 
-        GEOGRAPHY(JDBCType.Category.GEOGRAPHY, EnumSet.of(JDBCType.Category.GEOGRAPHY));
+        GEOGRAPHY(JDBCType.Category.GEOGRAPHY, EnumSet.of(JDBCType.Category.GEOGRAPHY)),
+
+        VECTOR(JDBCType.Category.VECTOR, EnumSet.of(JDBCType.Category.VECTOR, JDBCType.Category.CHARACTER));
```
divang: (similar to above) Do we need JDBCType.Category.VECTOR in GetterConversion for the CHARACTER type, and vice versa?

src/main/java/microsoft/sql/Vector.java
machavan: This is done to keep this similar to how it is defined on server side and how it would be specified when one defines a vector column in TSQL syntax in future when a support for float16 is added...

src/main/java/com/microsoft/sqlserver/jdbc/SimpleInputStream.java
```diff
@@ -362,4 +362,4 @@ final byte[] getBytes() throws SQLServerException {
         return value;
     }
 
-}
+}
```
divang, line 365: remove new line

src/main/java/com/microsoft/sqlserver/jdbc/DataTypes.java
muskan124947: Resolving this, already discussed

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
```diff
@@ -1386,6 +1394,8 @@ private String getDestTypeFromSrcType(int srcColIndx, int destColIndx,
                 else
                     return SSType.VARBINARY.toString() + "(" + bulkPrecision + ")";
 
+            case microsoft.sql.Types.VECTOR:
+                return SSType.VECTOR.toString() + "(" + bulkPrecision + ")";
```
machavan, line 1398: Are we generating type definition here? Should we call the VectorUtils method to do so for Vector type ?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -6745,6 +6835,24 @@ private void onFeatureExtAck(byte featureId, byte[] data) throws SQLServerExcept
                 sessionRecovery.setConnectionRecoveryPossible(true);
                 break;
             }
+
+            case TDS.TDS_FEATURE_EXT_VECTORSUPPORT: {
+                if (connectionlogger.isLoggable(Level.FINE)) {
+                    connectionlogger.fine(toString() + " Received feature extension acknowledgement for Vector Support.");
```
machavan: We should log the received byte in the ack response in this trace message

divang (dismissed):

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResource.java
```diff
@@ -313,6 +313,7 @@ protected Object[][] getContents() {
         {"R_StreamingDataTypeAE", "Data of length greater than {0} is not supported in encrypted {1} column."},
         {"R_AE_NotSupportedByServer", "SQL Server in use does not support column encryption."},
         {"R_InvalidAEVersionNumber", "Received invalid version number \"{0}\" for Always Encrypted."}, // From server
+        {"R_InvalidVectorVersionNumber", "Received invalid version number \"{0}\" for Vector feature negotiation."},
```
machavan: Change "Vector" to "vector"

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResource.java
```diff
@@ -339,6 +340,12 @@ protected Object[][] getContents() {
         {"R_unsupportedStmtColEncSetting", "SQLServerStatementColumnEncryptionSetting cannot be null."},
         {"R_unsupportedConversionAE", "The conversion from {0} to {1} is unsupported for encrypted column."},
         {"R_InvalidDataForAE", "The given value of type {0} from the data source cannot be converted to type {1} of the specified target column {2}."},
+        {"R_vectorTypeSupportPropertyDescription", "Determines the Vector support feature negotiation during connection initialization. Valid values are \"off\" or \"v1\". Default is \"v1\"."},
```
machavan: Change "Vector" to "vector"

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResource.java
```diff
@@ -480,6 +487,7 @@ protected Object[][] getContents() {
         {"R_InvalidDataClsVersionNumber", "Invalid version number {0} for Data Classification."}, // From Server
         {"R_unknownUTF8SupportValue", "Unknown value for UTF8 support."},
         {"R_unknownAzureSQLDNSCachingValue", "Unknown value for Azure SQL DNS Caching."},
+        {"R_unknownVectorSupportValue", "Unexpected version value received for Vector support feature negotiation."},
```
machavan: change "Vector" to "vector"

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResource.java
```diff
@@ -562,6 +570,9 @@ protected Object[][] getContents() {
         {"R_UnableToFindClass", "Unable to locate specified class: {0}"},
         {"R_ibmModuleNotFound", "com.ibm.security.auth.module.Krb5LoginModule module was not found."},
         {"R_moduleNotFound", "Neither com.sun.security.auth.module.Krb5LoginModule nor com.ibm.security.auth.module.Krb5LoginModule was found."},
+        {"R_vectorByteArrayLength", "Byte array length must be at least 8 bytes."},
```
machavan: Vector byte array length must be at least 8 bytes.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResource.java
```diff
@@ -562,6 +570,9 @@ protected Object[][] getContents() {
...
+        {"R_invalidVectorData", "The provided type of data is not supported for vector."},
+        {"R_vectorByteArrayMultipleOfBytesPerDimension", "Byte array length must be a multiple of {0}."},
```
machavan: Byte array length must be multiple of {0} for vector of type {1}

src/main/java/com/microsoft/sqlserver/jdbc/VectorUtils.java
```diff
@@ -0,0 +1,256 @@
...
+        ByteBuffer buffer = ByteBuffer.allocate(payloadSize).order(ByteOrder.LITTLE_ENDIAN);
+
+        buffer.put((byte) 0xA9);
+        buffer.put((byte) 0x01);
+        buffer.putShort((short) (vector.getDimensionCount()));
+        buffer.put(getScaleByte(vector.getVectorDimensionType())); // 0x00 for FLOAT32, 0x01 for FLOAT16
+        buffer.put(new byte[3]);
+
+        for (Object value : vector.getData()) {
+            switch (vector.getVectorDimensionType()) {
```
machavan: The switch case is done for each object.

Instead, the switch case can be outside and done once and then for each type we can iterate on objects in a tight loop to populate buffer.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
muskan124947: there we generate precision from vector object (dimension count) and outscale/valuelength.
Here, at this point we don't have vector object.

machavan (approved):

divang (approved):

PR #2641, opened 2025-04-03
pom.xml: use newer bundle plugin 6.0.0

ecki:
If compiling with maven/new JDK the generated OSGi metadata contains a require osgi.ee=UNKNOWN requirement, since the older bnd version does not know about the new Java profile. Updating to latest bundle plugin does also bring a new bnd version which has not this issue.

Alternative: provide `<_runee>Java-xx</_runee>` instruction.

X-Signed-Off: Bernd Eckenfels <b.eckenfels@seeburger.de>

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -518,7 +518,7 @@
 			<plugin>
 				<groupId>org.apache.felix</groupId>
 				<artifactId>maven-bundle-plugin</artifactId>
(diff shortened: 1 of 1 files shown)


ecki:
cc @MrEasy

MrEasy (approved):

PR #2652, opened 2025-05-06
correct processing of OffsetDateTime in SQLServerDataTable 

muskan124947:
When OffsetDateTime contains zero seconds, calling toString() produces the following format: 2025-04-26T19:42+02:00 (without seconds) which is rejected by the server with the message "Conversion failed when converting date and/or time from character string." (SQL Error: 241, SQLState: S0001)


muskan124947:
ADO pipeline -> https://sqlclientdrivers.visualstudio.com/mssql-jdbc/_build/results?buildId=114807&view=results

machavan (approved):

divang (approved):

Issue #2654, opened 2025-05-07
[FEATURE REQUEST] Allow up to 10 redirections per connection


David-Engel:
## Is your feature request related to a problem? If so, please give a short summary of the problem and how the feature would resolve it
Fabric has introduced a feature that can cause clients to be redirected more than once. The JDBC driver currently limits connections to one redirection.

## Describe the preferred solution
M.D.SqlClient has increased their limit to 10 redirections. Other drivers should follow suit and not error until redirection attempts exceed 10.

## Reference Implementation
https://github.com/dotnet/SqlClient/pull/2873

divang:
Handled above scenario: PR: https://github.com/microsoft/mssql-jdbc/pull/2659

PR #2659, opened 2025-05-09
Increased the max noOfRedirections and updated the message.

divang:
Issue: https://github.com/microsoft/mssql-jdbc/issues/2654

Details (copied from above issue)

Fabric has introduced a feature that can cause clients to be redirected more than once. The JDBC driver currently limits connections to one redirection.

Describe the preferred solution
M.D.SqlClient has increased their limit to 10 redirections. Other drivers should follow suit and not error until redirection attempts exceed 10.


David-Engel (approved):

divang:
ADO run is successful.

David-Engel (approved):

machavan (approved):

PR #2663, opened 2025-05-13
Handled sp_prepexec call for Column encryption enabled

divang:
**Description:**
- The incident involved unexpected behavior when inserting strings using Always Encrypted with Secure Enclaves in Azure SQL DB.
- Unexpected insert results when using Always Encrypted.
- Subsequent inserts into encrypted columns fail after initial successful inserts.
- Issue in details at code level:
The buildPreparedStrings method compares the last prepared statement with the current one to be executed. If both statements are identical, it signals that calling sp_prepexec is unnecessary. Otherwise, sp_prepexec will be invoked later in the driver's execution flow. However, multiple calls to buildPreparedStrings overwrite the preparedTypeDefinitions variable, which stores the current prepared statement. As a result, repeated invocations can lead to the loss of the original statement, affecting the logic that determines whether sp_prepexec should be called with updated data types and lengths.

- Resolution details:
Within the buildPreparedStrings method, the preparedTypeDefinitions variable is cached in preparedTypeDefinitionsPrev to prevent it from being overwritten. If isAEv2() is enabled, renewDefinition is false, and isInternalEncryptionQuery is true, then the driver caches the preparedTypeDefinitions value into preparedTypeDefinitionsPrev. In subsequent calls to buildPreparedStrings, if isAEv2() is still enabled, renewDefinition is true, and isInternalEncryptionQuery is false, the driver restores preparedTypeDefinitions from the cached preparedTypeDefinitionsPrev. This caching mechanism prevents the loss of the original statement and ensures the correct signal is returned regarding whether to invoke sp_prepexec.

**Issues:**
https://portal.microsofticm.com/imp/v5/incidents/details/619317008/summary

**Testing**
- Configure the SQL database with encryption enabled and create a table with column-level encryption. Below manual test method works well.
```
try (Connection conn = DriverManager.getConnection(url)) {
    try (PreparedStatement pstmt = conn.prepareStatement("insert into t1 values(?)")) {
        String data2 ="a"; 
        pstmt.setString(1, data2);
        pstmt.executeUpdate();
        String data3 =  data2 + "aa";
        pstmt.setString(1, data3);
        pstmt.executeUpdate();
        String data4 =  data3 + "aaa";
        pstmt.setString(1, data4);
        pstmt.executeUpdate();
    } catch (SQLException e) {
        e.printStackTrace();  
    }
} 

```
- Mocked the above use case via Mockito framework in the test case testBuildPreparedStringsForVaryLength()

### DIFF
--- a/src/test/java/com/microsoft/sqlserver/jdbc/AlwaysEncrypted/RegressionAlwaysEncryptedTest.java
+++ b/src/test/java/com/microsoft/sqlserver/jdbc/AlwaysEncrypted/RegressionAlwaysEncryptedTest.java
@@ -50,25 +50,27 @@ public class RegressionAlwaysEncryptedTest extends AESetup {
                 .getConnection(AETestConnectionString + ";columnEncryptionSetting=enabled;", AEInfo);
                 Statement stmt = connection.createStatement()) {
             dropTables(stmt);
+            try {        
+                createTable(NUMERIC_TABLE_AE, cekJks, numericTable);
 
-            createTable(NUMERIC_TABLE_AE, cekJks, numericTable);
+                populateNumericTable(connection);
+                verifyNumericTable(connection, false);
 
-            populateNumericTable(connection);
-            verifyNumericTable(connection, false);
+                dropTables(stmt);
+                createTable(DATE_TABLE_AE, cekJks, dateTable);
 
-            dropTables(stmt);
-            createTable(DATE_TABLE_AE, cekJks, dateTable);
+                populateDateTable(connection);
+                verifyDateTable(connection);
 
-            populateDateTable(connection);
-            verifyDateTable(connection);
+                dropTables(stmt);
+                createTable(NUMERIC_TABLE_AE, cekJks, numericTable);
 
-            dropTables(stmt);
-            createTable(NUMERIC_TABLE_AE, cekJks, numericTable);
-
-            populateNumericTableWithNull(connection);
-            verifyNumericTable(connection, true);
-
-            dropTables(stmt);
+                populateNumericTableWithNull(connection);
+                verifyNumericTable(connection, true);
+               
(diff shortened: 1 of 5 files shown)


divang:
ADO pipeline is successful.

David-Engel (commented):
No general objections to these changes. But we should add tests to cover the scenario this is fixing.

src/test/java/com/microsoft/sqlserver/jdbc/AlwaysEncrypted/RegressionAlwaysEncryptedTest.java
```diff
@@ -228,9 +230,43 @@ private void verifyNumericTable(Connection connection, boolean isNull) throws SQ
...
+    public void testStringColumnEncryptWithVaryLength(String serverName, String url, String protocol) throws Exception {
+        setAEConnectionString(serverName, url, protocol);
+        try (Connection connection = PrepUtil
+                .getConnection(AETestConnectionString + ";columnEncryptionSetting=enabled;", AEInfo);
+                Statement stmt = connection.createStatement()) {
+            dropTables(stmt);
+
+            createTable(VARY_STRING_TABLE_AE, cekJks, stringTable);
+
+            verifyStringTable(connection);
```
David-Engel, line 244: Make sure to clean up even if the test errors. Bonus points if you improve the two existing tests above to clean up on error, too.
```suggestion
            try {
                verifyStringTable(connection);
            } finally {
                dropTables(stmt);
            }
```

divang:
ADO pipeline is successful.

David-Engel (approved):

divang:
Successful run ADO pipeline

machavan (approved):

muskan124947 (approved):

PR #2671, opened 2025-06-04
POC: Improve JDBC driver performance using ByteBufferManager for TDS packet handling

Ananya2:
This POC introduces a ByteBufferManager and BufferPool implementation in the JDBC driver to optimize memory usage during TDS packet processing.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/BufferPool.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/BufferPool.java
@@ -1,0 +1,24 @@
+package com.microsoft.sqlserver.jdbc;
+
+import java.util.concurrent.ConcurrentLinkedQueue;
+
+public class BufferPool {
+    private final int bufferSize;
+    private final ConcurrentLinkedQueue<byte[]> pool = new ConcurrentLinkedQueue<>();
+
+    BufferPool(int bufferSize) {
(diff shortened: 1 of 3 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/BufferPool.java
```diff
@@ -0,0 +1,24 @@
...
+    }
+
+    public byte[] rent() {
+        byte[] buffer = pool.poll();
+        return buffer == null ? new byte[bufferSize] : buffer;
+    }
+
+    public void release(byte[] buffer) {
+        if (buffer.length == bufferSize) {
+            pool.offer(buffer);
```
David-Engel, line 21:
```suggestion
            Arrays.fill(buffer, (byte) 0); // Clear the array for re-use
            pool.offer(buffer);
```

src/main/java/com/microsoft/sqlserver/jdbc/BufferPool.java
```diff
@@ -0,0 +1,26 @@
...
+            Arrays.fill(buffer, (byte) 0); // Clear the array for re-use
```
mmimica, line 22: You need to clear the buffer first, then put it pool. You lose ownership once added to pool and another thread might take the buffer, you can't touch it any more.

PR #2674, opened 2025-06-05
Fixed New Type Param Definitions logic in doExecutePreparedStatement

divang:
**Description:**

The incident involved unexpected behavior when inserting using Always Encrypted with Secure Enclaves in Azure SQL DB.

Unexpected insert results when using Always Encrypted.

Subsequent inserts into encrypted columns fail after initial successful inserts.

**Issue in details at code level:**
The doExecutePreparedStatement() method invokes buildPreparedStrings(), which updates the preparedSQL and preparedTypeDefinitions, and also returns a flag indicating whether preparedTypeDefinitions was modified.

The issue arises because the method is expected to use only the result of the first buildPreparedStrings() call to determine the hasNewTypeDefinitions flag. This flag is crucial, as it helps decide whether to initiate the sp_prepexec call—based on its value in combination with other conditions.

However, the current implementation updates the hasNewTypeDefinitions variable on each buildPreparedStrings() call. This leads to incorrect behavior: if later calls do not indicate a type definition change, the earlier (valid) indication of change is overwritten, and the sp_prepexec call is incorrectly skipped when parameter type definitions actually change.

**Resolution details:**
The issue was resolved by ensuring that the hasNewTypeDefinitions flag is set only once, based on the result of the first call to buildPreparedStrings().

Subsequent calls to buildPreparedStrings() are now used solely to update the preparedSQL variable and are no longer allowed to overwrite the hasNewTypeDefinitions flag. This guarantees that any change in parameter type definitions detected during the first evaluation is preserved, ensuring that the sp_prepexec call is correctly triggered when required.

This change aligns with the intended design and prevents silent failures due to skipped sp_prepexec calls when parameter type definitions change.

**Issues:**
https://portal.microsofticm.com/imp/v5/incidents/details/619317008/summary

Testing

Configure the SQL database with encryption enabled and create a table with column-level encryption. Below manual test method works well.
```
try (Connection conn = DriverManager.getConnection(url)) {
    try (PreparedStatement pstmt = conn.prepareStatement("insert into t1 values(?)")) {
        String data2 ="a"; 
        pstmt.clearParameters();
        pstmt.setString(1, data2);
        pstmt.executeUpdate();
        String data3 =  data2 + "aa";
        pstmt.clearParameters();
        pstmt.setString(1, data3);
        pstmt.executeUpdate();
        String data4 =  data3 + "aaa";
        pstmt.clearParameters();
        pstmt.setString(1, data4);
        pstmt.executeUpdate();
    } catch (SQLException e) {
        e.printStackTrace();  
    }
} 
```


divang:
> I'm good with the changes. They make me a little nervous since I understand the AE code is complex, but I trust you understand the flow here and that this is the correct fix. Make sure the Fx tests pass, too.

Fx tests cases are passed. So we are good with these changes.

Issue #2669, opened 2025-05-21
Bulk copy for batch insert destroys data without error


literakl:
## Driver version
12.10.0.jre11

## SQL Server version
16 

## Client Operating System
Windows 11

## JAVA/JVM version
Java 18

## Table schema

create table states (
[Id] identity not null
[StateCode] nvarchar NOT NULL
)

## Problem description

var items = List.of(new State("IL"));
final var session = entityManager.unwrap(Session.class);
 session.doWork(connection -> {
     try (PreparedStatement ps = connection.prepareStatement("insert into states (StateCode) values (?)")) {
         for (T entity : items) {
             ps.setString(1, entity.getState());
             ps.addBatch();
         }
         ps.executeBatch();
}

## Expected behavior

Select StateCode from states;
OH

## Actual behavior

Select StateCode from states;
"[B@4b01d9e4

## Error message/stack trace

There is no error anywhere. I debugged the MS SQL JDBC driver up to com.microsoft.sqlserver.jdbc.AppDTVImpl#setValue.

When I turned off dataSource.setUseBulkCopyForBatchInsert(true);
then it started to work properly.

I found https://learn.microsoft.com/en-us/sql/connect/jdbc/use-bulk-copy-api-batch-insert-operation?view=sql-server-ver16. which states that identity columns are not supported. So that the array of bytes of the strings are persisted as an object reference instead of the array content, may be the reason, why it's not supported.

If this is the case and cannot be fixed, then please raise an exception when an identity column is detected. This took me hours to identify.

https://stackoverflow.com/questions/79632176/ms-sql-jdbc-driver-saves-string-as-byte-array-object-reference

muskan124947:
We are investigating the issue.
Could you also share the connection string used for above issue?

muskan124947:
Hi @literakl 
Issue has been resolved in below PR :
[Update batch insert for bulk copy operation to handle string correctly when sendStringParametersAsUnicode is set to false by muskan124947 · Pull Request #2704 · microsoft/mssql-jdbc](https://github.com/microsoft/mssql-jdbc/pull/2704)

viniciusxyz:
@muskan124947 Can you tell me when we'll see a version release with this fix applied? I just migrated a banking system to use batch inserts with bulk api and ran into this exact problem :/

muskan124947:
@viniciusxyz The milestone is added as 13.1.1.

literakl:
Thanks.

[...]

PR #2704, opened 2025-07-10
Update batch insert for bulk copy operation to handle string correctly when sendStringParametersAsUnicode is set to false

muskan124947:
## Description

This PR addresses GitHub reported issue : [Bulk copy for batch insert destroys data without error #2669](https://github.com/microsoft/mssql-jdbc/issues/2669)

When inserting string data using batch insert for bulk copy with below combinations:

```
- useBulkCopyForBatchInsert=true
- sendStringParametersAsUnicode=false
```

The data is inserted in byte array format. 

**Expected behavior**
Select StateCode from states -> "OH"

**Actual behavior**
Select StateCode from states; -> "[B@4b01d9e4"

## Root cause

The issue occurs because:
- When sendStringParametersAsUnicode=false, the setString() method stores parameter values as byte arrays in the DTV (Data Type Value) objects.
- During bulk copy batch insert, the SQLServerBulkBatchInsertRecord.convertValue() method retrieves these raw byte arrays from the parameter DTVs
- The method was missing logic to detect and convert byte arrays back to strings for string-type columns

## Testing
Added comprehensive test cases in BatchExecutionWithBulkCopyTest.java:

- testBulkInsertStringWhenSentAsUnicode():
Tests bulk insert with sendStringParametersAsUnicode=true (default behavior). Validates that string data is inserted correctly as Unicode

- testBulkInsertStringWhenNotSentAsUnicode():
Tests bulk insert with sendStringParametersAsUnicode=false (issue scenario)
Validates that byte arrays are properly converted back to strings

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkBatchInsertRecord.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkBatchInsertRecord.java
@@ -181,13 +181,24 @@ class SQLServerBulkBatchInsertRecord extends SQLServerBulkRecord {
                 return null;
             }
 
+            case Types.NCHAR:
+            case Types.NVARCHAR:
+            case Types.LONGNVARCHAR: {
+                /*
+                 * If string data comes in as a byte array through setString (and sendStringParametersAsUnicode = false)
+                 * through Bulk Copy for Batch Insert API, convert the byte array to a string.
+                 * If the data is already a string, return it as is.
+                 */
+                if (data instanceof byte[]) {
+                    return new String((byte[]) data);
+                }
+                return data;
+            }
+
             case Types.DATE:
             case Types.CHAR:
-            case Types.NCHAR:
             case Types.VARCHAR:
-            case Types.NVARCHAR:
             case Types.LONGVARCHAR:
-            case Types.LONGNVARCHAR:
             case Types.CLOB:
             default: {
                 // The string is copied as is.
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkBatchInsertRecord.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkBatchInsertRecord.java
@@ -181,6 +181,9 @@ class SQLServerBulkBatchInsertRecord extends SQLServerBulkRecord {
                 return null;
             }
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkBatchInsertRecord.java
```diff
@@ -181,13 +181,24 @@ else if (dateTimeFormatter != null)
                 return null;
             }
 
-            case Types.DATE:
             case Types.CHAR:
-            case Types.NCHAR:
             case Types.VARCHAR:
-            case Types.NVARCHAR:
```
machavan, line 188: For readability, can we keep the original cases for N* data types as is and add explicit cases for VARCHAR, CHAR and LONGVARCHAR which need this special treatment to convert back to string.

muskan124947: All the string type needs the conversion (from byte to string) when sendStringParametersAsUnicode=false in case of bulk copy api.

Ananya2 (approved):

machavan (approved):

Ananya2 (approved):

David-Engel (approved):

BiancaPuertaRocha:
@muskan124947 We are trying to use the new [13.1.1.jre11-preview](https://mvnrepository.com/artifact/com.microsoft.sqlserver/mssql-jdbc/13.1.1.jre11-preview), but we are getting an issue when handling accented characters. The image below shows names that were incorrectly saved in the database due to this problem:

names

This issue is critical for us, as our language makes extensive use of accent marks.

We are currently using the following configuration:
```
useBulkCopyForBatchInsert=true
sendStringParametersAsUnicode=false
```

We would appreciate any help with this. Thank you!

muskan124947:
Hi @BiancaPuertaRocha 
Will look into the issue.

BiancaPuertaRocha:
@muskan124947 Are you planning to reopen the issue [Bulk copy for batch insert destroys data without error #2669
](https://github.com/microsoft/mssql-jdbc/issues/2669) or would you prefer that I open a new issue to report this?

muskan124947:
@BiancaPuertaRocha Please create a new issue with the details.

PR #2706, opened 2025-07-15
Introducing Performance Loggers 

machavan:
Description:

This is foundational change for introducing performance loggers which can be used tracking performance of various activities within the mssql-jdbc driver e..g. connection time, login time, token acquisition time etc. 

A new java.util.Logging logger `com.microsoft.sqlserver.jdbc.PerformanceMetrics.Connection` is introduced here which can be subscribed to by applications. 
More loggers such as `com.microsoft.sqlserver.jdbc.PerformanceMetrics.Statement` will be introduced in future.

Various performance activities to be tracked are defined in a PerformanceActivity Enum which will be amended with new additions in future.

Testing:
-Manual testing of new logger functionality. 
-All existing tests ensure testing of newly added code.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/PerformanceActivity.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/PerformanceActivity.java
@@ -1,0 +1,23 @@
+package com.microsoft.sqlserver.jdbc;
+
+public enum PerformanceActivity {
+    CONNECTION("Connection"),
+    TOKEN_ACQUISITION("Token acquisition"),
+    LOGIN("Login"),
+    PRELOGIN("Prelogin");
+
+    private final String activity;
+
+    PerformanceActivity(String activity) {
+        this.activity = activity;
+    }
+
+    public String activity() {
+        return activity;
+    }
+
+    @Override
+    public String toString() {
+        return activity;
+    }
+}
--- a/src/main/java/com/microsoft/sqlserver/jdbc/PerformanceLog.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/PerformanceLog.java
@@ -1,6 +1,5 @@
 package com.microsoft.sqlserver.jdbc;
 
-import java.util.UUID;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
@@ -9,22 +8,24 @@ public class PerformanceLog {
     static final java.util.logging.Logger perfLoggerConnection = java.util.logging.Logger
             .getLogger("com.microsoft.sqlserver.jdbc.PerformanceMetrics.Connection");
 
+    //TODO
+    //More loggers to be added here e.g. com.microsoft.sqlserver.jdbc.PerformanceMetrics.Statement
+    
     public static class Scope implements AutoCloseable {
-        private final Logger logger;
-        private final UUID connId;
-        private final PerformanceActivity activity;
-        private final long startTime;
+        private Logger logger;
+        private String logPrefix;
+        private PerformanceActivity activity;
+        private long startTime;
         private final boolean enabled;
 
-        public Scope(Logger logger, UUID connId, PerformanceActivity activity) {
-            this.logger = logger;
-            this.connId = connId;
-            this.activity = activity;
+        public Scope(Logger logger, String logPrefix, PerformanceActivity activity) {
             this.enabled = logger.isLoggable(Level.INFO);
-            this.startTime = enabled ? System.currentTimeMillis() : 0;
-
             if (enabled) {
-                logger.info(String.format("%d %s %s start", startTime, connId, activity));
+                this.logger = logger;
+                this.logPrefix = logPrefix;
+                this.activity = activity;
+                this.startTime = enabled ? System.currentTimeMillis() : 0;
+                logger.info(String.format("%d %s %s start", startTime, logPrefix, activity));
             }
         }
 
@@ -33,12 +34,12 @@ public class PerformanceLog {
             if (enabled) {
                 long endTime = System.currentTimeMillis();
                 long duration = endTime - startTime;
-                logger.info(String.format("%d %s %s end, duration: %dms", endTime, connId, activity, duration));
+                logger.info(String.format("%d %s %s end, duration: %dms", endTime, logPrefix, activity, duration));
             }
         }
     }
 
-    public static Scope createScope(Logger logger, UUID connId, PerformanceActivity activity) {
-        return new Scope(logger, connId, activity);
+    public static Scope createScope(Logger logger, String logPrefix, PerformanceActivity activity) {
+        return new Scope(logger, logPrefix, activity);
     }
 }
--- a/src/main/java/com/microsoft/sqlserver/jdbc/PerformanceLog.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/PerformanceLog.java
@@ -1,0 +1,44 @@
+package com.microsoft.sqlserver.jdbc;
+
+import java.util.UUID;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+public class PerformanceLog {
+
+    static final java.util.logging.Logger perfLoggerConnection = java.util.logging.Logger
+            .getLogger("com.microsoft.sqlserver.jdbc.PerformanceMetrics.Connection");
+
+    public static class Scope implements AutoCloseable {
+        private final Logger logger;
+        private final UUID connId;
+        private final PerformanceActivity activity;
+        private final long startTime;
+        private final boolean enabled;
+
+        public Scope(Logger logger, UUID connId, PerformanceActivity activity) {
+            this.logger = logger;
+            this.connId = connId;
+            this.activity = activity;
+            this.enabled = logger.isLoggable(Level.INFO);
+            this.startTime = enabled ? System.currentTimeMillis() : 0;
+
+            if (enabled) {
+                logger.info(String.format("%d %s %s start", startTime, connId, activity));
+            }
+        }
+
+        @Override
+        public void close() {
+            if (enabled) {
+                long endTime = System.currentTimeMillis();
+                long duration = endTime - startTime;
+                logger.info(String.format("%d %s %s end, duration: %dms", endTime, connId, activity, duration));
+            }
+        }
+    }
+
+    public static Scope createScope(Logger logger, UUID connId, PerformanceActivity activity) {
+        return new Scope(logger, connId, activity);
+    }
+}
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
@@ -51,7 +51,6 @@ import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.logging.Level;
-import java.util.logging.Logger;
 
 import javax.sql.XAConnection;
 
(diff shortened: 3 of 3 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/PerformanceLog.java
```diff
@@ -0,0 +1,45 @@
...
+        private long startTime;
+        private final boolean enabled;
+
+        public Scope(Logger logger, String logPrefix, PerformanceActivity activity) {
+            this.enabled = logger.isLoggable(Level.INFO);
+            if (enabled) {
+                this.logger = logger;
+                this.logPrefix = logPrefix;
+                this.activity = activity;
+                this.startTime = enabled ? System.currentTimeMillis() : 0;
```
David-Engel: This can be simplified since it's now with the if block:
```suggestion
                this.startTime = System.currentTimeMillis();
```

src/main/java/com/microsoft/sqlserver/jdbc/PerformanceActivity.java
```diff
@@ -0,0 +1,23 @@
+package com.microsoft.sqlserver.jdbc;
+
+public enum PerformanceActivity {
```
David-Engel, line 11: I don't think we want to make these public.
```suggestion
enum PerformanceActivity {
```

src/main/java/com/microsoft/sqlserver/jdbc/PerformanceLog.java
```diff
@@ -0,0 +1,45 @@
+package com.microsoft.sqlserver.jdbc;
+
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+public class PerformanceLog {
```
David-Engel: Not public...
```suggestion
class PerformanceLog {
```

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -6459,190 +6465,192 @@ void onFedAuthInfo(SqlFedAuthInfo fedAuthInfo, TDSTokenHandler tdsTokenHandler)
...
-                // Break out of the retry loop in successful case.
-                break;
-            } else if (authenticationString
-                    .equalsIgnoreCase(SqlAuthentication.ACTIVE_DIRECTORY_MANAGED_IDENTITY.toString())) {
-
-                String managedIdentityClientId = activeConnectionProperties
-                        .getProperty(SQLServerDriverStringProperty.USER.toString());
-
-                if (null != managedIdentityClientId && !managedIdentityClientId.isEmpty()) {
+        try (PerformanceLog.Scope fedAuthScope = PerformanceLog.createScope(PerformanceLog.perfLoggerConnection, traceID, PerformanceActivity.TOKEN_ACQUISITION)) {
```
David-Engel: I would actually put this in onFedAuthInfo, so that the scope includes calls to any access token callbacks (but not the last two lines in there that send the token to the server).

src/main/java/com/microsoft/sqlserver/jdbc/PerformanceLog.java
```diff
@@ -0,0 +1,83 @@
...
+
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+public class PerformanceLog {
+
+    static final java.util.logging.Logger perfLoggerConnection = java.util.logging.Logger
+            .getLogger("com.microsoft.sqlserver.jdbc.PerformanceMetrics.Connection");
+
+    private static PerformanceLogCallback callback;
```
divang, line 16: Add volatile to the callback field declaration to ensure thread-safe visibility since setCallback can be called from multiple threads.

private static volatile PerformanceLogCallback callback;

src/main/java/com/microsoft/sqlserver/jdbc/PerformanceLogCallback.java
```diff
@@ -0,0 +1,22 @@
...
+public interface PerformanceLogCallback {
+
+    /**
+     * Publish a performance log entry.
+     *
+     * @param durationMs      The duration of the operation in milliseconds.
+     * @param activity        The type of activity being logged.
+     * @param exceptionMessage An exception message, if an error occurred.
+     */
+    void publish(long durationMs, PerformanceActivity activity, String exceptionMessage);
```
divang: Consider adding throws Exception to the publish method signature. The current design silently swallows callback failures which can mask critical issues like configuration errors, resource exhaustion, or compliance violations. Adding throws enables proper error handling, fail-fast behavior, and integration with monitoring systems for better system reliability.

void publish(.....) throws Exception;

src/main/java/com/microsoft/sqlserver/jdbc/PerformanceLogCallback.java
```diff
@@ -0,0 +1,22 @@
...
+
```
divang: Consider changing the parameter from String exceptionMessage to Exception exception. Passing the full Exception object provides richer context (stack traces, exception types, cause chains) for better debugging and observability rather than just the string message, allowing implementations to make more informed decisions about error handling.

void publish(long durationMs, PerformanceActivity activity, Exception exception) throws Exception;

src/main/java/com/microsoft/sqlserver/jdbc/PerformanceActivity.java
machavan: With the callback mechanism introduction, we need to make this enum public as we will sending this out to publish method.

src/main/java/com/microsoft/sqlserver/jdbc/PerformanceLog.java
machavan: Yes this should be package scope

src/main/java/com/microsoft/sqlserver/jdbc/PerformanceLog.java
machavan: We should follow below existing mechanism that is exposed for installing messageHandler

```
    /** Message handler */
    private transient ISQLServerMessageHandler serverMessageHandler;

    /**
     * Set current message handler
     * 
     * @param messageHandler
     *        message handler
     * @return The previously installed message handler (null if none)
     */
    @Override
    public ISQLServerMessageHandler setServerMessageHandler(ISQLServerMessageHandler messageHandler) {
        ISQLServerMessageHandler installedMessageHandler = this.serverMessageHandler;
        this.serverMessageHandler = messageHandler;
        return installedMessageHandler;
    }
```

Our guideline would be to set such callbacks once on application startups and not really on multiple threads at runtime.

src/main/java/com/microsoft/sqlserver/jdbc/PerformanceLogCallback.java
machavan: here also let us refer `ISQLServerMessageHandler`  that is an existing mechanism

src/main/java/com/microsoft/sqlserver/jdbc/PerformanceLogCallback.java
machavan: yes, we should pass full exception object

src/main/java/com/microsoft/sqlserver/jdbc/PerformanceLog.java
divang: The serverMessageHandler is an instance variable on the connection, not a static/global handler. This means each SQLServerConnection can have its own message handler, and it is set via the setServerMessageHandler method you referenced.
In contrast, the logger (like PerLogger) is a static reference, shared across all instances. This allows all parts of the application to write to the same log, and if the logger is reconfigured (for example, log level or output destination is changed), the change should be visible to all threads.

machavan: Right, this one is global and static. I think we may need something like below:

```
 private static String value;  // no assignment yet
    private static boolean initialized = false;

    // public method to set it only once
    public static synchronized void setValue(String v) {
        if (initialized) {
            throw new IllegalStateException("Value already set!");
        }
        value = v;
        initialized = true;
    }    ```

Comment (divang) [reply src/main/java/com/microsoft/sqlserver/jdbc/PerformanceLog.java:16]: If we want to allow the callback/value to be set only once (e.g., at application startup, never changed), then above approach is good.

If we want to allow changing the callback/value at runtime, then before approach is good:
private static volatile PerformanceLogCallback callback;

File: src/main/java/com/microsoft/sqlserver/jdbc/PerformanceLog.java [hunk 8]
```diff
@@ -0,0 +1,83 @@
...
+
+    public static void setCallback(PerformanceLogCallback cb) {
+        callback = cb;
```
Ananya2, line 28: The static callback field will retain its reference indefinitely, which may cause resource retention. Could we add a clearCallback() (or similar unregister/reset mechanism) with proper thread-safety to allow releasing the reference when it’s no longer needed?

src/main/java/com/microsoft/sqlserver/jdbc/PerformanceLog.java
```diff
@@ -0,0 +1,83 @@
...
+        }
+
+        @Override
+        public void close() {
+
+            if (!enabled) {
+                return;
+            }
+
+            long endTime = System.currentTimeMillis();
```
Ananya2, line 91: Instead of logging endTime as a raw epoch millis, can we format it using DateTimeFormatter into a human-readable timestamp (e.g., ISO-8601) before logging? This would make the logs easier to interpret while still keeping duration in milliseconds.

src/main/java/com/microsoft/sqlserver/jdbc/PerformanceLog.java
machavan: Yes, the callback should not be changed arbitrarily while the application is up and running.

src/main/java/com/microsoft/sqlserver/jdbc/PerformanceLog.java
machavan: We don't need to log the endtime actually, the timestamp will be added by the logger itself.

src/main/java/com/microsoft/sqlserver/jdbc/PerformanceLog.java
machavan: Yes, a public API to unregister the callback can be added.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -2228,129 +2228,147 @@ private void sleepForInterval(long interval) {
...
-
-        // Interactive auth may involve MFA which require longer timeout
-        if (SqlAuthentication.ACTIVE_DIRECTORY_INTERACTIVE.toString().equalsIgnoreCase(authenticationString)) {
-            loginTimeoutSeconds *= 10;
-        }
-
-        long elapsedSeconds = 0;
-        long start = System.currentTimeMillis();
+    public int getConnectionID() {
+        return Integer.parseInt(traceID.replace("ConnectionID:", ""));
```
David-Engel: We should persist the connectionID in its own variable instead of parsing it out of the string multiple times.

src/main/java/com/microsoft/sqlserver/jdbc/PerformanceLog.java
```diff
@@ -0,0 +1,105 @@
...
+                    logger.info(String.format("%s %s, duration: %dms, exception: %s", getTraceId(), activity, duration, exception.getMessage()));
+                } else {
+                    logger.info(String.format("%s %s, duration: %dms", getTraceId(), activity, duration));
+                }
+            }
+
+        }
+    }
+
+    public static Scope createScope(Logger logger, int connectionId, PerformanceActivity activity) {
```
David-Engel, line 121: Should we also add statementId for when we add activities in the context of Statements?

divang (approved):

Ananya2 (approved):

PR #2710, opened 2025-07-17
Enhance test coverage for SQLServerResultSet

muskan124947:


src/test/java/com/microsoft/sqlserver/jdbc/resultset/ResultSetTest.java
```diff
@@ -709,6 +724,855 @@ public void testResultSetClientCursorInitializerSqlErrorState() {
         }
     }
 
+    /**
+     * Test getObject() with unsupported type conversion that throws SQLException
+     */
+    @Test
+    @Tag(Constants.xAzureSQLDW)
```
muskan124947: is this required?

PR #2727, opened 2025-08-05
Fix to handle accented characters in database when using useBulkCopyForBatchInsert=true and sendStringParametersAsUnicode=false

muskan124947:


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkBatchInsertRecord.java
```diff
@@ -193,7 +196,7 @@ else if (dateTimeFormatter != null)
                  * If the data is already a string, return it as is.
                  */
                 if (data instanceof byte[]) {
-                    return new String((byte[]) data);
+                    return new String((byte[]) data, Charset.forName(encoding));
```
David-Engel: I would tweak the changes in this file so that we don't have to do this Charset lookup. It would be more efficient to:

1. Change the `String encoding` parameter to `Charset charset` and save it (replace the encoding member variable with charset).
2. The reference on line 52 changes from `encoding` to `charset.name()`. (We will only ever do this if logging is enabled.)
3. This line simply becomes `return new String((byte[]) data, charset);`
4. The parameter in constructor calls to SQLServerBulkBatchInsertRecord in SQLServerPreparedStatement simplifies from `connection.getDatabaseCollation().getCharset().name()` to just `connection.getDatabaseCollation().getCharset()`.

src/test/java/com/microsoft/sqlserver/jdbc/preparedStatement/BatchExecutionWithBulkCopyTest.java
```diff
@@ -1489,6 +1489,47 @@ public void testBulkInsertStringWhenNotSentAsUnicode() throws Exception {
         }
     }
 
+    /**
```
David-Engel, line 1427: This is very duplicative, as were the previous two tests added in #2704.

This new code can be removed and the previous two tests improved. They can exercise accented characters in their test data since those are supported in LATIN collations. Additionally, think about how the test can demonstrate/validate any expected behavior when inserting Unicode characters into the various column types with the varying connection options (be sure to use Unicode characters in the test data). Also, define the test strings as string constants to avoid string duplication in the code.

machavan (approved):

divang (approved):

Ananya2 (approved):

Issue #2732, opened 2025-08-08
Pre-release 13.1 is slower


JDBC-0:
## Driver version
13.1.1

## SQL Server version
Microsoft SQL Server 2019 (RTM-CU32-GDR) (KB5058722) - 15.0.4435.7 (X64)   Jun  9 2025 18:36:12   Copyright (C) 2019 Microsoft Corporation  Developer Edition (64-bit) on Windows Server 2019 Standard 10.0 <X64> (Build 17763: ) (Hypervisor) 

## Client Operating System
Windows 2000

## JAVA/JVM version
OpenJDK 21.0.5 LTS

## Problem description
Testing driver version 13.1.1 pre-release with bulk copy feature. Compared to 12.10.0 the driver is slower:
- about 10 % slower by just exchanging the driver
- about 50 % slower when enabling bulkCopy feature
(These are tests with our big application saving thousands ... to millions of records.)

## Expected behavior
Especially bulk copy wshould increase the speed.

## Actual behavior
Even withotu bulkCopy enabled, the new driver is slower.

## Error message/stack trace
none

## Any other details that can be helpful
See https://github.com/microsoft/mssql-jdbc/issues/2527#issuecomment-3164834761

machavan:
Hi @JDBC-0 ,

For Bulk Copy API functionality, the change 13.1.1 preview release that includes support for DATETIME datatype, which was missing previously, that should not really cause a performance regression, but we do want to investigate this further.    

Can you share a representative application which demonstrates this slowness?

JDBC-0:
[testPerformance.txt](https://github.com/user-attachments/files/21712161/testPerformance.txt)

Well, I reused the test that I wrote for the problem witgh "useFlexibleCallableStatements":  see "testPerformance.txt".

With this artificial test, I can see a performance improvement with 13.1.1 and useBulkCopyForBatchInsert=true for the "INSERT batch prepared" test.

I wonder why the new driver with bulkFeature enabled in our real world application is slower....



=> "Duration INSERT batch prepared"  ist faster with 13.1.1 and bulkcopy feature enabled ... in this artificial test - but not in our real world application.  :-/

A problem is that we don't see if bulk copy was used or not. Perhaps our real world application has still column types that are not supported and therefore the driver quietly doesn't use bulk copy?

JDBC-0:
[...]

Issue #2721, opened 2025-07-29
Accented characters saved incorrectly in database when using useBulkCopyForBatchInsert=true and sendStringParametersAsUnicode=false


BiancaPuertaRocha:
## Driver version
13.1.1.jre11-preview

## SQL Server version
Microsoft SQL Server 2022 (RTM) - 16.0.1000.6 (X64)   Oct  8 2022 05:58:25   Copyright (C) 2022 Microsoft Corporation  Developer Edition (64-bit)

## Client Operating System
Windows 10 Home Single Language 10.0 <X64> (Build 19045: ) 

## JAVA/JVM version
21

## Table schema
```
CREATE TABLE USUARIO (
    ID BIGINT IDENTITY(1,1) PRIMARY KEY,
    NOME_COMPLETO VARCHAR(100) NOT NULL,
    EMAIL VARCHAR(100) NOT NULL,
    GENERO CHAR(1) CHECK (GENERO IN ('M', 'F', 'O')),
    CPF CHAR(11) NOT NULL,
    TELEFONE VARCHAR(20),
    ENDERECO VARCHAR(200),
    DATA_NASCIMENTO DATETIME2,
    DATA_CRIACAO DATETIME2 DEFAULT SYSDATETIME(),
    RENDA_MENSAL DECIMAL(10, 2),
    STATUS CHAR(1) DEFAULT 'A' CHECK (STATUS IN ('A', 'I'))
);
```

## Problem description

When saving a register in database and using `useBulkCopyForBatchInsert=true` and `sendStringParametersAsUnicode=false` in the connection URL, the names with accented characters are saved incorrectly. This issue is critical for us, as our language makes extensive use of accent marks.

## Expected behavior
The names should be saved like this.
<img width="104" height="81" alt="Image" src="https://github.com/user-attachments/assets/55101079-d605-4dd7-b972-101572665310" />

## Actual behavior
The names are being saved like whis.
<img width="101" height="82" alt="Image" src="https://github.com/user-attachments/assets/1a8da664-fbc6-4b5c-be55-c01e55426501" />

## Error message/stack trace
No error message or stack trace.

## Any other details that can be helpful
The issue [Bulk copy for batch insert destroys data without error #2669](https://github.com/microsoft/mssql-jdbc/issues/2669) was helpful and has been resolved, but I am now experiencing a new and different error that appeared after the previous issue was fixed.

## JDBC trace logs
No errors

muskan124947:
[...]

PR #2735, opened 2025-08-11
Addressed performance issue on Bulk Copy API with batch insert

muskan124947:
## Description

A GitHub issue was raised highlighting a performance degradation in the Bulk Copy API when using batch inserts
[Pre-release 13.1 is slower #2732](https://github.com/microsoft/mssql-jdbc/issues/2732)

The original issue reported was [Bulk copy for batch insert destroys data without error #2669](https://github.com/microsoft/mssql-jdbc/issues/2669)
This occurred when inserting string data using the following settings:
```
- useBulkCopyForBatchInsert=true
- sendStringParametersAsUnicode=false
```
Under these conditions, data was inserted in byte array format rather than as readable strings.

Expected behavior
Select StateCode from states -> "OH"

Actual behavior
Select StateCode from states; -> "[B@4b01d9e4"

On top of this a new issue was reported [Accented characters saved incorrectly in database when using useBulkCopyForBatchInsert=true and sendStringParametersAsUnicode=false #2721](https://github.com/microsoft/mssql-jdbc/issues/2721).
When inserting data using useBulkCopyForBatchInsert=true and sendStringParametersAsUnicode=false, names with accented characters were saved incorrectly in the database.

## Root Cause
When inserting string data using batch insert for bulk copy with below combinations:

```
- useBulkCopyForBatchInsert=true
- sendStringParametersAsUnicode=false
```
[Fix#2704](https://github.com/microsoft/mssql-jdbc/pull/2704) and [Fix#2727](https://github.com/microsoft/mssql-jdbc/pull/2727) introduced a performance slowdown because of unnecessary conversion from byte array back to string during the bulk copy batch insert operation.

## Fix
Modified the logic to avoid sending data as a byte array in the first place when useBulkCopyForBatchInsert=true. As data inserted to destination table is being taken from dtv.value which is byte array.
Post this fix performance has improved significantly.

## Testing
- Added test cases for all combinations of useBulkCopyForBatchInsert and sendStringParametersAsUnicode to validate data insertion (including Unicode and accented characters) without performance degradation.
- Added a test case to simulate an unsupported INSERT-SELECT batch insert with useBulkCopyForBatchInsert=true and sendStringParametersAsUnicode=false, verifying that the fallback to normal execution occurs and the data is inserted correctly.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/dtv.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/dtv.java
@@ -2064,7 +2064,7 @@ final class AppDTVImpl extends DTVImpl {
             // SSType (i.e. VARCHAR vs. TEXT/VARCHAR(max)) is based on the exact length of
             // the MBCS value (in bytes).
             // If useBulkCopyForBatchInsert is true, conversion to byte array is not done due to performance
-            else if ((con.getUseBulkCopyForBatchInsert() == false) && null != collation && (JDBCType.CHAR == type || JDBCType.VARCHAR == type
+            else if (!con.getUseBulkCopyForBatchInsert() && null != collation && (JDBCType.CHAR == type || JDBCType.VARCHAR == type
                     || JDBCType.LONGVARCHAR == type || JDBCType.CLOB == type)) {
                 byte[] nativeEncoding = null;
 
--- a/src/main/java/com/microsoft/sqlserver/jdbc/dtv.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/dtv.java
@@ -2063,7 +2063,8 @@ final class AppDTVImpl extends DTVImpl {
             // then do the conversion now so that the decision to use a "short" or "long"
             // SSType (i.e. VARCHAR vs. TEXT/VARCHAR(max)) is based on the exact length of
             // the MBCS value (in bytes).
-            else if (null != collation && (JDBCType.CHAR == type || JDBCType.VARCHAR == type
+            // If useBulkCopyForBatchInsert is true, conversion to byte array is not done due to performance
+            else if ((con.getUseBulkCopyForBatchInsert() == false) && null != collation && (JDBCType.CHAR == type || JDBCType.VARCHAR == type
                     || JDBCType.LONGVARCHAR == type || JDBCType.CLOB == type)) {
                 byte[] nativeEncoding = null;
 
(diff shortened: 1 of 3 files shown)


Ananya2 (dismissed):

src/main/java/com/microsoft/sqlserver/jdbc/dtv.java
```diff
@@ -2063,7 +2063,8 @@ else if (type.isBinary()) {
             // then do the conversion now so that the decision to use a "short" or "long"
             // SSType (i.e. VARCHAR vs. TEXT/VARCHAR(max)) is based on the exact length of
             // the MBCS value (in bytes).
-            else if (null != collation && (JDBCType.CHAR == type || JDBCType.VARCHAR == type
+            // If useBulkCopyForBatchInsert is true, conversion to byte array is not done due to performance
+            else if ((con.getUseBulkCopyForBatchInsert() == false) && null != collation && (JDBCType.CHAR == type || JDBCType.VARCHAR == type
```
divang: Instead of else if ((con.getUseBulkCopyForBatchInsert() == false)), we can use:

```
else if ((!con.getUseBulkCopyForBatchInsert()) 
        && collation != null 
        && (JDBCType.CHAR == type || JDBCType.VARCHAR == type))

```

src/test/java/com/microsoft/sqlserver/jdbc/preparedStatement/BatchExecutionWithBulkCopyTest.java
```diff
@@ -1408,82 +1409,124 @@ private void getCreateTableTemporalSQL(String tableName) throws SQLException {
...
+            stmt.execute(createDestinationTableSQL);
+
+            // Attempt unsupported INSERT-SELECT query for bulk copy api
+            String insertSelectSQL = "INSERT INTO " + tableNameDestination + " SELECT * FROM " + tableNameSource
+                    + " WHERE id = ?";
+
+            try (SQLServerPreparedStatement pstmt = (SQLServerPreparedStatement) connection
+                    .prepareStatement(insertSelectSQL)) {
+                pstmt.setInt(1, 1);
+                pstmt.addBatch();
```
machavan: This is not really setting any String parameter in the prepared statement and will not exercise the code path we intend to verify with this. Basically we want the dtv to be created and no string to byte array conversion should happen and yet it should work fine. 

So we can change this select to `select ... where somestringcol = ?` ...

divang (approved):

Ananya2 (approved):

machavan (approved):

Issue #2722, opened 2025-07-29
Concatenated select SQL Statement returns null in JDBC Driver 12.10


psmacalp:
## Driver version
JDBC Driver 12.10

## SQL Server version
Provide the output of executing `SELECT @@VERSION` on your target SQL Server.
```
Microsoft SQL Server 2022 (RTM-GDR) (KB5058712) - 16.0.1140.6 (X64)   Jun 19 2025 11:40:25   Copyright (C) 2022 Microsoft Corporation  Developer Edition (64-bit) on Windows 10 Enterprise 10.0 <X64> (Build 26100: ) (Hypervisor)
```

## Client Operating System
Windows 11 Enterprise 24H2 

## JAVA/JVM version
IBM Java 8 SDK (running on IBM WebSphere Application Server traditional)

## Table schema
```
USE [OS1]
GO
/****** Object:  Table [dbo].[testTable]    Script Date: 7/24/2025 10:36:36 AM ******/
CREATE TABLE [dbo].[testTable](
	[c1] [int] NULL,
	[c2] [smallint] NULL
) ON [PRIMARY]
```

## Problem description
Starting with JDBC driver 12.10, the following resultset will be null and seems to ignore the select statement at the end. This is a regression from previous JDBC driver versions before 12.10. It is causing an issue in our product that has been deployed to many of our customers and prevents them from upgrading to JDBC driver 12.10.
```
PreparedStatement ps= conn.prepareStatement("delete from TestTable; INSERT INTO TestTable (c1,c2) VALUES (?, ?); select * from TestTable ");
ResultSet resultset = ps.execute();
```

We discovered the workaround here is to not append the select statement at the end of the SQL query.  However, when we do this we cannot include the select SQL Statement in a batch, which means that there is a performance hit due to this regression.  So, while we do have a workaround, we would very much prefer a fix to restore the previous behavior to prevent a performance regression from this workaround.

## Expected behavior
Expected behavior is that the concatenated select statement returns a value and does not return null.

## Actual behavior
The actual behavior is that the concatenated select statement is returning null for the ResultSet.

## Error message/stack trace
There is no error message, just a change in behavior causing the returned ResultSet to be null when it should not be null.

## Any other details that can be helpful
We have had multiple customers in the field report this issue when using Microsoft SQL JDBC Driver 12.10 with our product.  We reproduced the issue internally and independently of our product. We have simplified the reproduction steps in a way that does not require our product.

[...]

PR #2554, opened 2024-12-04
Issue#2550 - Fixed getGeneratedKeys functionality for execute API

machavan:
Description:

The execute API for INSERT was not reading off the explicit TDS_DONE token that actually contains the update count returned by the server.  As a result, it was not able to correctly report back the update count for getUpdateCount and also the subsequent resultsets on getMoreResults.

Note that this works correctly for excecuteUpdate API.

Testing:
- All existing tests pass
- Introduced a number of tests testing various combinations of execute/executeUpdate API, INSERT/UPDATE/DELETE/MERGE DML operations, followed by explicit SELECT statement or getGeneratedKeys call.

PR #2737, opened 2025-08-12
Fix update count handling for multi-statement queries executed via PreparedStatement. (#2722)

Ananya2:
**Description**
This PR addresses an inconsistency in how PreparedStatement handles update counts for multi-statement queries, particularly when INSERT operations are involved.

**Issue**
When executing multi-statement queries containing INSERT operations, PreparedStatement returned different update counts compared to Statement due to differences in Tabular Data Stream (TDS) token processing:

- Statement: Requires reading an additional TDS_DONE token for INSERT operations to obtain the correct update count.
- PreparedStatement: Does not require this additional token processing, but was incorrectly using the same logic as Statement, leading to inaccurate update counts.

Example of inconsistent behavior:
`PreparedStatement ps = conn.prepareStatement(
    "DELETE FROM TestTable; " +
    "INSERT INTO TestTable (c1,c2) VALUES (?, ?); " +
    "INSERT INTO TestTable (c1,c2) VALUES (?, ?); " +
    "UPDATE TestTable SET C1 = 1; " +
    "INSERT INTO TestTable (c1,c2) VALUES (?, ?); " +
    "SELECT * FROM TestTable"
);
do {
    boolean hasResults = ps.getMoreResults();
    System.out.println("update count: " + ps.getUpdateCount() + ", has result: " + hasResults);
} while (true);
`

- Driver 12.10 output:
`update count: 3, has result :false  
update count: 1, has result :false  
update count: 1, has result :false  
update count: 2, has result :false  
update count: 1, has result :false  
update count: -1, has result :true `
- Driver 13.1 output:
`update count: 3, has result :false  
update count: 2, has result :false  
update count: 1, has result :false  
update count: -1, has result :true`

**Root Cause**
In PR [#2554](https://github.com/microsoft/mssql-jdbc/pull/2554/files), the following logic was added to handle INSERT operations:
`if ((StreamDone.CMD_INSERT == doneToken.getCurCmd()) &&
    (-1 != doneToken.getUpdateCount()) &&
    EXECUTE == executeMethod) {
    return true;
}`
This logic is correct for Statement but not for PreparedStatement.
For PreparedStatement, the execute API for INSERT does not require reading an additional explicit TDS_DONE token — applying the same logic caused incorrect update counts.

**Fix**
Added overridden methods to correctly handle update count retrieval in both PreparedStatement and SQLServerStatement, ensuring that each uses the appropriate TDS token processing logic.

**Testing**
- Verified all existing tests pass.
- Added new test cases covering various combinations of execute API calls via PreparedStatement to validate update counts for INSERT, UPDATE, DELETE, and MERGE operations, followed by explicit SELECT statements.

**Closes** #2722

### DIFF
--- a/src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/StatementTest.java
+++ b/src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/StatementTest.java
@@ -3094,10 +3094,10 @@ public class StatementTest extends AbstractTest {
         /**
          * Tests multi-statement PreparedStatement with loop to process all results
          *
-         * @throws Exception
+         * @throws SQLException
          */
         @Test
-        public void testMultiStatementPreparedStatementLoopResults() {
+        public void testMultiStatementPreparedStatementLoopResults() throws SQLException {
             try (Connection con = getConnection()) {
                 try (PreparedStatement ps = con.prepareStatement("DELETE FROM " + tableName + " " +
                         "INSERT INTO " + tableName + " (NAME) VALUES (?) " +
@@ -3137,18 +3137,16 @@ public class StatementTest extends AbstractTest {
                         retval = ps.getMoreResults();
                     } while (true);
                 }
-            } catch (SQLException e) {
-                fail(TestResource.getResource("R_unexpectedException") + e.getMessage());
             }
         }
 
         /**
          * Tests PreparedStatement execute for Insert followed by select
          *
-         * @throws Exception
+         * @throws SQLException
          */
         @Test
-        public void testPreparedStatementExecuteInsertAndSelect() {
+        public void testPreparedStatementExecuteInsertAndSelect() throws SQLException {
             try (Connection con = getConnection()) {
                 String sql = "INSERT INTO " + tableName + " (NAME) VALUES(?) " +
                             "SELECT NAME FROM " + tableName + " WHERE ID = 1";
@@ -3176,18 +3174,16 @@ public class StatementTest extends AbstractTest {
                         retval = ps.getMoreResults();
                     } while (true);
                 }
-            } catch (SQLException e) {
-                fail(TestResource.getResource("R_unexpectedException") + e.getMessage());
             }
         }
 
         /**
          * Tests PreparedStatement execute for Merge followed by select
          *
-         * @throws Exception
+         * @throws SQLException
          */
         @Test
-        public void testPreparedStatementExecuteMergeAndSelect() {
+        public void testPreparedStatementExecuteMergeAndSelect() throws SQLException {
             try (Connection con = getConnection()) {
                 String sql = "MERGE INTO " + tableName + " AS target " +
                             "USING (VALUES (?)) AS source (name) " +
@@ -3219,18 +3215,16 @@ public class StatementTest extends AbstractTest {
                         retval = ps.getMoreResults();
                     } while (true);
                 }
-            } catch (SQLException e) {
-                fail(TestResource.getResource("R_unexpectedException") + e.getMessage());
             }
         }
 
         /**
          * Tests PreparedStatement execute two Inserts followed by select
          *
-         * @throws Exception
+         * @throws SQLException
          */
         @Test
-        public void testPreparedStatementExecuteTwoInsertsRowsAndSelect() {
+        public void testPreparedStatementExecuteTwoInsertsRowsAndSelect() throws SQLException {
             try (Connection con = getConnection()) {
                 try (PreparedStatement ps = con.prepareStatement("INSERT INTO " + tableName + " (NAME) VALUES(?) INSERT INTO " + tableName + " (NAME) VALUES(?) SELECT NAME from " + tableName + " WHERE ID = 1")) {
                     ps.setString(1, "test");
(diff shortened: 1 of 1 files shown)


src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/StatementTest.java
```diff
@@ -3091,6 +3091,244 @@ public void testExecuteDelAndSelect() {
...
+
+        /**
+         * Tests PreparedStatement execute for Insert followed by select
+         *
+         * @throws Exception
+         */
+        @Test
+        public void testPreparedStatementExecuteInsertAndSelect() {
+            try (Connection con = getConnection()) {
+                try (PreparedStatement ps = con.prepareStatement("INSERT INTO " + tableName + " (NAME) VALUES(?) SELECT NAME FROM " + tableName + " WHERE ID = 1")) {
```
divang: NIT: Can we split this query into multiple lines for better readability?
try (PreparedStatement ps = con.prepareStatement(
        "INSERT INTO " + tableName + " (NAME) VALUES(?) " +
        "SELECT NAME FROM " + tableName + " WHERE ID = 1")) {
    // ...
}

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/StatementTest.java
```diff
@@ -3091,6 +3091,244 @@ public void testExecuteDelAndSelect() {
...
+
+        /**
+         * Tests PreparedStatement execute for Merge followed by select
+         *
+         * @throws Exception
+         */
+        @Test
+        public void testPreparedStatementExecuteMergeAndSelect() {
+            try (Connection con = getConnection()) {
+                try (PreparedStatement ps = con.prepareStatement("MERGE INTO " + tableName + " AS target USING (VALUES (?)) AS source (name) ON target.name = source.name WHEN NOT MATCHED THEN INSERT (name) VALUES (?); SELECT NAME FROM " + tableName + " WHERE ID = 1")) {
```
divang: NIT: break into multiple lines:
String sql = 
    "MERGE INTO " + tableName + " AS target " +
    "USING (VALUES (?)) AS source (name) " +
    "ON target.name = source.name " +
    "WHEN NOT MATCHED THEN INSERT (name) VALUES (?); " +
    "SELECT NAME FROM " + tableName + " WHERE ID = 1";

try (PreparedStatement ps = con.prepareStatement(sql)) {

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerStatement.java
```diff
@@ -1845,6 +1844,24 @@ boolean consumeExecOutParam(TDSReader tdsReader) throws SQLServerException {
...
+     * the actual update count. This method can be overridden by subclasses to customize 
+     * TDS token processing behavior.
+     * 
+     * @param doneToken The current DONE token being processed
+     * @param executeMethod The execution method used
+     * @return true to continue processing more tokens to get the actual update count, 
+     *         false to stop and return this token
+     */
+    protected boolean hasUpdateCountTDSToken(StreamDone doneToken, int executeMethod) {
+        // For Insert, we must fetch additional TDS_DONE token that comes with the actual update count
```
machavan, line 1857: To be more explciit, hasUpdateCountTDSTokenForInsertCmd ?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerStatement.java
```diff
@@ -1845,6 +1844,24 @@ boolean consumeExecOutParam(TDSReader tdsReader) throws SQLServerException {
...
+        return (StreamDone.CMD_INSERT == doneToken.getCurCmd()) &&
+               (-1 != doneToken.getUpdateCount()) &&
+               (EXECUTE == executeMethod);
```
machavan: I think we should just return true from here and have the original logic as is in the caller with an additional check for hasUpdateCountTDSTokenForInsertCmd.

machavan (approved):

muskan124947 (approved):

divang (approved):

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/StatementTest.java
```diff
@@ -3091,6 +3091,251 @@ public void testExecuteDelAndSelect() {
...
+                                    rowCount++;
+                                }
+                                assertEquals(4, rowCount, "Expected 4 rows in result set");
+                            }
+                        }
+                        retval = ps.getMoreResults();
+                    } while (true);
+                }
+            } catch (SQLException e) {
+                fail(TestResource.getResource("R_unexpectedException") + e.getMessage());
```
David-Engel: I realize it's repeated from existing tests in this file, but this pattern is unnecessary here and in the rest of the new tests. Unhandled exceptions will fail the test. All this catch does is reduce the amount of failure information provided to junit to just the exception message. Without this catch, junit will be able to report exactly where the exception was thrown, call stack, inner exceptions, etc.

Also do this in the 5 additional places below.
```suggestion
```

src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/StatementTest.java
```diff
@@ -3091,6 +3091,251 @@ public void testExecuteDelAndSelect() {
             }
         }
 
+        /**
+         * Tests multi-statement PreparedStatement with loop to process all results
+         *
+         * @throws Exception
+         */
+        @Test
+        public void testMultiStatementPreparedStatementLoopResults() {
```
David-Engel: This is also needed in the 5 additional places below.
```suggestion
         * @throws SQLException
         */
        @Test
        public void testMultiStatementPreparedStatementLoopResults() throws SQLException {
```

muskan124947 (approved):

machavan (approved):

PR #2741, opened 2025-08-19
Fixed TDS meesage severity handling

divang:


src/main/java/com/microsoft/sqlserver/jdbc/tdsparser.java
```diff
@@ -245,12 +245,19 @@ boolean onRetValue(TDSReader tdsReader) throws SQLServerException {
...
     boolean onDone(TDSReader tdsReader) throws SQLServerException {
+        short status = tdsReader.peekStatusFlag();
         StreamDone doneToken = new StreamDone();
         doneToken.setFromTDS(tdsReader);
         if (doneToken.isFinal()) {
             // Response is completely processed hence decrement unprocessed response count.
             tdsReader.getConnection().getSessionRecovery().decrementUnprocessedResponseCount();
         }
+
+        if ((status & TDS.DONE_ERROR) != 0 || (status & TDS.DONE_SRVERROR) != 0) {
```
David-Engel: SqlClient treats DONE_SRVERROR as a fatal error:
https://github.com/dotnet/SqlClient/blob/v6.1.1/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs#L3167

Fatal errors ultimately close the connection and signal up to connection pools about the error. We need to plumb through something similar. See makeFromDatabaseError in SQLServerException.java. We don't have the database error, so we will need to figure something out there.

divang: I think, adding Error Severity 20 will make sure the notifyPooledConnection() call.
SQLServerError syntheticError = new SQLServerError();
syntheticError.setErrorMessage(SQLServerException.getErrString("R_severeError"));
**syntheticError.setErrorSeverity((byte) 20);** // A severity of 20 indicates a fatal error in the current
                                            // process.
                                            
 Ref: makeFromDatabaseError in SQLServerException.java:
// Close the connection if we get a severity 20 or higher error class (nClass is severity of error).
if ((sqlServerError.getErrorSeverity() >= 20) && (null != con)) {
    con.notifyPooledConnection(theException);
    con.close();
}

machavan (dismissed):

machavan (approved):

Issue #2739, opened 2025-08-15
Cannot relate procedures to procedure columns (same with functions)


sualeh:
## Driver version
13.2.0

## SQL Server version
Microsoft SQL Server 2019 (RTM-CU32-GDR) (KB5058722) - 15.0.4435.7 (X64)   Jun  9 2025 18:36:12   Copyright (C) 2019 Microsoft Corporation  Enterprise Edition: Core-based Licensing (64-bit) on Windows Server 2019 Datacenter 10.0 <X64> (Build 17763: ) (Hypervisor) 

## Client Operating System
Windows 11

## JAVA/JVM version
Java 21

## Problem description

Since
`getProcedures` returns
- `PROCEDURE_NAME` without the numeric portion (for example, without ";1")
- `SPECIFIC_NAME` is null

it is not possible to relate the parameters using

`getProcedureColumns` which returns
- `PROCEDURE_NAME` **with** the numeric portion (for example, including ";1")
- `SPECIFIC_NAME` is null

and the same for functions.

## Suggested behavior

`getProcedures` would return
- `PROCEDURE_NAME` **without** the numeric portion (for example, without ";1")
- `SPECIFIC_NAME` as procedure name **with** the numeric portion (for example, including ";1")

and relate parameters with

`getProcedureColumns` which would return
- `PROCEDURE_NAME` **without** the numeric portion (for example, without ";1")
- `SPECIFIC_NAME` as procedure name **with** the numeric portion (for example, including ";1")

and the same for functions.

sualeh:
The suggested change in how the `SPECIFIC_NAME` is returned (compatible with how the `PROCEDURE_NAME` was returned by older versions of the driver) will allow programmers to programmatically maintain backward compatibility where needed.

sualeh:
See https://github.com/microsoft/mssql-jdbc/issues/448

I had made a similar suggestion in that issue too.

muskan124947:
@sualeh , Earlier, getProcedures() and getFunctions() did not return correct results. You can refer to the issue : [DatabaseMetadata.getProcedures and getFunctions return both functions and procedures #2690](https://github.com/microsoft/mssql-jdbc/issues/2690)

[...]

PR #2705, opened 2025-07-15
Updated getFunctions() and getProcedures() to use sys.all_objects to give correct results post filtering

muskan124947:

PR #2743, opened 2025-08-20
Update getProcedureColumns() and getFunctionColumns() to return name without numbered prefixes

muskan124947:
## Description
This PR addresses GitHub reported issue : [Cannot relate procedures to procedure columns (same with functions) #2739](https://github.com/microsoft/mssql-jdbc/issues/2739)

Post fixing the issue [Use sys.all_objects for accurate function and procedure filtering #2705](https://github.com/microsoft/mssql-jdbc/pull/2705) : where both DatabaseMetadata.getProcedures and DatabaseMetadata.getFunctions returns correct result. With this change object names got updated as mentioned below:
```
Old behaviour: sp_testProc;1
New behaviour: sp_testProc
```

So, DatabaseMetadata.getProcedureColumns and DatabaseMetadata.getFunctionColumns should also return similar pattern.
Currently it returns the resultset where procedure and function names were appended with a numbered suffix (;1).

As per the [documentation](https://learn.microsoft.com/en-us/previous-versions/sql/sql-server-2008/ms187926(v=sql.100)?redirectedfrom=MSDN#:~:text=This%20feature%20will%20be%20removed%20in%20a%20future%20version%20of%20Microsoft%20SQL%20Server.%20Avoid%20using%20this%20feature%20in%20new%20development%20work%2C%20and%20plan%20to%20modify%20applications%20that%20currently%20use%20this%20feature.) : “Numbered procedures are deprecated. Avoid using them in new development work, and plan to modify applications that currently use them.”

## Fix
Updated the implementation of getProcedureColumns() and getFunctionColumns() to filter out numbered suffixes from procedure/function names.
This ensures that the returned column metadata reflects only the expected procedure/function names.

## Testing
- Manual testing: Verified locally with custom scripts to ensure that getProcedureColumns() and getFunctionColumns() return accurate object names without numbered suffixes.
- Automated tests: Added testGetProcedureColumnsWithValidation and testGetFunctionColumnsWithValidation to confirm that procedure and function column names are correctly validated.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
@@ -1130,6 +1130,7 @@ public final class SQLServerDatabaseMetaData implements java.sql.DatabaseMetaDat
                 rs.getColumn(9).setFilter(new ZeroFixupFilter());
                 rs.getColumn(17).setFilter(new ZeroFixupFilter());
             }
+            rs.getColumn(3).setFilter(new ColumnNameFilter());
         }
         return rs;
     }
@@ -2948,14 +2949,14 @@ class ZeroFixupFilter extends IntColumnFilter {
 }
 
 /**
- * Provides filter to remove numbered prefixes from procedure names.
+ * Provides filter to remove numbered prefixes from procedure and function names.
  */
 class ColumnNameFilter extends ColumnFilter {
     @Override
     public Object apply(Object value, JDBCType asJDBCType) throws SQLServerException {
         if (value instanceof String) {
             String name = (String) value;
-            int idx = name.indexOf(';');
+            int idx = name.lastIndexOf(';');
             return (idx > 0) ? name.substring(0, idx) : name;
         }
         return value;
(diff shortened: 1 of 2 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
```diff
@@ -2946,6 +2947,20 @@ int oneValueToAnother(int precl) {
     }
 }
 
+/**
+ * Provides filter to remove numbered prefixes from procedure names.
+ */
+class ColumnNameFilter extends ColumnFilter {
```
machavan, line 2954: can rename to ProcNameFilter

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
```diff
@@ -2946,6 +2947,20 @@ int oneValueToAnother(int precl) {
...
+    @Override
+    public Object apply(Object value, JDBCType asJDBCType) throws SQLServerException {
+        if (value instanceof String) {
+            String name = (String) value;
+            int idx = name.indexOf(';');
```
machavan: can change to use lastIndexOf

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
muskan124947: It will be used for getFunctionColumns() also

PR #2747, opened 2025-08-21
Fix issue#2745

prrvchr:
If the driver supports primary key and foreign key relationships then normally it should respond `true` regarding the `DatabaseMetaData.supportsIntegrityEnhancementFacility()` method.


prrvchr:
Thanks to whoever took the trouble to explain why it was important for the driver to respond positively to `supportsIntegrityEnhancementFacility()`.

divang:
Thanks for contributing! The changes are part of https://github.com/microsoft/mssql-jdbc/pull/2828 and now part of the 13.3.0 release. You can find the release details here: https://github.com/microsoft/mssql-jdbc/releases

PR #2795, opened 2025-10-08
Fix bulk insert parsing of isolated quotes in tab-delimited data (#2792)

### LINKED ISSUE #2792: Bulk insert does not handle " properly
## Driver version
13.2.0.jre11

## SQL Server version
Microsoft SQL Server 2019 (RTM-CU32-GDR) (KB5065222) - 15.0.4445.1 (X64)  	Aug 13 2025 11:07:18  	Copyright (C) 2019 Microsoft Corporation 	Developer Edition (64-bit) on Windows 10 Pro 10.0 <X64> (Build 19045: ) 

## Client Operating System
Ubuntu 24.04.3 LTS

## JAVA/JVM version
Java 21 (openjdk 21.0.8 2025-07-15)

## Problem description
Bulk insert does not handle string containing ". 

I call the MsSqlServer:doBulkInsert with these parameters: 
_bulkInsertBase = {BulkInsertBase@12886} 
 rowCount = 17562
 ignoreCount = 0
 fieldsTerminatedBy = "\t"
  value = {byte[1]@12923} [9]
  coder = 0
  hash = 9
  hashIsZero = false
 tableName = "syncphrase"
 forceEmpty = false
 recreateTable = true
 keepIdentity = false
 tableLock = false
 fields = {Arrays$ArrayList@12911}  size = 5
  0 = {StringField@12914} "Name: namespace, DbName: namespace, Pos: 7, Type: java.lang.String, Database type: java.lang.String, Allow null: no, Size: 500, Truncated: false, Fulltext indexed: false, Default value: , Collation: default"
  1 = {StringField@12915} "Name: key, DbName: key\_, Pos: 6, Type: java.lang.String, Database type: java.lang.String, Allow null: no, Size: 500, Truncated: false, Fulltext indexed: false, Default value: , Collation: default"
  2 = {StringField@12916} "Name: phrase_en, DbName: phrase_en, Pos: 4, Type: java.lang.String, Database type: java.lang.String, Allow null: no, Size: 2000, Truncated: false, Fulltext indexed: false, Default value: , Collation: default"
  3 = {StringField@12917} "Name: taskId, DbName: taskId, Pos: 10, Type: java.lang.String, Database type: java.lang.String, Allow null: no, Size: 20, Truncated: false, Fulltext indexed: false, Default value: , Collation: default"
  4 = {EnumField@12918} "Name: type, DbName: type, Pos: 11, Type: se.netset.phrases.po.AteaPersistentPhrase$PhraseType, Database type: se.netset.phrases.po.AteaPersistentPhrase$PhraseType, Allow null: no"
rowTerminator = "\n"
file = "/tmp/bulktemp4584663740698327138.tmp"_

The file contains a lot of rows, here is a snippet where line 2 and 3 are the ones causing the problem:
...
com.company.test	REMOVE_PRODUCT	Remove product	22451	0
com.company.test	REMOVE_PRODUCT_START	Do you wish to remove the product "	22451	1
com.company.test	REMOVE_PRODUCT_END	" from the report?	22451	1
com.company.test	REPORT	Report	22451	0
...

Row 2 and 3 contains " and should be treated as part of the string to be inserted to the table. 
But it results in an IndexOutOfBoundsException: 
at _com.microsoft.sqlserver.jdbc.SQLServerBulkCSVFileRecord.getRowData(SQLServerBulkCSVFileRecord.java:325)
at com.microsoft.sqlserver.jdbc.SQLServerBulkCopy.writeBatchData(SQLServerBulkCopy.java:3717)
at com.microsoft.sqlserver.jdbc.SQLServerBulkCopy.doInsertBulk(SQLServerBulkCopy.java:1590)
at com.microsoft.sqlserver.jdbc.SQLServerBulkCopy$1InsertBulk.doExecute(SQLServerBulkCopy.java:681)
at com.microsoft.sqlserver.jdbc.TDSCommand.execute(IOBuffer.java:7745)
at com.microsoft.sqlserver.jdbc.SQLServerConnection.executeCommand(SQLServerConnection.java:4700)
at com.microsoft.sqlserver.jdbc.SQLServerBulkCopy.sendBulkLoadBCP(SQLServerBulkCopy.java:715)
at com.microsoft.sqlserver.jdbc.SQLServerBulkCopy.writeToServer(SQLServerBulkCopy.java:1678)
at com.microsoft.sqlserver.jdbc.SQLServerBulkCopy.writeToServer(SQLServerBulkCopy.java:638)
...._

This is because data[] only contains 3 strings instead of 5 as expected. So the problem is that SQLServerBulkCSVFileRecord:parseString(String buffer, String delimiter)  is only returning an array of 3 instead of 5 Strings. 

## Expected behavior
It should be possible to bulk insert strings that contain "

## Actual behavior
Described above.

My code works as expected if I toggle to using versions prior to v12.7.1, when this commit was merged https://github.com/microsoft/mssql-jdbc/commit/722c910a0398bf6c8c37622dc470367053e73ff1

Ananya2:
**Overview**
This PR fixes bulk insert parsing of isolated quotes in tab-delimited data by removing problematic global quote state tracking from the `parseString` method in `SQLServerBulkCSVFileRecord`. The fix ensures that isolated quote characters are treated as literal data rather than field boundary markers, resolving IndexOutOfBoundsException errors during bulk copy operations.

**Problem Description**
The current implementation uses a global `quoted` boolean state in the `parseString` method that toggles on every quote character encounter. This causes issues when tab-delimited data contains isolated quotes within fields:
`if (buffer.charAt(i) == doubleQuoteChar) `{`
    quoted = !quoted;
} else if (!quoted && /* delimiter found */) {
    // Process delimiter
}`

When parsing data like "Do you wish to remove the product "\t22451\t1", the isolated quote incorrectly toggles the quoted state, causing subsequent tab delimiters to be ignored. This results in:
Expected: 5 fields parsed correctly
Actual: 3 fields parsed, causing `IndexOutOfBoundsException`

**Root Cause**
PR #2434 introduced quote handling logic to fix stack overflow issues in CSV parsing. While the fix successfully resolved the stack overflow problem for CSV files, it created a new issue where isolated quotes in tab-delimited data are treated as field boundary markers instead of literal characters.

**Solution**
Reverted to using `currentLine.split(delimiter, -1)` instead of `parseString(currentLine, delimiter)` for simple delimiter-based parsing.
Maintained stack overflow fix from PR #2434 while fixing the quote parsing regression.
Added comprehensive test coverage with the exact problematic data patterns from issue #2792

**Testing**

- Added testBulkCopyTabDelimitedWithQuotes() test case with problematic data from issue #2792

Closes #2792

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCSVFileRecord.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCSVFileRecord.java
@@ -22,6 +22,7 @@ import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.Map.Entry;
+import java.util.regex.Pattern;
 
 import microsoft.sql.Vector;
 
@@ -211,7 +212,7 @@ public class SQLServerBulkCSVFileRecord extends SQLServerBulkRecord implements j
                 columnNames = (escapeDelimiters && currentLine.contains("\""))
                                                                                ? escapeQuotesRFC4180(parseString(
                                                                                        currentLine, delimiter))
-                                                                               : currentLine.split(delimiter, -1);
+                                                                               : currentLine.split(Pattern.quote(delimiter), -1);
             }
         }
     }
@@ -299,7 +300,7 @@ public class SQLServerBulkCSVFileRecord extends SQLServerBulkRecord implements j
             String[] data = (escapeDelimiters && currentLine.contains("\""))
                                                                              ? escapeQuotesRFC4180(parseString(
                                                                                      currentLine, delimiter))
-                                                                             : currentLine.split(delimiter, -1);
+                                                                             : currentLine.split(Pattern.quote(delimiter), -1);
 
             // Cannot go directly from String[] to Object[] and expect it to act as an array.
 


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCSVFileRecord.java
```diff
@@ -299,7 +299,7 @@ public Object[] getRowData() throws SQLServerException {
             String[] data = (escapeDelimiters && currentLine.contains("\""))
                                                                              ? escapeQuotesRFC4180(parseString(
```
machavan, line 301: So , does this mean The call to parseString from escaleQuotesRFC4180 is what is required to address the stackoverflow issue in the FIx from PR#2434. The else part of it when the line does not contain quote character can be dealt with original logic of curreneLine.split. ?

Ananya2: Yes, and also updated to handle separators like "|" by using Pattern.quote(delimiter) to prevent regex issues.

abbrev:
This removes support for a regex delimiter.

After digging through old versions of the code, it appears that the original support for a regex delimiter was an accidental feature, but it might not have worked properly in some cases (especially when `escapeDelimiters` is true and there's a double quote in the line). Also, a note about regex was added to comments in #1711 in response to #1691.

At any rate, a delimiter cannot be a regex as of this merge, so all I propose now is to remove the mention of regex in the following comment (it appears twice in SQLServerBulkCSVFileRecord.java) since it is no longer true:

```java
     *        Delimiter to used to separate each column. Regex characters must be escaped with double backslashes.
```

Ananya2:
@abbrev - Thanks for the detailed context.
Since this pull request removes support for regex delimiters and the update you’re suggesting is about correcting comments /documentation rather than code behavior addressed here, could you please open a new issue to track the documentation change?

PR #2819, opened 2025-10-21
Allow minus sign in point wkt

### LINKED ISSUE #2820: Creating a Geography with very small coordinates causes a NumberFormatException
## Driver version
12.10.1 & latest main branch

## SQL Server version
2019 (but does not matter)

## Client Operating System
macOS / Linux

## JAVA/JVM version
Java 21

## Table schema
-

## Problem description
When creating a `Geography` object for a spacial column, the parsing of the values fail with an `NumberFormatException` when using very small coordinates that have 3 or more zeros after the decimal dot:

```
Geography g = Geography.point(0.0001234, 1.234, 4326);
```

Java converts this double value to the string `1.234E-4`, which cannot be parsed correctly by `SQLServerSpatialDatatype.readPointWkt`. When iterating the wkt to find the end of the number, it does not take the `-` into account and tries to parse `1.234E` only. `BigDecimal` cannot parse this and throws the exception.

## Expected behavior

`g` should be instantiated correctly.

## Actual behavior

The code crashed with a `NumberFormatException`.

I have already analysed and fixed the issue :) PR: https://github.com/microsoft/mssql-jdbc/pull/2819

## JDBC trace logs
n/a

amuttsch:
Very small double values that contain three or more zeros after the dot (like 0.000123) have a string representation of 1.23E-4. The readPointWkt did not account for this and set the end pos to the E, thus the BigDecimal tried to parse "1.23E". This resulted in a NumberFormatException.

This fixes the issue by also allowing the minus sign when determining the end of the number.

Please backport this fix to version 12, thanks!

Fixes #2820

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerSpatialDatatype.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerSpatialDatatype.java
@@ -1484,7 +1484,8 @@ abstract class SQLServerSpatialDatatype {
 
             while (currentWktPos < wkt.length()
                     && (Character.isDigit(wkt.charAt(currentWktPos)) || wkt.charAt(currentWktPos) == '.'
-                            || wkt.charAt(currentWktPos) == 'E' || wkt.charAt(currentWktPos) == 'e')) {
+                            || wkt.charAt(currentWktPos) == 'E' || wkt.charAt(currentWktPos) == 'e'
+                            || wkt.charAt(currentWktPos) == '-')) {
                 currentWktPos++;
             }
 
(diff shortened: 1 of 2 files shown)


divang:
Thank you @amuttsch for contributing to this fix.
For community PRs, we follow a process where, instead of merging directly, we create internal PRs based on your contributions. This is because we run a set of internal tests which cannot be executed with external PRs. We use our template format for the PR description, validate the tests, and then proceed with the merge.

Your contribution is valuable and is reflected in the internal PR linked (https://github.com/microsoft/mssql-jdbc/pull/2837).

Thank you again for your valued input!

amuttsch:
Awesome, thank you @divang !

Issue #2745, opened 2025-08-21
DatabaseMetaData.supportsIntegrityEnhancementFacility() should return true


prrvchr:
## Driver version
mssql-jdbc-13.2.0.jre11.jar

## SQL Server version
Microsoft SQL Server 2025 (CTP2.1) - 17.0.800.3 (X64) 
	Jun 12 2025 14:47:57 
	Copyright (C) 2025 Microsoft Corporation
	Standard Developer Edition (64-bit) on Windows 10 Pro 10.0 <X64> (Build 19045: )

## Client Operating System
Lubuntu 22.4

## JAVA/JVM version
Java 17.0.15

## Table schema
Standart table schema.

## Problem description
The DatabaseMetaData.supportsIntegrityEnhancementFacility() return false.
## Expected behavior
The DatabaseMetaData.supportsIntegrityEnhancementFacility() should return true.
## Actual behavior
DatabaseMetaData.supportsIntegrityEnhancementFacility() output a false value.
## Error message/stack trace
No error message.
## Any other details that can be helpful
I can submit a PR?

## JDBC trace logs
No logs.

divang:
@prrvchr 
Thanks for reporting this issue—we appreciate your contribution. Please share your PR, as it would be very helpful. In the meantime, we are also investigating it.

prrvchr:
@divang OK done [PR#2747](https://github.com/microsoft/mssql-jdbc/pull/2747)

divang:
@prrvchr 
Please include the relevant test cases impacted by this change. The test cases should be designed to be use-case driven.

divang:
Thanks for contributing @prrvchr! The changes from https://github.com/microsoft/mssql-jdbc/pull/2828
 are now part of the 13.3.0 release. You can find the release details here: https://github.com/microsoft/mssql-jdbc/releases

PR #2828, opened 2025-11-04
SQL Server integrity enhancement support in supportsIntegrityEnhancementFacility #2745

divang:
**Problem**:
The DatabaseMetaData.supportsIntegrityEnhancementFacility() method incorrectly returns false for SQL Server databases. This method is part of the JDBC specification and indicates whether the database supports integrity enhancement facilities such as primary key and foreign key constraints. SQL Server has comprehensive support for these integrity features, but the driver was not accurately reporting this capability to applications. This could lead to applications incorrectly assuming that SQL Server lacks integrity constraint support, potentially affecting application logic and database schema decisions.

**Root Cause**:
The supportsIntegrityEnhancementFacility() method in SQLServerDatabaseMetaData.java was hardcoded to return false. This was an incorrect implementation as SQL Server fully supports integrity enhancement facilities including:
Primary key constraints
Foreign key constraints
Check constraints
Unique constraints
NOT NULL constraints

The method implementation did not reflect the actual capabilities of SQL Server, causing a mismatch between the database's features and what the driver reported.

**Solution**:
Updated the supportsIntegrityEnhancementFacility() method in SQLServerDatabaseMetaData.java to return true instead of false. This change aligns the driver's reported capabilities with SQL Server's actual support for integrity enhancement facilities. The fix ensures that applications using this metadata method will receive accurate information about SQL Server's constraint support capabilities.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java
@@ -2289,7 +2289,7 @@ public final class SQLServerDatabaseMetaData implements java.sql.DatabaseMetaDat
     @Override
     public boolean supportsIntegrityEnhancementFacility() throws SQLServerException {
         checkClosed();
-        return false;
+        return true;
     }
 
     @Override


Ananya2 (approved):

machavan:
@prrvchr ,

Thank you for contributing to this fix. 

Your contribution does show up in this PR as above.



image


We need to follow this process for community PRs (we have done the same for previous set of community contributions) as we are not in position to run a set of internal tests to validate the PR with your contributed PRs. Hence, we create PRs out of your PRs with PR description in our template format, validate the tests and go ahead with the PR merge. 

Your contribution is valuable and it does show up in the PR above.

Thanks

PR #2837, opened 2025-11-12
Fix parsing of Geography coordinates with scientific notation #2819

divang:
**Problem**
When creating Geography objects with very small coordinate values (e.g., 0.0001234), a NumberFormatException is thrown. This occurs because coordinates with three or more zeros after the decimal point are represented in scientific notation (e.g., "1.23E-4") when converted to strings.

**Root Cause**
The readPointWkt() method in SQLServerSpatialDatatype.java parses Well-Known Text (WKT) coordinate strings. The parsing loop that identifies the end position of a numeric value checks for digits, decimal points, and the exponent indicators 'E' and 'e', but does not account for the minus sign (-) that appears in negative exponents.

When parsing "1.23E-4", the loop terminates at the 'E' character instead of continuing through the full scientific notation. This causes BigDecimal to attempt parsing "1.23E" (an incomplete scientific notation), resulting in a NumberFormatException.

**Solution**
Added support for the minus sign (-) in the character validation loop within readPointWkt(). This allows the parser to correctly identify the complete scientific notation string including negative exponents.

**Test**
A new test case testGeographySmallCoordinates() has been added to verify that Geography objects can be created with small coordinate values that result in scientific notation.

### DIFF
--- a/src/test/java/com/microsoft/sqlserver/jdbc/datatypes/SQLServerSpatialDatatypeTest.java
+++ b/src/test/java/com/microsoft/sqlserver/jdbc/datatypes/SQLServerSpatialDatatypeTest.java
@@ -2128,91 +2128,35 @@ public class SQLServerSpatialDatatypeTest extends AbstractTest {
      */
     @Test
     public void testGeographySmallCoordinates() throws SQLException {
-        // Test 1: Positive small latitude (original test case)
-        Geography g1 = Geography.point(0.0001234, 1.234, 4326);
-        assertEquals(0.0001234, g1.getLatitude(), 1e-10, "Latitude should match for positive small value");
-        assertEquals(1.234, g1.getLongitude(), 1e-10, "Longitude should match");
+        // Scientific notation with 'e' (lowercase)
+        Geography g1 = Geography.point(1.234e-1, 5.678e-1, 4326);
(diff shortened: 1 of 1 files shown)


divang:
Rerun CI pipeline was successfully - https://sqlclientdrivers.visualstudio.com/public/_build/results?buildId=130697&view=results

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/SQLServerSpatialDatatypeTest.java
```diff
@@ -2122,6 +2122,53 @@ public void testLargeCases() throws SQLException {
...
 
+    /**
+     * Tests Geography almost zero coordinates like 0.0001234. The string representation is "1.234E-4", which
+     * caused a bug when creating a Geography object.
+     */
+    @Test
+    public void testGeographySmallCoordinates() throws SQLException {
+        // Test 1: Positive small latitude (original test case)
+        Geography g1 = Geography.point(0.0001234, 1.234, 4326);
+        assertEquals(0.0001234, g1.getLatitude(), 1e-10);
```
machavan: Fix the message in the assertEquals calls.
Add - and e in values for some of the test cases.

muskan124947 (approved):

Issue #2825, opened 2025-10-24
PreparedStatement.executeBatch() fails for VARBINARY function parameters with bulk copy enabled


pranav-j-zstch1447:
## Driver version

Microsoft JDBC Driver for SQL Server 13.2.0

## SQL Server version

Microsoft SQL Server 2019 (RTM)

## Client Operating System

Ubuntu 22.04 LTS (x86_64)

## JAVA/JVM version

1.8.0_352

## Table schema
```sql
CREATE TABLE FunctionInsertTest (
ID int PRIMARY KEY,
Data varbinary(max)
);
```

## Problem description

When using Bulk Copy for Batch Insert with `setUseBulkCopyForBatchInsert(true)`, inserting data into a VARBINARY(MAX) column via `PreparedStatement` with SQL functions like `encryptbykey()` fails. The failure occurs when using `setString()` for the parameter that is passed to the SQL function.

## Reproduction Code
```java
import com.microsoft.sqlserver.jdbc.SQLServerConnection;
import java.sql.*;

public class FunctionInsert {
    public static void main(String[] args) throws SQLException {
        String url = "jdbc:sqlserver://<HOST>:1433;databaseName=<DB>;trustServerCertificate=true;";
        String insertSQL = "INSERT INTO FunctionInsertTest (ID, Data) VALUES (?, encryptbykey(key_guid('SYMM_KEY'),?));";

        try (Connection connection = DriverManager.getConnection(url, "sa", "<PASSWORD>")) {
            SQLServerConnection sqlServerConnection = (SQLServerConnection) connection;
            sqlServerConnection.setUseBulkCopyForBatchInsert(true);
            sqlServerConnection.setBulkCopyForBatchInsertBatchSize(1);

            PreparedStatement statement = connection.prepareStatement(insertSQL);
            statement.setObject(1, 4);
            statement.setString(2, "MySecretKey123");  // <--- causes failure

            statement.addBatch();
            statement.executeBatch();
        } catch (Exception e) {
            e.printStackTrace();
        }
    }
}
```

## Expected behavior

`PreparedStatement.executeBatch()` should work correctly when a String parameter is passed to a SQL Server function or expression that expects VARBINARY input, even when bulk copy for batch insert is enabled. If bulk copy does not support certain data types or functions, it should fallback to the normal batch insert approach to ensure execution succeeds without errors.

## Actual behavior

Bulk copy fails with an exception, and the data is not inserted.

## Error message/stack trace

```
[...]

PR #2845, opened 2025-11-19
Handling proper fallback for SQL functions in bulk copy batch insert operations

muskan124947:
## Description
GitHub issue : [PreparedStatement.executeBatch() fails for VARBINARY function parameters with bulk copy enabled #2825](https://github.com/microsoft/mssql-jdbc/issues/2825)

This PR addresses a critical bug where PreparedStatement.executeBatch() fails when SQL functions are used as parameters in INSERT statements with bulk copy enabled (useBulkCopyForBatchInsert=true). The driver now properly detects incompatible SQL constructs and automatically falls back to standard batch execution.

## Problem
When bulk copy for batch insert is enabled, the driver attempts to process all batch operations through the Bulk Copy API, including those containing SQL functions like len(?), encryptbykey(), or other expressions wrapping parameters. However, as documented in the [Bulk Copy API limitations](https://learn.microsoft.com/en-us/sql/connect/jdbc/use-bulk-copy-api-batch-insert-operation?view=sql-server-ver17#:~:text=Wildcards%20(%3F)%20are%20the%20only%20supported%20parameters%20for%20this%20function.), wildcards (?) are the only supported parameters for this functionality.

This incompatibility resulted in various failures:

- "Parsing user's Batch Insert SQL Query failed: Number of provided columns X does not match the table definition Y"
- Invalid data type conversion errors
- "Unable to retrieve data from the source"
- Complete batch operation failures without proper fallback mechanism

## Fix
Enhanced the fallback detection logic to identify when SQL functions or unsupported expressions are present in parameterized INSERT statements during bulk copy batch operations. When such cases are detected:

- Bulk copy processing is bypassed
- The driver automatically switches to standard PreparedStatement batch execution
- Operations complete successfully without data loss or user intervention

## Testing
- Added comprehensive test in `BatchExecutionWithBulkCopyTest` to validat fallback behavior for SQL function scenarios
- Confirmed no regression for standard bulk copy operations
- Verified proper logging and error messaging during fallback detection

This fix ensures reliable batch operations for all SQL statement types while maintaining optimal performance for compatible bulk copy scenarios.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
@@ -2855,12 +2855,16 @@ public class SQLServerPreparedStatement extends SQLServerStatement implements IS
                 continue;
             }
             if (localUserSQL.charAt(0) == ',' || localUserSQL.charAt(0) == ')') {
+
+                if (!"?".equals(sb.toString())) {
+                    // throw IllegalArgumentException and fallback to original logic for batch insert
+                    // Wildcards (?) are the only supported parameters for this functionality
+                    // Does not support functions or literals (e.g. len(), 'string', 1, etc.)
+                    throw new IllegalArgumentException(SQLServerException.getErrString("R_onlyFullParamAllowed"));
+                }
+
                 if (localUserSQL.charAt(0) == ',') {
                     localUserSQL = localUserSQL.substring(1);
-                    if (!"?".equals(sb.toString())) {
-                        // throw IllegalArgumentException and fallback to original logic for batch insert
-                        throw new IllegalArgumentException(SQLServerException.getErrString("R_onlyFullParamAllowed"));
-                    }
                     listOfValues.add(sb.toString());
                     sb.setLength(0);
                 } else {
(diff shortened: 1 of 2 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -2855,19 +2855,24 @@ private ArrayList<String> parseUserSQLForValueListDWHelper(ArrayList<String> lis
                 continue;
             }
             if (localUserSQL.charAt(0) == ',' || localUserSQL.charAt(0) == ')') {
+
+                if (!"?".equals(sb.toString())) {
```
machavan, line 2859: We can consider adding tests with ? in the data in say first argument and second argument as the first characters

e.g. 
insert into t1 values('?abc', ?);
insert into t1 values(?, '?abc');


Do these qualify for bulk copy? if not do they fallback to normal prep stmt execution

muskan124947: As per docs, only placeholder (?) is allowed as parameter for bulk copy batch insert
This will fall back to normal execution.

machavan (approved):

divang (approved):

PR #2868, opened 2025-12-22
[VECTOR_FLOAT16] Implement feature extension and version negotiation for Vector v2 support

muskan124947:
## Description

For upcoming VECTOR_FLOAT16 support in the Microsoft JDBC Driver for SQL Server, this PR introduces the initial framework for vector type support handling on the client side, enabling controlled negotiation of vector capabilities between client and server.

The changes focus on making vector support explicit, extensible, and future-proof, in preparation for additional vector versions and data types.

## Vector Version Negotiation Flow

- The client vector setting is parsed and validated:
```
off – vector support disabled
v1 – float32 vector support
v2 – float32 and float16 vector support
```
- If either the client disables vectors or the server does not support vectors, vector support is disabled.
- The client and server determine their maximum supported vector versions.
- When both support vectors, the negotiated version is the minimum of the client and server versions.
- The negotiated version is used for all vector-related operations for the connection.

## Impact

- Backward compatibility: Clients using only float32 continue to work without changes.
- Performance: Float16 support reduces memory usage and network payload for supported operations.
- Extensibility: Framework allows easy addition of future vector types and versions.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
@@ -7125,7 +7125,8 @@ public class SQLServerConnection implements ISQLServerConnection, java.io.Serial
 
             case TDS.TDS_FEATURE_EXT_VECTORSUPPORT: {
                 if (connectionlogger.isLoggable(Level.FINE)) {
-                    connectionlogger.fine(toString() + " Received feature extension acknowledgement for vector support. Received byte: " + data[0]);
+                    connectionlogger.fine(
+                            toString() + " Received feature extension acknowledgement for Vector Support.");
                 }
 
                 if (1 != data.length) {
@@ -7183,39 +7184,11 @@ public class SQLServerConnection implements ISQLServerConnection, java.io.Serial
     }
 
     /**
-     * Negotiates the vector version between client and server based on the
-     * following rules:
-     * - If either client or server is "off", negotiated version is "off"
-     * - If both support v2, negotiated version is v2
-     * - If both support v1, negotiated version is v1
-     * - Otherwise, use the minimum supported version
-     *
-     * @param clientVectorSupportEnum The client's vector type support setting
-     * @param serverVersion           The server's supported vector version
-     * @return The negotiated vector version
-     */
-    private byte negotiateVectorVersion(VectorTypeSupport clientVectorSupportEnum, byte serverVersion) {
-
-        // If server doesn't support vectors, negotiation is off
-        if (serverVersion == TDS.VECTORSUPPORT_NOT_SUPPORTED) {
-            return TDS.VECTORSUPPORT_NOT_SUPPORTED;
-        }
-
-        if (clientVectorSupportEnum == VectorTypeSupport.OFF) {
-            return TDS.VECTORSUPPORT_NOT_SUPPORTED;
-        }
-
-        byte clientMaxVersion = clientVectorSupportEnum.getTdsValue();
-
-        // Negotiate using the minimum supported version
-        return (byte) Math.min(clientMaxVersion, serverVersion);
-
-    }
-
-    /**
      * Returns the negotiated vector version between client and server.
+     * Returns 0 if vectorTypeSupport is 'off' or the server did not acknowledge vector support.
+     * Valid negotiated values from the server are 1 (v1) or 2 (v2).
      *
-     * @return The negotiated vector version (0 = off, 1 = v1, 2 = v2)
+     * @return The negotiated vector version (0 = not negotiated, 1 = v1, 2 = v2)
      */
     public byte getNegotiatedVectorVersion() {
         return negotiatedVectorVersion;
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDriver.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDriver.java
@@ -547,8 +547,10 @@ enum VectorTypeSupport {
                 return VectorTypeSupport.V2;
             default:
                 throwInvalid(value);
-                return VectorTypeSupport.OFF; // Never reached
         }
+
+        // Unreachable; throwInvalid() always throws
+        return VectorTypeSupport.OFF;
     }
 
     private static void throwInvalid(String value) throws SQLServerException {
--- a/src/test/java/com/microsoft/sqlserver/jdbc/datatypes/VectorTest.java
+++ b/src/test/java/com/microsoft/sqlserver/jdbc/datatypes/VectorTest.java
@@ -1746,107 +1746,15 @@ public class VectorTest extends AbstractTest {
      * Nested test class for vector negotiation tests.
      * Tests different combinations of vectorTypeSupport settings (v1, v2, off)
      * and validates error handling for invalid values.
-     * 
-     * Also includes direct unit tests for the negotiateVectorVersion method
-     * to validate negotiation logic for all key combinations of client and server versions.
      */
     @RunWith(JUnitPlatform.class)
     @DisplayName("Vector Negotiation Tests")
     @Tag(Constants.vectorFloat16Test)
     public static class VectorNegotiationTest extends AbstractTest {
 
-        // TDS constants for vector support versions (mirroring TDS class values)
-        private static final byte VECTORSUPPORT_NOT_SUPPORTED = 0x00;
-        private static final byte VECTORSUPPORT_VERSION_1 = 0x01;
-        private static final byte VECTORSUPPORT_VERSION_2 = 0x02;
-
-        // VectorTypeSupport enum values accessed via reflection (package-private enum)
-        private static Object VECTOR_TYPE_OFF;
-        private static Object VECTOR_TYPE_V1;
-        private static Object VECTOR_TYPE_V2;
-        private static Class<?> vectorTypeSupportClass;
-
         @BeforeAll
         public static void setupNegotiationTests() throws Exception {
-
             setConnection();
-
-            // Load VectorTypeSupport enum via reflection (it's package-private)
-            vectorTypeSupportClass = Class.forName("com.microsoft.sqlserver.jdbc.VectorTypeSupport");
-            VECTOR_TYPE_OFF = Enum.valueOf((Class<Enum>) vectorTypeSupportClass, "OFF");
-            VECTOR_TYPE_V1 = Enum.valueOf((Class<Enum>) vectorTypeSupportClass, "V1");
-            VECTOR_TYPE_V2 = Enum.valueOf((Class<Enum>) vectorTypeSupportClass, "V2");
-        }
-
-        // ============================================================================
-        // Direct Unit Tests for negotiateVectorVersion method
-        // These tests use reflection to directly invoke the private method and
-        // validate all key combinations of client and server versions.
-        // ============================================================================
-
-        /**
-         * Helper method to invoke the private negotiateVectorVersion method via reflection.
-         * 
-         * @param conn The SQLServerConnection instance
-         * @param clientVectorSupport The client's VectorTypeSupport enum value (as Object)
-         * @param serverVersion The server's supported vector version byte
-         * @return The negotiated vector version
-         */
-        private byte invokeNegotiateVectorVersion(SQLServerConnection conn, 
-                Object clientVectorSupport, byte serverVersion) throws Exception {
-            java.lang.reflect.Method method = SQLServerConnection.class.getDeclaredMethod(
-                    "negotiateVectorVersion", vectorTypeSupportClass, byte.class);
-            method.setAccessible(true);
-            return (byte) method.invoke(conn, clientVectorSupport, serverVersion);
-        }
-
-        /**
-         * Test: Comprehensive matrix test for all client/server version combinations.
-         * This test validates the negotiation logic truth table covering all 9 combinations:
-         * - Client OFF vs Server (NOT_SUPPORTED, V1, V2) -> always OFF
-         * - Client V1 vs Server (NOT_SUPPORTED, V1, V2) -> OFF, V1, V1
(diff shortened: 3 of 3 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -7152,6 +7193,57 @@ private void onFeatureExtAck(byte featureId, byte[] data) throws SQLServerExcept
...
+     * - If either client or server is "off", negotiated version is "off"
+     * - If both support v2, negotiated version is v2
+     * - If both support v1, negotiated version is v1
+     * - Otherwise, use the minimum supported version
+     *
+     * @param clientVectorSupport The client's vector type support setting
+     * @param serverVersion       The server's supported vector version
+     * @return The negotiated vector version
+     */
+    private byte negotiateVectorVersion(String clientVectorSupport, byte serverVersion) {
```
machavan: We should check and validate connection string value for vectorSupport option as a string only once and then map it to an enum value and use it everywhere instead of string values.

When we send the vector FE to server as part of pre-login, we should compare the connection string value of vectorSupport and driver's current vector version, do the min of it and send that to server. This simplifies trhe negoation.

src/main/java/com/microsoft/sqlserver/jdbc/ISQLServerConnection.java
```diff
@@ -658,7 +658,10 @@ CallableStatement prepareCall(String sql, int nType, int nConcur, int nHold,
      *
      * @param vectorTypeSupport
      * A string that indicates the vector type support during connection initialization.
-     * Valid values are "off" (vector types are returned as strings) and "v1" (vectors of type FLOAT32 are returned as vectors). 
+     * Valid values are :
```
machavan, line 661: Could we map the user supplied value from connection string to an enum type that has each vector type support byte defined. Once we map this then we can use that byte value directly to send to server without needing additional switch cases.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -5899,14 +5908,28 @@ int writeUserAgentFeatureRequest(boolean write, /* if false just calculates the
...
     int writeVectorSupportFeatureRequest(boolean write,
             TDSWriter tdsWriter) throws SQLServerException {
-        if (VECTOR_SUPPORT_OFF.equalsIgnoreCase(vectorTypeSupport)) {
+
+        // Initialize vectorTypeSupportEnum if not already done.
+        if (vectorTypeSupportEnum == null) {
+            try {
+                vectorTypeSupportEnum = VectorTypeSupport.valueOfString(vectorTypeSupport);
+            } catch (SQLServerException e) {
+                // Fallback to value OFF if invalid value is provided.
```
machavan: we should just throw the error back to user instead of silently falling back

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -7118,7 +7141,20 @@ private void onFeatureExtAck(byte featureId, byte[] data) throws SQLServerExcept
                 if (0 == serverSupportedVectorVersion || serverSupportedVectorVersion > TDS.MAX_VECTORSUPPORT_VERSION) {
                     throw new SQLServerException(SQLServerException.getErrString("R_InvalidVectorVersionNumber"), null);
                 }
-                serverSupportsVector = true;
+                // Negotiate the vector version between client and server
+                negotiatedVectorVersion = negotiateVectorVersion(vectorTypeSupportEnum, serverSupportedVectorVersion);
```
machavan: The server would have already done the negotiation, do we really need to do negotiation again or is it just a  conservative/safety code ?

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/VectorTest.java
```diff
@@ -1733,21 +1733,280 @@ void validatePreparedStatementMetaData() throws SQLException {
...
+         * - Case variations (uppercase, mixed case)
+         * - Default behavior when not specified
+         */
+        @Test
+        @DisplayName("Test valid vectorTypeSupport edge cases")
+        public void testValidVectorTypeSupportEdgeCases() throws SQLException {
+            // Test cases: [connectionStringValue, description, expectedNormalizedValue]
+            // The driver trims whitespace and is case-insensitive
+            String[][] validCases = {
+                {" v1 ", "whitespace-padded v1", "v1"},
```
machavan, line 1915: trailing spaces ?

machavan (dismissed):

divang (commented):
Do we need a way to identify different statement types: Statement, PreparedStatement, and CallableStatement , so that users can analyze and compare the performance differences among them?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -7154,6 +7182,45 @@ private void onFeatureExtAck(byte featureId, byte[] data) throws SQLServerExcept
...
+     * - If either client or server is "off", negotiated version is "off"
+     * - If both support v2, negotiated version is v2
+     * - If both support v1, negotiated version is v1
+     * - Otherwise, use the minimum supported version
+     *
+     * @param clientVectorSupportEnum The client's vector type support setting
+     * @param serverVersion           The server's supported vector version
+     * @return The negotiated vector version
+     */
+    private byte negotiateVectorVersion(VectorTypeSupport clientVectorSupportEnum, byte serverVersion) {
```
divang: Is this method in use negotiateVectorVersion() ?

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/VectorTest.java
```diff
@@ -1733,21 +1733,281 @@ void validatePreparedStatementMetaData() throws SQLException {
...
+
+            // Load VectorTypeSupport enum via reflection (it's package-private)
+            vectorTypeSupportClass = Class.forName("com.microsoft.sqlserver.jdbc.VectorTypeSupport");
+            VECTOR_TYPE_OFF = Enum.valueOf((Class<Enum>) vectorTypeSupportClass, "OFF");
+            VECTOR_TYPE_V1 = Enum.valueOf((Class<Enum>) vectorTypeSupportClass, "V1");
+            VECTOR_TYPE_V2 = Enum.valueOf((Class<Enum>) vectorTypeSupportClass, "V2");
+        }
+
+        // ============================================================================
+        // Direct Unit Tests for negotiateVectorVersion method
```
divang: If negotiateVectorVersion() method not is use then we can remove the test cases for that.

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/VectorTest.java
```diff
@@ -773,8 +773,8 @@ void testInvalidVectorTypeSupportConnectionProperty() throws SQLException {
                 fail("Expected IllegalArgumentException for invalid vectorTypeSupport value, but none was thrown.");
             } catch (IllegalArgumentException e) {
                 exceptionThrown = true;
-                assertTrue(
-                        e.getMessage().contains("Incorrect connection string property for vectorTypeSupport: 1"),
+                String expectedMessage = "Invalid value for vectorTypeSupport: 1. Valid values are \"off\", \"v1\", or \"v2\".";
```
divang, line 776: Do we need to match with R_invalidVectorTypeSupport constant?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -5899,14 +5908,23 @@ int writeUserAgentFeatureRequest(boolean write, /* if false just calculates the
      */
     int writeVectorSupportFeatureRequest(boolean write,
             TDSWriter tdsWriter) throws SQLServerException {
-        if (VECTOR_SUPPORT_OFF.equalsIgnoreCase(vectorTypeSupport)) {
+
+        // Initialize vectorTypeSupportEnum if not already done.
+        if (vectorTypeSupportEnum == null) {
+            vectorTypeSupportEnum = VectorTypeSupport.valueOfString(vectorTypeSupport);
+        }
```
divang, line 5915: why do need null check? if it is already initialized by default. 
private VectorTypeSupport vectorTypeSupportEnum = VectorTypeSupport.V1

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -1136,15 +1146,14 @@ public void setVectorTypeSupport(String vectorTypeSupport) {
...
-        switch (vectorTypeSupport.trim().toLowerCase()) {
-            case VECTOR_SUPPORT_OFF:
-            case VECTOR_SUPPORT_V1:
-                this.vectorTypeSupport = vectorTypeSupport.toLowerCase();
-                break;
-            default:
-                MessageFormat form = new MessageFormat(SQLServerException.getErrString("R_invalidVectorTypeSupport"));
-                Object[] msgArgs = { vectorTypeSupport };
-                throw new IllegalArgumentException(form.format(msgArgs));
+        String normalizedValue = vectorTypeSupport.trim().toLowerCase(Locale.US);
```
Ananya2, line 1149: Consider using Locale.ROOT instead of Locale.US here, since this is protocol-level normalization and should be culture-neutral and language-independent.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -7114,11 +7132,21 @@ private void onFeatureExtAck(byte featureId, byte[] data) throws SQLServerExcept
                     throw new SQLServerException(SQLServerException.getErrString("R_unknownVectorSupportValue"), null);
                 }
 
+                // The server's FEATUREEXTACK response already contains the negotiated version
+                // (i.e. min(clientRequested, serverMax)), so we accept data[0] directly
+                // as the negotiated version — no client-side re-negotiation needed.
                 serverSupportedVectorVersion = data[0];
-                if (0 == serverSupportedVectorVersion || serverSupportedVectorVersion > TDS.MAX_VECTORSUPPORT_VERSION) {
+                negotiatedVectorVersion = data[0];
+                if (0 == negotiatedVectorVersion || negotiatedVectorVersion > TDS.MAX_VECTORSUPPORT_VERSION) {
```
Ananya2, line 7141: getNegotiatedVectorVersion() documents 0 = off, but the ACK logic treats 0 as invalid. Please align the validation with the comment, either allow 0 or update the comment.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
Ananya2: If data.length == 0, accessing data[0] will throw ArrayIndexOutOfBoundsException before the intended validation. Please validate the length first, then log data[0] (lines 7131–7133).

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
Ananya2: negotiateVectorVersion() is currently unused. Are we planning to integrate it into the ACK handling logic?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDriver.java
```diff
@@ -501,6 +501,64 @@ static DatetimeType valueOfString(String value) throws SQLServerException {
...
+        switch (value.trim().toLowerCase(Locale.US)) {
+            case "off":
+                return VectorTypeSupport.OFF;
+            case "v1":
+                return VectorTypeSupport.V1;
+            case "v2":
+                return VectorTypeSupport.V2;
+            default:
+                throwInvalid(value);
+                return VectorTypeSupport.OFF; // Never reached
```
Ananya2: remove the unreachable return VectorTypeSupport.OFF after throwInvalid()

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDriver.java
```diff
@@ -501,6 +501,64 @@ static DatetimeType valueOfString(String value) throws SQLServerException {
...
+    public String getType() {
+        return type;
+    }
+
+    static VectorTypeSupport valueOfString(String value) throws SQLServerException {
+        if (value == null) {
+            throwInvalid(value);
+        }
+
+        switch (value.trim().toLowerCase(Locale.US)) {
```
Ananya2, line 541: Consider using Locale.ROOT instead of Locale.US here.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -7154,6 +7182,45 @@ private void onFeatureExtAck(byte featureId, byte[] data) throws SQLServerExcept
...
+
+        // Negotiate using the minimum supported version
+        return (byte) Math.min(clientMaxVersion, serverVersion);
+
+    }
+
+    /**
+     * Returns the negotiated vector version between client and server.
+     *
+     * @return The negotiated vector version (0 = off, 1 = v1, 2 = v2)
```
Ananya2: getNegotiatedVectorVersion() documents 0 = off, but the ACK logic treats 0 as invalid. Please align the validation with the comment, either allow 0 or update the comment.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDriver.java
muskan124947: Moved the unreachable [return VectorTypeSupport.OFF] out of the default case to after the switch, with a comment explaining it's only there to satisfy the compiler since [throwInvalid()] always throws

PR #2872, opened 2026-01-05
Fix incorrect NULL values in TABLE_CATALOG returned by getSchemas() for built-in schemas.

### LINKED ISSUE #2863: DatabaseMetaData getSchemas returns only one "dbo" schema with a null TABLE_CATALOG
## Driver version
13.2.1.jre11

## SQL Server version
```
Microsoft SQL Server 2022 (RTM-CU17) (KB5048038) - 16.0.4175.1 (X64) 
	Dec 13 2024 09:01:53 
	Express Edition (64-bit) on Windows Server 2016 Datacenter 10.0 <X64> (Build 14393: ) (Hypervisor)
```

## Client Operating System
Microsoft Windows 11 Enterprise

## JAVA/JVM version
17

## Table schema
dbo

## Problem description
When there are multiple databases/catalogs, `connection.getMetaData().getSchemas(connection.getCatalog(), "dbo")` and `connection.getMetaData().getSchemas(null, "dbo")` return only one `dbo` with a null `TABLE_CATALOG`.

```java
String[] catalogs = new String[]{connection.getCatalog(), null};
for (String catalog: catalogs) {
	System.out.println("catalog: " + catalog + "\n");
	try (ResultSet rs = connection.getMetaData().getSchemas(catalog, "dbo")) {
		int rows = 0;
		while (rs.next()) {
			for (int col = 1; col <= rs.getMetaData().getColumnCount(); col++) {
				System.out.println(String.format("%d. %s : \"%s\"",
						col, rs.getMetaData().getColumnName(col), rs.getString(col)
				));
			}
			System.out.println();
			rows++;
		}
		System.out.println("Found " + rows + " row(s).\n");
	}
}
```

## Expected behavior
1. Schema `dbo` should have a non-empty catalog.
2. There should be a `dbo` schema for every catalog.

## Actual behavior
(When there are multiple databases/catalogs each with a `dbo` schema)

```
catalog: TestDb

1. TABLE_SCHEM : "dbo"
2. TABLE_CATALOG : "null"

Found 1 row(s).

catalog: null

1. TABLE_SCHEM : "dbo"
2. TABLE_CATALOG : "null"

Found 1 row(s).
```

Ananya2:
### Summary
Fix incorrect NULL values in TABLE_CATALOG returned by getSchemas() for built-in schemas.

### Problem
DatabaseMetaData.getSchemas() was returning NULL for the TABLE_CATALOG column when querying built-in schemas such as dbo, guest, sys, INFORMATION_SCHEMA, and database roles. 
According to the JDBC specification, TABLE_CATALOG should contain the name of the catalog (database) in which the schema resides and should only be NULL when the value is not applicable. 

### Root Cause
The issue originated from the query logic backing getSchemas(). The implementation contained a CASE statement that explicitly set TABLE_CATALOG to NULL for schemas included in an internal constSchemas list representing built-in schemas. While intended to special-case system schemas, this logic unintentionally suppressed valid catalog values. As a result, built-in schemas were returned as null.

### Solution
The fix removes the special CASE handling entirely and simplifies the catalog derivation logic. Instead of conditionally nulling the catalog name, the method now always resolves the correct catalog based on the input parameters:

- when no catalog is specified, the current database name is returned (DB_NAME())
- when a catalog is provided, that value is returned
- when an empty catalog string is supplied, the catalog is treated as not applicable and NULL is returned for cross-database scenarios

With this change, built-in schemas are treated consistently with user schemas, and the driver now accurately reflects catalog ownership as required by the JDBC specification.

Closes #2863

### DIFF
--- a/src/test/java/com/microsoft/sqlserver/jdbc/databasemetadata/DatabaseMetaDataTest.java
+++ b/src/test/java/com/microsoft/sqlserver/jdbc/databasemetadata/DatabaseMetaDataTest.java
@@ -1126,6 +1126,153 @@ public class DatabaseMetaDataTest extends AbstractTest {
     }
 
     /**
+     * Validates that when a non-null catalog is specified, TABLE_CATALOG matches the specified catalog
+     * 
+     * @throws SQLException
+     */
+    @Test
+    @Tag(Constants.xAzureSQLDW)
+    @Tag(Constants.xAzureSQLDB)
+    public void testGetSchemasWithSpecificCatalogForConstSchema() throws SQLException {
+        UUID id = UUID.randomUUID();
+        String testDb = "TestDbConst_" + id;
+
+        try (Connection connection = getConnection(); Statement stmt = connection.createStatement()) {
+            TestUtils.dropDatabaseIfExists(testDb, connectionString);
+            
+            stmt.execute(String.format("CREATE DATABASE [%s]", testDb));
+            stmt.execute(String.format("USE [%s]", testDb));
+
+            // Test with specific catalog for dbo (const schema)
+            try (ResultSet rs = connection.getMetaData().getSchemas(testDb, "dbo")) {
+                boolean foundDbo = false;
+                while (rs.next()) {
+                    String schemaName = rs.getString("TABLE_SCHEM");
+                    String catalogName = rs.getString("TABLE_CATALOG");
+                    
+                    if ("dbo".equals(schemaName)) {
+                        foundDbo = true;
+                        assertNotNull(catalogName, 
+                            "TABLE_CATALOG should not be null for const schema 'dbo' with specific catalog");
+                        assertEquals(testDb, catalogName,
+                            "TABLE_CATALOG should match specified catalog for const schema 'dbo'");
+                    }
+                }
+                assertTrue(foundDbo, "dbo schema should be found in specified catalog");
+            }
+
+        } finally {
+            TestUtils.dropDatabaseIfExists(testDb, connectionString);
+        }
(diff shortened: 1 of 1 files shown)


src/test/java/com/microsoft/sqlserver/jdbc/databasemetadata/DatabaseMetaDataTest.java
```diff
@@ -1087,6 +1087,44 @@ public void testGetSchemasWithAndWithoutCatalog() throws SQLException {
...
+        String testDb2 = "TestDb2_" + id;
+
+        try (Connection connection = getConnection(); Statement stmt = connection.createStatement()) {
+            TestUtils.dropDatabaseIfExists(testDb1, connectionString);
+            TestUtils.dropDatabaseIfExists(testDb2, connectionString);
+            
+            stmt.execute(String.format("CREATE DATABASE [%s]", testDb1));
+            stmt.execute(String.format("CREATE DATABASE [%s]", testDb2));
+
+            try (ResultSet rs = connection.getMetaData().getSchemas(null, "dbo")) {
```
machavan, line 1111: also a testcase when a non-null catalog is specified for a const schema.
Also a testcase when some other const schema other than dbo is specified.
and a test case when some valid catalog and valid non-const schema is specified

Ananya2: Added all test scenarios. Please review.

machavan (approved):

PR #2879, opened 2026-01-14
Remove outdated comment about escaping delimiter

abbrev:
Pull request #2795 removed all remaining support for regex delimiters in `SQLServerBulkCSVFileRecord`.

After digging through old versions of the code, it appears that the original support for a regex delimiter was an accidental feature. Even then it might not have worked properly—or it was broken at some point—when `escapeDelimiters` is true and a line contains a double quote. Also, a note about needing to escape regex characters in the delimiter was added to javadoc in #1711 in response to #1691, but this only documented existing (at the time) behavior.

In any case, a delimiter cannot be a regex pattern in the latest version. The delimiter is currently used as literal text.


Ananya2:
Hi @abbrev! Thanks for the contribution! We’ve merged your change as part of https://github.com/microsoft/mssql-jdbc/pull/2880. Please go ahead and close this PR.

machavan:
Closing this PR since it has been merged. Thank you @abbrev for your contribution !

PR #2885, opened 2026-01-27
Introducing Performance Logger metrics - Statement/PreparedStatement

muskan124947:
## Summary
This PR extends the existing Performance Logger work by adding **statement-level metrics** for **`Statement`** and **`PreparedStatement`** execution paths. (Callback/logging design and connection metrics were added in a previous PR.)
[Introducing Performance Logger and callback handling mechanism#2706](https://github.com/microsoft/mssql-jdbc/pull/2706)

## Statement activities tracked
Metrics are emitted via `PerformanceLogCallback` (statement overload with `connectionId` + `statementId`) and/or `com.microsoft.sqlserver.jdbc.PerformanceMetrics.Statement` logger.

- **`STATEMENT_REQUEST_BUILD`** *(timing-only)*: client-side request build time (SQL processing, parameter binding, TDS packet construction). No exception tracking.
- **`STATEMENT_FIRST_SERVER_RESPONSE`** *(timing-only)*: time from request send to first server response (`startResponse()`). No exception tracking.
- **`STATEMENT_PREPARE`**: `sp_prepare` time when `prepareMethod=prepare`. Tracks exceptions.
- **`STATEMENT_PREPEXEC`**: combined prepare+execute time via `sp_prepexec` (default reuse path). Tracks exceptions.
- **`STATEMENT_EXECUTE`**: execution time for Statement + PreparedStatement variants (`sp_executesql`, `sp_execute`, direct SQL, batch). Tracks exceptions.

## Execution mapping (high level)
- `Statement`: REQUEST_BUILD → FIRST_SERVER_RESPONSE → EXECUTE
- `PreparedStatement`: REQUEST_BUILD → FIRST_SERVER_RESPONSE → (PREPEXEC or EXECUTE) + PREPARE (when `prepareMethod=prepare`)

## Testing
| Test | Validates |
|------|-----------|
| `testPublishIsCalledAndLogFileCreated` | Callback is invoked for connection-level activities; `performance.log` file is created |
| `testStatementExecutionPerformanceLogging` | Statement and PreparedStatement execution triggers the callback |
| `testPerformanceOverhead` | Measures overhead of enabling callback+logging vs disabled (configurable iterations) |
| `testHighVolumeStatementPerformance` | High-volume single-connection test (10k+ iterations) to validate no performance regression |
| `testBatchStatementPerformanceLogging` | `PreparedStatement.executeBatch()` with parameterized inserts triggers the callback |
| `testAllStatementActivities` | End-to-end coverage of all 5 `prepareMethod` variants: Statement (direct SQL), `prepexec`, `prepare`, `none`, `scopeTempTablesToConnection` |
| `testPrepExecActivityTrackedOnlyOnSecondExecution` | Verifies default flow: 1st→EXECUTE (`sp_executesql`), 2nd→PREPEXEC (`sp_prepexec`), 3rd→EXECUTE (`sp_execute`) |
| `testPrepExecActivityTrackedOnFirstExecutionWhenEagerPrepareEnabled` | Verifies `enablePrepareOnFirstPreparedStatementCall=true` flow: 1st→PREPEXEC, 2nd→EXECUTE, 3rd→EXECUTE |

## Sample Log Output
```
ConnectionID:1, StatementID:1 Request build time, duration: 8ms
ConnectionID:1, StatementID:1 First server response, duration: 17ms
ConnectionID:1, StatementID:1 Statement execute, duration: 75ms      ← 1st call: sp_executesql
ConnectionID:1, StatementID:1 Request build time, duration: 9ms
ConnectionID:1, StatementID:1 First server response, duration: 0ms
ConnectionID:1, StatementID:1 Statement prepexec, duration: 0ms      ← 2nd call: sp_prepexec
ConnectionID:1, StatementID:1 Request build time, duration: 0ms
ConnectionID:1, StatementID:1 First server response, duration: 0ms
ConnectionID:1, StatementID:1 Statement execute, duration: 0ms       ← 3rd call: sp_execute
```


src/main/java/com/microsoft/sqlserver/jdbc/PerformanceActivity.java
```diff
@@ -6,13 +6,111 @@
...
+ * 
+ * Connection-Level Activities:
+ * - CONNECTION: Total connection establishment time
+ * - PRELOGIN: TDS prelogin negotiation time
+ * - LOGIN: Authentication/login handshake time
+ * - TOKEN_ACQUISITION: Azure AD token acquisition time
+ * 
+ * Statement-Level Activities:
+ * - STATEMENT_REQUEST_BUILD: Client-side request building time
+ * - STATEMENT_SERVER_ROUNDTRIP: Time from packet sent to first response received
```
muskan124947: Update Name : FIRST_SERVER_RESPONSE

src/main/java/com/microsoft/sqlserver/jdbc/PerformanceActivity.java
```diff
@@ -6,13 +6,111 @@
...
+     * Start: startResponse() begins - packet sent to server
+     * End: First response packet received from server
+     * 
+     * In non-adaptive (full buffering) mode this therefore reflects the time to
+     * fully buffer the entire response on the client, not just the first-byte latency.
+     * 
+     * This represents the "server-side" time from client's perspective,
+     * including network transit in both directions.
+     */
+    STATEMENT_SERVER_ROUNDTRIP("Server roundtrip time"),
```
muskan124947: Update Name : FIRST_SERVER_RESPONSE

docs/performance-metrics.md
```diff
@@ -0,0 +1,212 @@
+# Performance Metrics for Microsoft JDBC Driver for SQL Server
```
muskan124947, line 1: TODO ::

> md file location
> test with azuredb

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -1229,6 +1258,9 @@ boolean containsTemporaryTableOperations(String sql) {
     private boolean doPrepExec(TDSWriter tdsWriter, Parameter[] params, boolean hasNewTypeDefinitions,
             boolean hasExistingTypeDefinitions, TDSCommand command) throws SQLServerException {
 
+        // Reset usedPrepExec to default before determining execution path
```
machavan: The PR will need to be rebased carefully once we merge the prepareMethod=none change.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -1271,11 +1303,14 @@ private boolean doPrepExec(TDSWriter tdsWriter, Parameter[] params, boolean hasN
...
                 buildExecSQLParams(tdsWriter);
                 isExecutedAtLeastOnce = true;
+                usedPrepExec = false;
             } else if (needsPrepare) { // Second execution, use prepared statements since we seem to be re-using it.
                 if (isPrepareMethodSpPrepExec) { // If true, we're using sp_prepexec.
                     buildPrepExecParams(tdsWriter);
+                    usedPrepExec = true;
                 } else { // Otherwise, we're using sp_prepare instead of sp_prepexec.
                     isSpPrepareExecuted = true;
+                    usedPrepExec = false;
```
machavan: We need to revisit this resetting of the usedPrepExec in the context of the ongoing change for prepareMethod=none

src/test/java/com/microsoft/sqlserver/jdbc/PerformanceLogCallbackTest.java
```diff
@@ -85,9 +103,396 @@ public void publish(PerformanceActivity activity, int connectionId, int statemen
...
+    /**
+     * Test to validate statement-level performance metrics are logged when executing statements.
+     */
+    @Test
+    void testStatementExecutionPerformanceLogging() throws Exception {
+        AtomicBoolean called = new AtomicBoolean(false);
+
+        PerformanceLogCallback callbackInstance = new PerformanceLogCallback() {
+            @Override
+            public void publish(PerformanceActivity activity, int connectionId, long durationNanos, Exception exception) {
```
machavan: durationNanos is it a right name ? should it be durationMillis?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
muskan124947: For metrics:

[usePrepExec] : "does the driver intend to use sp_prepexec?" (configuration)
[usedPrepExec] : "did this specific call actually use sp_prepexec?" (runtime)

- usePrepExec | set inside Constructor | true (config says "use prepexec strategy")
- usedPrepExec | Each execution | Call 1: false, Call 2: true, Call 3: false

_usedPrepExec ? PerformanceActivity.STATEMENT_PREPEXEC : PerformanceActivity.STATEMENT_EXECUTE_
This gives accurate metrics per execution. Using usePrepExec instead would report PREPEXEC for all three calls above, even though only the second one truly used sp_prepexec.

```
// User creates a PreparedStatement with default settings (prepareMethod=prepexec)
PreparedStatement ps = conn.prepareStatement("SELECT * FROM users WHERE id = ?");

// FIRST CALL
ps.setInt(1, 100);
ps.executeQuery();  // Calls sp_executesql (NOT sp_prepexec!)
                    // Why? Driver optimizes first call assuming single use

// SECOND CALL  
ps.setInt(1, 200);
ps.executeQuery();  // NOW calls sp_prepexec (combined prepare+execute)
                    // Why? Reuse detected, worth preparing

// THIRD CALL
ps.setInt(1, 300);
ps.executeQuery();  // Calls sp_execute (just execute, already prepared)
                    // Why? Handle cached from previous call
```

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -733,9 +736,31 @@ final void doExecutePreparedStatement(PrepStmtExecCmd command) throws SQLServerE
...
-                getNextResult(true);
+                // End request build time tracking
+                endCreationToFirstPacketTracking();
+
+                // Track statement execution (PREPEXEC or EXECUTE based on execution path)
+                try (PerformanceLog.Scope executeScope = PerformanceLog.createScope(
+                        PerformanceLog.perfLoggerStatement,
+                        connection.getConnectionID(),
+                        getStatementID(),
+                        usedPrepExec ? PerformanceActivity.STATEMENT_PREPEXEC : PerformanceActivity.STATEMENT_EXECUTE)) {
```
machavan: consider renaming usedPrepExec to wasPrepExecUsed

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -3130,217 +3189,247 @@ final void doExecutePreparedStatementBatch(PrepStmtBatchExecCmd batchCommand)
...
-                for (Parameter aBatchParam : batchParam) {
-                    cryptoMetaBatch.add(aBatchParam.cryptoMeta);
-                }
-            } else {
-                // cryptoMetaBatch will be empty for non-AE connections/statements.
-                for (int i = 0; i < cryptoMetaBatch.size(); i++) {
-                    batchParam[i].cryptoMeta = cryptoMetaBatch.get(i);
-                }
-            }
+                    // Get the encryption metadata for the first batch only.
```
machavan, line 3230: looks like a tab character instead of spaces

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -3130,217 +3189,247 @@ final void doExecutePreparedStatementBatch(PrepStmtBatchExecCmd batchCommand)
...
+                                            processBatch();
+                                            continue;
+                                        }
+                                    }
 
-                            // In batch execution, we have a special update count
-                            // to indicate that no information was returned
-                            batchCommand.updateCounts[numBatchesExecuted] = (-1 == updateCount) ? Statement.SUCCESS_NO_INFO
-                                                                                                : updateCount;
+                                    // In batch execution, we have a special update count
```
machavan, line 3390: tabs

src/test/java/com/microsoft/sqlserver/jdbc/PerformanceLogCallbackTest.java
```diff
@@ -85,9 +106,485 @@ public void publish(PerformanceActivity activity, int connectionId, int statemen
...
+            }
+        }
+        return System.nanoTime() - startTime;
+    }
+
+    /**
+     * Helper method to print performance comparison between disabled and enabled logging.
+     * Uncomment the method body to see results during testing.
+     */
+    private void printPerformanceComparison(long disabledDuration, long enabledDuration, int iterations, String description) {
```
machavan, line 652: just comment the method instead

src/test/java/com/microsoft/sqlserver/jdbc/PerformanceLogCallbackTest.java
```diff
@@ -85,9 +106,485 @@ public void publish(PerformanceActivity activity, int connectionId, int statemen
...
+                    try (ResultSet rs = pstmt.executeQuery()) {
+                    }
+                }
+            }
+        }
+        long scenario3Duration = System.nanoTime() - scenario3Start;
+
+        // Scenario 4: prepareMethod=none - direct SQL execution (STATEMENT_EXECUTE)
+        long scenario4Start = System.nanoTime();
+        try (Connection con = PrepUtil.getConnection(connectionString + ";prepareMethod=none")) {
```
machavan, line 411: Can Add another one for prepareMethod=none

docs/performance-metrics.md
```diff
@@ -0,0 +1,244 @@
...
+
+---
+
+## Files Modified
+
+- `SQLServerConnection.java` - Connection-level activities
+- `SQLServerStatement.java` - Base statement activities and tracking helper methods
+- `SQLServerPreparedStatement.java` - Prepared statement activities
+- `PerformanceActivity.java` - Activity enum definitions
+- `PerformanceLog.java` - Logging infrastructure
```
divang, line 244: Why do we this comment?

divang (commented):
Do we need a way to identify different statement types: Statement, PreparedStatement, and CallableStatement , so that users can analyze and compare the performance differences among them?
 
image

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
muskan124947: used VSCode Format selection; as we added try block, so formatting was required

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
muskan124947: used VSCode Format selection; as we added try block, so formatting was required

docs/performance-metrics.md
muskan124947: this is added for copilot context

divang (approved):

machavan (approved):

JDBC-0:
OK, the log output works for log level FINE. But is there another way/API to get thse values? Because we don't want to set our log level to FINE and browse endloss log files. Instead we want to "listen" to these metrics, inspect them and check if specific value thresholds are hit to create our own log events (to specific metric systems). 
Is there such an API, Interface, Listerm Handler?

muskan124947:
Hi @JDBC-0 
FINE-level logging is not required. The driver already provides a programmatic callback API for exactly this use case.

Register a PerformanceLogCallback to receive metrics in-process — no log files involved:
(This is just a sample example)
```
    @Override
    public void publish(PerformanceActivity activity, int connectionId,
                        long durationMs, Exception exception) {
        // Connection-level metrics (CONNECTION, PRELOGIN, LOGIN, TOKEN_ACQUISITION)
        if (durationMs > 500) {
            myMetrics.record("jdbc.connection." + activity, durationMs);
        }
    }

    @Override
    public void publish(PerformanceActivity activity, int connectionId,
                        int statementId, long durationMs, Exception exception) {
        // Statement-level metrics (REQUEST_BUILD, FIRST_SERVER_RESPONSE, EXECUTE, etc.)
        if (durationMs > 200) {
            myAlertSystem.warn("Slow op: conn=" + connectionId
                + " stmt=" + statementId + " " + activity + " " + durationMs + "ms");
        }
    }
});
```
The callback fires independently of the log level — it's triggered whenever a performance scope closes, as long as a callback is registered. You get the activity type, connection/statement IDs, duration in ms, and any exception — everything you need to apply threshold checks and forward to your own metrics/alerting system.

To unregister later: SQLServerDriver.unregisterPerformanceLogCallback().

For more info, you can refer to: [Performance logger and callback](https://learn.microsoft.com/en-us/sql/connect/jdbc/performance-logger-callback?view=sql-server-ver17)

JDBC-0:
That's great to have an API, but this PerformanceActivity just has numeric statementId so the result

JDBC performance (ConId: 9, StmtId: 128) 'First server response': 8507 ms

is unreadable/unusable because the statement name is missing. How can I know what "statement 128" is?

It would be great to have the statement as a "readable" name (with parameter values, if possible).

muskan124947:
@JDBC-0 ,
The callback is invoked synchronously on the same application thread that executed the statement — not on a background thread. So the application already has full context about which SQL triggered the callback at the point it's received, and can make out what it executed to receive these callback metrics.

JDBC-0:
Hm, what IDs are thes: ConnectionId, statementId? On the application context I don't have any access when I call a JDBC statement .... only statements with a resultsset might have such metainformation somewhere hidden in the JDBC driver API. So how can you get/inspect a "statementId" that is set in the PerformanceAcrtivity? The whole idea of this is to have a single point of configuration to get this info, e.g.:


```
    static {

    	//#31980: register MS SQL JDBC driver "PerformanceLogCallback" to listen to performance metrics:
        try {
                @Override
                public void publish(PerformanceActivity activity, int connectionId, long durationMs, Exception exception) {
                    if (durationMs > 60_000L) { //longer than 1 minute?
                        logPerf(activity, connectionId, -1, durationMs, exception);
                    }
                }

                @Override
                public void publish(PerformanceActivity activity, int connectionId, int statementId, long durationMs, Exception exception) {
                    if (durationMs > 60_000L) { //longer than 1 minute?
                        logPerf(activity, connectionId, statementId, durationMs, exception);
                    }
                }

                private void logPerf(PerformanceActivity activity, int connId, int stmtId, long ms, Exception ex) {
                    final String msg = String.format("JDBC performance (ConId: %d, StmtId: %d) '%s': %d ms", connId, stmtId, activity, ms);
                    LOGGER.log(Level.WARNING, msg, ex);
                }
            });
        } catch (Throwable t) {
            LOGGER.config("Could not register MS-SQL Performance Callback: " + t.getMessage());
        }

    }//end of static
```


And this has no context to the generig threads of the applications that concurrently call misc JDBC methods. The only workaround I can think of is to samehow "get" this driver internal statementId when the statement is created, build an internal HashMap to store these and when the PerformanceLogCallback triggers, I could read this global HashMap to get information from it ... which would be possible within the driver itself, too.

I tried write such a global hashmap but I can't get the internal ID (statementId) from com.microsoft.sqlserver.jdbc.SQLServerStatement - there is no public method to read it!  (getStatementID() is not public)

PR #2886, opened 2026-01-28
 Fix: Exception chaining for nested stored procedure errors (#2115)

### LINKED ISSUE #2115: Exception chaining doesn't work in JDBC driver

Ananya2:
### Problem
When nested stored procedures raise multiple errors using RAISERROR, the JDBC driver only surfaces the first exception.
Although SQL Server sends all errors in the TDS stream, the driver stops parsing early and:
- SQLException.getNextException() returns null
- Only the first error is visible to applications
- Behavior diverges from SSMS and SQL Server semantics
Example: 
`-- P2 (inner procedure) raises error and returns`
`-- P1 (outer procedure) calls P2, then raises its own error`
`EXEC P1`
Expected: 2 exceptions
Actual: only the inner exception is visible

### Root Cause
The JDBC driver stops parsing the TDS stream early when it encounters a DONEINPROC token with the error flag set. This causes:
- Parsing to terminate before subsequent error tokens are read
- The driver to throw the first exception immediately
- Remaining errors to be left unread in the stream

Modifying the TDS parsing state machine to continue parsing was explored, but doing so breaks existing driver behavior, especially when mixed with SELECT result sets inside stored procedures.

### Solution
Instead of modifying the core TDS parsing behavior, this PR introduces lazy exception chaining by overriding getNextException() in SQLServerException to resume TDS parsing via Statement.getMoreResults().

With this approach, the driver continues parsing the TDS stream only when the application explicitly requests the next exception. When invoked, getNextException() safely resumes parsing, reads the next error token already sent by SQL Server, throws it internally, and chains it to the original exception in the correct order.

This ensures that all errors from nested stored procedures, including 3-4 or more levels, are reliably surfaced, while preserving backward compatibility and existing driver behavior.

**Testing**

- testExceptionChaining – Validates exception chaining with 2 nested stored procedures
- testExceptionChainingWith4NestedProcs – Validates chaining across 4 nested stored procedures
- testExceptionChainingWithGetMoreResults – Verifies exception retrieval using a manual getMoreResults() loop
- testExceptionChainingWith3NestedProcs– Tests exception chaining via getNextException() with 3 nested stored procedures.
- testExceptionChainingWithDifferentErrorTypes – Verifies exception chaining across 3 nested stored procedures when each procedure throws a different SQL error type (division by zero and distinct RAISERRORs), ensuring chaining works beyond RAISERROR-only scenarios.
- testExceptionChainingWithClosedStatement – Verifies that getNextException() returns null when the statement is closed after an exception is thrown, instead of throwing a “statement is closed” error that would mask the original exception.

Part1: https://github.com/microsoft/mssql-jdbc/pull/2251

Fixes #2115


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerException.java
```diff
@@ -107,6 +110,51 @@ final void setDriverErrorCode(int value) {
...
+        
+        if (sourceStatement != null) {
+            try {
+                if (sourceStatement.isClosed()) {
+                    sourceStatement = null;
+                    return null;
+                }
+                
+                sourceStatement.getMoreResults();
+                sourceStatement = null;
```
machavan: what if there are multiple exceptions to get ( more than 2) should we set sourceStatement to null here itself ? 

Also lets have a testcase ( if not already) that generates more than 2 exceptions

Ananya2: We’ve validated this scenario: the driver correctly handles more than two exceptions, and we’re able to retrieve 3–4 nested exceptions via getNextException(). We’ve added the following test cases to cover this:

- testExceptionChainingWith4NestedProcs – Validates exception chaining across 4 nested stored procedures
- testExceptionChainingWith3NestedProcs – Verifies exception chaining via getNextException() with 3 nested stored procedures
- testExceptionChainingWithDifferentErrorTypes – Confirms chaining works across different SQL error types (division by zero and distinct RAISERRORs), ensuring it’s not limited to RAISERROR-only scenarios

machavan (dismissed):

muskan124947 (dismissed):

muskan124947 (dismissed):

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerException.java
```diff
@@ -107,6 +110,58 @@ final void setDriverErrorCode(int value) {
...
+    @Override
+    public java.sql.SQLException getNextException() {
+        java.sql.SQLException nextException = super.getNextException();
+        if (nextException != null) {
+            return nextException;
+        }
+        
+        if (sourceStatement != null) {
+            try {
+                if (sourceStatement.isClosed()) {
```
divang: There is race condition when multiple threads call getNextException() concurrently.

`import java.util.concurrent.atomic.AtomicReference;

/**
 * Reference to the statement that threw this exception (for lazy exception chaining via getMoreResults).
 * Uses AtomicReference to ensure thread-safe access when multiple threads call getNextException()
 * concurrently. The getAndSet(null) operation atomically retrieves and clears the reference,
 * preventing race conditions where multiple threads could process the same statement.
 */
private transient AtomicReference<SQLServerStatement> sourceStatement = new AtomicReference<>();

final void setSourceStatement(SQLServerStatement stmt) {
    this.sourceStatement.set(stmt);
}

@Override
public java.sql.SQLException getNextException() {
    java.sql.SQLException nextException = super.getNextException();
    if (nextException != null) {
        return nextException;
    }

    // Atomically retrieve and clear the statement reference to ensure thread-safety.
    // getAndSet(null) is a single atomic operation that prevents race conditions where
    // multiple threads calling getNextException() concurrently could both obtain the
    // same statement reference and attempt to process it simultaneously.
    SQLServerStatement localStmt = sourceStatement.getAndSet(null);

    if (localStmt != null) {
        try {
            if (localStmt.isClosed()) {
                return null;
            }

            while (localStmt.getMoreResults() || localStmt.getUpdateCount() != -1) {
                // Continue processing results
            }

            return null;
        } catch (java.sql.SQLException e) {
            if (e instanceof SQLServerException) {
                ((SQLServerException) e).setSourceStatement(localStmt);
            }
            super.setNextException(e);
            return e;
        }
    }

    return null;
}`

David-Engel (dismissed):
Approved from a general functionality/API perspective.

muskan124947 (approved):

Issue #1882, opened 2022-07-29
[JDBC] Execute Stored Procedure with arguments gives error when called stored procedure resides in another DB in same SQL instance


arpanpatel11:
## Driver version
com.microsoft.sqlserver:mssql-jdbc:9.2.1.jre8

## SQL Server version
Microsoft SQL Server 2019

## Client Operating System
Window

## JAVA/JVM version
 version "11.0.3"

## Table schema
Executing stored procedures ("db2.dbo.testSP");

## Problem description
connection string - jdbc:sqlserver://dbserver;databaseName=db1
Connect to the database db1 using the JDBC driver.
Stored Procedure of another database (db2) having arguments (Present in same SQL instance).
When we trying to executed stored procedure other database using name parameters it throws an error "The database name component of the object qualifier must be the name of the current database."
The error throws from method setString(String parameterName, String value) when it try to find column index from parameter name.

findColumn method try to execute stored procedure "exec sp_sproc_columns" at that time it throws the error.

## Expected behavior
We would expect the stored procedure successfully execute from other database present in same sql instance.

## Actual behavior
it throws the error. 

## Error message/stack trace
com.microsoft.sqlserver.jdbc.SQLServerException: com.microsoft.sqlserver.jdbc.SQLServerException: The database name component of the object qualifier must be the name of the current database.
	at com.microsoft.sqlserver.jdbc.SQLServerException.makeFromDriverError(SQLServerException.java:234) 
	at com.microsoft.sqlserver.jdbc.SQLServerCallableStatement.findColumn(SQLServerCallableStatement.java:1340) 
	at com.microsoft.sqlserver.jdbc.SQLServerCallableStatement.setString(SQLServerCallableStatement.java:1937) 

## JDBC trace logs
Provide the JDBC driver trace logs.  Instructions can be found here: https://docs.microsoft.com/sql/connect/jdbc/tracing-driver-operation

tkyc:
I'll try and reproduce this on my end and investigate. In the mean time, if possible could you provide a repro app of the issue to expedite the investigation? Thanks.

arpanpatel11:
attaching stored procedure and test java program to reproduce issue.

[mssql_jdbc_sp_issue.zip](https://github.com/microsoft/mssql-jdbc/files/9231268/mssql_jdbc_sp_issue.zip)

lilgreenbird:
hi @arpanpatel11 

[...]

PR #2895, opened 2026-02-09
Fix cross-database stored procedure execution with named parameters

muskan124947:
### Description
Fixes [[JDBC] Execute Stored Procedure with arguments gives error when called stored procedure resides in another DB in same SQL instance #1882](https://github.com/microsoft/mssql-jdbc/issues/1882?reload=1?reload=1)

When calling a stored procedure from a different database context using named parameters (e.g., [callableStatement.setString("Name", "value")]), the driver internally calls sp_sproc_columns to resolve parameter metadata. Previously, sp_sproc_columns was called without the target database prefix, causing the error:
```
com.microsoft.sqlserver.jdbc.SQLServerException: The database name component of the object qualifier must be the name of the current database.
at com.microsoft.sqlserver.jdbc.SQLServerCallableStatement.findColumn(...)
at com.microsoft.sqlserver.jdbc.SQLServerCallableStatement.setString(...)
```
This PR modifies the driver to always qualify sp_sproc_columns with the database name and sys schema (e.g., DB2.sys.sp_sproc_columns), ensuring metadata queries execute in the correct database context.

Using sys schema instead of dbo prevents name-squatting security vulnerabilities:

- Users cannot create objects in the sys schema (SQL Server protects it)
- If using dbo, a malicious user could create a custom schema (e.g., SCH1) with a procedure SCH1.sp_sproc_columns that would be invoked instead of the system procedure
- The sys schema ensures we always call the real system stored procedure

### Added below test scenarios in `CallableStatementTest.java`

| Test Name | Description |
|-----------|-------------|
| `testCrossDatabaseStoredProcedureWithNamedParams` | Verifies calling a stored procedure in a different database using named parameters works correctly. |
| `testSameDatabaseStoredProcedureWithNamedParams` | Regression test ensuring same-database calls still work (with and without `@` prefix). |
| `testGitHubIssue1882` | Exact reproduction of the reported GitHub issue scenario. |
| `testDatabaseNameCaseInsensitiveComparison` | Verifies case-insensitive database name handling (e.g., `MYDB` vs `mydb`). |
| `testSysSchemaProtectsAgainstNameSquatting` | Security test verifying that user-created `sp_sproc_columns` in custom schemas cannot intercept metadata queries. |

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCallableStatement.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCallableStatement.java
@@ -1344,7 +1344,10 @@ public class SQLServerCallableStatement extends SQLServerPreparedStatement imple
                 // sys schema, preventing any possibility of name-squatting with a user-defined
                 // sp_sproc_columns procedure.
                 String currentDb = connection.getCatalog();
-                String targetDb = threePartName.getDatabasePart() != null ? threePartName.getDatabasePart() : currentDb;
+                // When using getCatalog(), we need to escape the identifier since it returns
+                // an unescaped database name. The database part from ThreePartName preserves
+                // its original format (may already be bracketed from user input).
+                String targetDb = threePartName.getDatabasePart() != null ? threePartName.getDatabasePart() : Util.escapeSQLId(currentDb);
                 metaQuery.append(targetDb);
                 metaQuery.append(".sys.");
                 


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCallableStatement.java
```diff
@@ -1337,7 +1337,23 @@ private int findColumn(String columnName) throws SQLServerException {
                 // is not a security issue.
 
                 ThreePartName threePartName = ThreePartName.parse(procedureName);
-                StringBuilder metaQuery = new StringBuilder("exec sp_sproc_columns ");
+                StringBuilder metaQuery = new StringBuilder("exec ");
+                
+                // Check if the stored procedure is in a different database than the current connection
+                String currentDb = connection.getCatalog();
+                String targetDb = threePartName.getDatabasePart();
+                boolean isCrossDatabase = targetDb != null && !targetDb.equals(currentDb);
```
machavan: Do we have to worry about case mismatch ? if yes then a test case will be good

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCallableStatement.java
```diff
@@ -1337,7 +1337,23 @@ private int findColumn(String columnName) throws SQLServerException {
...
+                
+                // When calling a stored procedure from a different database context, we need to
+                // qualify sp_sproc_columns with the target database name. Using 'sys' schema is
+                // safest because users cannot create objects in the sys schema, preventing any
+                // possibility of name-squatting with a user-defined sp_sproc_columns procedure.
+                if (isCrossDatabase) {
+                    metaQuery.append(targetDb);
```
David-Engel: Is the targetDb part of the three-part-name escaped? Do we need to worry about escaping it here? I'd guess not, if we are already relying on the user/app to have escaped it and getDatabasePart() isn't removing []s, if the user has already escaped it.

src/test/java/com/microsoft/sqlserver/jdbc/callablestatement/CallableStatementTest.java
```diff
@@ -1833,6 +1833,235 @@ public void testRegisterOutParameterWithPrecisionAndScale() throws SQLException
...
+     * This tests the fix for GitHub issue #1882 where sp_sproc_columns fails when the stored procedure
+     * is in a different database than the current connection context.
+     * 
+     * @throws SQLException
+     */
+    @Test
+    @Tag(Constants.xAzureSQLDW)
+    @Tag(Constants.xAzureSQLDB)
+    public void testCrossDatabaseStoredProcedureWithNamedParams() throws SQLException {
+        String crossDbName = "CrossDbTest_" + UUID.randomUUID().toString().replace("-", "");
```
David-Engel: We have RandomUtil.getIdentifier() to get names like this. (+ Several more UUID occurrences throughout the new test code.)

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCallableStatement.java
David-Engel: (GitHub is acting weird. Hopefully this comment isn't duplicated.)
Remember, while SQL databases themselves (stuff in them) can be case-sensitive, or case in-sensitive, database names must be unique regardless of case. (The names themselves are case insensitive.)

src/test/java/com/microsoft/sqlserver/jdbc/callablestatement/CallableStatementTest.java
muskan124947: Have updated to below
```
String shortUuid = UUID.randomUUID().toString().substring(0, 8);
        String db2 = "DB2_" + shortUuid;
```
combined length of schema + procedure could exceed SQL Server's 128-character limit in case of RandomUtil.getIdentifier()

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCallableStatement.java
muskan124947: We have below code:
```
if (null != threePartName.getDatabasePart()) {
    metaQuery.append("@procedure_qualifier=");
    metaQuery.append(threePartName.getDatabasePart());  // Used directly here too
    metaQuery.append(", ");
}
```
new code is consistent with the existing pattern - we're just reusing the same value that's already being used for @procedure_qualifier. If the user provided {call [My Database].dbo.MyProc(?,?)}, then:

- getDatabasePart() returns [My Database]
- Our query becomes [My Database].sys.sp_sproc_columns
- The @procedure_qualifier=[My Database] 

If the user didn't escape a name that needs escaping, both our code and the existing @procedure_qualifier code might fail - but that's the user's responsibility.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCallableStatement.java
muskan124947: In other places, driver uses case-sensitive comparison. (eg : SQLServerDatabaseMetaData#switchCatalogs).

machavan: we can avoid the comparison and always use the targetDb for the system proc call. 

Will need to edit 1344 to 

String targetDb = threePartName.getDatabasePart() != null ? threePartName.getDatabasePart(): currentDb

divang (approved):

PR #2899, opened 2026-02-23
[VECTOR_FLOAT16] Add support for vector float16 data type

muskan124947:
## Description
Building on the existing VECTOR(FLOAT32) support, this PR adds full end-to-end support for the **VECTOR(FLOAT16)** data type in the Microsoft JDBC Driver for SQL Server.

FLOAT16 (IEEE 754 half-precision, 16-bit) vectors use 2 bytes per dimension instead of 4, halving the storage footprint and enabling up to **3996 dimensions per vector** (vs. 1998 for FLOAT32) within the 8000-byte TDS limit. This is especially valuable for large-scale AI/ML workloads such as similarity search and embedding storage where memory efficiency matters.

FLOAT16 vector support is enabled through the existing vector feature extension **(FE identifier 0x0E) with a version 2** handshake: the client requests v2 via the vectorTypeSupport connection property, and the server acknowledges FLOAT16 capability in the feature-ack response.

Since the standard JDBC specification does not include native support for the VECTOR type, the driver uses the existing `microsoft.sql.Vector` class, now extended with the FLOAT16 dimension type:

The vector object includes number of elements, a type indicator, and the actual data, as defined below:

```
    public enum VectorDimensionType {
    FLOAT16, // 16-bit (half precision) float  ← NEW
    FLOAT32  // 32-bit (single precision) float
    }
    private VectorDimensionType vectorType; 
    private int dimensionCount; 
    private Object[] data;
```

This enhancement introduces support for a number of VECTOR data type operations, including inserts, selections, stored procedures, bulk inserts with useBulkCopyForBatchInsert, table-to-table bulk copies, CSV-to-table bulk copies, and table-valued parameters (TVP).

1. Insertion of vector data into table
```
@Test 
void insertVectorData() throws SQLException { 
        String insertSql = "INSERT INTO " + AbstractSQLGenerator.escapeIdentifier(tableName) + " (v) VALUES (?, ?)"; 
        Object[] data= new Float[] { 0.45f, 7.9f, 63.0f }; 
        Vector vector = new Vector(3, VectorDimensionType.FLOAT16, data); 
        //Vector vector = new Vector(3, 2, data); 
       //Vector nullVector = new Vector(3,2,null);
        try (PreparedStatement pstmt = connection.prepareStatement(insertSql))         { 
            pstmt.setObject(1, vector, microsoft.sql.Types.VECTOR);
            pstmt.executeUpdate(); 
        } 
}
```

2. Select vector data from table
```
@Test
    public void getVectorData() throws SQLException {
        String query = "SELECT v FROM " + AbstractSQLGenerator.escapeIdentifier(tableName);
        try (PreparedStatement stmt = connection.prepareStatement(query)) {
            try (ResultSet rs = stmt.executeQuery()) {
                assertTrue(rs.next(), "No result found for inserted vector.");
                ResultSetMetaData meta = rs.getMetaData();
                int columnCount = meta.getColumnCount();

                while (rs.next()) {
                    for (int i = 1; i <= columnCount; i++) {
                        String columnName = meta.getColumnName(i);
                        int columnType = meta.getColumnType(i); // from java.sql.Types

                        Object value = null;
                        switch (columnType) {
                            case Types.VARCHAR:
                            case Types.NVARCHAR:
                                value = rs.getString(i);
                                break;
                            case microsoft.sql.Types.VECTOR:
                                value = rs.getObject(i, microsoft.sql.Vector.class);
                        }

                        System.out.println(columnName + " = " + value + 
                        " (type: " + columnType + ")" + " , (dimensionType: " +  value.getVectorDimensionType() + ")");
                    }
                    System.out.println("---");
                }
            }
        }
    }
```
 
3. Stored Procedure Call
```
@Test 
public void testVectorStoredProcedureInputOutput() throws SQLException { 
        createProcedure(); 
        try (SQLServerCallableStatement cstmt = (SQLServerCallableStatement) connection.prepareCall(call)) { 
            Vector vector= new Vector(3, VectorDimensionType.FLOAT16, new Float[]{0.5f, 1.0f, 1.5f}); 
            cstmt.setObject(1, vector, microsoft.sql.Types.VECTOR); 
            cstmt.registerOutParameter(2, microsoft.sql.Types.VECTOR, 3, 2);
            cstmt.execute(); 
        } 
   } 
```

4. Bulk Copy from CSV to table
```
@Test
public void testBulkCopyVectorFromCSV() throws SQLException { 
        String dstTable = RandomUtil.getIdentifier("dstTable"); 
        String fileName = filePath + vectorInputCsvFile; 
                SQLServerBulkCSVFileRecord fileRecord = new SQLServerBulkCSVFileRecord(fileName, null, ",", true)) { 
            stmt.executeUpdate( 
                    "CREATE TABLE " + dstTable + " (id INT, vectorCol VECTOR(3, float16));"); 
            fileRecord.addColumnMetadata(1, "vectoxrCol", microsoft.sql.Types.VECTOR, 3, 2);
        } 
} 
```

## Backward Compatibility

- **Fully backward compatible**: Existing behavior for FLOAT32 vectors and all other data types is unchanged.
- **Opt-in via connection property**: FLOAT16 requires vectorTypeSupport=v2. The default v1 continues to support FLOAT32 only. "off" disables vector support entirely (server sends vector types as JSON string data).
- **Version validation**: If the server only supports v1, attempting to send a FLOAT16 vector now throws a clear IllegalArgumentException before any data reaches the wire, instead of failing silently or producing a cryptic server error. Similarly, if vectorTypeSupport=off, any vector type (FLOAT32 or FLOAT16) is rejected with a clear error.
- **No API break**: Changes are purely additive — [FLOAT16] added to the existing VectorDimensionType enum, existing constructors and methods continue to work unchanged.

## Validation guards

**VectorUtils.getTypeDefinition** now validates the negotiated vector version before building the type definition:

- negotiatedVersion == 0 (vector support disabled): throws IllegalArgumentException with _"Vector type is not supported by the server"_ for any vector type (FLOAT32 or FLOAT16).
- negotiatedVersion == 1 + FLOAT16 vector: throws IllegalArgumentException with _"FLOAT16 vector type is not supported by the server"_.
This prevents silent data corruption or cryptic server errors when the connection doesn't support the requested vector type.

## Testing

- **Primary validation**: Executed end-to-end against a SQL Server build with FLOAT16 vector support, using the server-team cloud build.
- **Gating**: FLOAT16 tests are tagged with vectorFloat16Test and excluded from default CI/ADO runs. Enabling the tag runs the full parameterized vector test suite for FLOAT16.
- **Version validation tests**: Both VectorFloat32Test and VectorFloat16Test include integration tests that create connections with vectorTypeSupport=off and vectorTypeSupport=v1, then verify that setObject + executeUpdate throws the expected IllegalArgumentException for unsupported vector type / version combinations.
- All tests have been verified locally using a SQL Server build with vector support.
- Code coverage will increase once **vectorFloat16Test** tag is enabled.

**`Note: Currently, tests tagged with vectorFloat16Test are excluded from ADO runs and Github checks. These will be enabled post the official build rollout.`**

## Changes made

- **Core support**: Added handling for VECTOR(FLOAT16) across read + write paths (ResultSet retrieval, parameter binding, CallableStatement usage, and other related operations).
- **Tests**: Refactored existing vector tests to run for FLOAT16 in addition to current vector element types.
- **Gating**: Introduced/used the vectorfloat16 tag to keep FLOAT16 tests disabled by default while still keeping coverage in-tree.

## Related PRs
- [[VECTOR_FLOAT16] Implement serialization and deserialization logic
#2853](https://github.com/microsoft/mssql-jdbc/pull/2853)
- [[VECTOR_FLOAT16] Implement feature extension and version negotiation for Vector v2 support
#2868](https://github.com/microsoft/mssql-jdbc/pull/2868)
- [[VECTOR_FLOAT16] Add bulk copy support for vector float16 and added tests
#2898](https://github.com/microsoft/mssql-jdbc/pull/2898)

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/VectorUtils.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/VectorUtils.java
@@ -17,9 +17,9 @@ import microsoft.sql.Vector.VectorDimensionType;
 class VectorUtils {
 
     private static final int VECTOR_HEADER_LENGTH = 8;
-    private static final int BYTES_PER_FLOAT = 4;
+    private static final int BYTES_PER_FLOAT32 = 4;
     private static final byte SCALE_BYTE_FLOAT32 = 0x00;
-    private static final int BYTES_PER_SHORT = 2;
+    private static final int BYTES_PER_FLOAT16 = 2;
     private static final byte SCALE_BYTE_FLOAT16 = 0x01;
 
     /**
@@ -111,7 +111,7 @@ class VectorUtils {
     }
 
     static int getDefaultPrecision() {
-        return BYTES_PER_FLOAT; // FLOAT32
+        return BYTES_PER_FLOAT32; // FLOAT32
     }
 
     static int getHeaderLength() {
@@ -148,11 +148,11 @@ class VectorUtils {
     static int getBytesPerDimensionFromScale(int scaleByte) {
         switch (scaleByte) {
             case SCALE_BYTE_FLOAT32:
-                return BYTES_PER_FLOAT;
+                return BYTES_PER_FLOAT32;
             case SCALE_BYTE_FLOAT16:
-                return BYTES_PER_SHORT;
+                return BYTES_PER_FLOAT16;
             default:
-                return BYTES_PER_FLOAT;
+                return BYTES_PER_FLOAT32;
         }
     }
 
@@ -163,10 +163,10 @@ class VectorUtils {
     static int getBytesPerDimensionFromScale(VectorDimensionType vectorType) {
         switch (vectorType) {
             case FLOAT16:
-                return BYTES_PER_SHORT;
+                return BYTES_PER_FLOAT16;
             case FLOAT32:
             default:
-                return BYTES_PER_FLOAT;
+                return BYTES_PER_FLOAT32;
         }
     }
 
@@ -190,9 +190,9 @@ class VectorUtils {
      */
     static byte getScaleByte(int scale) {
         switch (scale) {
-            case BYTES_PER_SHORT:
+            case BYTES_PER_FLOAT16:
                 return SCALE_BYTE_FLOAT16;
-            case BYTES_PER_FLOAT:
+            case BYTES_PER_FLOAT32:
             default:
                 return SCALE_BYTE_FLOAT32;
         }
@@ -206,11 +206,11 @@ class VectorUtils {
         int bytesPerDimension;
         switch (vector.getVectorDimensionType()) {
             case FLOAT16:
-                bytesPerDimension = BYTES_PER_SHORT;
+                bytesPerDimension = BYTES_PER_FLOAT16;
                 break;
             case FLOAT32:
             default:
-                bytesPerDimension = BYTES_PER_FLOAT;
+                bytesPerDimension = BYTES_PER_FLOAT32;
                 break;
         }
         return getHeaderLength() + bytesPerDimension * vector.getDimensionCount(); // 8-byte header + dimension payload
@@ -248,11 +248,20 @@ class VectorUtils {
 
         int precision = (vector != null) ? vector.getDimensionCount() : valueLength;
 
-        // V1: Only FLOAT32 is supported, use VECTOR(dimensionCount)
-        // V2: Explicitly specify type for both FLOAT32 and FLOAT16
-        //     VECTOR(dimensionCount, FLOAT32) or VECTOR(dimensionCount, FLOAT16)
+        // Version 0: Server does not support vector at all
+        if (negotiatedVersion <= 0) {
+            throw vectorException("R_vectorNotSupported");
+        }
+
+        // Version 1: Only FLOAT32 is supported; reject FLOAT16
+        if (negotiatedVersion == 1 && scale == BYTES_PER_FLOAT16) {
+            throw vectorException("R_float16VectorNotSupported");
+        }
+
+        // V2+: Explicitly specify type for both FLOAT32 and FLOAT16
+        //      VECTOR(dimensionCount, FLOAT32) or VECTOR(dimensionCount, FLOAT16)
         if (negotiatedVersion >= 2) {
-            if (scale == BYTES_PER_SHORT) {
+            if (scale == BYTES_PER_FLOAT16) {
                 return "VECTOR(" + precision + ", FLOAT16)";
             } else {
                 return "VECTOR(" + precision + ", FLOAT32)";
--- a/.github/instructions/glossary.instructions.md
+++ b/.github/instructions/glossary.instructions.md
@@ -607,6 +607,8 @@ Error messages defined in `SQLServerResource.java` as `R_` prefixed keys. Used w
 | `R_VectorDimensionCountMismatch` | "Mismatch between vector dimension count and provided data." |
 | `R_vectorByteArrayLength` | "Vector byte array length must be at least 8 bytes." |
 | `R_invalidVectorTypeSupport` | "Invalid value for vectorTypeSupport: {0}. Valid values are off, v1, or v2." |
+| `R_vectorNotSupported` | "Vector type is not supported by the server. Enable vector support by setting the vectorTypeSupport connection property." |
+| `R_float16VectorNotSupported` | "FLOAT16 vector type is not supported by the server. The server only supports FLOAT32 vectors (vector version 1)." |
 
 ### Connection Resiliency Errors
 
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResource.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResource.java
@@ -575,6 +575,8 @@ public final class SQLServerResource extends ListResourceBundle {
         {"R_vectorByteArrayLength", "Vector byte array length must be at least 8 bytes."},
         {"R_invalidVectorData", "The provided type of data is not supported for vector."},
         {"R_vectorByteArrayMultipleOfBytesPerDimension", "Byte array length must be a multiple of {0} for vector of type {1}."},
+        {"R_vectorNotSupported", "Vector type is not supported by the server. Enable vector support by setting the vectorTypeSupport connection property."},
+        {"R_float16VectorNotSupported", "FLOAT16 vector type is not supported by the server. The server only supports FLOAT32 vectors (vector version 1)."},
         {"R_severeError", "A severe error occurred on the current command.  The results, if any, should be discarded."},
            
     };
--- a/src/test/java/com/microsoft/sqlserver/jdbc/datatypes/vector/VectorFloat16Test.java
+++ b/src/test/java/com/microsoft/sqlserver/jdbc/datatypes/vector/VectorFloat16Test.java
@@ -10,16 +10,19 @@ import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.junit.jupiter.api.Assertions.fail;
 
+import java.sql.DriverManager;
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 
+import org.junit.jupiter.api.Assumptions;
 import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
 import org.junit.platform.runner.JUnitPlatform;
 import org.junit.runner.RunWith;
 
+import com.microsoft.sqlserver.jdbc.SQLServerConnection;
 import com.microsoft.sqlserver.jdbc.SQLServerException;
 import com.microsoft.sqlserver.testframework.AbstractSQLGenerator;
 import com.microsoft.sqlserver.testframework.Constants;
@@ -88,6 +91,93 @@ public class VectorFloat16Test extends VectorTest {
     // ============================================================================
     // FLOAT16-Specific Tests
     // ============================================================================
+
+    /**
+     * Tests that inserting a FLOAT32 vector is rejected on a v0 (off) connection.
+     * When vectorTypeSupport is "off", the server does not negotiate vector support
+     * and any vector insert should throw an IllegalArgumentException.
+     */
+    @Test
+    public void testFloat32VectorRejectedOnV0Connection() throws SQLException {
+        String connStr = connectionString + ";vectorTypeSupport=off";
+        try (SQLServerConnection v0Connection = (SQLServerConnection) DriverManager.getConnection(connStr)) {
+            Assumptions.assumeTrue(v0Connection.getNegotiatedVectorVersion() == 0,
+                    "Expected negotiated vector version 0 for 'off' connection. Skipping test.");
+
+            String insertSql = "INSERT INTO " + AbstractSQLGenerator.escapeIdentifier(tableName)
+                    + " (id, v) VALUES (?, ?)";
+            Float[] data = createTestData(1.0f, 2.0f, 3.0f);
+            Vector float32Vector = new Vector(3, VectorDimensionType.FLOAT32, data);
+
+            try (PreparedStatement pstmt = v0Connection.prepareStatement(insertSql)) {
+                pstmt.setInt(1, 1);
+                pstmt.setObject(2, float32Vector, microsoft.sql.Types.VECTOR);
+                pstmt.executeUpdate();
+                fail("Expected IllegalArgumentException when inserting FLOAT32 vector on v0 connection.");
+            } catch (IllegalArgumentException e) {
+                assertTrue(e.getMessage().contains("Vector type is not supported by the server"),
+                        "Expected vectorNotSupported error, but got: " + e.getMessage());
+            }
+        }
+    }
+
+    /**
+     * Tests that inserting a FLOAT16 vector is rejected on a v0 (off) connection.
+     * When vectorTypeSupport is "off", the server does not negotiate vector support
+     * and any vector insert should throw an IllegalArgumentException.
+     */
+    @Test
+    public void testFloat16VectorRejectedOnV0Connection() throws SQLException {
+        String connStr = connectionString + ";vectorTypeSupport=off";
+        try (SQLServerConnection v0Connection = (SQLServerConnection) DriverManager.getConnection(connStr)) {
+            Assumptions.assumeTrue(v0Connection.getNegotiatedVectorVersion() == 0,
+                    "Expected negotiated vector version 0 for 'off' connection. Skipping test.");
+
+            String insertSql = "INSERT INTO " + AbstractSQLGenerator.escapeIdentifier(tableName)
+                    + " (id, v) VALUES (?, ?)";
+            Float[] data = createTestData(1.0f, 2.0f, 3.0f);
+            Vector float16Vector = new Vector(3, VectorDimensionType.FLOAT16, data);
+
+            try (PreparedStatement pstmt = v0Connection.prepareStatement(insertSql)) {
+                pstmt.setInt(1, 1);
+                pstmt.setObject(2, float16Vector, microsoft.sql.Types.VECTOR);
+                pstmt.executeUpdate();
+                fail("Expected IllegalArgumentException when inserting FLOAT16 vector on v0 connection.");
+            } catch (IllegalArgumentException e) {
+                assertTrue(e.getMessage().contains("Vector type is not supported by the server"),
+                        "Expected vectorNotSupported error, but got: " + e.getMessage());
+            }
+        }
+    }
+
+    /**
+     * Tests that inserting a FLOAT16 vector is rejected when using a v1 connection.
+     * Even though this test class normally uses a v2 connection, this test explicitly
+     * creates a v1 connection to verify the FLOAT16 validation guard in getTypeDefinition.
+     */
+    @Test
+    public void testFloat16VectorRejectedOnV1Connection() throws SQLException {
+        String connStr = connectionString + ";vectorTypeSupport=v1";
+        try (SQLServerConnection v1Connection = (SQLServerConnection) DriverManager.getConnection(connStr)) {
+            Assumptions.assumeTrue(v1Connection.getNegotiatedVectorVersion() >= 1,
+                    "Server did not negotiate vector version 1. Skipping test.");
+
+            String insertSql = "INSERT INTO " + AbstractSQLGenerator.escapeIdentifier(tableName)
+                    + " (id, v) VALUES (?, ?)";
+            Float[] data = createTestData(1.0f, 2.0f, 3.0f);
+            Vector float16Vector = new Vector(3, VectorDimensionType.FLOAT16, data);
+
+            try (PreparedStatement pstmt = v1Connection.prepareStatement(insertSql)) {
+                pstmt.setInt(1, 1);
+                pstmt.setObject(2, float16Vector, microsoft.sql.Types.VECTOR);
+                pstmt.executeUpdate();
+                fail("Expected IllegalArgumentException when inserting FLOAT16 vector on v1 connection.");
+            } catch (IllegalArgumentException e) {
+                assertTrue(e.getMessage().contains("FLOAT16 vector type is not supported"),
+                        "Expected FLOAT16 not supported error, but got: " + e.getMessage());
+            }
+        }
+    }
 
     /**
      * Validates basic vector data insert and retrieve for FLOAT16.
--- a/src/test/java/com/microsoft/sqlserver/jdbc/datatypes/vector/VectorFloat32Test.java
+++ b/src/test/java/com/microsoft/sqlserver/jdbc/datatypes/vector/VectorFloat32Test.java
@@ -5,6 +5,7 @@
 
 package com.microsoft.sqlserver.jdbc.datatypes.vector;
 
+import java.sql.DriverManager;
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
@@ -13,7 +14,9 @@ import java.sql.Statement;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.fail;
 
+import org.junit.jupiter.api.Assumptions;
 import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
@@ -21,6 +24,7 @@ import org.junit.platform.runner.JUnitPlatform;
 import org.junit.runner.RunWith;
 
 import com.microsoft.sqlserver.jdbc.RandomUtil;
+import com.microsoft.sqlserver.jdbc.SQLServerConnection;
 import com.microsoft.sqlserver.jdbc.TestUtils;
 import com.microsoft.sqlserver.testframework.AbstractSQLGenerator;
 import com.microsoft.sqlserver.testframework.Constants;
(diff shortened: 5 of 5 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/Parameter.java
```diff
@@ -626,9 +626,24 @@ private void setTypeDefinition(DTV dtv) {
                     break;
 
                 case VECTOR:
+                    // Determine the bytes-per-dimension (scale) for the vector type definition.
+                    // For non-null vectors, derive the scale from the vector's dimension type
+                    // (e.g. FLOAT32 = 4 bytes, FLOAT16 = 2 bytes).
+                    // For null or output parameters, fall back to the registered output scale
+                    // or the default precision if neither is available.
+                    Vector vectorValue = (Vector) dtv.getSetterValue();
```
machavan: Consider moving this vector specific logic in VectorUtils

src/main/java/com/microsoft/sqlserver/jdbc/VectorUtils.java
```diff
@@ -152,8 +149,8 @@ static int getBytesPerDimensionFromScale(int scaleByte) {
         switch (scaleByte) {
             case 0:
```
machavan: We have SCALE_BYTE_FLOAT16 and SCALE_BYTE_FLOAT32 defined as const, can we use them for switch case on scale everywhere in this file ?

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/vector/AbstractVectorTest.java
```diff
@@ -0,0 +1,1472 @@
...
+ * - {@code getTypeName()} - returns the type name for display purposes
+ * - {@code getMaxDimensionCount()} - returns the maximum dimension count for the vector type
+ * - {@code getRequiredVectorTypeSupport()} - returns the required vectorTypeSupport connection property value
+ *
+ * Example concrete implementations:
+ * - {@link VectorFloat32Test} - Tests for FLOAT32 vector type (v1, scale=4, max 1998 dimensions)
+ * - {@link VectorFloat16Test} - Tests for FLOAT16 vector type (v2, scale=2, max 3996 dimensions)
+ */
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+public abstract class AbstractVectorTest extends AbstractTest {
```
machavan: should we just name it VectorTest since it is alreayd defined as abstract ?

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/vector/AbstractVectorTest.java
```diff
@@ -0,0 +1,1472 @@
...
+     */
+    protected void assertVectorDataEquals(Object[] expected, Object[] actual, String message) {
+        if (expected == null) {
+            assertNull(actual, message + " - expected null but actual is not null");
+            return;
+        }
+        assertNotNull(actual, message + " - actual is null");
+        assertEquals(expected.length, actual.length, message + " - length mismatch");
+        
+        // FLOAT16 has a 10-bit mantissa, so its rounding step grows with magnitude
```
machavan: Are we ok to assume this tolerance ?

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/vector/VectorTest.java
```diff
@@ -0,0 +1,1472 @@
...
+            pstmt.executeUpdate();
+        }
+
+        String query = "SELECT id, v1, v2 FROM " + AbstractSQLGenerator.escapeIdentifier(tableNameWithMultipleVectorColumn) + " WHERE id = ?";
+        try (PreparedStatement stmt = connection.prepareStatement(query)) {
+            stmt.setInt(1, 1);
+            try (ResultSet rs = stmt.executeQuery()) {
+                assertTrue(rs.next(), "No result found for inserted vector.");
+
+                Vector resultVector1 = rs.getObject("v1", Vector.class);
```
machavan, line 485: It would be better to have test code check the object type first and then proceed with processing, like we have documented the usage.

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/vector/AbstractVectorTest.java
```diff
@@ -0,0 +1,1472 @@
...
+    protected String tableName;
+    protected String tableNameWithMultipleVectorColumn;
+    protected String maxVectorDataTableName;
+    protected String procedureName;
+    protected String functionName;
+    protected String functionTvpName;
+    protected String vectorUdf;
+    protected String TABLE_NAME;
+    protected String TVP_NAME;
+    protected String TVP;
```
divang: ABLE_NAME, TVP_NAME, and TVP are not constants, should it be camelCase?

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/vector/AbstractVectorTest.java
```diff
@@ -0,0 +1,1437 @@
...
+        connection = (SQLServerConnection) java.sql.DriverManager.getConnection(connStr);
+        
+        // Verify the server negotiated the required version
+        byte negotiatedVersion = connection.getNegotiatedVectorVersion();
+        byte requiredVersion = "v2".equals(vectorSupport) ? (byte) 2 : (byte) 1;
+        if (negotiatedVersion < requiredVersion) {
+            System.out.println("WARNING: Server negotiated vector version " + negotiatedVersion + 
+                ", but " + getTypeName() + " tests require version " + requiredVersion + 
+                ". Some tests may be skipped.");
+        }
```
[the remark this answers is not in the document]
divang: ". Some tests may be skipped." 
How are we skipping few test? If not, then we need to fix this logging statement.

-------------
Replaced the misleading System.out.println warning with [Assumptions.assumeTrue(...)].
 Now when the server doesn't negotiate the required vector version, the entire test class will be reported as skipped by JUnit rather than failing with confusing server errors.

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/vector/VectorTest.java
muskan124947: We have test case testValidateVectorWhenVectorFEIsDisabled which does this
Last time we decided to have 2 test separately to use type check, and rest processed as it is

src/main/java/microsoft/sql/Vector.java
```diff
@@ -13,7 +13,7 @@
 public final class Vector implements java.io.Serializable {
```
divang, line 29: This is a public API class - could we add a class-level Javadoc?

/**
Represents a SQL Server {@ code VECTOR} data type value.
<p>A vector stores a fixed-length array of floating-point values,
with support for {@ link VectorDimensionType#FLOAT32 FLOAT32} (32-bit)
and {@ link VectorDimensionType#FLOAT16 FLOAT16} (16-bit half-precision) element types.</p>
<p>Example usage:</p>
{ @ code
Float[] data = new Float[] { 0.1f, 0.2f, 0.3f };
Vector vector = new Vector(3, VectorDimensionType.FLOAT32, data);
pstmt.setObject(1, vector, microsoft.sql.Types.VECTOR);
}
@ since 13.3.2
*/

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/vector/AbstractVectorTest.java
muskan124947: updated from 0.1 to 0.05%
// FLOAT16 has a 10-bit mantissa so its rounding step (1 ULP) scales with
        // magnitude. The only lossy conversion is float32 -> float16 (round-to-nearest-even);
        // the reverse (float16 -> float32) is exact. Maximum error is therefore
        // 0.5 ULP of float16 = 2^-11 ~ 0.000488 relative. We use 0.05% (0.0005)
        // as a tight bound just above that theoretical worst case.

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/vector/AbstractVectorTest.java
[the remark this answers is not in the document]
muskan124947: Replaced the misleading System.out.println warning with [Assumptions.assumeTrue(...)].
Now when the server doesn't negotiate the required vector version, the entire test class will be reported as skipped by JUnit rather than failing with confusing server errors.

src/main/java/microsoft/sql/Vector.java
```diff
@@ -121,8 +121,8 @@ private static VectorDimensionType getVectorDimensionTypeFromScaleValue(int scal
         switch (scale) {
             case 4:
                 return VectorDimensionType.FLOAT32;
-            // case 2:
-            // return VectorDimensionType.FLOAT16;
+            case 2:
+                return VectorDimensionType.FLOAT16;
```
divang, line 140: a lot many place, there are formatting issues

src/main/java/microsoft/sql/Vector.java
divang: No need to check - (data != null ?) ...  Arrays.toString is already caring the null and return as "null"
return "VECTOR(" + vectorType + ", " + dimensionCount + ") : " + Arrays.toString(data);

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/vector/AbstractVectorTest.java
muskan124947: done and made it private
Only [tableName] and [functionTvpName] are accessed by subclasses and need to stay protected.

src/main/java/microsoft/sql/Vector.java
```diff
@@ -121,8 +121,8 @@ private static VectorDimensionType getVectorDimensionTypeFromScaleValue(int scal
         switch (scale) {
             case 4:
                 return VectorDimensionType.FLOAT32;
-            // case 2:
-            // return VectorDimensionType.FLOAT16;
+            case 2:
+                return VectorDimensionType.FLOAT16;
             default:
                 return VectorDimensionType.FLOAT32;
```
divang, line 142: Can we combine case 4 and default?

case 2:
return VectorDimensionType.FLOAT16;
case 4:
default:
return VectorDimensionType.FLOAT32;

src/main/java/com/microsoft/sqlserver/jdbc/Parameter.java
```diff
@@ -626,9 +626,24 @@ private void setTypeDefinition(DTV dtv) {
...
+                    int vectorScale = (vectorValue != null)
+                            ? VectorUtils.getBytesPerDimensionFromScale(vectorValue.getVectorDimensionType())
+                            : (param.isOutput() ? param.getOutScale()
+                                    : VectorUtils.getDefaultPrecision());
```
divang: nit: more readability code 
int vectorScale;
if (vectorValue != null) {
    vectorScale = VectorUtils.getBytesPerDimensionFromScale(vectorValue.getVectorDimensionType());
} else if (param.isOutput()) {
    vectorScale = param.getOutScale();
} else {
    vectorScale = VectorUtils.getBytesPerDimensionFromScale(VectorDimensionType.FLOAT32);
}

src/main/java/microsoft/sql/Vector.java
muskan124947: formatting is correct, somehow tab spaces are reflected

src/main/java/com/microsoft/sqlserver/jdbc/Parameter.java
muskan124947: this has been updated as vector related logic is dhifted to vectorUtils

src/main/java/microsoft/sql/Vector.java
muskan124947: In future we will have more cases, so we might remove default and throw error

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/vector/VectorTest.java
```diff
@@ -0,0 +1,1473 @@
...
+
+    /**
+     * Test for inserting a vector with Double[] data type. 
+     * This validates that Double[] throws an error when FLOAT32/FLOAT16 expects Float[].
+     */
+    @Test
+    public void testInsertDoubleVectorDataThrowsError() throws SQLException {
+        Object[] originalData = new Double[] { 0.45, 7.9, 63.0 };
+        try {
+            new microsoft.sql.Vector(3, VectorDimensionType.FLOAT32, originalData);
```
divang, line 506: nit: Consider using Float[] instead of Object[] for the data parameter
The constructor accepts Object[] but validation immediately rejects anything that isn't Float[]:

Vector's validateVectorParameters() method:
if (!(data instanceof Float[])) {
    throw vectorException("R_VectorDataTypeMismatch");
}

Since both FLOAT32 and FLOAT16 vectors only accept Float[], declaring the parameter as Float[] would catch misuse at compile time rather than runtime. For example, passing new Double[]{...} compiles fine today but would be caught by the compiler if the signature were Float[] data.

If Object[] is for forward compatibility, please add a doc comment explaining that design choice.

/**
 * Constructor for Vector with dimension count and vector type.
 * .....
@ param data           The object array representing the vector data.
 *                       Must be {@code Float[]} or {@code null}.

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/vector/VectorTest.java
```diff
@@ -0,0 +1,1473 @@
...
+        }
+
+        String query = "SELECT id, v FROM " + AbstractSQLGenerator.escapeIdentifier(tableName) + " WHERE id = ?";
+        try (PreparedStatement stmt = connection.prepareStatement(query)) {
+            stmt.setInt(1, 1);
+            try (ResultSet rs = stmt.executeQuery()) {
+                assertTrue(rs.next(), "No result found for inserted vector.");
+
+                try {
+                    String vectorString = rs.getString("v");
```
divang: can it be only? String vectorString - not required
rs.getString("v");

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/vector/VectorTest.java
```diff
@@ -0,0 +1,1473 @@
...
+        try (Statement stmt = connection.createStatement()) {
+            String sql = "CREATE OR ALTER PROCEDURE " + AbstractSQLGenerator.escapeIdentifier(procedureName) + "\n" +
+                    "    @ioVector " + getColumnDefinition(3) + " OUTPUT\n" +
+                    "AS\n" +
+                    "BEGIN\n" +
+                    "    -- Simply pass the value back\n" +
+                    "    SET @ioVector = @ioVector\n" +
+                    "END";
+            stmt.execute(sql);
+        }
```
divang, line 819: why does it not in separate method like createProcedure() method or this method will handle both case?

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/vector/VectorTest.java
```diff
@@ -0,0 +1,1473 @@
...
+    @Test
+    public void testVectorStoredProcedureInOut() throws SQLException {
+        // Create a procedure with a single INOUT vector parameter
+        try (Statement stmt = connection.createStatement()) {
+            String sql = "CREATE OR ALTER PROCEDURE " + AbstractSQLGenerator.escapeIdentifier(procedureName) + "\n" +
+                    "    @ioVector " + getColumnDefinition(3) + " OUTPUT\n" +
+                    "AS\n" +
+                    "BEGIN\n" +
+                    "    -- Simply pass the value back\n" +
+                    "    SET @ioVector = @ioVector\n" +
```
divang: SET @ ioVector = @ ioVector is a no-op
this test echoes the input back unchanged, making it functionally identical to testVectorStoredProcedureInputOutput test method but with a single parameter.

This can't distinguish between - driver correctly read the server's output and driver accidentally returned the cached input. Consider having the SP assign a different vector:

SET @ ioVector = CAST('[4.0, 5.0, 6.0]' AS VECTOR(3))

Then assert the output differs from the input. That would actually exercise the INOUT round-trip, write one value in, get a different value back.

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/vector/VectorTest.java
```diff
@@ -0,0 +1,1473 @@
...
+            Vector result = cstmt.getObject(1, Vector.class);
+            assertNotNull(result, "Returned vector should not be null");
+            assertVectorDataEquals(inputVector.getData(), result.getData(), "INOUT vector data mismatch.");
+        }
+    }
+
+    /**
+     * Validates that a null vector can be passed as both input and output parameters in a stored procedure.
+     * This test creates a stored procedure that takes a VECTOR input parameter and an OUTPUT VECTOR parameter,
+     * calls the procedure with a null vector, and verifies that the output vector is also null.
```
divang: Javadoc should be like below

/**
 * Validates that a null vector can be passed as both input and output parameters in a stored procedure.
 * This test creates a stored procedure that takes a VECTOR input parameter and an OUTPUT VECTOR parameter,
 * calls the procedure with a null-data vector, and verifies that the returned Vector object is non-null
 * but contains null data (i.e., the metadata envelope is preserved while the payload is null).
 */

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/vector/VectorTest.java
```diff
@@ -0,0 +1,1473 @@
...
+        String call = "{call " + AbstractSQLGenerator.escapeIdentifier(procedureName) + "(?, ?)}";
+        try (SQLServerCallableStatement cstmt = (SQLServerCallableStatement) connection.prepareCall(call)) {
+            Vector inputVector = new Vector(3, getVectorDimensionType(), null);
+
+            cstmt.setObject(1, inputVector, microsoft.sql.Types.VECTOR);
+            cstmt.registerOutParameter(2, microsoft.sql.Types.VECTOR, 3, getScale());
+            cstmt.execute();
+
+            Vector result = cstmt.getObject(2, Vector.class);
+            assertNotNull(result, "Returned vector should not be null");
```
divang, line 858: Per the [[JDBC getObject specification]](https://learn.microsoft.com/en-us/sql/connect/jdbc/reference/getobject-method-java-lang-string?view=sql-server-ver17)

"If the value is a SQL NULL, the driver returns a Java null."

Either:
-The test should assert assertNull(result) to match the JDBC contract, or
-If wrapping SQL NULL in a Vector(dim, type, null) envelope is by design, the Javadoc should be corrected and this deviation from the JDBC spec should be documented.

muskan124947: It is expected for vector as null vector object doesn't have any relevance
so only data part will be null, rest values should be there (dimensioncount and vector type)

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/vector/VectorTest.java
```diff
@@ -0,0 +1,1473 @@
...
+    @Test
+    public void testVectorTVP() throws SQLException {
+        Vector expectedVector = new Vector(3, getVectorDimensionType(), createTestData(0.1f, 0.2f, 0.3f));
+
+        SQLServerDataTable tvp = new SQLServerDataTable();
+        tvp.addColumnMetadata("c1", microsoft.sql.Types.VECTOR);
+        tvp.addRow(expectedVector);
+
+        try (SQLServerPreparedStatement pstmt = (SQLServerPreparedStatement) connection.prepareStatement(
+                "INSERT INTO " + AbstractSQLGenerator.escapeIdentifier(tvpTableName) + " SELECT * FROM ?;")) {
```
divang: can be put in variable and use in other two places?
String tvpTableName = AbstractSQLGenerator.escapeIdentifier(tvpTableName)

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/vector/VectorTest.java
```diff
@@ -0,0 +1,1473 @@
...
+            pstmt.setStructured(1, tvpTypeName, tvp);
+            pstmt.execute();
+
+            try (Statement stmt = connection.createStatement();
+                    ResultSet rs = stmt.executeQuery(
+                            "SELECT c1 FROM " + AbstractSQLGenerator.escapeIdentifier(tvpTableName) + " ORDER BY rowId")) {
```
divang: order by is not required here?

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/vector/VectorTest.java
```diff
@@ -0,0 +1,1473 @@
...
+    // Transaction Tests
+    // ============================================================================
+
+    /**
+     * Validates that vector data is correctly rolled back in a transaction.
+     * This test creates a table with a VECTOR column, inserts data within a transaction, and then rolls back the transaction.
+     * It verifies that the data is not present after the rollback, ensuring that vector data is properly handled in transactions.
+     */
+    @Test
+    public void testTransactionRollbackForVector() throws SQLException {
```
divang, line 1275: Savepoint - rollback to a savepoint mid-transaction to verify partial rollback with vector data.

`@Test
public void testSavepointRollbackForVector() throws SQLException {
    String savepointTable = AbstractSQLGenerator.escapeIdentifier(
            RandomUtil.getIdentifier("savepointTable_" + uuid.substring(0, 8)));

    try (Statement outerStmt = connection.createStatement()) {
        TestUtils.dropTableIfExists(savepointTable, outerStmt);
        outerStmt.executeUpdate("CREATE TABLE " + savepointTable
                + " (id INT PRIMARY KEY, v " + getColumnDefinition(3) + ")");

        connection.setAutoCommit(false);
        try {
            // Insert first vector — this should survive
            Float[] data1 = createTestData(1.0f, 2.0f, 3.0f);
            Vector vector1 = new Vector(3, getVectorDimensionType(), data1);
            try (PreparedStatement pstmt = connection.prepareStatement(
                    "INSERT INTO " + savepointTable + " (id, v) VALUES (?, ?)")) {
                pstmt.setInt(1, 1);
                pstmt.setObject(2, vector1, microsoft.sql.Types.VECTOR);
                pstmt.executeUpdate();
            }

            // Create savepoint after first insert
            Savepoint sp = connection.setSavepoint("afterFirstInsert");

            // Insert second vector — this should be rolled back
            Float[] data2 = createTestData(4.0f, 5.0f, 6.0f);
            Vector vector2 = new Vector(3, getVectorDimensionType(), data2);
            try (PreparedStatement pstmt = connection.prepareStatement(
                    "INSERT INTO " + savepointTable + " (id, v) VALUES (?, ?)")) {
                pstmt.setInt(1, 2);
                pstmt.setObject(2, vector2, microsoft.sql.Types.VECTOR);
                pstmt.executeUpdate();
            }

            // Rollback to savepoint — undoes second insert, keeps first
            connection.rollback(sp);
            connection.commit();

        } catch (Exception e) {
            connection.rollback();
            throw e;
        } finally {
            connection.setAutoCommit(true);
        }

        // Verify: only row 1 exists
        try (ResultSet rs = outerStmt.executeQuery(
                "SELECT id, v FROM " + savepointTable + " ORDER BY id")) {
            assertTrue(rs.next(), "First row should exist after partial rollback.");
            assertEquals(1, rs.getInt("id"));
            Vector result = rs.getObject("v", Vector.class);
            assertNotNull(result, "Vector should not be null after commit.");
            assertVectorDataEquals(createTestData(1.0f, 2.0f, 3.0f), result.getData(),
                    "Vector data mismatch after savepoint rollback.");

            assertFalse(rs.next(), "Second row should not exist after savepoint rollback.");
        }
    } finally {
            TestUtils.dropTableIfExists(savepointTable, stmt);
        }
    }
}`

src/main/java/com/microsoft/sqlserver/jdbc/Parameter.java
```diff
@@ -626,9 +626,15 @@ private void setTypeDefinition(DTV dtv) {
                     break;
 
                 case VECTOR:
+                    // Delegate scale resolution and type definition building to VectorUtils.
```
machavan, line 629: Let us add code samples in PR description exactly like we did for FLOAT32

src/main/java/com/microsoft/sqlserver/jdbc/VectorUtils.java
```diff
@@ -19,8 +19,8 @@ class VectorUtils {
     private static final int VECTOR_HEADER_LENGTH = 8;
     private static final int BYTES_PER_FLOAT = 4;
     private static final byte SCALE_BYTE_FLOAT32 = 0x00;
-    // private static final int BYTES_PER_SHORT = 2;
-    // private static final byte SCALE_BYTE_FLOAT16 = 0x01;
+    private static final int BYTES_PER_SHORT = 2;
```
machavan: could we call these BYTES_PER_FLOAT32 and BYTES_PER_FLOAT16 for better clarity. In future we would have more types specifically integer type which could confuse this name BYERS_PER_SHORT

src/main/java/com/microsoft/sqlserver/jdbc/VectorUtils.java
```diff
@@ -229,23 +226,39 @@ static int getVectorLength(int scale, int precision) {
...
+        int scale = (vector != null)
+                ? getBytesPerDimensionFromScale(vector.getVectorDimensionType())
+                : (isOutput ? outScale : getDefaultPrecision());
+
+        int precision = (vector != null) ? vector.getDimensionCount() : valueLength;
+
+        // V1: Only FLOAT32 is supported, use VECTOR(dimensionCount)
+        // V2: Explicitly specify type for both FLOAT32 and FLOAT16
+        //     VECTOR(dimensionCount, FLOAT32) or VECTOR(dimensionCount, FLOAT16)
+        if (negotiatedVersion >= 2) {
```
machavan, line 263: We should throw error if the negotiated version is 1 and the user application is trying to send a Vector16 
We should throw error if the negotiated version is 0 and the user application is trying to send a Vector16/Vector32

PR #2907, opened 2026-03-05
Add comprehensive test suite for mixed-type vector columns (float32 and float16)

muskan124947:
## Description
Building on the existing VECTOR(FLOAT316) support, this PR expands the test suite to validate driver behavior when **FLOAT32** and **FLOAT16** vector columns coexist within the same schema. This requires `vectorTypeSupport=v2`.

The tests cover both **standard JDBC execution paths** and **bulk copy scenarios** to ensure correct functionality across mixed vector types.

## Changes
- Added a new CSV resource containing **mixed FLOAT32 and FLOAT16 vector column data** for bulk copy testing.
- Extended **FLOAT16 bulk-copy tests** to include mixed-type scenarios (FLOAT32 + FLOAT16) covering:
  - Table-to-table bulk copy
  - CSV bulk copy
  - `ISQLServerBulkData`
  - Batch insert with bulk copy
- Expanded **FLOAT16 vector tests** to support mixed-type table scenarios, including:
  - CRUD operations
  - Null handling
  - Metadata validation
  - Table-Valued Parameters (TVPs)
  - Stored procedures
  - Transactions
  - Temporary tables
  - Views
  - User-defined functions (UDFs)

## Related PRs
- [[VECTOR_FLOAT16] Implement serialization and deserialization logic
#2853](https://github.com/microsoft/mssql-jdbc/pull/2853)
- [[VECTOR_FLOAT16] Implement feature extension and version negotiation for Vector v2 support
#2868](https://github.com/microsoft/mssql-jdbc/pull/2868)
- [[VECTOR_FLOAT16] Add support for vector float16 data type
#2899](https://github.com/microsoft/mssql-jdbc/pull/2899)
- [[VECTOR_FLOAT16] Add bulk copy support for vector float16 and added tests
#2898](https://github.com/microsoft/mssql-jdbc/pull/2898)

## Performance Testing
_(Tests performed using FLOAT32 vectors with `dimensionCount = 1998`. Timings include both serialization and deserialization overhead.)_

| Number of Records | Insert operation (ms) | Insert with useBulkCopyForBatchInsert (ms) | Bulk Copy operation (ms) |
|-------------------|-----------------------|---------------------------------------------|---------------------------|
| 1                 | 50                    | 344                                         | 314                       |
| 100               | 238                   | 429                                         | 443                       |
| 1,000             | 833                   | 738                                         | 892                       |
| 10,000            | 5,800                 | 2,385                                       | 2,081                     |
| 100,000           | 48,354                | 14,098                                      | 14,881                    |
| 1,000,000         | 396,687               | 119,984                                     | 120,652                   |

### DIFF
--- a/src/test/java/com/microsoft/sqlserver/jdbc/datatypes/vector/VectorFloat16Test.java
+++ b/src/test/java/com/microsoft/sqlserver/jdbc/datatypes/vector/VectorFloat16Test.java
@@ -295,6 +295,188 @@ public class VectorFloat16Test extends VectorTest {
     }
 
     // ============================================================================
+    // Cross-Type Conversion Failure Tests (FLOAT32 <-> FLOAT16)
+    // ============================================================================
+
+    /**
+     * Tests that INSERT INTO a FLOAT32 table by SELECTing from a FLOAT16 table
+     * is rejected by the server with a conversion error.
+     *
+     * SQL Server does not allow implicit conversion between vector subtypes:
+     * "Conversion of vector from data type float16 to float32 is not allowed."
+     */
+    @Test
+    public void testInsertFloat32FromFloat16TableFails() throws SQLException {
+        String f16Table = AbstractSQLGenerator.escapeIdentifier(
+                RandomUtil.getIdentifier("F16_CrossType_" + uuid.substring(0, 8)));
+        String f32Table = AbstractSQLGenerator.escapeIdentifier(
+                RandomUtil.getIdentifier("F32_CrossType_" + uuid.substring(0, 8)));
+
+        try (Statement stmt = connection.createStatement()) {
+            stmt.executeUpdate("CREATE TABLE " + f16Table + " (f16_col VECTOR(3, float16))");
+            stmt.executeUpdate("CREATE TABLE " + f32Table + " (f32_col VECTOR(3))");
+
+            // Seed the FLOAT16 table
+            stmt.executeUpdate("INSERT INTO " + f16Table + " (f16_col) VALUES ('[0.3, 0.2, 0.1]')");
+
+            // Attempt cross-type insert: FLOAT16 -> FLOAT32 should fail
+            try {
+                stmt.executeUpdate("INSERT INTO " + f32Table + " (f32_col) SELECT f16_col FROM " + f16Table);
+                fail("Expected SQLException for cross-type vector conversion (float16 -> float32).");
+            } catch (SQLException e) {
+                assertTrue(e.getMessage().contains("Conversion of vector from data type float16 to float32 is not allowed."),
+                        "Expected vector conversion error, but got: " + e.getMessage());
+            }
+        } finally {
+            try (Statement stmt = connection.createStatement()) {
+                TestUtils.dropTableIfExists(f32Table, stmt);
+                TestUtils.dropTableIfExists(f16Table, stmt);
+            }
+        }
+    }
+
+    /**
+     * Tests that INSERT INTO a FLOAT16 table by SELECTing from a FLOAT32 table
+     * is rejected by the server with a conversion error.
+     *
+     * SQL Server does not allow implicit conversion between vector subtypes:
+     * "Conversion of vector from data type float32 to float16 is not allowed."
+     */
+    @Test
+    public void testInsertFloat16FromFloat32TableFails() throws SQLException {
+        String f16Table = AbstractSQLGenerator.escapeIdentifier(
+                RandomUtil.getIdentifier("F16_CrossType2_" + uuid.substring(0, 8)));
+        String f32Table = AbstractSQLGenerator.escapeIdentifier(
+                RandomUtil.getIdentifier("F32_CrossType2_" + uuid.substring(0, 8)));
+
+        try (Statement stmt = connection.createStatement()) {
+            stmt.executeUpdate("CREATE TABLE " + f16Table + " (f16_col VECTOR(3, float16))");
+            stmt.executeUpdate("CREATE TABLE " + f32Table + " (f32_col VECTOR(3))");
+
+            // Seed the FLOAT32 table
+            stmt.executeUpdate("INSERT INTO " + f32Table + " (f32_col) VALUES ('[0.3, 0.2, 0.1]')");
+
+            // Attempt cross-type insert: FLOAT32 -> FLOAT16 should fail
+            try {
+                stmt.executeUpdate("INSERT INTO " + f16Table + " (f16_col) SELECT f32_col FROM " + f32Table);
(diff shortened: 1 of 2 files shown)


src/test/java/com/microsoft/sqlserver/jdbc/datatypes/vector/bulkcopy/VectorFloat16BulkCopyTest.java
```diff
@@ -71,4 +112,1002 @@ protected int getMaxDimensionCount() {
...
+        // ================================================================
+        // ISQLServerBulkData
+        // ================================================================
+
+        /**
+         * Tests bulk copy via ISQLServerBulkData with two vector columns of
+         * different types (FLOAT32 + FLOAT16).
+         */
+        @Test
+        public void testMixedBulkCopyISQLServerBulkData() throws SQLException {
```
machavan, line 310: Does server support scenarios of insert select from float16 -> float32 columns and float32 -> float16 (with expected data)  ? If yes then we could have bulk copy scenarios to cover for these

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/vector/VectorFloat16Test.java
```diff
@@ -271,4 +294,1579 @@ public void testVectorWithFloat16RangeBoundary() throws SQLException {
...
+            }
+        }
+
+        // ====================================================================
+        // Insert and Retrieve Tests
+        // ====================================================================
+
+        /**
+         * Tests inserting and retrieving a row with both FLOAT32 and FLOAT16 vector columns.
+         * Validates that each column preserves its respective type and data.
```
machavan, line 606: reading a Vector object for a Float32 column in driver and using that object as is to do a setObject to insert it into a Float16 column 

reading a Vector object for a Float36 column in driver and using that object as is to do a setObject to insert it into a Float32 column 

are these valid scenarios ?

Ananya2 (approved):

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/vector/bulkcopy/VectorFloat16BulkCopyTest.java
```diff
@@ -71,4 +112,1096 @@ protected int getMaxDimensionCount() {
...
+            for (int i = 0; i < values.length; i++) {
+                arr[i] = values[i];
+            }
+            return arr;
+        }
+
+        /**
+         * Type-aware comparison: FLOAT16 columns get tolerance, FLOAT32 get exact match.
+         */
+        private void assertMixedVectorEquals(Object[] expected, Object[] actual,
```
divang, line 177: We can place the assertMixedVectorEquals method in a single location, for example in the BaseVectorTest abstract class, and have VectorBulkCopyTest extend it.

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/vector/VectorFloat16Test.java
```diff
@@ -271,4 +294,1761 @@ public void testVectorWithFloat16RangeBoundary() throws SQLException {
...
+        @AfterEach
+        public void cleanupAfterEachMixedTest() throws SQLException {
+            try (Statement stmt = mixedConnection.createStatement()) {
+                stmt.executeUpdate("DELETE FROM " + AbstractSQLGenerator.escapeIdentifier(mixedTableName));
+                stmt.executeUpdate("DELETE FROM " + AbstractSQLGenerator.escapeIdentifier(mixedTvpTableName));
+            }
+        }
+
+        // Helper to create Float[] test data
+        private Float[] data(float... values) {
```
divang, line 569: Duplicate method: Same as above comment. We can place the data method in a single location, for example in the BaseVectorTest abstract class, and have VectorBulkCopyTest extend it.

divang (commented):
Please check if there is any duplicate logic in the test cases, such as getMixedConnectionString() or other similar scenarios.

src/test/java/com/microsoft/sqlserver/jdbc/datatypes/vector/bulkcopy/VectorFloat16BulkCopyTest.java
muskan124947: VectorTest is already an abstract base class. However, VectorBulkCopyTest can't directly extend VectorTest because VectorTest contains [@Test] methods (for regular insert/retrieve, stored procedures, TVPs, metadata, etc.). 
If VectorBulkCopyTest extended VectorTest, the concrete classes like VectorFloat32BulkCopyTest and VectorFloat16BulkCopyTest would inherit and run all those non-bulk-copy tests as well, which isn't correct.

divang: we can have VectorUtilsTest kind of class which have the common methods,

divang: We can try this way:

package com.microsoft.sqlserver.jdbc.datatypes.vector;

import static org.junit.jupiter.api.Assertions.*;
import microsoft.sql.Vector.VectorDimensionType;
import com.microsoft.sqlserver.testframework.AbstractTest;

public abstract class AbstractMixedVectorTestBase extends AbstractTest {

    protected static Float[] data(float... values) {
        Float[] arr = new Float[values.length];
        for (int i = 0; i < values.length; i++) arr[i] = values[i];
        return arr;
    }

    protected static void assertMixedVectorEquals(Object[] expected, Object[] actual,
            VectorDimensionType type, String message) {
        if (expected == null) {
            assertNull(actual, message + " - expected null but actual is not null");
            return;
        }
        assertNotNull(actual, message + " - actual is null");
        assertEquals(expected.length, actual.length, message + " - length mismatch");
        boolean isFloat16 = (type == VectorDimensionType.FLOAT16);
        for (int i = 0; i < expected.length; i++) {
            float exp = (Float) expected[i], act = (Float) actual[i];
            float tol = isFloat16 ? Math.max(0.0005f * Math.abs(exp), 1e-7f) : 0.0f;
            assertEquals(exp, act, tol, message + " at index " + i);
        }
    }

    protected String getMixedConnectionString() {
        return connectionString + ";vectorTypeSupport=v2";
    }
}

// VectorTest.java
public abstract class VectorTest extends AbstractMixedVectorTestBase { ... }

// VectorBulkCopyTest.java
public abstract class VectorBulkCopyTest extends AbstractMixedVectorTestBase { ... }

PR #2909, opened 2026-03-06
Fix: Select sp_columns_170 based on vector feature negotiation instead of try/catch fallback

muskan124947:


vaclavbrodec-ataccama:
Opened #3013 for the user-visible consequence of the try/catch approach this PR set out to replace: because `getColumns()` probes `sp_columns_170` with no version or feature gate, every call against a server below SQL Server 2025 produces one failed request server-side, swallowed at `FINER`. Reproduced there on SQL Server 2019 and 2022, and bracketed against 13.2.1 which does not do it. Flagging here in case reviving this PR is the shortest path.

Issue #2902, opened 2026-02-27
[FEATURE REQUEST] JDBC URL supporting setting Transaction Isolation Level


advl03:
I have found that some tools connecting to SQL Server only allow to use the JDBC URL for configuration purposes. This means that the only way I have for configuring the connection must be done in this JDBC URL. One of these things is the Transaction Isolation Level, that in my environments, I need to set to READ UNCOMMITED. But the current driver version (13.3.1) does not allow to set the Transaction Isolation Level by using the JDBC URL.

## Describe the preferred solution
Include one new parameter in the JDBC URL which directly allows to set the Transaction Isolation Level.

## Describe alternatives you've considered
No other alternatives fulfilled my requirement.

## Reference Implementation
Implementation available in https://github.com/microsoft/mssql-jdbc/pull/2900. Successfully tested in a real environment. Even when I block de table, the app using the new JDBC URL format with transaction isolation level set to READ UNCOMMITTED can read the data.

Ananya2:
@advl03 Thanks for the proposal. The implementation approach itself looks fine, but instead of introducing a new property name like transactionIsolation, we should align with PostgreSQL JDBC, which exposes this functionality using the defaultTransactionIsolation connection property. Aligning with the PostgreSQL driver’s naming and behavior would help keep the configuration pattern consistent and familiar for users across drivers.

advl03:
Thank you very much for your reply and comments.
Could you please detail your proposal, @Ananya2 ? I'm not familiar with how PostgreSQL JDBC driver exposes this functionality, so:

- which naming should I use?
- what is the expected behavior?

Do you mean to use something like this?
`jdbc:postgresql://localhost:5432/db?options=-c%20default_transaction_isolation=serializable`
And the allowed values for that parameter:
- “read uncommitted”, “read committed”, “repeatable read”, or “serializable”. 

Is this naming what you are requesting?

Ananya2:
@advl03 Thanks for asking for clarification. 

The suggestion is to expose this as a connection property named defaultTransactionIsolation, which clearly indicates that it defines the isolation level applied when the connection is established.

**Property name**: defaultTransactionIsolation
**Accepted values (case-insensitive strings)**: READ_UNCOMMITTED, READ_COMMITTED, REPEATABLE_READ, SERIALIZABLE, and SNAPSHOT.
These would internally map to the corresponding java.sql.Connection.TRANSACTION_* constants.

[...]

PR #2900, opened 2026-02-23
Transaction isolation parameter

advl03:
### Overview

Solves issue https://github.com/microsoft/mssql-jdbc/issues/2902

This PR introduces support for the `defaultTransactionIsolation` connection property, allowing users to specify the initial transaction isolation level for a connection directly via the JDBC URL or through the `SQLServerDataSource` object. This property defines the isolation level applied immediately after the connection is established.

Using human-readable string values makes the configuration self-explanatory and aligns with modern JDBC driver practices. This change simplifies configuration, especially for environments where the connection string is the primary way to manage driver behavior (e.g., connection pools or third-party tools).

The implementation automatically executes the corresponding `SET TRANSACTION ISOLATION LEVEL` command upon connection establishment if the property is provided.

**Supported Values (Case-Insensitive Strings):**

- `READ_UNCOMMITTED`
- `READ_COMMITTED`
- `REPEATABLE_READ`
- `SERIALIZABLE`
- `SNAPSHOT`

### Example Usage
**Via JDBC URL:**

`jdbc:sqlserver://localhost;databaseName=testdb;defaultTransactionIsolation=READ_UNCOMMITTED;`

**Via SQLServerDataSource:**

```java
SQLServerDataSource ds = new SQLServerDataSource();
ds.setDefaultTransactionIsolation("READ_UNCOMMITTED");
Connection con = ds.getConnection();
```

### Testing
Unit tests have been included in:

`src/test/java/com/microsoft/sqlserver/jdbc/connection/TransactionIsolationPropertyTest.java`

to verify:
- Correct parsing of the `defaultTransactionIsolation` property from the JDBC URL.
- Validation of isolation level string values (ensuring only the 5 supported constants are accepted).
- Programmatic configuration via updated `SQLServerDataSource` getters and setters (`setDefaultTransactionIsolation` / `getDefaultTransactionIsolation`).
- Error handling for invalid or unsupported string values.

The changes have been verified to compile and pass tests using the JRE 11 profile (`mvn test -Pjre11 -Dtest=TransactionIsolationPropertyTest`).

### Instrumentation
Meaningful logging has been added to `SQLServerConnection`. When the property is set, the driver logs the isolation level being applied at the `FINE` level, facilitating debugging:

```java
logger.log(Level.FINE, "Setting default transaction isolation level from connection property: {0} ({1})", new Object[]{levelName, levelNum});
```

### Clean Diff
[...]

PR #2918, opened 2026-03-25
Transaction isolation parameter

Ananya2:
### Overview

Solves issue https://github.com/microsoft/mssql-jdbc/issues/2902
PR: https://github.com/microsoft/mssql-jdbc/pull/2900

This PR introduces support for the `defaultTransactionIsolation` connection property, allowing users to specify the initial transaction isolation level for a connection directly via the JDBC URL or through the `SQLServerDataSource` object. This property defines the isolation level applied immediately after the connection is established.

Using human-readable string values makes the configuration self-explanatory and aligns with modern JDBC driver practices. This change simplifies configuration, especially for environments where the connection string is the primary way to manage driver behavior (e.g., connection pools or third-party tools).

The implementation automatically executes the corresponding `SET TRANSACTION ISOLATION LEVEL` command upon connection establishment if the property is provided.

**Supported Values (Case-Insensitive Strings):**

- `READ_UNCOMMITTED`
- `READ_COMMITTED`
- `REPEATABLE_READ`
- `SERIALIZABLE`
- `SNAPSHOT`

### Example Usage
**Via JDBC URL:**

`jdbc:sqlserver://localhost;databaseName=testdb;defaultTransactionIsolation=READ_UNCOMMITTED;`

**Via SQLServerDataSource:**

```java
SQLServerDataSource ds = new SQLServerDataSource();
ds.setDefaultTransactionIsolation("READ_UNCOMMITTED");
Connection con = ds.getConnection();
```

### Testing
Unit tests have been included in:

`src/test/java/com/microsoft/sqlserver/jdbc/connection/TransactionIsolationPropertyTest.java`

to verify:
- Correct parsing of the `defaultTransactionIsolation` property from the JDBC URL.
- Validation of isolation level string values (ensuring only the 5 supported constants are accepted).
- Programmatic configuration via updated `SQLServerDataSource` getters and setters (`setDefaultTransactionIsolation` / `getDefaultTransactionIsolation`).
- Error handling for invalid or unsupported string values.

The changes have been verified to compile and pass tests using the JRE 11 profile (`mvn test -Pjre11 -Dtest=TransactionIsolationPropertyTest`).

### Instrumentation
Meaningful logging has been added to `SQLServerConnection`. When the property is set, the driver logs the isolation level being applied at the `FINE` level, facilitating debugging:

```java
logger.log(Level.FINE, "Setting default transaction isolation level from connection property: {0} ({1})", new Object[]{levelName, levelNum});
```
**Also fixes QUOTED_IDENTIFIER and CONCAT_NULL_YIELDS_NULL choices arrays from {OFF, OFF} to {ON, OFF} — copy-paste bug from PR #2618.**

### DIFF
--- a/src/test/java/com/microsoft/sqlserver/jdbc/connection/TransactionIsolationPropertyTest.java
+++ b/src/test/java/com/microsoft/sqlserver/jdbc/connection/TransactionIsolationPropertyTest.java
@@ -24,9 +24,11 @@ import org.junit.platform.runner.JUnitPlatform;
 import org.junit.runner.RunWith;
 
 import com.microsoft.sqlserver.jdbc.ISQLServerConnection;
+import com.microsoft.sqlserver.jdbc.RandomUtil;
 import com.microsoft.sqlserver.jdbc.SQLServerDataSource;
 import com.microsoft.sqlserver.jdbc.SQLServerDriver;
 import com.microsoft.sqlserver.jdbc.TestUtils;
+import com.microsoft.sqlserver.testframework.AbstractSQLGenerator;
 import com.microsoft.sqlserver.testframework.AbstractTest;
 import com.microsoft.sqlserver.testframework.Constants;
 import com.microsoft.sqlserver.testframework.PrepUtil;
@@ -39,10 +41,7 @@ public class TransactionIsolationPropertyTest extends AbstractTest {
         setConnection();
     }
 
-    /**
-     * Finds a {@link DriverPropertyInfo} by name (case-insensitive) from the given array.
-     * Fails the test if the property is not found.
-     */
+    /** Finds a DriverPropertyInfo by name (case-insensitive); fails if absent. */
     private static DriverPropertyInfo findProperty(DriverPropertyInfo[] props, String name) {
         for (DriverPropertyInfo prop : props) {
             if (name.equalsIgnoreCase(prop.name)) {
@@ -53,6 +52,7 @@ public class TransactionIsolationPropertyTest extends AbstractTest {
         return null;
     }
 
+    /** Verifies the property appears in DriverPropertyInfo with the correct value. */
     @Test
     public void testPropertyInfo() throws SQLException {
         SQLServerDriver driver = new SQLServerDriver();
@@ -66,6 +66,7 @@ public class TransactionIsolationPropertyTest extends AbstractTest {
         assertEquals("READ_COMMITTED", prop.value);
     }
 
+    /** Invalid value must throw R_InvalidConnectionSetting. */
     @Test
     public void testInvalidProperty() {
         String url = connectionString + ";defaultTransactionIsolation=invalid";
@@ -75,11 +76,12 @@ public class TransactionIsolationPropertyTest extends AbstractTest {
         });
 
         assertTrue(exception.getMessage().matches(TestUtils.formatErrorMsg("R_InvalidConnectionSetting")), 
-                "Exception message should match R_InvalidConnectionSetting. Actual message: " + exception.getMessage());
-        assertTrue(exception.getMessage().contains("defaultTransactionIsolation"), "Message should contain property name 'defaultTransactionIsolation'");
-        assertTrue(exception.getMessage().contains("invalid"), "Message should contain the invalid value 'invalid'");
+                "Actual message: " + exception.getMessage());
+        assertTrue(exception.getMessage().contains("defaultTransactionIsolation"));
+        assertTrue(exception.getMessage().contains("invalid"));
     }
 
+    /** Unsupported value must throw R_InvalidConnectionSetting. */
     @Test
     public void testOutOfRangeProperty() {
         String url = connectionString + ";defaultTransactionIsolation=NUMERIC_VALUE_UNSUPPORTED";
@@ -89,11 +91,12 @@ public class TransactionIsolationPropertyTest extends AbstractTest {
         });
 
         assertTrue(exception.getMessage().matches(TestUtils.formatErrorMsg("R_InvalidConnectionSetting")),
-                "Exception message should match R_InvalidConnectionSetting. Actual message: " + exception.getMessage());
-        assertTrue(exception.getMessage().contains("defaultTransactionIsolation"), "Message should contain property name 'defaultTransactionIsolation'");
-        assertTrue(exception.getMessage().contains("NUMERIC_VALUE_UNSUPPORTED"), "Message should contain the invalid value 'NUMERIC_VALUE_UNSUPPORTED'");
+                "Actual message: " + exception.getMessage());
+        assertTrue(exception.getMessage().contains("defaultTransactionIsolation"));
+        assertTrue(exception.getMessage().contains("NUMERIC_VALUE_UNSUPPORTED"));
     }
 
+    /** SNAPSHOT is a valid choice in DriverPropertyInfo. */
     @Test
     public void testSnapshotPropertyInfo() throws SQLException {
         SQLServerDriver driver = new SQLServerDriver();
@@ -104,9 +107,10 @@ public class TransactionIsolationPropertyTest extends AbstractTest {
         assertNotNull(propertyInfo);
 
         DriverPropertyInfo prop = findProperty(propertyInfo, "defaultTransactionIsolation");
-        assertEquals("SNAPSHOT", prop.value, "Transaction isolation should be SNAPSHOT");
+        assertEquals("SNAPSHOT", prop.value);
     }
 
+    /** DataSource getter/setter round-trips the value. */
     @Test
     public void testDataSourceProperty() {
         SQLServerDataSource ds = new SQLServerDataSource();
@@ -114,31 +118,30 @@ public class TransactionIsolationPropertyTest extends AbstractTest {
         assertEquals("READ_UNCOMMITTED", ds.getDefaultTransactionIsolation());
     }
 
+    /** DataSource returns null when the property is not set. */
     @Test
     public void testDataSourceDefaultProperty() {
         SQLServerDataSource ds = new SQLServerDataSource();
-        // By default, the property is null because no default value is defined in the property metadata.
-        // It falls back to driver defaults during connection establishment if not specified.
         assertNull(ds.getDefaultTransactionIsolation());
     }
 
+    /** Empty and whitespace-only values fall back to the default isolation level. */
     @Test
     public void testEdgeCasePropertyValues() throws Exception {
-        // Empty value — ignored, falls back to default.
+        // Empty value — falls back to default.
         String emptyUrl = connectionString + ";defaultTransactionIsolation=";
         try (Connection con = PrepUtil.getConnection(emptyUrl)) {
-            assertEquals(Connection.TRANSACTION_READ_COMMITTED, con.getTransactionIsolation(),
-                    "Empty property value should fall back to default isolation level.");
+            assertEquals(Connection.TRANSACTION_READ_COMMITTED, con.getTransactionIsolation());
         }
 
-        // Whitespace — trimmed to empty, also falls back to default.
+        // Whitespace — falls back to default.
         String whitespaceUrl = connectionString + ";defaultTransactionIsolation= ";
         try (Connection con = PrepUtil.getConnection(whitespaceUrl)) {
-            assertEquals(Connection.TRANSACTION_READ_COMMITTED, con.getTransactionIsolation(),
-                    "Whitespace-only property value should fall back to default isolation level.");
+            assertEquals(Connection.TRANSACTION_READ_COMMITTED, con.getTransactionIsolation());
         }
     }
 
+    /** Mixed-case input is normalized to uppercase in DriverPropertyInfo. */
     @Test
     public void testCaseInsensitiveProperty() throws SQLException {
         SQLServerDriver driver = new SQLServerDriver();
@@ -149,13 +152,10 @@ public class TransactionIsolationPropertyTest extends AbstractTest {
         assertNotNull(propertyInfo);
 
         DriverPropertyInfo prop = findProperty(propertyInfo, "defaultTransactionIsolation");
-        assertEquals("READ_COMMITTED", prop.value, "Transaction isolation should be normalized to READ_COMMITTED");
+        assertEquals("READ_COMMITTED", prop.value);
     }
 
-    /**
-     * Integration test to verify that the defaultTransactionIsolation property is correctly applied to a real connection.
-     * We iterate through all supported levels and verify con.getTransactionIsolation() matches.
-     */
+    /** All five supported levels are applied correctly to a live connection. */
     @Test
     public void testTransactionIsolationApplied() throws Exception {
         String[] levels = {"READ_UNCOMMITTED", "READ_COMMITTED", "REPEATABLE_READ", "SERIALIZABLE", "SNAPSHOT"};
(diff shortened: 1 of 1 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -5528,6 +5566,33 @@ public String getCatalog() throws SQLServerException {
...
+     *
+     * @param name the isolation level name (e.g. "READ_UNCOMMITTED", "SNAPSHOT")
+     * @return the JDBC transaction isolation level constant, or -1 if invalid
+     */
+    private static int mapTransactionIsolationName(String name) {
+        if (name == null) {
+            return -1;
+        }
+        switch (name.toUpperCase(Locale.US)) {
+            case "READ_UNCOMMITTED":
```
machavan, line 5581: check if PG uses same option values like these

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDriver.java
```diff
@@ -1115,10 +1125,13 @@ public final class SQLServerDriver implements java.sql.Driver {
                     Integer.toString(SQLServerDriverIntProperty.CONNECT_RETRY_INTERVAL.getDefaultValue()), false, null),
             new SQLServerDriverPropertyInfo(SQLServerDriverStringProperty.QUOTED_IDENTIFIER.toString(),
                     SQLServerDriverStringProperty.QUOTED_IDENTIFIER.getDefaultValue(), false,
-                    new String[] {OnOffOption.OFF.toString(), OnOffOption.OFF.toString()}),
+                    new String[] {OnOffOption.ON.toString(), OnOffOption.OFF.toString()}),
```
machavan, line 1128: why is this toggled ?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDriver.java
```diff
@@ -1115,10 +1125,13 @@ public final class SQLServerDriver implements java.sql.Driver {
...
             new SQLServerDriverPropertyInfo(SQLServerDriverStringProperty.CONCAT_NULL_YIELDS_NULL.toString(),
                     SQLServerDriverStringProperty.CONCAT_NULL_YIELDS_NULL.getDefaultValue(), false,
-                    new String[] {OnOffOption.OFF.toString(), OnOffOption.OFF.toString()}),};
+                    new String[] {OnOffOption.ON.toString(), OnOffOption.OFF.toString()}),
```
machavan, line 1131: same here

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResource.java
```diff
@@ -246,6 +246,7 @@ protected Object[][] getContents() {
         {"R_accessTokenCallbackPropertyDescription", "A SQLServerAccessTokenCallback object which is used to call a callback method to return an access token."},
         {"R_quotedIdentifierPropertyDescription", "Indicates whether quotedIdentifier property is set."},
         {"R_concatNullYieldsNullPropertyDescription", "Indicates whether concatNullYieldsNull property is set."},
+        {"R_defaultTransactionIsolationPropertyDescription", "The default transaction isolation level applied when the connection is established."},
```
machavan, line 249: R_transactionIsolationPropertyDescription  - to conform to other prop names ?

src/test/java/com/microsoft/sqlserver/jdbc/connection/TransactionIsolationPropertyTest.java
```diff
@@ -0,0 +1,190 @@
...
+    public static void setupTests() throws Exception {
+        // setConnection() is not called here to allow unit tests to run without a real database connection.
+        // Integration tests will check if connectionString is available before proceeding.
+    }
+
+    @Test
+    public void testPropertyInfo() throws SQLException {
+        SQLServerDriver driver = new SQLServerDriver();
+        Properties info = new Properties();
+        String url = "jdbc:sqlserver://localhost;defaultTransactionIsolation=READ_COMMITTED";
```
machavan: no need for URLs in the tests, tests handle them

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResource.java
Ananya2: The key R_defaultTransactionIsolationPropertyDescription follows the existing convention where keys align with the exact connection property name (e.g., columnEncryptionSetting, multiSubnetFailover, applicationName). Since the property here is defaultTransactionIsolation, keeping the current key maintains consistency.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
Ananya2: @machavan - PG uses space-separated values (e.g., READ COMMITTED) and no dedicated connection property, only the options parameter.
Here, underscore-separated values (READ_COMMITTED, etc.) follow JDBC constants and are easier for connection strings (no quoting). SNAPSHOT is SQL Server-specific and not in PG.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDriver.java
advl03: This is feedback from Copilot during its review of the first version of the PR:

> In DRIVER_PROPERTIES, the available choices for quotedIdentifier are listed as OFF twice. This prevents tools that rely on DriverPropertyInfo.choices from discovering the valid ON value. Update the choices array to include both ON and OFF (matching OnOffOption).

This is the reason why in the new commit you have the following line:
`new SQLServerDriverPropertyInfo(SQLServerDriverStringProperty.QUOTED_IDENTIFIER.toString(),
                    SQLServerDriverStringProperty.QUOTED_IDENTIFIER.getDefaultValue(), false,
                    new String[] {OnOffOption.ON.toString(), OnOffOption.OFF.toString()}),`

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDriver.java
advl03: Similar to the previous answer. The feedback from copilot said:

> In DRIVER_PROPERTIES, the available choices for concatNullYieldsNull are listed as OFF twice. This should enumerate both ON and OFF so DriverPropertyInfo.choices is accurate for UIs/introspection.

This is why, in the new commit, you have the following line:
`new SQLServerDriverPropertyInfo(SQLServerDriverStringProperty.CONCAT_NULL_YIELDS_NULL.toString(),
                    SQLServerDriverStringProperty.CONCAT_NULL_YIELDS_NULL.getDefaultValue(), false,
                    new String[] {OnOffOption.ON.toString(), OnOffOption.OFF.toString()}),`

src/test/java/com/microsoft/sqlserver/jdbc/connection/TransactionIsolationPropertyTest.java
advl03: OK. Fixed in the new commit for PR #2900 : https://github.com/microsoft/mssql-jdbc/pull/2900/changes/03c22b1f64545b90b3bbedd482fcbf5c4a236b1f

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDriver.java
Ananya2: Good catch, this was a pre-existing copy-paste bug introduced in https://github.com/microsoft/mssql-jdbc/pull/2618. Both choices were OFF instead of ON and OFF. The contributor spotted it while adding the DEFAULT_TRANSACTION_ISOLATION entry. Fixed it in the latest commit.

src/main/java/com/microsoft/sqlserver/jdbc/ISQLServerDataSource.java
```diff
@@ -1561,4 +1561,19 @@ public interface ISQLServerDataSource extends javax.sql.CommonDataSource {
...
      */
     void setConcatNullYieldsNull(String concatNullYieldsNull);
+
+    /**
+     * Sets the default transaction isolation level for the connection.
+     * 
+     * @param level
+     *        A String that contains the transaction isolation level (e.g. "READ_COMMITTED").
+     */
+    void setDefaultTransactionIsolation(String level);
```
muskan124947, line 1571: Can we rename the methods to setTransactionIsolation() and getTransactionIsolation() to maintain consistency with other properties? Most existing methods do not include “default” in their naming.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -3782,6 +3799,27 @@ private void setSessionProperties() throws SQLServerException {
                     break;
             }
         }
+
+        // Apply the defaultTransactionIsolation connection property if set.
+        // This log block helps confirm from the logs what isolation level was requested during the handshake.
+        String defaultTxnIsolationProperty = SQLServerDriverStringProperty.DEFAULT_TRANSACTION_ISOLATION.toString();
+        String defaultTxnIsolationValue = activeConnectionProperties.getProperty(defaultTxnIsolationProperty);
```
muskan124947, line 3806: mapTransactionIsolationName is called twice — once during validation and again while applying the property.
Since the value is already validated and mapped earlier, can we avoid re-mapping it again and reuse the computed transactionIsolationLevel instead? This would remove redundant work and ensure consistency.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDataSource.java
```diff
@@ -1859,6 +1859,16 @@ private Object writeReplace() throws java.io.ObjectStreamException {
         return new SerializationProxy(this);
     }
 
+    @Override
+    public void setDefaultTransactionIsolation(String level) {
+        setStringProperty(connectionProps, SQLServerDriverStringProperty.DEFAULT_TRANSACTION_ISOLATION.toString(), level);
```
muskan124947, line 1864: Same :: Can we rename the methods to setTransactionIsolation() and getTransactionIsolation() to maintain consistency with other properties? Most existing methods do not include “default” in their naming.

src/test/java/com/microsoft/sqlserver/jdbc/connection/TransactionIsolationPropertyTest.java
```diff
@@ -0,0 +1,175 @@
...
+    public void testCaseInsensitiveProperty() throws SQLException {
+        SQLServerDriver driver = new SQLServerDriver();
+        Properties info = new Properties();
+        String url = connectionString + ";defaultTransactionIsolation=read_Committed";
+
+        DriverPropertyInfo[] propertyInfo = driver.getPropertyInfo(url, info);
+        assertNotNull(propertyInfo);
+
+        boolean found = false;
+        for (DriverPropertyInfo prop : propertyInfo) {
```
muskan124947: can we - Extract a helper method to reduce duplication and improve readability.
```
private DriverPropertyInfo getProperty(DriverPropertyInfo[] props, String name) {
    for (DriverPropertyInfo prop : props) {
        if (name.equalsIgnoreCase(prop.name)) {
            return prop;
        }
    }
    return null;
}
```

```
DriverPropertyInfo prop = getProperty(propertyInfo, "defaultTransactionIsolation");
assertNotNull(prop);
assertEquals("READ_COMMITTED", prop.value);
```

src/test/java/com/microsoft/sqlserver/jdbc/connection/TransactionIsolationPropertyTest.java
```diff
@@ -0,0 +1,175 @@
...
+        ds.setDefaultTransactionIsolation("READ_UNCOMMITTED");
+        assertEquals("READ_UNCOMMITTED", ds.getDefaultTransactionIsolation());
+    }
+
+    @Test
+    public void testDataSourceDefaultProperty() {
+        com.microsoft.sqlserver.jdbc.SQLServerDataSource ds = new com.microsoft.sqlserver.jdbc.SQLServerDataSource();
+        // By default, the property is null because no default value is defined in the property metadata.
+        // It falls back to driver defaults during connection establishment if not specified.
+        assertTrue(ds.getDefaultTransactionIsolation() == null);
```
muskan124947: assertNull(ds.getDefaultTransactionIsolation());

src/test/java/com/microsoft/sqlserver/jdbc/connection/TransactionIsolationPropertyTest.java
```diff
@@ -0,0 +1,175 @@
...
+
+    @Test
+    public void testOutOfRangeProperty() {
+        String url = connectionString + ";defaultTransactionIsolation=NUMERIC_VALUE_UNSUPPORTED";
+
+        SQLException exception = assertThrows(SQLException.class, () -> {
+            PrepUtil.getConnection(url);
+        });
+
+        assertTrue(exception.getMessage().matches(TestUtils.formatErrorMsg("R_InvalidConnectionSetting")),
```
muskan124947: Instead of having 3 asserts we can match with exact error
assertTrue(exception.getMessage().contains("R_InvalidConnectionSetting"));

src/test/java/com/microsoft/sqlserver/jdbc/connection/TransactionIsolationPropertyTest.java
```diff
@@ -0,0 +1,175 @@
...
+
+    @Test
+    public void testInvalidProperty() {
+        String url = connectionString + ";defaultTransactionIsolation=invalid";
+
+        SQLException exception = assertThrows(SQLException.class, () -> {
+            PrepUtil.getConnection(url);
+        });
+
+        assertTrue(exception.getMessage().matches(TestUtils.formatErrorMsg("R_InvalidConnectionSetting")),
```
muskan124947: Instead of having 3 asserts we can match with exact error
assertTrue(exception.getMessage().contains("R_InvalidConnectionSetting"));

src/test/java/com/microsoft/sqlserver/jdbc/connection/TransactionIsolationPropertyTest.java
```diff
@@ -0,0 +1,175 @@
...
+                found = true;
+                break;
+            }
+        }
+        assertTrue(found, "Property 'defaultTransactionIsolation' not found in DriverPropertyInfo");
+    }
+
+    @Test
+    public void testDataSourceProperty() {
+        com.microsoft.sqlserver.jdbc.SQLServerDataSource ds = new com.microsoft.sqlserver.jdbc.SQLServerDataSource();
```
muskan124947: multiple places fully classified class name is used - can we just import and use for readability

src/test/java/com/microsoft/sqlserver/jdbc/connection/TransactionIsolationPropertyTest.java
```diff
@@ -0,0 +1,175 @@
...
+
+import com.microsoft.sqlserver.jdbc.ISQLServerConnection;
+import com.microsoft.sqlserver.jdbc.SQLServerDriver;
+import com.microsoft.sqlserver.jdbc.TestUtils;
+import com.microsoft.sqlserver.testframework.AbstractTest;
+import com.microsoft.sqlserver.testframework.Constants;
+import com.microsoft.sqlserver.testframework.PrepUtil;
+
+@RunWith(JUnitPlatform.class)
+public class TransactionIsolationPropertyTest extends AbstractTest {
```
muskan124947: Adding edge cases like empty values or whitespace would further strengthen coverage.

src/main/java/com/microsoft/sqlserver/jdbc/ISQLServerDataSource.java
Ananya2: We can’t rename these: setTransactionIsolation() / getTransactionIsolation() are already defined in java.sql.Connection for live connection state (int-based). Reusing them here would be misleading.

src/test/java/com/microsoft/sqlserver/jdbc/connection/TransactionIsolationPropertyTest.java
Ananya2: R_InvalidConnectionSetting is a resource key, not the actual message, so contains() won’t work.
TestUtils.formatErrorMsg() correctly matches the resolved message, and the extra contains() checks validate parameters.

src/test/java/com/microsoft/sqlserver/jdbc/connection/TransactionIsolationPropertyTest.java
```diff
@@ -0,0 +1,188 @@
...
+            try (Connection con = PrepUtil.getConnection(url)) {
+                assertEquals(expectedConstants[i], con.getTransactionIsolation(),
+                        "Isolation level " + levels[i] + " was not applied correctly.");
+            }
+        }
+    }
+
+    /**
+     * Integration test to verify case-insensitivity when applying to a real connection.
+     */
```
machavan: test that sets defaultTransactionIsolation to say SERIALIZABLE on connection string ... but explicitly overrides it to say READ_UNCOMMITTED using a statement   and then the test verifies this isolation level semantically.

A vice versa of this or some other combination with the test actual validating the effect of the overriden isolation level

Ananya2: done

PR #2935, opened 2026-04-13
Add Enhanced Routing support for Hyperscale reader endpoints (TDS FEATUREEXT 0x0F / ENVCHANGE  type 21)

Ananya2:
## Overview

Hyperscale is introducing reader endpoints that load balance connections across named replicas. Unlike standard gateway routing (ENVCHANGE type 20 0x14) which only carries server and port, reader endpoints need to route connections to both a target network location **and a specific database** — since the client doesn't know the database ahead of time.

This PR implements the full TDS protocol changes required for the JDBC driver to support this scenario.

## Changes

**Feature negotiation:**
The driver now always sends `FEATUREEXT 0x0F` (EnhancedRouting) in the Login7 feature extension block, regardless of the target server. When the server acknowledges with a 1-byte payload (`data[0]=1`), the driver enables enhanced routing support for the session. This ensures backward compatibility — older servers that don't recognize the token simply ignore it.

**ENVCHANGE 0x21 parsing:**
A new ENVCHANGE type 21 (EnhancedRouting) is now parsed alongside the existing type 20 (Routing). The enhanced payload includes all standard routing fields plus a required `AlternateDatabase` field (≤128 characters). The driver validates port number and database name, and updates `hostNameInCertificate` for the routed server domain.

**Routing guard in login():**
When the server sends ENVCHANGE type 21 0X15 with a database name but did **not** acknowledge the enhanced routing feature, the driver discards the routing info and stays on the original server. This is the critical safety mechanism that prevents an unverified server from redirecting the client to an arbitrary database.

**Login7 database name override:**
On the reconnection to the routed target, the driver uses the database name from the routing info (not the connection string) in the Login7 packet. This ensures the routed replica receives the correct database context.

**ServerPortPlaceHolder extension:**
The routing placeholder now carries a `databaseName` field, enabling the routing info to flow through the existing connection lifecycle without structural changes.

## Test coverage
13 tests organized into 3 groups, using crafted TDS packets injected via mocked `TDSReader`/`TDSPacket`:

**Routing behavior**:
- Enhanced routing enabled — server acks feature + sends ENVCHANGE 21 → routing info carries routed server, port, and database
- No feature ack — login() guard discards routing, original database preserved
- Feature ack disabled — login() guard discards routing, original database preserved
- Legacy ENVCHANGE 20 — backward-compatible routing with server/port only

**Feature negotiation**:
- Enabled — ack `[1]` sets flag true
- Disabled — ack `[0]` leaves flag false
- DoNotAcknowledge — no ack, flag stays false
- Invalid ack data — empty/extra bytes throw with exact error message
- Feature ack filter — non-DNS/non-ER acks silently skipped during routing

**ENVCHANGE 21 validation**:
- Port 0 → throws
- Empty database name → throws
- `hostNameInCertificate` updated for routed server domain

### Manual testing

Validated end-to-end against an Azure Hyperscale database on a OneBox server with 3 named replicas behind a reader endpoint:
- 15 consecutive connections successfully load balanced across all 3 replica servers
- Verified FEATUREEXT 0x0F negotiation and ENVCHANGE type 21 0x15 parsing in TDS traffic
- Confirmed routed database name matches `DB_NAME()` on each target replica

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
@@ -3937,6 +3937,9 @@ public class SQLServerConnection implements ISQLServerConnection, java.io.Serial
                 while (true) {
                     clientConnectionId = null;
                     state = State.INITIALIZED;
+                    // Reset enhanced routing state for each connection attempt so a stale
+                    // ack from a prior failed attempt does not carry over to a different server.
+                    serverSupportsEnhancedRouting = false;
 
                     try {
                         if (isDBMirroring && useFailoverHost) {
@@ -4060,8 +4063,6 @@ public class SQLServerConnection implements ISQLServerConnection, java.io.Serial
                             } else {
                                 // set isRoutedInCurrentAttempt to false for the next attempt
                                 isRoutedInCurrentAttempt = false;
-                                // Reset so the routed server must re-negotiate enhanced routing
-                                serverSupportsEnhancedRouting = false;
 
                                 continue;
                             }
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
@@ -1343,6 +1343,13 @@ public class SQLServerConnection implements ISQLServerConnection, java.io.Serial
         return serverSupportsJSON;
     }
 
+    /** whether server supports Enhanced Routing */
+    private boolean serverSupportsEnhancedRouting = false;
+
+    boolean getServerSupportsEnhancedRouting() {
+        return serverSupportsEnhancedRouting;
+    }
+
     /** Boolean that indicates whether LOB objects created by this connection should be loaded into memory */
     private boolean delayLoadingLobs = SQLServerDriverBooleanProperty.DELAY_LOADING_LOBS.getDefaultValue();
 
@@ -6025,6 +6032,24 @@ public class SQLServerConnection implements ISQLServerConnection, java.io.Serial
         return len;
     }
 
+    /**
+     * Writes the Enhanced Routing feature request.
+     * 
+     * @param write true to write the request, false to just calculate the length
+     * @param tdsWriter the TDS writer
+     * @return The length of the feature request in bytes
+     * @throws SQLServerException
+     */
+    int writeEnhancedRoutingFeatureRequest(boolean write, /* if false just calculates the length */
+            TDSWriter tdsWriter) throws SQLServerException {
+        int len = 5; // 1byte = featureID, 4bytes = featureData length (0 bytes, no payload)
+        if (write) {
+            tdsWriter.writeByte(TDS.TDS_FEATURE_EXT_ENHANCEDROUTING);
+            tdsWriter.writeInt(0); // No payload for enhanced routing
+        }
+        return len;
+    }
+
     int writeIdleConnectionResiliencyRequest(boolean write, TDSWriter tdsWriter) throws SQLServerException {
         SessionStateTable ssTable = sessionRecovery.getSessionStateTable();
         int len = 1;
@@ -6255,6 +6280,7 @@ public class SQLServerConnection implements ISQLServerConnection, java.io.Serial
     private static final int ENVCHANGE_RESET_COMPLETE = 18;
     private static final int ENVCHANGE_USER_INFO = 19;
     private static final int ENVCHANGE_ROUTING = 20;
+    private static final int ENVCHANGE_ENHANCED_ROUTING = 21;
 
     final void processEnvChange(TDSReader tdsReader) throws SQLServerException {
         tdsReader.readUnsignedByte(); // token type
@@ -6369,18 +6395,25 @@ public class SQLServerConnection implements ISQLServerConnection, java.io.Serial
                     connectionlogger.finer(toString() + " Ignored env change: " + envchange);
                 break;
             case ENVCHANGE_ROUTING:
-
-                // initialize to invalid values
-                int routingDataValueLength, routingProtocol, routingPortNumber, routingServerNameLength;
+            case ENVCHANGE_ENHANCED_ROUTING:
+                // Enhanced routing should only be processed if the feature was successfully negotiated
+                boolean isEnhancedRouting = (envchange == ENVCHANGE_ENHANCED_ROUTING);
+                if (isEnhancedRouting && !serverSupportsEnhancedRouting) {
+                    if (connectionlogger.isLoggable(Level.WARNING)) {
+                        connectionlogger.warning(toString() + " Received enhanced routing ENVCHANGE but feature was not negotiated");
+                    }
+                    throwInvalidTDS();
+                }
+                int routingDataValueLength, routingProtocol, routingPortNumber, routingServerNameLength, routingDatabaseNameLength = -1;
                 routingDataValueLength = routingProtocol = routingPortNumber = routingServerNameLength = -1;
 
                 String routingServerName = null;
+                String routingDatabaseName = null;
 
                 try {
                     routingDataValueLength = tdsReader.readUnsignedShort();
-                    if (routingDataValueLength <= 5)// (5 is the no of bytes in protocol + port number+ length field of
-                                                    // server name)
-                    {
+                    int minDataLength = isEnhancedRouting ? 7 : 5; // Enhanced routing requires additional database name field
+                    if (routingDataValueLength <= minDataLength) {
                         throwInvalidTDS();
                     }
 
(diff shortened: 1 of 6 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -3941,13 +3948,24 @@ private void login(String primary, String primaryInstanceName, int primaryPortNu
...
                         } else {
                             if (routingInfo != null) {
-                                if (loggerRedirection.isLoggable(Level.FINE)) {
-                                    loggerRedirection
-                                            .fine(toString() + " Connection open - redirecting to server and instance: "
-                                                    + routingInfo.getFullServerName());
+                                // If we received enhanced routing info (with database name) but the server
+                                // did not acknowledge the enhanced routing feature, discard the routing info
+                                // and fall through to connect to the current server instead.
+                                if (routingInfo.getDatabaseName() != null && !serverSupportsEnhancedRouting) {
```
machavan, line 3954: So we have various conditions : 

Condition1 : C1 - server supports enhanced routing 
Condition2: C2 - database name received 

If (C1)
{
  if (C2)
   {
       -  We are good to accept the routingInfo with databseName
   } 
   else 
   {
       - ? What do we do here? Redirect in old way without databaseName?
   }
}
else 
{
  if (C2)
  {
      - Discard since this could be malicious 
  }
  else 
  {
     - Redirect in old way without databaseName ?
  } 
}


Can we confirm the behavior with other (.NET) driver too?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -6369,18 +6406,24 @@ final void processEnvChange(TDSReader tdsReader) throws SQLServerException {
...
+                String routingDatabaseName = null;
 
                 try {
                     routingDataValueLength = tdsReader.readUnsignedShort();
-                    if (routingDataValueLength <= 5)// (5 is the no of bytes in protocol + port number+ length field of
-                                                    // server name)
-                    {
+                    int minDataLength = isEnhancedRouting ? 7 : 5; // Enhanced routing adds a database name field
+                    if (routingDataValueLength <= minDataLength) {
                         throwInvalidTDS();
```
machavan, line 6429: So if databaseName is invalid we are throwing or should we be discarding ? What is .NET driver doing ?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -7249,6 +7310,25 @@ private void onFeatureExtAck(byte featureId, byte[] data) throws SQLServerExcept
...
+            case TDS.TDS_FEATURE_EXT_ENHANCEDROUTING: {
+                if (connectionlogger.isLoggable(Level.FINER)) {
+                    connectionlogger.fine(toString() + " Received feature extension acknowledgement for Enhanced Routing.");
+                }
+
+                // Enhanced Routing feature extension ack should contain exactly 1 byte:
+                // data[0] == 1 means the server supports enhanced routing
+                // data[0] == 0 means the server does not support enhanced routing
+                if (1 != data.length) {
+                    throw new SQLServerException(SQLServerException.getErrString("R_enhancedRoutingFeatureAckContainsExtraData"), null);
```
machavan, line 7324: Does .NET driver throw or ignore the FE

Ananya2: Yes the .NET throws, the exception. It validates data.Length == 1; otherwise it raises a parsing error (SqlException for corrupted TDS stream).

https://github.com/dotnet/SqlClient/blob/8964a1786ce53c9e6228511bdadd8403e90aad17/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Connection/SqlConnectionInternal.cs#L1749

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -7249,6 +7310,25 @@ private void onFeatureExtAck(byte featureId, byte[] data) throws SQLServerExcept
                 break;
             }
 
+            case TDS.TDS_FEATURE_EXT_ENHANCEDROUTING: {
+                if (connectionlogger.isLoggable(Level.FINER)) {
+                    connectionlogger.fine(toString() + " Received feature extension acknowledgement for Enhanced Routing.");
```
divang: if (connectionlogger.isLoggable(Level.FINER)), 
connectionlogger.finer level or change the isLoggable

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -6369,18 +6406,24 @@ final void processEnvChange(TDSReader tdsReader) throws SQLServerException {
                     connectionlogger.finer(toString() + " Ignored env change: " + envchange);
                 break;
             case ENVCHANGE_ROUTING:
-
-                // initialize to invalid values
-                int routingDataValueLength, routingProtocol, routingPortNumber, routingServerNameLength;
+            case ENVCHANGE_ENHANCED_ROUTING:
```
divang, line 6411: Add comment for this new case

divang (commented):
minor comment

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
Ananya2: Neither .NET nor JDBC validate the database name content beyond null/empty. Looking at the SqlClient code I just pulled up: https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Connection/SqlConnectionInternal.cs#L1259

That's the entire validation. SqlClient only checks:

ServerName is not null/empty
DatabaseName is not null/empty
Protocol == 0
Port != 0

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
Ananya2: C1 = true, C2 = true: The server acknowledged enhanced routing and provided a routed DB, so it is used in Login7.

C1 = true, C2 = false: The server acknowledged enhanced routing but did not send a DB name, so redirect uses server/port and Login7 keeps the original DB.

C1 = false, C2 = true: A routed DB name without enhanced routing ack is unsupported, so log a warning, set routingInfo = null, and connect to the current server.
.Net does the same:
https://github.com/dotnet/SqlClient/blob/8964a1786ce53c9e6228511bdadd8403e90aad17/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Connection/SqlConnectionInternal.cs#L3299

C1 = false, C2 = false: This is standard server/port-only routing and works independently of enhanced routing.

All scenarios are also covered with test cases using mocked APIs.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -1343,6 +1343,13 @@ boolean getServerSupportsJSON() {
         return serverSupportsJSON;
     }
 
+    /** whether server supports Enhanced Routing */
+    private boolean serverSupportsEnhancedRouting = false;
```
muskan124947, line 1347: Shall we reset at start of each login attempt ?

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
Ananya2: This matches the existing pattern across the driver, where checks use FINER while logging at FINE. This is safe because enabling FINER also enables FINE. Kept as is for consistency and can be cleaned up across the driver in a follow-up.

muskan124947: level mismatch

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
machavan, line 1347: ---

### 1. `serverSupportsEnhancedRouting` not reset between login retry hops — security concern

**Location:** `SQLServerConnection.java`, `onFeatureExtAck` / `login()` routing guard.

The flag is declared as a single instance field:

```java
private boolean serverSupportsEnhancedRouting = false;
```

On the first login attempt (to the primary), if the server acks with `data[0] = 1`, the flag becomes `true`. After routing, the driver reconnects to the routed server. If that server sends `ENVCHANGE 21` but does **not** send `FEATUREEXTACK 0x0F` (e.g., an older replica that doesn't know about the feature), `serverSupportsEnhancedRouting` stays `true` from the first hop. The guard in `login()`:

```java
if (routingInfo.getDatabaseName() != null && !serverSupportsEnhancedRouting) {
    // discard routing
}
```

…would then **pass through**, incorrectly accepting the second-hop routing. The fix is to reset `serverSupportsEnhancedRouting = false` at the start of each login attempt loop iteration, just as `isRoutedInCurrentAttempt` is reset each iteration. This is the same structural gap that would affect a malicious intermediate hop.

---

### 2. `routingDatabaseNameLength` initialization is split inconsistently

**Location:** `SQLServerConnection.java`, `processEnvChange`, `case ENVCHANGE_ENHANCED_ROUTING:`.

```java
int routingDataValueLength, routingProtocol, routingPortNumber, routingServerNameLength, routingDatabaseNameLength = -1;
routingDataValueLength = routingProtocol = routingPortNumber = routingServerNameLength = -1;
```

`routingDatabaseNameLength` is initialized in the declaration, but the rest are initialized on the separate line. This is misleading — a reader might think `routingDatabaseNameLength` was skipped. All five should be initialized consistently (either all in the declaration, or all via the compound-assignment line).

---

### 3. Database name validation throws a different exception type than server/port validation

**Location:** `SQLServerConnection.java`, `processEnvChange`.

Port and server-name checks use `throwInvalidTDS()`:

```java
if (routingPortNumber <= 0 || routingPortNumber > 65535) {
    throwInvalidTDS();
}
```

But database-name length validation throws directly:

```java
if (routingDatabaseNameLength <= 0 || routingDatabaseNameLength > 128) {
    throw new SQLServerException(
        SQLServerException.getErrString("R_invalidEnhancedRoutingInfo"), null);
}
```

This inconsistency means enhanced routing produces a different error path and message format than all other `processEnvChange` validation failures. Either adopt `throwInvalidTDS()` here too, or add a specific check to `throwInvalidTDS()` — but either way the style should be consistent within the method.

---

### 4. `IOBuffer.getTokenName` new case has 0% test coverage

**Location:** `IOBuffer.java`, `getTokenName` switch.

Codecov reports `0.00%` on the new lines in `IOBuffer.java`. The new case:

```java
case TDS_FEATURE_EXT_ENHANCEDROUTING:
    return "TDS_FEATURE_EXT_ENHANCEDROUTING (0x0F)";
```

…is never exercised. Adding a single assertion in the existing test class that calls `TDS.getTokenName(TDS.TDS_FEATURE_EXT_ENHANCEDROUTING)` and expects the string would close this gap.

---

Ananya2: done

machavan (dismissed):

muskan124947 (dismissed):

divang (approved):

machavan (approved):

PR #2944, opened 2026-05-07
Add nanosecond granularity option to PerformanceLogCallback

machavan:
# Add nanosecond granularity option to PerformanceLogCallback

## Summary

Adds an opt-in nanosecond duration mode to the `PerformanceLogCallback` interface, allowing applications to receive high-resolution timing data without any breaking changes to the existing API.

## Motivation

Millisecond resolution is insufficient for profiling fast operations (e.g., cached statement execution, connection pool checkout). Applications that need sub-millisecond visibility can now opt into nanosecond granularity by overriding a single default method.

## Design

- **No API signature changes** — the `publish()` methods and `registerPerformanceLogCallback()` remain identical.
- **Opt-in via default method** — `PerformanceLogCallback` gains a `default boolean useNanoseconds()` that returns `false`. Applications override it to return `true` to receive nanosecond values in the existing `duration` parameter.
- **Correct time source** — when `useNanoseconds()` is `true`, the driver uses `System.nanoTime()` for both start and end; otherwise it uses `System.currentTimeMillis()`.
- **Zero impact on existing users** — the default behavior is unchanged (milliseconds).

## Changes

| File | Change |
|------|--------|
| `PerformanceLogCallback.java` | Added `default boolean useNanoseconds()` with Javadoc |
| `PerformanceLog.java` | `Scope` reads `useNanoseconds()` at construction, selects `nanoTime()` or `currentTimeMillis()` accordingly; logger output includes correct unit label ("ms" or "ns") |
| `PerformanceLogCallbackTest.java` | Added `testPublishWithNanosecondGranularity` test |

## Usage

```java
    @Override
    public boolean useNanoseconds() {
        return true; // opt-in to nanosecond durations
    }

    @Override
    public void publish(PerformanceActivity activity, int connectionId, long duration, Exception ex) {
        // duration is now in nanoseconds
        System.out.printf("%s conn=%d duration=%dns%n", activity, connectionId, duration);
    }

    @Override
    public void publish(PerformanceActivity activity, int connectionId, int statementId, long duration, Exception ex) {
        System.out.printf("%s conn=%d stmt=%d duration=%dns%n", activity, connectionId, statementId, duration);
    }
});
```

## Testing

- All 11 existing `PerformanceLogCallbackTest` tests pass.
- New test validates nanosecond values are within expected range (> 1000ns for a 1ms+ operation).

## Backward Compatibility

Fully backward compatible. Existing callbacks that do not override `useNanoseconds()` continue to receive millisecond durations with no code changes required.

## Performance Considerations

When useNanoseconds() returns true, an **additional overhead** is expected compared to millisecond mode. This is because System.nanoTime() have higher invocation cost than System.currentTimeMillis() on some platforms. This overhead only applies when a callback is registered and logging is enabled — there is no impact when both are disabled.

## Sample Output

```
May 13, 2026 3:21:25 PM com.microsoft.sqlserver.jdbc.PerformanceLog$Scope close
FINE: ConnectionID:1 Prelogin, duration: 16078800ns
May 13, 2026 3:21:25 PM com.microsoft.sqlserver.jdbc.PerformanceLog$Scope close
FINE: ConnectionID:1 Login, duration: 736384700ns
May 13, 2026 3:21:25 PM com.microsoft.sqlserver.jdbc.PerformanceLog$Scope close
FINE: ConnectionID:1 Connection, duration: 746256200ns
May 13, 2026 3:21:25 PM com.microsoft.sqlserver.jdbc.PerformanceLog$Scope close
FINE: ConnectionID:1, StatementID:1 Request build time, duration: 1098600ns
May 13, 2026 3:21:25 PM com.microsoft.sqlserver.jdbc.PerformanceLog$Scope close
FINE: ConnectionID:1, StatementID:1 First server response, duration: 1334200ns
May 13, 2026 3:21:25 PM com.microsoft.sqlserver.jdbc.PerformanceLog$Scope close
FINE: ConnectionID:1, StatementID:1 Statement execute, duration: 48569800ns
```

### DIFF
--- a/.github/instructions/performance-metrics.instructions.md
+++ b/.github/instructions/performance-metrics.instructions.md
@@ -34,6 +34,39 @@ SQLServerDriver.registerPerformanceLogCallback(new PerformanceLogCallback() {
 });
 ```
 
+#### Nanosecond Granularity
+
+By default, the `duration` parameter is reported in **milliseconds**. To receive
+**nanosecond** granularity instead, override `useNanoseconds()` to return `true`:
+
+```java
+SQLServerDriver.registerPerformanceLogCallback(new PerformanceLogCallback() {
+    @Override
+    public boolean useNanoseconds() {
+        return true; // duration values will be in nanoseconds
+    }
+
+    @Override
+    public void publish(PerformanceActivity activity, int connectionId, long durationNs, 
+            Exception exception) {
+        System.out.printf("Activity: %s, Connection: %d, Duration: %d ns%n",
+                activity, connectionId, durationNs);
+    }
+
+    @Override
+    public void publish(PerformanceActivity activity, int connectionId, int statementId, 
+            long durationNs, Exception exception) {
+        System.out.printf("Activity: %s, Connection: %d, Statement: %d, Duration: %d ns%n",
+                activity, connectionId, statementId, durationNs);
+    }
+});
+```
+
+When `useNanoseconds()` returns `true`:
+- Timing uses `System.nanoTime()` instead of `System.currentTimeMillis()`
+- Log output uses `ns` as the unit suffix instead of `ms`
+- All publish calls (connection-level and statement-level) use nanosecond values
+
 ### 2. Java Logging Configuration
 
 Configure `java.util.logging` for the performance metrics loggers at `FINE` level:
@@ -242,3 +275,4 @@ try {
 - `SQLServerPreparedStatement.java` - Prepared statement activities
 - `PerformanceActivity.java` - Activity enum definitions
 - `PerformanceLog.java` - Logging infrastructure
+- `PerformanceLogCallback.java` - Callback interface (includes `useNanoseconds()` for nanosecond granularity)
--- a/.github/instructions/performance-metrics.instructions.md
+++ b/.github/instructions/performance-metrics.instructions.md
@@ -37,12 +37,12 @@ SQLServerDriver.registerPerformanceLogCallback(new PerformanceLogCallback() {
 #### Nanosecond Granularity
 
 By default, the `duration` parameter is reported in **milliseconds**. To receive
-**nanosecond** granularity instead, override `useNanoseconds()` to return `true`:
+**nanosecond** granularity instead, override `useNanoSeconds()` to return `true`:
 
 ```java
 SQLServerDriver.registerPerformanceLogCallback(new PerformanceLogCallback() {
     @Override
-    public boolean useNanoseconds() {
+    public boolean useNanoSeconds() {
         return true; // duration values will be in nanoseconds
     }
 
@@ -62,7 +62,11 @@ SQLServerDriver.registerPerformanceLogCallback(new PerformanceLogCallback() {
 });
 ```
 
-When `useNanoseconds()` returns `true`:
+The value of `useNanoSeconds()` is captured once at registration time and remains fixed
+for the lifetime of the callback. To change the duration unit, unregister and re-register
+with the new setting.
+
+When `useNanoSeconds()` returns `true`:
 - Timing uses `System.nanoTime()` instead of `System.currentTimeMillis()`
 - Log output uses `ns` as the unit suffix instead of `ms`
 - All publish calls (connection-level and statement-level) use nanosecond values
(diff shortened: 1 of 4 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/PerformanceLog.java
```diff
@@ -50,6 +50,7 @@ public static class Scope implements AutoCloseable {
         private PerformanceActivity activity;
         private long startTime;
         private final boolean enabled;
+        private final boolean nanos;
```
divang: useNanos should be better name here too.

src/main/java/com/microsoft/sqlserver/jdbc/PerformanceLog.java
```diff
@@ -60,16 +61,23 @@ public Scope(Logger logger, int connectionId, PerformanceActivity activity) {
 
         // Constructor for statement-level activities
         public Scope(Logger logger, int connectionId, int statementId, PerformanceActivity activity) {
-
-            // Check if logging is enabled
             this.enabled = logger.isLoggable(Level.FINE) || (callback != null);
+            boolean useNanos = false;
+            if (callback != null) {
+                try {
+                    useNanos = callback.useNanoseconds();
```
Ananya2: Cache useNanoseconds() once during registerPerformanceLogCallback(...) instead of calling it for every Scope.

Right now, if the callback implementation changes its return value at runtime, the same log stream could contain a mix of ms and ns durations, while publish(..., long duration) itself does not carry unit information. 

Caching the value at registration will keep the duration unit consistent for the lifetime of the callback, makes unit changes an explicit re-registration step, and also removes an extra virtual call from this hot path.

Also, please add a test case to validate this behavior, for example:

```
class MyCb implements PerformanceLogCallback {
    boolean nanos;                                  
    @Override public boolean useNanoseconds() { return nanos; }
    @Override public void publish(...) { ... }
}

MyCb cb = new MyCb();
cb.nanos = false;
SQLServerDriver.registerPerformanceLogCallback(cb);
// ... later in same JVM ...
cb.nanos = true;        // altered
```

The expected behavior should remain stable based on the value captured during registration.

.github/instructions/performance-metrics.instructions.md
```diff
@@ -34,6 +34,39 @@ SQLServerDriver.registerPerformanceLogCallback(new PerformanceLogCallback() {
...
+#### Nanosecond Granularity
+
+By default, the `duration` parameter is reported in **milliseconds**. To receive
+**nanosecond** granularity instead, override `useNanoseconds()` to return `true`:
+
+```java
+SQLServerDriver.registerPerformanceLogCallback(new PerformanceLogCallback() {
+    @Override
+    public boolean useNanoseconds() {
+        return true; // duration values will be in nanoseconds
```
Ananya2, line 46: useNanoseconds() is a boolean only supports two units. If we later want to leverage for microseconds, you'll need yet another opt-in method or another boolean.

Suggestion: use java.util.concurrent.TimeUnit instead:
`default TimeUnit getDurationUnit() { return TimeUnit.MILLISECONDS; }`

src/main/java/com/microsoft/sqlserver/jdbc/PerformanceLog.java
```diff
@@ -18,9 +18,13 @@ class PerformanceLog {
 
     private static PerformanceLogCallback callback;
     private static boolean callbackInitialized = false;
+    private static boolean cachedUseNanos = false;
```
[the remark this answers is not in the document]
muskan124947: Registration is a one-time startup operation — registerCallback() is called once before any JDBC work begins. By the time any Scope is created, the registration has long completed. There's a natural happens-before in any real application (setup thread completes → worker threads start making JDBC calls).

The race window is theoretical — It would only matter if a thread creates a Scope at the exact instant another thread is mid-registration. Even then, the worst case is one metric uses milliseconds instead of nanoseconds, or one callback invocation is missed.

volatile adds cost on the hot path — Every Scope creation reads callback and cachedUseNanos. Adding volatile forces a memory barrier on every read — that's a real overhead for every single JDBC operation, to guard against a race that essentially never happens in practice.

The synchronized on register/unregister already provides ordering between those operations themselves, which is the important part (no torn writes between callback and cachedUseNanos).

divang (approved):

muskan124947 (approved):

Ananya2 (approved):

PR #2953, opened 2026-05-14
Fix CI-MACOS flakiness in connection-timeout tests by giving login phase adequate headroom

Ananya2:
Fix CI-MACOS flakiness in connection-timeout tests: Three tests were consistently failing on the MacOS CI pipeline with "Read timed out" errors while passing on Linux and Windows:
  - ConfigurableRetryLogicTest.connectionTimingTest
  - ReflectiveTests.testDefaultRetry
  - TimeoutTest.testDefaultSocketTimeoutUnlimitedAfterLogin

Root cause
----------
All three tests assumed that prelogin + TLS handshake + LOGIN7 could complete within a short loginTimeout (5s, or the default 15s). Hosted MacOS runners are noticeably slower than Linux/Windows agents, so the JDBC socket-read deadline expired before the server's clean response arrived. The thrown SocketTimeoutException either bypassed the inner
assertion (ReflectiveTests) or produced an unexpected error message (ConfigurableRetryLogicTest), and in TimeoutTest the login phase itself timed out before the post-login behavior could be verified.

This was not caused by a driver regression. PR #2927 (IOBuffer login socket timeout fix) correctly tightened login-phase socket deadlines; the tests just had insufficient headroom to ride out slow CI agents.

Fixes
-----
ReflectiveTests.testDefaultRetry
  loginTimeout: 5  -> 15  (allow initial getConnection on slow agents)
  bound:        6000 -> 32000 ms
  Bound now matches the actual retry math:  (retryCount + 1) * loginTimeout + slack = 0 + 15s + 5s = 20s. Tight enough to catch an accidental extra retry.
  The previous 6000ms bound was inconsistent with retry semantics and only passed before #2927 because the login socket lacked a deadline.

ConfigurableRetryLogicTest.connectionTimingTest
  Added loginTimeout=20 to all three retry connection strings so the TDS 4060 "cannot open database" round-trip always completes before the JDBC socket read deadline, even on slow CI agents.

TimeoutTest.testDefaultSocketTimeoutUnlimitedAfterLogin
  loginTimeout:   5s  -> 15s
  WAITFOR DELAY:  8s  -> 20s
  The test verifies post-login socket behavior; raising both keeps the WAITFOR safely above loginTimeout while giving the login phase enough headroom to complete on any CI agent.

These changes are platform-agnostic - no @DisabledOnOs gates, no OS-specific branches, no relaxed error-message matching. The under-test assertions remain intact and will still fail fast on real regressions.

### DIFF
--- a/src/test/java/com/microsoft/sqlserver/jdbc/configurableretry/ConfigurableRetryLogicTest.java
+++ b/src/test/java/com/microsoft/sqlserver/jdbc/configurableretry/ConfigurableRetryLogicTest.java
@@ -498,9 +498,11 @@ public class ConfigurableRetryLogicTest extends AbstractTest {
         long totalTime;
         long timerStart = System.currentTimeMillis();
 
-        // No retries since CRL rules override, expected time ~1 second
+        // No retries since CRL rules override, expected time ~1 second.
+        // loginTimeout=30 ensures the server's TDS 4060 "cannot open database" round-trip
+        // always completes before the JDBC socket read deadline, even on slow CI agents.
         try {
-            testConnectionRetry("blah", "retryConn={9999};");
+            testConnectionRetry("blah", "loginTimeout=30;retryConn={9999};");
         } catch (Exception e) {
             assertTrue(
                     (e.getMessage().toLowerCase()
@@ -517,7 +519,8 @@ public class ConfigurableRetryLogicTest extends AbstractTest {
 
         // (0s attempt + 0s attempt + 10s wait + 0s attempt) = expected 10s execution time
         try {
-            testConnectionRetry("blah", "retryConn={4060,4070};connectRetryCount=2;connectRetryInterval=10");
+            testConnectionRetry("blah",
+                    "loginTimeout=30;retryConn={4060,4070};connectRetryCount=2;connectRetryInterval=10");
         } catch (Exception e) {
             assertTrue(
                     (e.getMessage().toLowerCase()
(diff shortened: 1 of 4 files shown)


src/test/java/com/microsoft/sqlserver/jdbc/configurableretry/ConfigurableRetryLogicTest.java
```diff
@@ -498,9 +498,11 @@ public void connectionTimingTest() {
         long totalTime;
         long timerStart = System.currentTimeMillis();
 
-        // No retries since CRL rules override, expected time ~1 second
+        // No retries since CRL rules override, expected time ~1 second.
+        // loginTimeout=30 ensures the server's TDS 4060 "cannot open database" round-trip
+        // always completes before the JDBC socket read deadline, even on slow CI agents.
         try {
-            testConnectionRetry("blah", "retryConn={9999};");
+            testConnectionRetry("blah", "loginTimeout=30;retryConn={9999};");
```
divang: Why loginTimeout=30 instead of loginTimeout=15 (which was used in the other two tests)?
30 seconds seems excessive if the TDS 4060 round-trip should complete quickly
Consider whether 15 or 20 seconds would be sufficient here for consistency.

divang (approved):

muskan124947 (approved):

machavan (approved):

PR #2958, opened 2026-05-19
Fix | Avoid extra whitespace around prepared statement parameter markers (#2946)

### LINKED ISSUE #2946: Unneeded spaces added around prepared statement parameters
## Driver version
12.4.1 - 13.4.0

## SQL Server version
Any but used Microsoft SQL Server 2016 (SP3-GDR) (KB5084821) - 13.0.6485.1 (X64)   Mar 13 2026 08:38:10   Copyright (c) Microsoft Corporation  Developer Edition (64-bit) on Windows 10 Pro 10.0 <X64> (Build 26200: ) (Hypervisor) 

## Client Operating System
Microsoft Windows [Version 10.0.26200.8246]

## JAVA/JVM version
Any but used openjdk 17.0.10 2024-01-16
OpenJDK Runtime Environment Temurin-17.0.10+7 (build 17.0.10+7)
OpenJDK 64-Bit Server VM Temurin-17.0.10+7 (build 17.0.10+7, mixed mode, sharing)

## Problem description
https://github.com/microsoft/mssql-jdbc/pull/2192 added spaces around prepared statement parameters (to auto-fix incorrect SQL looking at the test cases?)
MS SQL server plan guides match on the exact SQL and failed to match when the driver was updated from 11.2.4.jre11 to 13.4.0.jre11 due to the extra spaces added.
The original SQL (before handed to the JDBC driver) is generated by an ORM, which is why plan guide are used instead of fixing the SQL.

## Expected behavior
Prepared statement parameters are replaced without adding extra white space

## Actual behavior
    "SELECT t0.ID FROM SOME_TABLE t0 WHERE t0.ID = ?"
generates extra space before and after the parameter:

    "SELECT t0.ID FROM SOME_TABLE t0 WHERE t0.ID =  @P1 "

If the parameter is not at the end, the space after is removed:

    "SELECT t0.ID FROM SOME_TABLE t0 WHERE t0.ID = ? ORDER BY t0.ID"
generates only extra space before:

    "SELECT t0.ID FROM SOME_TABLE t0 WHERE t0.ID =  @P1 ORDER BY t0.ID"

The https://github.com/microsoft/mssql-jdbc/pull/2192 added test only tests if the query succeeds. It does not check the parameter replacement is correct (though SqlServerPreparedStatement#replaceParameterMarkers is package private, so hard to test stand-alone without change that). Ideally the test check that replacing parameter does not change the output, since plan guide (maybe other things) depend on the exact SQL statement text.

The current implementation always adds the spaces, then has some logic to remove the trailing space if followed by a space, but not the starting one. Better would be to only add the extra spacing if needed (if the driver should be auto-fixing invalid SQL in the first place) and remove the confusing boolean isPreparedSQL parameter for makeParamName. For example (not-tested):

    // add space before if none
    if(dstBegin > 0 && !sqlSrc.charAt(dstBegin - 1) != ' ') {
        sqlDst[dstBegin] = ' ';
        dstBegin++;
    }
    dstBegin += SQLServerConnection.makeParamName(nParam++, sqlDst, dstBegin);
    srcBegin = srcEnd + 1;
    // add space after if none
    if(srcBegin < sqlSrc.length() && !sqlSrc.charAt(srcBegin) != ' ') {
        sqlDst[dstBegin] = ' ';
        dstBegin++;
    }

SQLServerConnection.makeParamName could be moved to SqlServerPreparedStatement as only used there (when replaceParameterMarkers was moved from there, makeParamName was not)

machavan:
## Problem description

The driver unconditionally pads every substituted `@P<n>` parameter marker with a leading and trailing space when rewriting user SQL for a prepared statement. As a result, source SQL such as

```sql
WHERE t0.ID = ?
```

is sent to SQL Server as

```sql
WHERE t0.ID =  @P0 
```

This breaks any feature that matches on the exact normalized statement text:

- **SQL Server plan guides** (`sp_create_plan_guide`) silently stop matching, so DBA-applied hints are no longer injected at compile time.
- **Query Store forced plans** keyed on `query_hash` similarly fail to match.
- **Audit / Extended Events / Resource Governor classifier** filters that inspect `sys.dm_exec_sql_text` no longer match.
- **APM tools** that fingerprint SQL produce duplicate entries for the same logical query.

The regression appeared between mssql-jdbc 11.2.4 and 12.4.1 and was introduced in PR #2192, which solved a real bug (queries like `c1=?and c2=?` lacked a separator after parameter substitution) but did so by padding *every* marker rather than only those that needed a separator.

Fixes #2946.

## Fix Description

`SQLServerPreparedStatement.replaceParameterMarkers` now invokes `SQLServerConnection.makeParamName` with `isPreparedSQL = false` (which emits a bare `@P<n>` with no padding) and **conditionally** writes a single space *after* the marker (or after `@P<n> OUT` for output parameters) only when the next source character would otherwise extend the substituted text into a longer SQL identifier (i.e. when it is a letter, digit, or underscore -- `Character.isJavaIdentifierPart`). A leading separator is never needed because `@` always starts a fresh SQL token.

This preserves PR #2192's safety property -- the driver still inserts a separator when one is required for correctness (e.g. `?and` -> `@P0 and`) -- while leaving all other whitespace in the user SQL untouched.

Behavior summary:

| User SQL                | Rewritten SQL (without fix)        | Rewritten SQL (with fix)   | Behavior  |
|-------------------------|------------------------------------|----------------------------|-----------|
| `... ID = ?`            | `... ID =  @P0 `                   | `... ID = @P0`             | Changed   |
| `... ID = ? ORDER BY x` | `... ID =  @P0  ORDER BY x`        | `... ID = @P0 ORDER BY x`  | Changed   |
| `... ID=?`              | `... ID= @P0 `                     | `... ID=@P0`               | Changed   |
| `... ID=?and y=2`       | `... ID= @P0 and y=2`              | `... ID=@P0 and y=2`       | Changed (separator preserved) |
| `(?,?,?)`               | `( @P0 , @P1 , @P2 )`              | `(@P0,@P1,@P2)`            | Changed   |
| `SELECT ?`              | `SELECT  @P0 `                     | `SELECT @P0`               | Changed   |

The `?and` row keeps a single space between `@P0` and `and` because they would otherwise merge into the identifier `@P0and` -- that is the case PR #2192 was solving and the fix continues to handle it. All other rows show the spurious whitespace that broke exact-text matching.

Tests added in `PreparedStatementTest.testPreparedStatementParamNameSpacingNoExtraWhitespace` prepare each scenario, execute it, then reflect the private `preparedSQL` field on `SQLServerPreparedStatement` and assert on the exact rewritten string. The pre-existing PR #2192 test (`testPreparedStatementParamNameSpacingWithMultipleParams`) is kept and continues to cover the no-separator cases.

## New Public APIs

None.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
@@ -3946,21 +3946,26 @@ public class SQLServerPreparedStatement extends SQLServerStatement implements IS
             if (sqlSrc.length() == srcEnd)
                 break;
 
-            // Issue #2946: only insert whitespace around @P<n> when the surrounding source SQL
-            // does not already provide a separator. Preserves PR #2192's fix for cases like
-            // "=?" while avoiding extra spaces that break exact-text matching (plan guides).
-            if (dstBegin > 0 && !Character.isWhitespace(sqlDst[dstBegin - 1])) {
-                sqlDst[dstBegin++] = ' ';
-            }
+            // Issue #2946: avoid injecting extra whitespace around the substituted @P<n>
+            // marker, which broke exact-text matching used by SQL Server plan guides,
+            // Query Store forced plans, sql_text-based auditing, etc.  PR #2192 added
+            // unconditional padding to fix queries like "c1=?and c2=?" where the lack of
+            // a separator caused "@P<n>and" to be parsed as a single identifier.  We
+            // preserve that safety property by inserting a single space only when the
+            // *next* source character would extend the @P<n> (or @P<n> OUT) token into a
+            // longer identifier -- i.e. when it is a letter, digit, or underscore.  A
+            // leading separator is never required because '@' cannot continue a
+            // preceding token.
             dstBegin += SQLServerConnection.makeParamName(nParam++, sqlDst, dstBegin, false);
             srcBegin = srcEnd + 1;
-            if (srcBegin < sqlSrc.length() && !Character.isWhitespace(sqlSrc.charAt(srcBegin))) {
-                sqlDst[dstBegin++] = ' ';
-            }
 
             if (params[paramIndex++].isOutput() && (!isReturnValueSyntax || paramIndex > 1)) {
                 System.arraycopy(OUT, 0, sqlDst, dstBegin, OUT.length);
                 dstBegin += OUT.length;
+            }
+
+            if (srcBegin < sqlSrc.length() && Character.isJavaIdentifierPart(sqlSrc.charAt(srcBegin))) {
+                sqlDst[dstBegin++] = ' ';
             }
         }
 
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
@@ -3946,8 +3946,17 @@ public class SQLServerPreparedStatement extends SQLServerStatement implements IS
             if (sqlSrc.length() == srcEnd)
                 break;
 
-            dstBegin += SQLServerConnection.makeParamName(nParam++, sqlDst, dstBegin, true);
-            srcBegin = srcEnd + 1 <= sqlSrc.length() - 1 && sqlSrc.charAt(srcEnd + 1) == ' ' ? srcEnd + 2 : srcEnd + 1;
+            // Issue #2946: only insert whitespace around @P<n> when the surrounding source SQL
+            // does not already provide a separator. Preserves PR #2192's fix for cases like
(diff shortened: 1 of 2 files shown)


divang (dismissed):

muskan124947:
Can we add test for (@P0,@P1,@P2)
(?,?,?) → (@P0,@P1,@P2) (no inserted spaces), but the code logic suggests otherwise:

- Pre-space check: [dstBegin > 0 && !Character.isWhitespacesqlDst[dstBegin - 1])— the character before the first ? is (, which is not whitespace, so a space would be inserted.
- Post-space check: srcBegin < sqlSrc.length() && !Character.isWhitespace(sqlSrc.charAt(srcBegin)) — the character after each ? is , or ), which is not whitespace, so a space would be inserted.

It is still ( @P0 , @P1 , @P2 )

muskan124947 (approved):

Ananya2 (approved):

PR #2960, opened 2026-05-22
Introduce parameter length hints via defineParameterType and setObject

### LINKED ISSUE #2913: Why does String paramater being passed to query go as nvarchar(4000) or varchar(8000)
## Question
We have seen that when ever we are running a query using JPA and passing String parameter it is being sent as nvarchar(4000) even though the target column is varchar(40). So we tried using jdbcTemplate and setting parameter using MapSqlParameterSource's addValue with SqlParameterValue mentioning datatype as varchar and sacle as 40. Even then we are seeing on the sql server side it is being recived as nvarchar(4000) . Because of this Our query is looking for higher memory grant . Note our query has around 10 input params and all are string. Even with just 10 parameters the memory grant out query is looking for is around 9 GB. 

Here is the example
**With JPA we are sending as below**
@NativeQuery("Select * from Users where dep =: dep and tenure = :tenure ")
List<User> getUserByDepAndTenure(@Param("dep") String dep, @Param String("tenure") tenure);

**and in the Query store we can see that it is being shown as** 

(@P0 nvarchar(4000), @P1 nvarchar(4000) )Select * from Users where dep =@P0 and tenure = @P1

**Similarly when I use JDBCTemplate as below: ** 

String dep = "Dep";
String tenure = "Tenure";
String query = "Select * from Users where dep =: dep and tenure = :tenure";
MapSqlParameterSource params = new MapSqlParameterSource();
params.addValue("dep", new SqlParameterValue(Types.VACRHAR, 40, dep));
params.addValue("tenure", new SqlParameterValue(Types.VACRHAR, 4, tenure));

**and in the Query store we can see that it is being shown as** 

(@P0 varchar(8000), @P1 varchar(8000) )Select * from Users where dep =@P0 and tenure = @P1
 

Is there a way to tell sql-driver to use given datatype and scale rather than defaulting to navarchar(4000) or varchar(8000) ?

muskan124947:
## Summary

Adds a new public API, `defineParameterType(int parameterIndex, int sqlType, int maxLength)`, to `ISQLServerPreparedStatement` and `SQLServerPreparedStatement`, and also enables the existing `setObject` overloads that accept `scaleOrLength` to act as length hints for bounded short character and binary types.

This allows the driver to send tighter TDS type definitions on the wire, such as `nvarchar(50)` instead of `nvarchar(4000)`, which gives SQL Server much more accurate width information for memory grant calculations.

The change also brings the same behavior to callable statements, centralizes hint validation in `Parameter.java`, and adds comprehensive prepared and callable statement coverage.

## Motivation

When applications use `setString`, `setNString`, `setBytes`, or `setObject` for short character and binary values, the driver has historically declared the maximum bounded width on the wire:

- `varchar(8000)`
- `nvarchar(4000)`
- `varbinary(8000)`

SQL Server uses those declared widths when estimating row size and memory grants. For small actual values, this can lead to substantial over-allocation, especially for batches and concurrent workloads.

The JDBC API does not provide a dedicated standard method for declaring bounded short string and binary parameter widths, so this change adds an explicit driver API and also honors `scaleOrLength` on supported `setObject` paths.

## API

```
// On ISQLServerPreparedStatement / SQLServerPreparedStatement
void defineParameterType(int parameterIndex, int sqlType, int maxLength) throws SQLServerException;
```

### Supported Types

| `sqlType` | Wire type produced |
|-----------|-------------------|
| `Types.VARCHAR` | `varchar(N)` |
| `Types.CHAR` | `varchar(N)` (CHAR routes through VARCHAR on wire) |
| `Types.NVARCHAR` | `nvarchar(N)` |
| `Types.NCHAR` | `nvarchar(N)` (NCHAR routes through NVARCHAR on wire) |
| `Types.VARBINARY` | `varbinary(N)` |
| `Types.BINARY` | `varbinary(N)` (BINARY routes through VARBINARY on wire) |

### Behavior

- `defineParameterType` sets a maximum length hint for supported bounded short character and binary parameter types.
- The existing `setObject` overloads that accept `scaleOrLength` now also use that value as a length hint for the same supported types.
- If both are present, `defineParameterType` takes precedence over `setObject(scaleOrLength)`.
- Hints are applied when the driver computes the parameter type definition for execution.
- Non-positive hints passed to `defineParameterType` are rejected eagerly at call time with `R_invalidParameterLength`. Non-positive hints passed via `setObject(scaleOrLength)` are rejected at execution time.
- The `sqlType` argument to `defineParameterType` enforces type family compatibility at execution time. The setter must produce a type in the same family as the declared `sqlType` (character family: VARCHAR, CHAR, NVARCHAR, NCHAR; binary family: VARBINARY, BINARY). Cross-family usage fails with `R_defineParameterTypeTypeMismatch`.
- When Always Encrypted is active for a parameter, hints are silently ignored — the driver uses the exact type info derived from the actual value.
- Unsupported types passed to `defineParameterType` are rejected with `R_unsupportedTypeForDefineParamType`.
- If the actual value length exceeds the active hint, execution fails with `R_parameterTypeValueLengthExceedsHint`.
- Hints persist across `clearParameters()` and across batch rows for the statement parameter they were defined on.
- Hints are scoped per parameter and do not leak across statements.
- Existing behavior remains unchanged when no hint is provided.

## Boundary Rules

| Condition | Wire type |
|---|---|
| `VARCHAR` or `CHAR` hint greater than `8000` | `varchar(max)` |
| `NVARCHAR` or `NCHAR` hint greater than `4000` | `nvarchar(max)` |
| `VARBINARY` or `BINARY` hint greater than `8000` | `varbinary(max)` |
| Any supported hint less than or equal to `0` | rejected with `R_invalidParameterLength` |

## Validation Model

- **Type family enforcement** is performed at setter time in Parameter.setValue(). When defineParameterTypeCalled is true, validateSetterTypeFamily() ensures the setter's JDBC type family matches the declared family before any further processing.
- **Length hint resolution and value-length enforcement** are centralized in Parameter.GetTypeDefinitionOp when the driver resolves the effective parameter type definition. The length hint is stored in a dedicated defineParameterTypeLengthHint field (separate from valueLength/userProvidesPrecision used by DECIMAL/NUMERIC precision logic).

This keeps defineParameterType and setObject hint handling aligned and ensures the same rules apply across prepared and callable statement execution paths.

## Files Changed

| File | Description |
|---|---|
| `src/main/java/com/microsoft/sqlserver/jdbc/ISQLServerPreparedStatement.java` | Adds the public `defineParameterType` API and Javadoc |
| `src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java` | Stores `defineParameterType` hints and wires `setObject(scaleOrLength)` into the hint flow |
| `src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCallableStatement.java` | Brings the same hint behavior to callable statement `setObject` paths |
| `src/main/java/com/microsoft/sqlserver/jdbc/Parameter.java` | Resolves hint precedence, validates non-positive hints, and enforces value length against the active hint |
| `src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResource.java` | Adds error messages for validation (`R_invalidParameterLength`, `R_unsupportedTypeForDefineParamType`, `R_parameterTypeValueLengthExceedsHint`, `R_defineParameterTypeTypeMismatch`) |
| `src/test/java/com/microsoft/sqlserver/jdbc/preparedStatement/ParameterLengthHintTest.java` | Comprehensive prepared statement coverage |
| `src/test/java/com/microsoft/sqlserver/jdbc/callablestatement/CallableParameterLengthHintTest.java` | Callable statement coverage and parity checks |
| `.github/instructions/parameter-hints/README.md` | Overview and behavior documentation for parameter hints |
| `.github/instructions/parameter-hints/define-parameter-type.md` | API design and `defineParameterType` behavior details |
| `.github/instructions/parameter-hints/setObject-scaleOrLength.md` | `setObject(scaleOrLength)` hint behavior and precedence documentation |

### Prepared Statement Coverage
- `defineParameterType` hint-to-wire-type mapping for `VARCHAR`, `CHAR`, `NVARCHAR`, `NCHAR`, `VARBINARY`, and `BINARY`
- `setObject(scaleOrLength)` hint behavior for the same supported types
- precedence of `defineParameterType` over `setObject(scaleOrLength)`
- boundary promotion to `varchar(max)`, `nvarchar(max)`, and `varbinary(max)`
- null values with and without active hints
- empty string and empty binary payload behavior
- per-parameter independence on multi-parameter statements
- persistence across `clearParameters()`
- persistence across batch rows
- batch execution with mixed value sizes under the same hint

### Callable Statement Coverage
- callable statement parity for `setObject(scaleOrLength)` hint handling
- wire-type selection for supported bounded string and binary types
- precedence behavior where `defineParameterType` is also present
- validation and execution behavior matching prepared statements

### Validation and Error Coverage
- non-positive hint rejection with `R_invalidParameterLength`
- unsupported type rejection with `R_unsupportedTypeForDefineParamType`
- value length exceeding the active hint with `R_parameterTypeValueLengthExceedsHint`
- invalid parameter index handling
- closed statement behavior
- failure cases during batch execution when a row exceeds the active hint
- type family mismatch rejection with `R_defineParameterTypeTypeMismatch` (e.g. VARCHAR declared + setBytes)
- same-family compatibility validation (e.g. VARCHAR declared + setNString succeeds)

### Regression Coverage
- existing default behavior when no hint is provided
- no hint leakage across separate statement instances
- interaction with existing execution flows for prepared and callable statements
- documentation updated to match final runtime behavior and error keys

## Breaking Changes

**`setObject(..., scaleOrLength)` now enforces length for string/binary types.** The `setObject(int, Object, int, int)` method (and the callable statement `setObject(String, Object, int, int)` variants) now treats `scaleOrLength` as a maximum length constraint for character and binary target types (VARCHAR, CHAR, NVARCHAR, NCHAR, VARBINARY, BINARY). Previously, per the JDBC 4.3 specification, `scaleOrLength` was only defined for DECIMAL/NUMERIC (as scale) and InputStream/Reader (as stream length) — it was ignored for string and binary types.

With this change, if the actual value length exceeds the specified `scaleOrLength`, execution will fail with `R_parameterTypeValueLengthExceedsHint`. Applications that pass arbitrary or undersized `scaleOrLength` values to `setObject` for string/binary types must either:
- Increase the value to accommodate their largest expected payload, or
- Use a `setObject` overload that omits the length parameter

The `defineParameterType` API itself is purely additive — existing behavior is unchanged when it is not called.

Additionally, type family mismatches (e.g. defineParameterType(1, Types.VARCHAR, 50) followed by setBytes(1, ...)) now throw at setter time rather than at execution time. Code that catches these errors around executeUpdate() / executeBatch() must move the catch to the setter call.

Closes [Why does String paramater being passed to query go as nvarchar(4000) or varchar(8000) #2913](https://github.com/microsoft/mssql-jdbc/issues/2913)


src/main/java/com/microsoft/sqlserver/jdbc/Parameter.java
```diff
@@ -621,6 +720,15 @@ private void setTypeDefinition(DTV dtv) {
                         if (JDBCType.LONGVARBINARY == jdbcTypeSetByUser) {
                             param.typeDefinition = VARBINARY_MAX;
                         }
+                    } else if (null != applicationLengthHint) {
+                        // Application-provided hint from defineParameterType() or setObject(..., scaleOrLength):
+                        // declare the requested length for VARBINARY/BINARY.
+                        int hint = applicationLengthHint;
+                        if (hint > DataTypes.SHORT_VARTYPE_MAX_BYTES) {
```
Ananya2, line 815: Could you clarify how hints in the 4001-8000 range are handled when sendStringParametersAsUnicode=true? Since setString is emitted as NVARCHAR, I'd expect a hint such as 5000 to resolve to nvarchar(max) rather than nvarchar(5000). Please add test cases for 4001, 5000, and 7999 to validate this boundary behavior

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -2008,19 +2040,27 @@ public final void setObject(int parameterIndex, Object x, int targetSqlType,
...
-                        || Reader.class.isInstance(x)
-                        || microsoft.sql.Types.VECTOR == targetSqlType) ? scaleOrLength : null,
+        JDBCType targetJDBCType = JDBCType.of(targetSqlType);
+
+        setObject(setterGetParam(parameterIndex), x, JavaType.of(x), targetJDBCType,
+            (java.sql.Types.NUMERIC == targetSqlType || java.sql.Types.DECIMAL == targetSqlType
+                || java.sql.Types.TIMESTAMP == targetSqlType || java.sql.Types.TIME == targetSqlType
+                || microsoft.sql.Types.DATETIMEOFFSET == targetSqlType || InputStream.class.isInstance(x)
+                || Reader.class.isInstance(x) || microsoft.sql.Types.VECTOR == targetSqlType
+                || JDBCType.CHAR == targetJDBCType || JDBCType.VARCHAR == targetJDBCType
```
Ananya2, line 2085: This is a breaking behavioral change. setObject(..., scaleOrLength) previously ignored scaleOrLength for string/binary types (per JDBC 4.3), it now enforces it as a length constraint and can cause existing applications to fail with R_parameterTypeValueLengthExceedsHint. Please document this in the description instead of marking "Breaking Changes: None."

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -219,6 +219,38 @@ public int getPreparedStatementHandle() throws SQLServerException {
...
+                    typeName = String.valueOf(sqlType);
+                }
+                MessageFormat form = new MessageFormat(
+                        SQLServerException.getErrString("R_unsupportedTypeForDefineParamType"));
+                SQLServerException.makeFromDriverError(connection, this,
+                        form.format(new Object[] {typeName}), null, false);
+        }
+        Parameter param = setterGetParam(parameterIndex);
+        param.setDefineParameterTypeCalled(true);
+        param.setValueLength(maxLength); // also sets userProvidesPrecision = true
```
Ananya2: sqlType is validated but never used, the actual wire type is determined entirely by the setter. This allows defineParameterType(1, Types.VARCHAR, 50) followed by setBytes(...) to silently send VARBINARY(50), ignoring the declared type with no error. This gives a false sense of type safety. Either enforce that the setter matches the declared sqlType, or remove sqlType from the API since it doesn't affect the emitted type.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -219,6 +219,38 @@ public int getPreparedStatementHandle() throws SQLServerException {
         return prepStmtHandle;
     }
 
+    @Override
+    public void defineParameterType(int parameterIndex, int sqlType, int maxLength) throws SQLServerException {
+        checkClosed();
```
Ananya2, line 224: maxLength <= 0 should be validated in defineParameterType() instead of being deferred to execution, since the value is already known here.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCallableStatement.java
```diff
@@ -1994,7 +1994,17 @@ public void setObject(String parameterName, Object value, int sqlType, int decim
...
         checkClosed();
-        setObject(setterGetParam(findColumn(parameterName)), value, JavaType.of(value), JDBCType.of(sqlType), decimals,
+        // decimals - for java.sql.Types.DECIMAL or java.sql.Types.NUMERIC types,
+        // this is the number of digits after the decimal point.
+        // For supported short character/binary SQL types (VARCHAR, CHAR, NVARCHAR, NCHAR,
+        // VARBINARY, BINARY), this is treated as application-provided length hint.
+        // Precedence: defineParameterType() hint (if present) takes priority over this value.
+        // For all other types, this value will be ignored.
+        setObject(setterGetParam(findColumn(parameterName)), value, JavaType.of(value), JDBCType.of(sqlType),
+                (java.sql.Types.NUMERIC == sqlType || java.sql.Types.DECIMAL == sqlType
```
Ananya2, line 2022: Extract this duplicated 6-type check into a private helper (e.g., isLengthHintType(int sqlType)) so the CallableStatement and PreparedStatement setObject implementations share a single source of truth.

src/test/java/com/microsoft/sqlserver/jdbc/preparedStatement/ParameterLengthHintTest.java
```diff
@@ -0,0 +1,1261 @@
...
+                    Arguments.of("A", Types.VARCHAR, 1, "nvarchar(1)", "VARCHAR minimum meaningful hint"),
+                    Arguments.of("boundary", Types.VARCHAR, 8000, "nvarchar(max)", "VARCHAR hint=8000 exceeds nvarchar 4000-char limit"),
+                    Arguments.of("maxtest", Types.VARCHAR, 8001, "nvarchar(max)", "VARCHAR hint>8000 promotes to max"),
+
+                    Arguments.of("hello", Types.CHAR, 10, "nvarchar(10)", "CHAR hint equal to value length"),
+                    Arguments.of("hi", Types.CHAR, 100, "nvarchar(100)", "CHAR hint larger than value"),
+                    Arguments.of("A", Types.CHAR, 1, "nvarchar(1)", "CHAR minimum meaningful hint"),
+                    Arguments.of("boundary", Types.CHAR, 8000, "nvarchar(max)", "CHAR hint=8000 exceeds nvarchar 4000-char limit"),
+                    Arguments.of("maxtest", Types.CHAR, 8001, "nvarchar(max)", "CHAR hint>8000 promotes to max")
+            );
```
Ananya2, line 124: varcharHintCases() misses the critical 4001–7999 boundary with sendStringParametersAsUnicode=true, where VARCHAR hints are emitted as NVARCHAR. Please add cases for 4001, 5000, and 7999, and verify they emit nvarchar(max) to cover this boundary.

src/test/java/com/microsoft/sqlserver/jdbc/preparedStatement/ParameterLengthHintTest.java
```diff
@@ -0,0 +1,1261 @@
...
+                        pstmt.setNString(1, "batchrow" + i);
+                    } else {
+                        pstmt.setString(1, "batchrow" + i);
+                    }
+                    pstmt.addBatch();
+                }
+                assertEquals(expectedTypeDef, getTypeDefinition(pstmt, 1),
+                        "Expected type definition to persist across batch rows");
+                int[] counts = pstmt.executeBatch();
+                assertEquals(BATCH_SIZE, counts.length);
```
Ananya2, line 976: These batch tests verify the type definition and row counts, but never read the inserted values back to validate data integrity under the length hint. Since read-back helpers already exist, please use them to verify the stored values match the inserted values and catch potential truncation regressions.

src/test/java/com/microsoft/sqlserver/jdbc/preparedStatement/ParameterLengthHintTest.java
```diff
@@ -0,0 +1,1261 @@
...
+                pstmt.addBatch();
+                pstmt.setObject(1, value, sqlType, 40);
+                pstmt.addBatch();
+
+                assertEquals(expectedTypeDef, getTypeDefinition(pstmt, 1),
+                    "Expected type definition from setObject scaleOrLength hint");
+                int[] counts = pstmt.executeBatch();
+                assertEquals(2, counts.length);
+                }
+                assertEquals(2, countRowsSince(maxIdBefore));
```
Ananya2, line 1097: These batch tests verify the type definition and row counts, but never read the inserted values back to validate data integrity under the length hint. Since read-back helpers already exist, please use them to verify the stored values match the inserted values and catch potential truncation regressions.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -219,6 +219,38 @@ public int getPreparedStatementHandle() throws SQLServerException {
...
+            case java.sql.Types.CHAR:
+            case java.sql.Types.NVARCHAR:
+            case java.sql.Types.NCHAR:
+            case java.sql.Types.VARBINARY:
+            case java.sql.Types.BINARY:
+                break;
+            default:
+                String typeName;
+                try {
+                    typeName = java.sql.JDBCType.valueOf(sqlType).getName() + " (" + sqlType + ")";
```
machavan, line 246: The sqlType argument is checked against the list of supported types, but after that it's never used — the actual type sent to the server comes from the setter you call (e.g. setBytes, setString), not from sqlType.

So this compiles and runs without error, but sends varbinary(50):


ps.defineParameterType(1, Types.VARCHAR, 50); // says VARCHAR
ps.setBytes(1, someBytes);                    // actually sends VARBINARY


The caller thinks they declared VARCHAR, but the driver ignores that and uses the setter's type. Two options:

Enforce it — throw if the setter's type doesn't match the declared sqlType,
Separately: maxLength <= 0 is already known here but only rejected later at execution time. Reject it right here instead.> ps.defineParameterType(1, Types.VARCHAR, 50); // says VARCHAR
ps.setBytes(1, someBytes); // actually sends VARBINARY

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -2008,19 +2040,27 @@ public final void setObject(int parameterIndex, Object x, int targetSqlType,
...
 
-        setObject(setterGetParam(parameterIndex), x, JavaType.of(x), JDBCType.of(targetSqlType),
-                (java.sql.Types.NUMERIC == targetSqlType || java.sql.Types.DECIMAL == targetSqlType
-                        || java.sql.Types.TIMESTAMP == targetSqlType || java.sql.Types.TIME == targetSqlType
-                        || microsoft.sql.Types.DATETIMEOFFSET == targetSqlType || InputStream.class.isInstance(x)
-                        || Reader.class.isInstance(x)
-                        || microsoft.sql.Types.VECTOR == targetSqlType) ? scaleOrLength : null,
+        JDBCType targetJDBCType = JDBCType.of(targetSqlType);
+
+        setObject(setterGetParam(parameterIndex), x, JavaType.of(x), targetJDBCType,
```
machavan, line 2080: This enforces scaleOrLength, which was previously ignored for string/binary types per JDBC 4.3 (as the PR doc itself quotes). Existing apps passing an incidental scaleOrLength smaller than the value will now throw R_parameterTypeValueLengthExceedsHint. This is a behavioral break — the PR's "Breaking Changes: None" should be corrected, and given the driver's back-compat stance, consider an opt-in.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCallableStatement.java
muskan124947: Leaving it like this only for now due to existing behavior

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
```diff
@@ -219,6 +219,45 @@ public int getPreparedStatementHandle() throws SQLServerException {
...
+                MessageFormat form = new MessageFormat(
+                        SQLServerException.getErrString("R_unsupportedTypeForDefineParamType"));
+                SQLServerException.makeFromDriverError(connection, this,
+                        form.format(new Object[] {typeName}), null, false);
+        }
+        Parameter param = setterGetParam(parameterIndex);
+        param.setDefineParameterTypeCalled(true);
+        param.setDefineParameterTypeSqlType(sqlType);
+        param.setValueLength(maxLength); // also sets userProvidesPrecision = true
+    }
```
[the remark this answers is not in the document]
muskan124947: Summary of changes in this session:

Dedicated defineParameterTypeLengthHint field — The length hint from defineParameterType() is now stored in its own field, separate from valueLength/userProvidesPrecision, preventing the hint from being misinterpreted as DECIMAL precision.

Setter-time type family enforcement — validateSetterTypeFamily() in Parameter.setValue() throws immediately if a setter's JDBC type family (character vs binary) conflicts with the defineParameterType declaration.

Removed redundant execution-time check — validateDeclaredTypeFamily() in GetTypeDefinitionOp was deleted since the setter-time check already guarantees the family is correct before execution is reached.

AE bypass — Length-hint validation is skipped entirely when Always Encrypted is active, since AE requires exact type info from the actual value.

docs/parameter-hints/parameter-hints-overview.md
```diff
@@ -0,0 +1,196 @@
+# Parameter Length Hints for Variable-Length Types
```
divang, line 1: READEME.md is very generic, can we have feature specific file name.

docs/parameter-hints/define-parameter-type.md
```diff
@@ -0,0 +1,559 @@
+# `defineParameterType()` — Primary Parameter Length Hint API
```
divang, line 1: As discussed, please check the placement of these md files. Is it good fit for doc kind of folder..

src/main/java/com/microsoft/sqlserver/jdbc/ISQLServerPreparedStatement.java
```diff
@@ -110,6 +110,90 @@ void setObject(int parameterIndex, Object x, SQLType targetSqlType, Integer prec
...
+     * @throws SQLServerException
+     *                            if {@code parameterIndex} is out of range,
+     *                            {@code maxLength} is non-positive, {@code sqlType}
+     *                            is not a
+     *                            supported type, this method is called on a closed
+     *                            statement, or execution detects a value length
+     *                            that exceeds {@code maxLength} for supported short
+     *                            character/binary parameter types
+     */
+    void defineParameterType(int parameterIndex, int sqlType, int maxLength) throws SQLServerException;
```
divang, line 196: This is a brand-new public API method on a public interface. Should we add @ since <version> tag (e.g. @ since 13.5.1, matching the milestone)?

src/main/java/com/microsoft/sqlserver/jdbc/ISQLServerPreparedStatement.java
```diff
@@ -110,6 +110,90 @@ void setObject(int parameterIndex, Object x, SQLType targetSqlType, Integer prec
...
+    /**
+     * Specifies the SQL type and maximum character or byte length for a parameter,
+     * allowing the driver to use a
+     * tighter type declaration than the default (e.g. {@code varchar(50)} instead
+     * of {@code varchar(8000)}). This
+     * enables SQL Server to compute a more accurate memory grant for query
+     * execution plans, which is especially
+     * beneficial for large batch operations.
+     *
+     * Important: The {@code sqlType} argument identifies the intended type
```
divang, line 122: Wrap as < b >Important:< /b > / < b >Precedence:< /b > ?

muskan124947: In our repo, i have seen this pattern of html tags are not followed everywhere.
I see this added in 2-3 files only. so have kept it similar to others

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerCallableStatement.java
```diff
@@ -1988,19 +1988,55 @@ public void setObject(String parameterName, Object value, int sqlType) throws SQ
...
                 null, false, findColumn(parameterName), null);
         if (loggerExternal.isLoggable(Level.FINER)) {
             loggerExternal.exiting(getClassNameLogging(), "setObject");
         }
     }
 
+    /**
+     * {@inheritDoc}
+     *
+     * Breaking Behavioral Change: For character and binary target types (VARCHAR, CHAR,
```
divang, line 2035: In my opinion, we should avoid introducing breaking changes to an existing public API, as it could lead to unintended issues for customers. Instead, we could consider an alternative approach ...
 
keep this existing setObject(..., sqlType, decimals) overload backward-compatible, and put the new max-length behavior behind the new explicit API.
 
stmt.defineParameterType(1, java.sql.Types.VARCHAR, 50);
stmt.setObject(1, value, java.sql.Types.VARCHAR);
 
This is explicit. The caller is clearly saying “declare this parameter as varchar(50).” If the value is longer than 50, enforcing that is reasonable.

src/main/java/com/microsoft/sqlserver/jdbc/Parameter.java
```diff
@@ -63,6 +63,107 @@ final CryptoMetadata getCryptoMetadata() {
...
     private boolean forceEncryption = false;
 
+    // Set to true when defineParameterType() has been called for this parameter.
+    // When true, defineParameterTypeLengthHint holds the caller-supplied max-length hint
+    // and the type definition (e.g. varchar(N)) is built from that hint rather than the
+    // conservative driver default (e.g. varchar(8000) / nvarchar(4000) / varbinary(8000)).
+    // This is intentionally separate from valueLength/userProvidesPrecision, which are used
+    // by DECIMAL/NUMERIC precision logic and would misinterpret a character/binary length
+    // hint as numeric precision.
+    private boolean defineParameterTypeCalled = false;
```
machavan: check if this can be removed

machavan (approved):

Ananya2 (approved):

PR #2961, opened 2026-05-26
Improve JAAS configuration handling in Kerberos authentication

divang:
## Description

This PR hardens Kerberos authentication by adding defense-in-depth validation of the `java.security.auth.login.config` system property. It blocks remote JAAS config sources (JNDI injection vector, MSRC-117029) while preserving local file path behavior.

### Modified Files

- `src/main/java/com/microsoft/sqlserver/jdbc/KerbAuthentication.java`
  - Updated `initAuthInit()` to use `JaasConfiguration(null)` for the `useDefaultJaasConfig=true` path
  - Added `assertSafeJaasLoginConfigProperty()` to block remote schemes (http, https, ldap, rmi, ftp, etc.)
  - Added `failUnsafeJaasLoginConfig()` helper that logs at SEVERE and throws `SQLServerException`
  - Uses `java.net.URI` parsing to detect non-file schemes, with `java.nio.file.Paths.get()` fallback for values that aren't valid URIs
  - Handles JAAS `=` prefix syntax (forces single config source) by stripping before validation

- `src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResource.java`
  - Added `R_unsafeJaasLoginConfigProperty` error message

### Test Files

- `src/test/java/com/microsoft/sqlserver/jdbc/KerberosTest.java`
  - `testAssertSafeJaasLoginConfigProperty_allowsNullProperty()` — null/empty property passes
  - `testAssertSafeJaasLoginConfigProperty_rejectsRemoteUrls()` — blocks http/https/ldap/ftp
  - `testAssertSafeJaasLoginConfigProperty_allowsLocalPaths()` — permits file paths and file: URIs
  - `testMSRC117029_pocPayloadsBlocked()` — proves exact PoC payloads from the vulnerability report are blocked (including `=`-prefixed variants)
  - `testAssertSafeJaasLoginConfigProperty_rejectsMalformedRemoteUrls()` — catches malformed remote URLs that bypass URI parsing but would still be fetched by `java.net.URL`

## Key Behavior

- **Allowed**: local file paths (`/etc/jaas.conf`, `C:\config\jaas.conf`), `file:` URIs, `=`-prefixed local paths
- **Blocked**: any non-file remote scheme (`http://`, `https://`, `ldap://`, `rmi://`, `ftp://`, etc.)
- **Fail-closed**: if `SecurityManager` denies reading the property, throws rather than proceeding blind
- **Fail-fast**: throws `SQLServerException` immediately — no retries, no network activity

## Benefits

1. **JNDI Injection Prevention**: Remote JAAS configs can declare arbitrary LoginModules (e.g. JndiLoginModule) triggering JNDI/LDAP lookups during login, leading to RCE
2. **Defense in Depth**: Path-based fallback catches malformed URLs that bypass URI parsing but may still be opened by JAAS `ConfigFile` (which uses `java.net.URL` internally)
3. **No Deprecated APIs**: Uses `java.net.URI` + `java.nio.file.Paths` instead of deprecated `java.net.URL(String)`
4. **JAAS Syntax Aware**: Handles the `=` prefix that forces Java to use only the specified config source

## Testing

- Unit tests validate accept/reject behavior via reflection on the package-private method
- Tests cover exact MSRC-117029 PoC payloads to prove the vulnerability is mitigated
- All existing Kerberos integration tests pass without modification

## Backward Compatibility

- Standard local JAAS config files remain fully supported (absolute/relative paths, UNC paths, file: URIs)
- `useDefaultJaasConfig=true` continues to work with built-in `JaasConfiguration`
- No changes to connection string properties or public API

## Backward Compatibility

Fully backward compatible
- Existing connection strings work without changes
- Standard Krb5LoginModule authentication remains fully supported
- No changes to public API or connection properties

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/KerbAuthentication.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/KerbAuthentication.java
@@ -13,6 +13,7 @@ import java.text.MessageFormat;
 import java.util.logging.Level;
 
 import javax.security.auth.Subject;
+import javax.security.auth.login.AppConfigurationEntry;
 import javax.security.auth.login.Configuration;
 import javax.security.auth.login.LoginContext;
 import javax.security.auth.login.LoginException;
@@ -107,11 +108,14 @@ final class KerbAuthentication extends SSPIAuthentication {
                     }
 
                     if (null == currentSubject) {
-                        if (useDefaultJaas) {
-                            lc = new LoginContext(configName, null, callback, new JaasConfiguration(null));
-                        } else {
-                            lc = new LoginContext(configName, callback);
-                        }
+                        // SECURITY FIX: Always use JaasConfiguration with null delegate
+                        // This prevents loading remote JAAS configs from system property
+                        JaasConfiguration secureConfig = new JaasConfiguration(null);
+
+                        // SECURITY FIX: Validate no JndiLoginModule present
+                        validateNoJndiLoginModule(secureConfig, configName);
+
+                        lc = new LoginContext(configName, null, callback, secureConfig);
                         lc.login();
                         // per documentation LoginContext will instantiate a new subject.
                         currentSubject = lc.getSubject();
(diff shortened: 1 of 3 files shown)


machavan (dismissed):

src/test/java/com/microsoft/sqlserver/jdbc/KerberosTest.java
```diff
@@ -132,6 +132,196 @@ public void testJaasConfigurationConstructor() {
...
+        java.lang.reflect.Method assertMethod = kerbAuthClass.getDeclaredMethod("assertSafeJaasLoginConfigProperty");
+        assertMethod.setAccessible(true);
+
+        // Exact payloads from the MSRC-117029 PoC reproduction scripts
+        String[] pocPayloads = {
+                "http://remote.example.com/evil.jaas",
+                "=http://remote.example.com/evil.jaas",
+                "https://remote.example.com/evil.jaas",
+                "ldap://remote.example.com/cn=Evil",
+                "rmi://remote.example.com/Exploit"
```
machavan, line 221: lets add few positive scenarios of valid FS paths

machavan (dismissed):

src/test/java/com/microsoft/sqlserver/jdbc/KerberosTest.java
```diff
@@ -132,6 +132,238 @@ public void testJaasConfigurationConstructor() {
...
+    }
+
+    /**
+     * Test that assertSafeJaasLoginConfigProperty allows mounted drive paths,
+     * UNC paths, mapped network drives, relative paths, paths with spaces,
+     * and paths using the JAAS '=' prefix syntax.
+     */
+    @Test
+    public void testAssertSafeJaasLoginConfigPropertyAllowsMountedAndPositiveFilePaths() throws Exception {
+        Class<?> kerbAuthClass = Class.forName("com.microsoft.sqlserver.jdbc.KerbAuthentication");
```
muskan124947: every test is repeating this
can we extract it to BeforeAll or helper function

src/main/java/com/microsoft/sqlserver/jdbc/KerbAuthentication.java
```diff
@@ -178,6 +185,79 @@ private void initAuthInit() throws SQLServerException {
...
+     * Rejects values of the {@code java.security.auth.login.config} system property that do not
+     * resolve to a local filesystem file. Prevents RCE via attacker-controlled JAAS configuration
+     * (e.g. a remote jaas.conf declaring JndiLoginModule, fetched through http/https/ldap/jar/rmi
+     * or any custom URLStreamHandler).
+     *
+     * <p>Allow-list: unset/empty values, anything parseable as a local {@link java.nio.file.Path}
+     * (including Windows drive-letter paths and UNC paths), and {@code file:} URIs that resolve
+     * to a {@code Path}. Everything else is rejected.
+     */
+    private static void assertSafeJaasLoginConfigProperty() throws SQLServerException {
```
muskan124947, line 198: The assertSafeJaasLoginConfigProperty() method has a gap when new URI(value) throws URISyntaxException (e.g., spaces in URL). It falls through to Paths.get(value), which on Linux succeeds for values like http://evil.com/path with spaces/evil.jaas because colons and spaces are valid in Linux paths. This means malformed remote URLs pass validation on Linux — exactly the class of input the defense-in-depth test testAssertSafeJaasLoginConfigProperty_rejectsMalformedRemoteUrls tries to block.

machavan (approved):

Ananya2 (approved):

muskan124947 (approved):

PR #2963, opened 2026-05-29
Add useColumnTypeSizing connection property to size string/binary parameters to the target column (#2913)

### LINKED ISSUE #2913: Why does String paramater being passed to query go as nvarchar(4000) or varchar(8000)

dfn-itech:
# Add `useColumnTypeSizing` connection property to size string/binary parameters to the target column

Fixes #2913. Relates to #2960.

## Problem

When an application binds a variable-length string/binary parameter without an explicit size (`setString`, `setNString`, `setBytes`), the driver declares it at a fixed maximum bucket — `varchar(8000)` / `nvarchar(4000)` / `varbinary(8000)` — regardless of the value or the target column. This causes three independent production problems (listed by impact):

- **Lost partition elimination (length-driven) — the primary motivation.** On a table partitioned on a narrower `varchar(N)` column, a parameter declared `varchar(8000)` still **probes every partition** even though the plan is a Seek: the optimizer cannot statically resolve the over-wide value to a single partition, so it seeks *within each* partition. The query then runs far slower than it should because the plan does not use the partitioning (or a narrow index) effectively. Declaring the parameter at the **column's actual width** is what lets the optimizer eliminate to the one matching partition — and `sendStringParametersAsUnicode=false` alone does **not** achieve this. This is the case the property exists to fix.
- **Lost index seek (type-driven).** Under the default `sendStringParametersAsUnicode=true` a parameter is declared `nvarchar`; compared against a `varchar`/`char` column this forces an implicit convert **on the column** (type precedence), making the predicate non-SARGable → Scan instead of Seek. This one is fixed by the parameter *type* (`sendStringParametersAsUnicode=false`); the length is what then unlocks partition/index narrowing on top of the restored seek.
- **Memory-grant over-allocation.** SQL Server sizes a query's memory grant from the *declared* width (~half per row). An 8000-wide declaration reserves memory as if every row were ~4 KB → `RESOURCE_SEMAPHORE` waits, reduced concurrency, tempdb spills.

`varchar`/`char` columns suffer the seek/partition problems and the memory problem; `nvarchar`/`nchar` columns suffer **only** the memory-grant problem (an `nvarchar` parameter already matches a unicode column, so its seek is not lost).

## What this PR does

Adds an **opt-in** boolean connection property **`useColumnTypeSizing`** (default `false`). When enabled, for an unsized, non-output string/binary input parameter whose target column the server can resolve, the driver declares it sized to the **column's** length — discovered once per statement via `sp_describe_undeclared_parameters` and cached — instead of the fixed bucket.

Key safety properties:
- The declared SQL **type keyword is unchanged** (stays the runtime arm's `varchar`/`nvarchar`/`varbinary`), so the **TDS value encoding is identical** and query results cannot change.
- Only the declared **length** is tightened. If a bound value is longer than the column, the declaration **snaps to `(max)`** so the operand is never truncated and the plan cache stays bounded (≤ 2 value-independent declarations).
- Sizing to the **column** (not the value) keeps the prepared-statement / plan-cache key stable across executions — no per-value fragmentation.
- Default off ⇒ behavior is **byte-for-byte unchanged**.

Clean fallback to the legacy default for: range/expression/ambiguous parameters (the server returns the wide type for those), Always Encrypted connections (the AE precise-sizing path is preserved and inference is not attempted), the FMTONLY path, pre-2012 servers, and any metadata-lookup failure (best-effort, never propagated).

The round-trip runs **at most once per statement object** (independent of statement pooling); the inner `sp_describe` statement is flagged internal so it cannot recurse.

## Type vs. length — why both knobs matter

The two are independent. The **type** (`varchar` vs `nvarchar`) decides whether a `varchar`/`char` column gets a seek at all — restored by `sendStringParametersAsUnicode=false`. The **length** decides whether the optimizer can narrow that seek to the matching partition (or use a narrow index efficiently) — restored by `useColumnTypeSizing=true`. Setting only the type leaves the seek probing every partition; setting only the length (while still `nvarchar`) leaves a Scan. For a `varchar`/`char`-partitioned column you need **both**.

## Evidence (SQL Server 2022)

Canonical case `WHERE region_code = ?` against a `varchar(8)`-partitioned column, measured via the driver. The middle row is the key one: a correct `varchar` type that is still too **wide** seeks but **fails to eliminate partitions**.

| Mode | Declared | Plan | Partitions | Reads |
|------|----------|------|-----------|-------|
| default (today) | `nvarchar(4000)` | Clustered Index **Scan** | 13 (all) | 715 |
| `sendStringParametersAsUnicode=false` only (no this feature) | `varchar(8000)` | Clustered Index **Seek** | **13 (all!)** | — |
| `useColumnTypeSizing=true;sendStringParametersAsUnicode=false` | `varchar(8)` | Clustered Index **Seek** | **1** | **29** |

On a non-partitioned `varchar(17)` index, `nvarchar(4000)` did 1548 logical reads while both `varchar(8000)` and `varchar(17)` did 3 — so on a plain index the *type* restores the seek and the length only tightens the memory grant; the length becomes decisive once partitioning is involved. Memory grant for a parameter feeding a sort: **1,108,256 KB → 70,296 KB (~15.8×)** when `varchar(8000)` → `varchar(8)`. Repeated executions with different values reuse a **single** cached plan.

## Files

- New property end-to-end: `SQLServerDriver` (enum + `DRIVER_PROPERTIES`), `SQLServerConnection` (+ `ISQLServerConnection`, `SQLServerConnectionPoolProxy`), `SQLServerDataSource` (+ `ISQLServerDataSource`), `SQLServerResource`.
- Inference + sizing: `SQLServerPreparedStatement` (once-per-statement describe + apply, single & batch paths), `Parameter` (bind-time value capture, `cloneForBatch`, the new sizing rung in the three var-length arms), `SQLServerParameterMetaData` + `SQLServerStatement` (internal-query recursion guard).
- Tests: `ColumnTypeSizingTest` (24 tests: per-type sizing incl. CHAR/NCHAR/NVARCHAR/VARBINARY/LOB, SSPAU true vs false, clamp/snap-to-max + exact-length boundary, range no-truncation, non-string fallback, SELECT, explicit-column INSERT + batch INSERT persistence, **partition elimination requires the column-sized length** (the primary-motivation scenario), UPDATE (SET + WHERE) and DELETE (incl. over-length DELETE that snaps to max so no wrong row is deleted), re-execution snap-to-max transition, multi-parameter ordinal alignment, declaration stability, NULL, no-regression off, and Connection/DataSource accessors).
- Docs: `CHANGELOG.md`, `.github/instructions/use-column-type-sizing.md`.

## Compatibility

Opt-in, default off. No public API removed. Existing prepared-statement tests pass unchanged.


dfn-itech:
@microsoft-github-policy-service agree company="icaria Technology"

muskan124947:
@dfn-itech , Thank you for contributing.
Closing this PR. 
Will be taking this forward as discussed below :
[Introduce parameter length hints via defineParameterType and setObject #2960](https://github.com/microsoft/mssql-jdbc/pull/2960)
[Why does String paramater being passed to query go as nvarchar(4000) or varchar(8000) #2913](https://github.com/microsoft/mssql-jdbc/issues/2913)

PR #2964, opened 2026-06-01
Fix #2957: route enclave AE CEK lookup through SQLServerSymmetricKeyCache

### LINKED ISSUE #2957: Performance: Always Encrypted enclave queries always hit Azure Key Vault — provider CEK cache disabled at registration, enclave path bypasses SQLServerSymmetricKeyCache
## Driver version

`13.4.0.jre11` (latest stable as of writing). Bug also confirmed by source inspection in `12.6.4`, `12.10.1`, and `13.2.0` — relevant lines appear at identical positions in all four releases.

## SQL Server version

```
Microsoft SQL Azure (RTM) - 12.0.2000.8
    Apr 14 2026 20:27:12
    Copyright (C) 2025 Microsoft Corporation
```

## Client Operating System

```
Linux version 7.0.3-arch1-2 (linux@archlinux) (gcc (GCC) 16.1.1 20260430, GNU ld (GNU Binutils) 2.46.0) #1 SMP PREEMPT_DYNAMIC Fri, 01 May 2026 15:49:22 +0000
```

The bug is not OS-specific; it is a logic bug in the driver's CEK-cache management.

## JAVA/JVM version

Eclipse Temurin 21.0.11.

## Table schema

Minimal schema. The encrypted column must be `RANDOMIZED` with a `BIN2` collation so that predicates engage the enclave path.

```sql
CREATE COLUMN MASTER KEY [ReproCMK]
WITH (
    KEY_STORE_PROVIDER_NAME = N'AZURE_KEY_VAULT',
    KEY_PATH = N'https://<vault>.vault.azure.net/keys/<keyname>/<version>',
    ENCLAVE_COMPUTATIONS (SIGNATURE = 0x<from New-SqlAzureKeyVaultColumnMasterKeySettings>)
);

CREATE COLUMN ENCRYPTION KEY [ReproCEK]
WITH VALUES (
    COLUMN_MASTER_KEY = [ReproCMK],
    ALGORITHM = 'RSA_OAEP',
    ENCRYPTED_VALUE = 0x<from New-SqlColumnEncryptionKey>
);

CREATE TABLE dbo.EnclaveCachingRepro (
    Id          INT IDENTITY(1, 1) PRIMARY KEY,
    Description NVARCHAR(200) COLLATE Latin1_General_BIN2
        ENCRYPTED WITH (
            ENCRYPTION_TYPE = RANDOMIZED,
            ALGORITHM       = 'AEAD_AES_256_CBC_HMAC_SHA_256',
            COLUMN_ENCRYPTION_KEY = [ReproCEK]
        ) NOT NULL
);
```

Connection uses `columnEncryptionSetting=Enabled` with an enclave attestation URL/protocol. The Azure Key Vault provider is registered globally via `SQLServerConnection.registerColumnEncryptionKeyStoreProviders` — `keyVaultProviderClientId` / `keyVaultProviderClientKey` are *not* set on the JDBC URL (the recommended pattern per the Always-Encrypted-with-secure-enclaves docs).

## Problem description

**This is a performance bug, not a functional one. No exception is thrown.** In any Always Encrypted + secure enclaves workload that uses the recommended global-registration pattern, every enclave query with an encrypted predicate pays a full set of Azure Key Vault round-trips, indefinitely. The Azure provider's internal `SimpleTtlCache` is permanently disabled, and the enclave query path bypasses `SQLServerSymmetricKeyCache`, so there is no functioning CEK cache for enclave queries at all.

Four locations in the driver source combine to produce the bug. Line numbers are stable across `12.6.4`, `12.10.1`, `13.2.0`, and `13.4.0`:

**1. `SQLServerConnection.java:1511` — registration-time poison.**

In `registerColumnEncryptionKeyStoreProviders`:

```java
// Global providers should not use their own CEK caches.
provider.setColumnEncryptionCacheTtl(Duration.ZERO);
globalCustomColumnEncryptionKeyStoreProviders.put(providerName, provider);
```

Fires the moment an application calls `SQLServerConnection.registerColumnEncryptionKeyStoreProviders` — before any query has run. Disables the provider's internal CEK cache.

**2. `SQLServerSymmetricKeyCache.java:100` — per-query poison.**

In `getKey()`, inside the cache-miss branch:

```java
/*
 * When provider decrypt Column Encryption Key, it can cache the decrypted key if cacheTTL > 0.
 * To prevent conflicts between CEK caches, system providers and global providers should not use their own CEK caches.
 */
provider.setColumnEncryptionCacheTtl(Duration.ZERO);
plaintextKey = provider.decryptColumnEncryptionKey(...);
```

Re-fires the same call on every symmetric-key-cache miss. Redundant with `#1` in steady state, but it re-poisons the provider's cache if anything restores the TTL.

**3. `ISQLServerEnclaveProvider.java:262` — enclave path routes through site `#2` for the parameter CEK.**

In `processSDPEv1`, for each encrypted query parameter:

```java
SQLServerSecurityUtility.decryptSymmetricKey(params[paramIndex].cryptoMeta, connection, statement);
```

`decryptSymmetricKey` calls `SQLServerSymmetricKeyCache.getInstance().getKey(...)`, which re-fires the line-100 poison on every cold-cache lookup. **So even pure-enclave applications, which one might expect to never reach the standard path, trigger the per-query poison whenever they have an encrypted parameter.**

**4. `ISQLServerEnclaveProvider.java:229` — enclave path bypasses the symmetric cache for the enclave's CEK.**

```java
aev2CekEntry.put(provider.decryptColumnEncryptionKey(keyPath, algo, encryptedKey));
```

Direct call to the provider with no caching layer. Once the provider's own cache is disabled, every enclave query pays a full Key Vault round-trip here for the CEK that will be passed to the enclave.

The combination is doubly-locked-in: the cache is disabled at registration *and* re-disabled by every cold cache miss for any encrypted parameter. Manual TTL restoration after registration is undone again on the next enclave query.

## Expected behavior

The Azure provider's `SimpleTtlCache` (default TTL: 2 hours) should serve subsequent enclave queries from cache. Key Vault round-trips for a given CEK should happen once per ~2 hours, not once per query. Steady-state per-query latency should be local-only (parameter encryption + plumbing, no Key Vault traffic).

## Actual behavior

Every enclave query with an encrypted parameter pays ~3–4 HTTPS round-trips to Azure Key Vault:

- 1 × `KeyGet` (sometimes 2; Azure SDK `CryptographyClient` lazy-fetches the key for version validation before each crypto op)
- 1 × `KeyVerify` (RS256 signature check on the encrypted CEK)
- 1 × `KeyUnwrap` (RSA-OAEP unwrap of the CEK)

This translates to several hundred milliseconds of added latency per query in any non-co-located deployment, and consumes ~3–4 ops per query from the Key Vault per-vault-per-region rate-limit budget.

## Error message/stack trace

None. This is a silent performance regression — there is no exception. Under sustained load above the Key Vault throttling threshold, the driver eventually surfaces HTTP `429` as a `SQLServerException`, but at that point the cause (cache disabled) is far removed from the symptom (rate-limit failure).

### Customer impact

1. **Latency.** ~300–500 ms per enclave query in observed environments. For interactive applications this is user-visible.
2. **Throttling exposure.** From [Key Vault service limits](https://learn.microsoft.com/en-us/azure/key-vault/general/service-limits): RSA 2048 software keys get 4,000 ops/10 s per vault per region; RSA 2048 HSM keys get 2,000/10 s; RSA 4096 HSM keys get just 250/10 s. With this bug consuming 3–4 ops per enclave query, an application using HSM-protected RSA 4096 keys hits the throttling ceiling at ~6 enclave queries/sec per vault. Above that, requests get `429`.
3. **Operating cost.** Azure Key Vault charges per operation. The bug multiplies billable KV traffic by 3–4 × the enclave-query rate where a working cache would have collapsed it to near zero.

### Scope

Affects every application using:

- `SQLServerConnection.registerColumnEncryptionKeyStoreProviders` (the recommended global registration pattern)
- Always Encrypted with secure enclaves
- Any enclave-using query (i.e. a predicate against a randomized-encrypted column with `ENCLAVE_COMPUTATIONS`)

This includes **pure-enclave workloads** — the registration-time poison fires regardless of query mix.

### Caches not affected

The Azure provider's CMK signature verification cache (`cmkMetadataSignatureVerificationCache`, default TTL 10 days) is *not* touched by either poison call and continues to function. So the bug affects only the decrypt path; `verifyColumnMasterKeyMetadata` calls usually serve from cache.

### Minimal reproducer

A self-contained Maven project that demonstrates the bug end-to-end is at <https://github.com/drabikp/enclave-caching-bug-reproducer>.

The reproducer instruments two observation layers:

- **JDBC-provider counters** (`decryptColumnEncryptionKey`, `verifyColumnMasterKeyMetadata`, `setColumnEncryptionCacheTtl`) — counts at the registered provider level.
- **HTTP-level counters** via an Azure SDK `HttpClientProvider` SPI — counts actual Azure Key Vault HTTPS requests by category (`key-get`, `unwrapkey`, `verify`, `sign`, `oauth-token`).

Three measured phases isolate the cost components:

```
=== Phase 1: Registration-time poison (SQLServerConnection.java:1511) ===
  Azure delegate TTL before registration: PT2H
  Azure delegate TTL after registration:  PT0S  <-- poison fired
  Manually restored TTL to PT2H so phases 2a/2b can measure cache-working behavior.

=== Phase 2a: SQL round-trip floor (non-encrypted query) ===
  Query: SELECT 1
  Measured 100 iterations
  Provider counters: Counts[decrypt=0, encrypt=0, verify=0, setTtl=0]
  KV HTTP counts:    {}
  Total elapsed: 9370.0 ms (avg 93.7 ms per query)

=== Phase 2b: AE non-enclave query (caching DOES work for this) ===
  Query: SELECT Description FROM ... WHERE Id = ?
  Azure delegate TTL at phase start: PT2H
  After 2 warmup iteration(s):
    Provider counters: Counts[decrypt=1, encrypt=0, verify=0, setTtl=1]
    KV HTTP counts:    {key-get=6, unwrapkey=2, verify=2}
    Azure delegate TTL: PT0S  <-- per-query poison fired during warmup
  Measured 100 iterations
  Provider counters: Counts[decrypt=0, encrypt=0, verify=0, setTtl=0]
  KV HTTP counts:    {key-get=0, unwrapkey=0, verify=0}
  Total elapsed: 9085.6 ms (avg 90.9 ms per query)

=== Phase 3: Enclave query (the bug) ===
  Query: SELECT Id FROM ... WHERE Description LIKE ?
  Azure delegate TTL at phase start: PT0S (was poisoned by phase 2b — not restored)
  Measured 100 iterations
  Provider counters: Counts[decrypt=100, encrypt=0, verify=100, setTtl=0]  <-- every query bypassed all caches
  KV HTTP counts:    {key-get=200, sign=0, unwrapkey=200, verify=200}
  Total elapsed: 45425.0 ms (avg 454.3 ms per query)

=== Summary ===
  Phase 2a — SQL floor (non-encrypted):     avg 93.7 ms per query
  Phase 2b — AE non-enclave (cache works):  avg 90.9 ms per query  (-2.8 ms vs 2a: AE driver overhead is ~0)
  Phase 3  — AE enclave (the bug):          avg 454.3 ms per query  (+363.4 ms vs 2b: enclave-bypass KV cost)
```

The contrast between phase 2b and phase 3 is the smoking gun: same provider, same CEK, same JVM. Phase 2b's measured loop emits zero Key Vault HTTPS traffic (the driver's symmetric key cache absorbs everything after the warmup miss). Phase 3's measured loop emits 600 HTTPS round-trips to Key Vault for 100 enclave queries — exactly the per-iteration cost predicted by site `#4` firing on every query with no cache underneath.

The ~363 ms of per-query overhead in phase 3 is entirely Key Vault round-trip cost. It is invariant to SQL Server latency (the slowdown factor depends on the SQL floor; the bug-attributable overhead does not).

### Diagnosability

This bug is silent at every JDBC trace level (verified at `FINER`). In a 5,860-line trace covering the reproducer's full run, there are **zero** log entries containing `setColumnEncryptionCacheTtl`, `decryptColumnEncryptionKey`, `verifyColumnMasterKeyMetadata`, or any reference to the provider's CEK cache being disabled. `SQLServerSymmetricKeyCache.getKey` emits `Checking trusted master key path...` and `Checking Symmetric key cache...` around the poison call, but the poison call itself produces no log output. The Azure Key Vault provider's methods are not traced at `FINER` either. The only signal that 2+ seconds of latency was a Key Vault round-trip is the wall-clock gap between two adjacent log lines.

**A trivial fix that would help all customers**, independent of the cache-bug fix itself: add `aeLogger.fine(...)` next to both `setColumnEncryptionCacheTtl(Duration.ZERO)` call sites, and add `aeLogger.fine(...)` entries inside `SQLServerColumnEncryptionAzureKeyVaultProvider.decryptColumnEncryptionKey` / `verifyColumnMasterKeyMetadata` / `setColumnEncryptionCacheTtl`. This would surface the bug — and any similar future bug — in standard JDBC trace output.

### Possible fix shape

Routing the line-229 CEK lookup through `SQLServerSymmetricKeyCache.getKey(...)` — the same path the standard query path and `ISQLServerEnclaveProvider.java:262` already use — would let both existing poison calls remain in place while restoring caching for enclave queries. The driver's original comments (*"Global providers should not use their own CEK caches"* / *"To prevent conflicts between CEK caches"*) become correct again because `SQLServerSymmetricKeyCache` is then the sole cache layer by design. A one-line behavior change in `processSDPEv1`, plus the bonus of `setTrustedMasterKeyPaths` enforcement also covering the enclave path (currently it does not). Lock contention on `SQLServerSymmetricKeyCache.getKey`'s global `ReentrantLock` is a follow-up consideration, not a blocker.

## JDBC trace logs

[mssql-jdbc-trace.log](https://github.com/user-attachments/files/27799097/mssql-jdbc-trace.log)

Attached: `mssql-jdbc-trace.log`, generated by the reproducer with `repro.captureJdbcTrace=true`, FINER level on the `com.microsoft.sqlserver.jdbc` logger. Captured against driver version `13.4.0.jre11`.

machavan:
## Problem description

Always Encrypted queries that target columns with `ENCLAVE_COMPUTATIONS` (AE v2) issue an Azure Key Vault (or other CMK store) round-trip on **every** execution, even when the same column encryption key (CEK) was just resolved milliseconds earlier. With AKV this adds ~150–400 ms per query and consumes vault throttling quota; under load it dominates query latency and surfaces as `429 Too Many Requests` from AKV.

The non-enclave AE path does not have this problem because it goes through `SQLServerSymmetricKeyCache` via `SQLServerSecurityUtility.decryptSymmetricKey`. The enclave path in `ISQLServerEnclaveProvider.processSDPEv1` instead resolves the enclave-requested CEK by calling `provider.decryptColumnEncryptionKey(...)` directly, bypassing the cache.

In addition, when a CMK provider is registered system-wide or globally (the recommended pattern for AKV), the provider's internal `SimpleTtlCache` is intentionally disabled — `SQLServerSymmetricKeyCache` is meant to be the single caching layer. With the enclave path bypassing that single layer, no cache is in effect for enclave queries at all.

Fixes #2957

## Fix Description

`ISQLServerEnclaveProvider.processSDPEv1` ("Site 4") now resolves the enclave-requested CEK through the same path as the non-enclave code:

- Build an `EncryptionKeyInfo` from the `sp_describe_parameter_encryption` row.
- When `connection.getServerColumnEncryptionVersion()` indicates instance-level providers are in scope (`shouldUseInstanceLevelProviderFlow`), call `SQLServerSecurityUtility.getKeyFromLocalProviders(...)`.
- Otherwise, call `SQLServerSymmetricKeyCache.getInstance().getKey(keyInfo, connection)`.
- Pass `symKey.getRootKey()` into `aev2CekEntry.put(...)` for delivery to the enclave.

Effects:

- Repeat enclave queries reuse the cached plaintext CEK; AKV is hit once per `(server, encrypted-CEK, key-store)` triple within the cache TTL.
- The enclave path now also goes through `setTrustedMasterKeyPaths` enforcement, closing a minor inconsistency with the non-enclave path.
- No public API change. No behavior change for callers that were not using AE v2.

Test coverage added:

- `src/test/java/com/microsoft/sqlserver/jdbc/EnclaveCacheUnitTest.java` — pure unit test, no SQL Server required. Uses Mockito to drive `processSDPEv1` with a fabricated `sp_describe_parameter_encryption` result set and a counting CMK provider.
  - `nonEnclavePath_hitsCacheAfterFirstLookup`: 50 lookups of the same wrapped CEK → asserts the provider's `decryptColumnEncryptionKey` is called exactly once.
  - `enclavePath_hitsCacheAfterFirstLookup`: 25 invocations of `processSDPEv1` for the same enclave-requested CEK → asserts the provider's `decryptColumnEncryptionKey` is called exactly once. Without this fix the same test produces ~26 calls.

Verified locally with `mvn -P jre25 -Dtest=EnclaveCacheUnitTest test` — both tests pass.

## New Public APIs

None.

## Performance Validation

Tested against Azure SQL Database with VBS enclave, AKV-backed CMK, randomized encryption. Query: `SELECT * FROM dbo.TestEnclave WHERE SSN LIKE ?` (enclave-required LIKE on randomized column).

| Run | Without fix (ms) | With fix (ms) |
|-----|-----------------|---------------|
| 0   | 3426            | 1763          |
| 1   | 1992            | 415           |
| 2   | 2503            | 411           |
| 3   | 1953            | 413           |
| 4   | 2056            | 413           |
| 5   | 1926            | 413           |
| 6   | 2662            | 412           |
| 7   | 2458            | 411           |
| 8   | 1944            | 421           |
| 9   | 2380            | 412           |

Without the fix, every run pays the AKV round-trip (~2000+ ms). With the fix, only Run 0 hits AKV; Run 1+ serve from cache (~415 ms, matching non-enclave baseline).

| Metric | Without fix | With fix | Improvement |
|--------|-------------|----------|-------------|
| Avg time (Run 1–9) | ~2200 ms | ~413 ms | **~5.3x faster** |
| AKV calls per 10 queries | 10 | 1 | **90% reduction** |

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/ISQLServerEnclaveProvider.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/ISQLServerEnclaveProvider.java
@@ -224,9 +224,18 @@ interface ISQLServerEnclaveProvider {
                 aev2CekEntry.put(mdVer);
                 aev2CekEntry.putShort((short) keyID);
 
-                SQLServerColumnEncryptionKeyStoreProvider provider = SQLServerSecurityUtility
-                        .getColumnEncryptionKeyStoreProvider(keyStoreName, connection, statement);
-                aev2CekEntry.put(provider.decryptColumnEncryptionKey(keyPath, algo, encryptedKey));
+                // Issue #2957: route the enclave CEK lookup through SQLServerSymmetricKeyCache
+                // (or the local-provider flow for per-connection/per-statement providers) so that
+                // enclave queries don't hit the CMK key store on every call. Mirrors the lookup
+                // SQLServerSecurityUtility.decryptSymmetricKey performs for non-enclave params.
+                EncryptionKeyInfo keyInfo = new EncryptionKeyInfo(encryptedKey, dbID, keyID,
+                        rs.getInt(DescribeParameterEncryptionResultSet1.KEYVERSION.value()), mdVer, keyPath,
+                        keyStoreName, algo);
+                SQLServerSymmetricKey symKey = SQLServerSecurityUtility
+                        .shouldUseInstanceLevelProviderFlow(keyStoreName, connection, statement)
+                                ? SQLServerSecurityUtility.getKeyFromLocalProviders(keyInfo, connection, statement)
+                                : SQLServerSymmetricKeyCache.getInstance().getKey(keyInfo, connection);
+                aev2CekEntry.put(symKey.getRootKey());
                 enclaveRequestedCEKs.add(aev2CekEntry.array());
             }
         }
--- a/src/test/java/com/microsoft/sqlserver/jdbc/EnclaveCacheUnitTest.java
+++ b/src/test/java/com/microsoft/sqlserver/jdbc/EnclaveCacheUnitTest.java
@@ -8,6 +8,9 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.util.ArrayList;
 import java.util.Collections;
@@ -15,7 +18,11 @@ import java.util.concurrent.atomic.AtomicInteger;
 
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
+
+import com.microsoft.sqlserver.testframework.Constants;
 
 
 /**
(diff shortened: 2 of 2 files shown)


Ananya2 (dismissed):

divang (dismissed):

David-Engel (approved):
Nice! I bet that hidden gem was fun to find. 🙂

Ananya2 (approved):

PR #2965, opened 2026-06-03
Introduce getCurrentUserSql and getCurrentStatementType in PerformanceLogCallback

muskan124947:
### Motivation

Performance monitoring consumers need visibility into **which SQL statement** triggered a performance event and **what type of statement** it was. Previously, the `publish()` callback only provided `activity`, `connectionId`, `statementId`, `duration`, and `exception` — making it impossible to correlate performance data back to specific application queries without maintaining external state.

### What This PR Does

Adds two `default` methods to the public `PerformanceLogCallback` interface that expose SQL text and statement type during the `publish()` callback window:

- **`getCurrentUserSql()`** — returns the SQL text submitted by the application
- **`getCurrentStatementType()`** — returns a Statement type enum -> `"Statement"`, `"PreparedStatement"`, or `"CallableStatement"`

These are only valid inside a `publish()` invocation and return `null` otherwise.

### Benefits

- **Zero breaking changes** — default methods on existing interface; existing consumers unaffected
- **Pull-based** — consumers only retrieve data they need; no extra parameters on `publish()`
- **Zero-copy** — same String reference flows through; no `new String()` allocations
- **Minimal overhead** — ThreadLocal set/remove only when a callback is registered (a few nanoseconds)
- **Secure lifecycle** — data only exists during `publish()` scope, cleared in `finally` block

### Consumer Usage

```java
    @Override
    public void publish(PerformanceActivity activity, int connectionId, int statementId,
                        long duration, Exception exception) {
        String sql = getCurrentUserSql();        // e.g. "SELECT ? FROM users WHERE id = ?"
        Statement type = getCurrentStatementType(); // e.g. StatementType.PREPARED_STATEMENT
        logger.info("[{}] {} | {} | {}ms", type, activity, sql, duration);
    }

    @Override
    public void publish(PerformanceActivity activity, int connectionId, long duration,
                        Exception exception) {
        // Connection-level: getCurrentUserSql() returns null
    }
});
```

### File Changes

| File | Change |
|------|--------|
| `src/main/java/com/microsoft/sqlserver/jdbc/PerformanceLog.java` | Added package-private ThreadLocals (`currentUserSql`, `currentStatementType`); consolidated to single statement-level constructor; set/remove ThreadLocals around `publish()` in `close()`; added `deriveStatementType()` helper |
| `src/main/java/com/microsoft/sqlserver/jdbc/PerformanceLogCallback.java` | Added `default getCurrentUserSql()` and `default getCurrentStatementType()` methods |
| `src/main/java/com/microsoft/sqlserver/jdbc/SQLServerStatement.java` | Updated 3 call sites (`doExecuteStatement`, `doExecuteStatementBatch`, `doExecuteCursored`) to pass `this` + SQL; sub-activities pass `null, null` |
| `src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java` | Updated 4 call sites (`doExecutePreparedStatement`, `doPrep`, `executeBatchInternal` x2) to pass `this` + `userSQL` |
| `.github/instructions/performance-metrics.instructions.md` | Documented `getCurrentUserSql()` / `getCurrentStatementType()` usage and behavior |
| `src/main/java/com/microsoft/sqlserver/jdbc/StatementType.java` | New enum with STATEMENT , PREPARED_STATEMENT , CALLABLE_STATEMENT values |

### Testing

Added the following tests to `PerformanceLogCallbackTest.java`:

| Test | Validates |
|------|-----------|
| `testUserSqlAndStatementTypeInCallback` | SQL and type reported correctly for Statement and PreparedStatement |
| `testUserSqlInBatchExecution` | SQL captured during `executeBatch()` for PreparedStatement |
| `testUserSqlConsistentAcrossPreparedStatementReuse` | Same SQL across sp_executesql → sp_prepexec → sp_execute |
| `testUserSqlNullForConnectionLevelActivities` | Returns `null` in connection-level `publish()` |
| `testUserSqlNullForSubActivities` | Returns `null` for REQUEST_BUILD / FIRST_SERVER_RESPONSE |
| `testConcurrentConnectionsReceiveCorrectSql` | 100 connections firing simultaneously — callback receives correct SQL per thread, verified via log file |
| `testSingleConnectionMultipleThreadsReceiveCorrectSql` |  Single connection, 100 threads — callback SQL isolation verified via thread markers in log |

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/PerformanceLog.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/PerformanceLog.java
@@ -22,7 +22,7 @@ class PerformanceLog {
 
     // ThreadLocal to hold current SQL text and statement type for the duration of a publish callback
     static final ThreadLocal<String> currentUserSql = new ThreadLocal<>();
-    static final ThreadLocal<String> currentStatementType = new ThreadLocal<>();
+    static final ThreadLocal<StatementType> currentStatementType = new ThreadLocal<>();
 
     /**
      * Register a callback for performance log events.
@@ -153,15 +153,15 @@ class PerformanceLog {
         return new Scope(logger, connectionId, statementId, stmt, userSql, activity);
     }
 
-    // Helper method to derive statement type string based on the statement class
-    private static String deriveStatementType(SQLServerStatement stmt) {
+    // Helper method to derive statement type based on the statement class
+    private static StatementType deriveStatementType(SQLServerStatement stmt) {
         if (stmt instanceof SQLServerCallableStatement) {
-            return "CallableStatement";
+            return StatementType.CALLABLE_STATEMENT;
         }
         if (stmt instanceof SQLServerPreparedStatement) {
-            return "PreparedStatement";
+            return StatementType.PREPARED_STATEMENT;
         }
-        return "Statement";
+        return StatementType.STATEMENT;
     }
 
 }
--- a/src/main/java/com/microsoft/sqlserver/jdbc/PerformanceLog.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/PerformanceLog.java
@@ -20,6 +20,10 @@ class PerformanceLog {
     private static boolean callbackInitialized = false;
     private static boolean cachedUseNanos = false;
 
+    // ThreadLocal to hold current SQL text and statement type for the duration of a publish callback
+    static final ThreadLocal<String> currentUserSql = new ThreadLocal<>();
+    static final ThreadLocal<String> currentStatementType = new ThreadLocal<>();
+
     /**
      * Register a callback for performance log events.
      * The value of {@link PerformanceLogCallback#useNanoseconds()} is captured at registration
@@ -46,9 +50,6 @@ class PerformanceLog {
         callbackInitialized = false;
     }
 
-    //TODO
-    //More loggers to be added here e.g. com.microsoft.sqlserver.jdbc.PerformanceMetrics.Statement
-    
     public static class Scope implements AutoCloseable {
         private Logger logger;
         private int connectionId;
@@ -59,14 +60,17 @@ class PerformanceLog {
         private final boolean useNanos;
 
         private Exception exception;
+        private SQLServerStatement stmtHandle;
+        private String userSql;
 
         // Constructor for connection-level activities
         public Scope(Logger logger, int connectionId, PerformanceActivity activity) {
-            this(logger, connectionId, 0, activity);
+            this(logger, connectionId, 0, null, null, activity);
         }
 
         // Constructor for statement-level activities
-        public Scope(Logger logger, int connectionId, int statementId, PerformanceActivity activity) {
+        public Scope(Logger logger, int connectionId, int statementId,
+                     SQLServerStatement stmt, String userSql, PerformanceActivity activity) {
             this.enabled = logger.isLoggable(Level.FINE) || (callback != null);
             this.useNanos = cachedUseNanos;
 
@@ -76,6 +80,12 @@ class PerformanceLog {
                 this.statementId = statementId;
                 this.activity = activity;
                 this.startTime = useNanos ? System.nanoTime() : System.currentTimeMillis();
+
+                // If we have a callback and statement info, capture it for use during publish
+                if (callback != null && stmt != null) {
+                    this.stmtHandle = stmt;
+                    this.userSql = userSql;
+                }
             }
         }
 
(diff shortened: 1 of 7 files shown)


machavan (dismissed):

src/main/java/com/microsoft/sqlserver/jdbc/PerformanceLogCallback.java
```diff
@@ -49,4 +49,28 @@ default boolean useNanoseconds() {
...
+    /**
+     * Returns the statement type for the current performance event.
+     * Only valid inside a {@link #publish(PerformanceActivity, int, int, long, Exception)} callback invocation.

+     * Only valid inside a {@link #publish} callback invocation.
+     * Returns {@code null} for connection-level activities or when called outside {@code publish()}.
+     *
+     * @return the statement type string, or null if not applicable.
+     */
+    default String getCurrentStatementType() {
```
Ananya2: Returning a bare String hard-codes magic literals into the API. It sacrifices compile-time safety, silently allows consumer typos, and makes future additions or renames of statement types a breaking change. This should return an `enum` (preferred) or expose public constants instead of raw strings.

src/main/java/com/microsoft/sqlserver/jdbc/PerformanceLog.java
```diff
@@ -101,13 +111,25 @@ public void close() {
 
             if (callback != null) {
```
divang, line 112: Would it be more appropriate to use the callbackInitialized flag here rather than relying on if (callback != null)?

src/main/java/com/microsoft/sqlserver/jdbc/PerformanceLog.java
divang, line 78: The logger assignment seems to be independent of:
Am I missing any dependency between them?

src/main/java/com/microsoft/sqlserver/jdbc/PerformanceLog.java
muskan124947: checking the actual reference which we intend to use is the more defensive pattern.

src/main/java/com/microsoft/sqlserver/jdbc/PerformanceLog.java
muskan124947: we have if check
 if (enabled) {
                this.logger = logger;

divang (approved):

Ananya2 (approved):

muskan124947:
Hi @JDBC-0
Just wanted to let you know that we merged https://github.com/microsoft/mssql-jdbc/pull/2885, which adds getCurrentUserSql and getCurrentStatementType to PerformanceLogCallback.
Since you sometimes do performance testing, thought this might be interesting for you. Feel free to take a look and try it out if you think it's useful.
We'd love to hear any feedback. 🙂

JDBC-0:
> Hi @JDBC-0 Just wanted to let you know that we merged #2885, which adds getCurrentUserSql and getCurrentStatementType to PerformanceLogCallback. Since you sometimes do performance testing, thought this might be interesting for you. Feel free to take a look and try it out if you think it's useful. We'd love to hear any feedback. 🙂

Thank you very much for this info - this helps very much in automatic identification of long running statements! Looking forward to the next release.  

You know, we can't get enough ... may I ask if it would be possible to get a ResultSet row count in addition?  We have mass data problems and try to workaround them with a initial check if there are too many data rows in a ResultSet. So far, we need to cycle through the ResultSet to get the row count.

muskan124947:
@JDBC-0 
Thanks for the suggestion!
Unfortunately, this isn't something we can currently expose through the callback. The event is published as soon as we receive the first response packet from the server for a Statement/PreparedStatement execution. At that point, the full ResultSet hasn't been processed yet, so the final row count is not available.

JDBC-0:
Yeah ... just a fallback idea. A better solution would be that the application would do this already on the database level. But this would be an enormous task to change in alle stored procedes, views, selects. The basic problem is:

If a statement produces a large amount of rows, how can we check this amount as soon as possible to avoid "freezing", blocking, OutOfMemory problems. I hoped to find an easy solution that could be integrated at a single central location (here : JDBC driver response for a statement call).

muskan124947:
@JDBC-0 , Understood — that's a common pain point with large result sets.

Unfortunately this isn't feasible at the driver level. TDS uses a streaming protocol: rows arrive incrementally as the server produces them, and the driver hands them to the application one at a time via ResultSet.next(). There's no "total row count" available upfront — the server itself may not know the final count until it finishes producing all rows (e.g., for complex joins or filtered queries).

JDBC-0:
OK, so the only way is to implement this on the application level - i.e. changing alle stored procedures to first return a meta data result with the number of rows, the following ResultSet will have.

Issue #2929, opened 2026-03-30
[FEATURE REQUEST] support set  nocount off/on switch as part of jdbc connection string


qmarsun:
## Is your feature request related to a problem? If so, please give a short summary of the problem and how the feature would resolve it

 support set  nocount off/on switch as part of jdbc connection string
https://learn.microsoft.com/en-us/sql/connect/jdbc/setting-the-connection-properties?view=sql-server-ver17
Property
Type
set_nocount value on/off == when on "set nocount on " or when on off 
"set nocount off" for all the sql's getting executed this connection.

## Describe the preferred solution

MSSQL by default is not returning affect row counts.
we need to set set nocount off ,  then start executing the sql or stroed proc  
## Describe alternatives you've considered

 as part of jdbc connecction string : set_nocount=off/on if you take set 
it that all the subsequent connection is not required 
jdbc::.....?set_nocount='on'|'off'
then all the execute stmt ... just to prefix 
"set nocount on/off"+  <sql> 

based on the flag from jdbc connection string
https://github.com/microsoft/mssql-jdbc/blob/main/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java#L291
sql ="set nocount off + sql ;
or 
sql = "set ncount on "+ sql 
this way code can get updated rows. correctly and there is no change to existing code,

## Additional context

this is to support large java code base move from Sybase to MSSQL

machavan:
Thanks @qmarsun for submitting this feature request. While we evaluate and prioritize incoming requests, we welcome contributions from the community. If you’d like to work on this, we’re happy to collaborate, review your PR, and help get it into an upcoming release.

Please refer to our contribution guidelines for more details:
https://github.com/microsoft/mssql-jdbc/blob/main/CONTRIBUTING.md

qmarsun:
@machavan 
I will work on this. I may be have few questions .  Is there way to setup a google meet with you?

qmarsun:
## Summary

This PR adds support for a new JDBC connection property:

    set_nocount=on|off

## Behavior

- Parses `set_nocount` from JDBC URL
- Stores it in `SQLServerConnection`
- Applies session setting using:

    SET NOCOUNT ON | OFF

before statement execution

## Example

```java
Connection conn = DriverManager.getConnection(
    "jdbc:sqlserver://host;databaseName=test;set_nocount=on",
    "user",
    "password"
);

Implementation Details

[...]

PR #2972, opened 2026-06-17
support set_nocount swith from jdbc url property

qmarsun:
To support set_nocount propert as part URL jdbc connection string 
jdbc:sqlserver://localhost:1433;databaseName=testdb;encrypt=true;trustServerCertificate=true;set_nocount=on
return -1

and 
jdbc:sqlserver://localhost:1433;databaseName=testdb;encrypt=true;trustServerCertificate=true;set_nocount=off
that will automatically return rows
Connected to SQL Server successfully.
101 | John | sk | IT | 75000.0
102 | Mary | Johnson | HR | 68000.0
103 | David | Brown | Finance | 82000.0
104 | Lisa | Davis | Operations | 71000.0
105 | James | Wilson | Support | 65000.0
106 | John | sk | it | 100000.0
107 | John | sk | it | 100000.0
UPDATE Employee SET LastName ='sk' where FirstName = ?
Rows affected: 3

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
@@ -94,7 +94,22 @@ import mssql.googlecode.concurrentlinkedhashmap.EvictionListener;
  * SQLServerConnectionPoolProxy
  * 
  */
+
 public class SQLServerConnection implements ISQLServerConnection, java.io.Serializable {
+    
+ /**
+  * This enum is used to represent On and Off values for connection properties that support that. 
+  * It is used instead of a boolean to allow for a third "uninitialized" state, since some connection
+  *  properties have values that are determined
+  * after the connection is established and cannot be determined at connection initialization time 
+  * when the connection  properties are being parsed and set.    
+  */
+    
+    private OnOffOption setNoCount = OnOffOption.OFF;
+
+    final OnOffOption getSetNoCount() {
+        return setNoCount;
+    }
 
     /**
      * Always refresh SerialVersionUID when prompted
@@ -103,6 +118,7 @@ public class SQLServerConnection implements ISQLServerConnection, java.io.Serial
 
     /**
      * A random netAddress for this process to send during LOGIN7
+     * 
      */
     private static final byte[] netAddress = getRandomNetAddress();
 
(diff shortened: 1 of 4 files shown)


qmarsun:
I have tested using simple java code 
c:\forkumar\test>java MSSQLConnectionExample
jdbc:sqlserver://localhost:1433;databaseName=testdb;encrypt=true;trustServerCertificate=true;set_nocount=off
Connected to SQL Server successfully.
101 | John | sk | IT | 75000.0
102 | Mary | Johnson | HR | 68000.0
103 | David | Brown | Finance | 82000.0
104 | Lisa | Davis | Operations | 71000.0
105 | James | Wilson | Support | 65000.0
106 | John | sk | it | 100000.0
107 | John | sk | it | 100000.0
UPDATE Employee SET LastName ='sk' where FirstName = ?
Rows affected: 3

return 3

divang:
@qmarsun 
Thanks for the contribution and for adding support for set_nocount via JDBC properties.

We are currently reviewing the PR. Could you please add test cases for this change as well, covering:

Property parsing and validation for ON/OFF values.
Default behavior when the property is not provided.
Runtime behavior impact on statement execution/update counts.
Once tests are added, we can continue with the review.

divang:
@qmarsun 
Hope you are working on the test cases for this PR.
Could you share the exact Sybase-to-SQL Server migration use case behind this change? I am trying to understand what customer behavior breaks today and why exposing SET NOCOUNT through a JDBC URL property is needed.
[Ref](https://github.com/microsoft/mssql-jdbc/issues/2929)

qmarsun:
default behavior is set_nocount on. or jdbc connection does not have property passed 
in that case. 
return -1

divang:
@qmarsun 
Need one more help, could you also share a concrete migration or application scenario where this functionality is required?

It would be very helpful if you could provide a minimal repro/sample application that depends on configuring SET NOCOUNT through the JDBC connection string. Understanding the real-world use case will help us validate both the necessity of the feature and the proposed implementation.

divang:
@qmarsun 
We have simulated the MYSQL and Sybase DB behavior with MS-JDBC Driver and jConnect Driver.
| Configuration | SAP ASE with jConnect | SQL Server `main` | SQL Server PR #2972 |
|---|---:|---:|---:|
| Property omitted | `3` | `3` | `3` |
| URL requests `NOCOUNT ON` | `0` | `3` (property ignored) | `-1` |
| URL requests `NOCOUNT OFF` | `3` | `3` (property ignored) | `3` |

Could you please confirm the exact behavior in your environment?

The statement “MSSQL by default is not returning affected-row counts” does not appear to be universally true. On an unmodified (main branch code) Microsoft JDBC driver, a new SQL Server session returned the expected affected-row count when NOCOUNT was OFF.

Missing counts may instead be caused by environment or execution context, such as:

- Server-level **user options**
- A login trigger
- Application or connection-pool initialization
- Session state retained by a pooled connection
- A stored procedure containing SET NOCOUNT ON
- We also confirmed that SAP jConnect supports URL-level session initialization through (please confirm this too)

The Microsoft JDBC driver does not currently provide an equivalent generic URL property; applications must execute SET NOCOUNT OFF after opening each connection or use a framework/pool initialization hook. 

Your PR #2972 appears to address this migration gap through the narrower set_nocount=on|off property.

Could you confirm:
Whether the Sybase application currently uses SQLInitString=SET NOCOUNT OFF.

PR #2974, opened 2026-06-19
Perf: Optimize the ResultSet read path and reduce allocation overhead

Ananya2:
## Summary

This PR introduces six targeted ResultSet read-path optimizations for allocation-heavy wide-row workloads. The changes reduce allocation churn, streamline common reader operations, and eliminate unnecessary wrapper objects for synchronous `getString()` calls while preserving the existing lazy decoding architecture and specialized code paths.

ResultSet bulk-read workloads incur substantial per-column allocation overhead from transient reader state and wrapper objects. On wide result sets, these short-lived allocations create significant young-generation GC pressure and reduce throughput.

**Measured result** on a 601-column × 1M-row × 200-iteration scan with 400 runs (default + fetchSize=512):

|---|---|---|---|
| Average fetch (ms) | 52,158 | **47,651** | **8.64 % faster** |
| Median fetch (ms) | 51,423 | **46,976** | **8.65 % faster** |
| P95 fetch (ms) | 58,239 | **52,756** | **9.41 % faster** |
| Max fetch (ms) | 80,500 | **62,374** | **22.52 % faster** |
| Standard deviation (ms) | 3,344 | **2,704** | **19.14 % lower** |
| Spikes > 55 s | 53 | **15** | **72 % fewer** |
| Spikes > 60 s | 15 | **5** | **67 % fewer** |
| Spikes > 65 s | 3 | **0** | **100 % fewer** |
| Throughput (rows/sec) | 19,173 | **20,986** | **9.5 % higher** |

The tail-latency improvement is a notable secondary benefit: fewer allocations slow young-generation growth, reduce minor GC frequency, and shrink the GC-induced latency tail. The largest baseline outlier in this dataset reached 80.5 seconds; this PR's worst case was 62.4 seconds,18 seconds lower. The observed latency range narrowed from 33.4 seconds to 18.4 seconds (−45%).

## Key improvements

* Reuse a single `ServerDTVImpl` instance per column across rows.
* Add inline fast paths for common `IOBuffer` operations.
* Eliminate unnecessary stream wrapper allocations for common synchronous `getString()` calls.
* Reduce logging overhead in byte-copy loops.
* Harden `ResultSet` row cleanup bounds.
* Preserve existing behavior for streaming, adaptive, encrypted, and cursor-based code paths.

---

### 1. Reuse a single `ServerDTVImpl` instance per-column (`dtv.java`)

Every column's `DTV` instance allocates a fresh `ServerDTVImpl` on first access (`DTV.getValue()`, `DTV.skipValue()`, `DTV.initFromCompressedNull()`), then discards it when `DTV.clear()` sets `impl = null` at end of row. On the benchmark workload, this translates into tens of billions of short-lived ServerDTVImpl allocations.

A single reusable-instance field (`reusableServerImpl`) on each `DTV` holds one `ServerDTVImpl` for the column's lifetime. The helper `acquireServerImpl()` creates it on first use and, on every reuse, calls `reset()` (zeroing its four wire-state fields) before handing it out, so it is always clean. During an active read, `impl` points at this same instance. `DTV.clear()` simply resets `impl` to `null` at end of row (the reusable instance stays put), and `setImpl()` needs no special handling on the `ServerDTVImpl → AppDTVImpl` swap because the instance is already retained.

```java
private ServerDTVImpl reusableServerImpl;

final void clear() {
    impl = null;                                      // preserve isInitialized() contract; instance stays retained
}

private ServerDTVImpl acquireServerImpl() {
    ServerDTVImpl s = reusableServerImpl;
    if (null != s) {
        s.reset();                                    // zero wire-state before reuse
        return s;
    }
    reusableServerImpl = new ServerDTVImpl();          // create once on first use
    return reusableServerImpl;
}

void setImpl(DTVImpl impl) {
    this.impl = impl;                                 // instance already held in reusableServerImpl
}
```

`ServerDTVImpl.reset()` zeros the same four fields the constructor initializes (`valueLength`, `valueMark`, `isNull`, `internalVariant`), so a reused instance is externally indistinguishable from a fresh one.

**Net effect:** one reusable ServerDTVImpl per DTV instead of one allocation per accessed column per row.

**Safety.** The external contract of `isInitialized()` is preserved because `impl` is still cleared to `null` after `clear()`; the reusable instance is only returned on the *next* read-side call, and it is `reset()` at that single point before use. The reusable instance is per-DTV (one per column per ResultSet), so there is no cross-column or cross-row aliasing and no thread sharing. Only `ServerDTVImpl` is reused; `AppDTVImpl` is not, on the updatable-row swap, `impl` is replaced with a fresh `AppDTVImpl` while the `ServerDTVImpl` remains retained for the next read.

---

### 2. Inline fast path for `readUnsignedByte()` (`IOBuffer.java`)

Every `BYTELENTYPE` column's length prefix (e.g. `char(1)`, `varchar(64)`, nullable `int`/`float`/`decimal`) goes through `readUnsignedByte()`, which unconditionally invokes `ensurePayload()`. In the common case the current packet already contains payload data, making
`ensurePayload()` a no-op. Add an inline `payloadOffset < currentPacket.payloadLength` check before the call, mirroring the pattern already in place for `readShort()`, `readInt()`, and `readLong()`:

```java
final int readUnsignedByte() throws SQLServerException {
    if (payloadOffset < currentPacket.payloadLength) {
        return currentPacket.payload[payloadOffset++] & 0xFF;
    }
    if (!ensurePayload()) throwInvalidTDS();
    return currentPacket.payload[payloadOffset++] & 0xFF;
}
```

**Safety.** Identical semantics. The slow path is preserved verbatim for packet-boundary cases.

---

### 3. Zero-copy fast path for `readDaysIntoCE()` (`IOBuffer.java`)

`readDaysIntoCE()` is called for every `DATE`, `DATETIME2`, `DATETIMEOFFSET` column. The original allocates a `byte[3]`, calls `readBytes()` to fill it, then loops to reassemble the 3 bytes into an `int`, a method-call chain and an allocation per call. Read the 3 bytes directly from the current packet payload when available; fall back to the existing cross-packet path otherwise:

```java
private int readDaysIntoCE() throws SQLServerException {
    final int length = TDS.DAYS_INTO_CE_LENGTH;     // 3
    int daysIntoCE;
    if (payloadOffset + length <= currentPacket.payloadLength) {
        final byte[] p = currentPacket.payload;
        final int off = payloadOffset;
        daysIntoCE = (p[off]     & 0xFF)
                   | ((p[off + 1] & 0xFF) <<  8)
                   | ((p[off + 2] & 0xFF) << 16);
        payloadOffset += length;
    } else {
        // Slow path: same as before. Uses the shared per-reader scratch buffer,
        // so neither path allocates per call.
        final byte[] value = readWrappedBytes(length);
        daysIntoCE = 0;
        for (int i = 0; i < length; i++)
            daysIntoCE |= ((value[i] & 0xFF) << (8 * i));
    }
    if (daysIntoCE < 0) throwInvalidTDS();
    return daysIntoCE;
}
```

**Safety.** Output is equivalent to the existing implementation. The fallback is the original code unchanged.

---

### 4. Hoist `isLoggable()` out of the `readBytes()` and `readSkipBytes()` loops (`IOBuffer.java`)

`readBytes()` and `readSkipBytes()` are the byte-copy and byte-skip loops used by every variable-length column read and skip. Both inner bodies did `if (logger.isLoggable(Level.FINEST))` on every iteration, one volatile read per packet copied or skipped, never constant-folded by the JIT. Hoist the check to a `final boolean` local before the loop in both methods, avoiding repeated logger.isLoggable() checks within the loop.

```java
final void readBytes(byte[] value, int valueOffset, int valueLength) throws SQLServerException {
    final boolean isLogging = logger.isLoggable(Level.FINEST);  // hoisted
    for (int bytesRead = 0; bytesRead < valueLength;) {
        ...
        if (isLogging) logger.finest(...);
        ...
    }
}

final void readSkipBytes(int valueLength) throws SQLServerException {
    final boolean isLogging = logger.isLoggable(Level.FINEST);  // hoisted
    for (int bytesSkipped = 0; bytesSkipped < valueLength;) {
        ...
        if (isLogging) logger.finest(...);
        ...
    }
}
```

**Safety.** The only observable difference is that FINEST logging toggled *during* a single call to either method would not affect subsequent iterations of that call. Logging configuration is set at startup in any real deployment; this is the standard Java logging-performance idiom and the two methods are now symmetric.

---

### 5. Tighter loop bound in `discardCurrentRow()` (`SQLServerResultSet.java`)

When `next()` advances to the next row, `discardCurrentRow()` iterates through previously-accessed columns calling `column.clear()`. The original loop bound was `lastColumnIndex` without a defensive check against `columns.length`, leaving open an `ArrayIndexOutOfBoundsException` in edge cases (server cursor with rowstat column under unusual access patterns). Use `Math.min(lastColumnIndex, columns.length + 1)` as the loop bound:

```java
int clearUpTo = Math.min(lastColumnIndex, columns.length + 1);
for (int columnIndex = 1; columnIndex < clearUpTo; ++columnIndex)
    getColumn(columnIndex).clear();
```

**Safety.** Identical work in the normal case, strictly safer bounds.

---

### 6. Bypass `SimpleInputStream` for small synchronous string reads (`dtv.java`)

A major source of transient allocation pressure on this workload. For `rs.getString()` on a CHAR/VARCHAR/NCHAR/NVARCHAR column, `ServerDTVImpl.getValue()` allocates:

1. `new SimpleInputStream(tdsReader, valueLength, ...)` : ~48 bytes including `BaseInputStream` fields
2. A `TDSReaderMark` inside `BaseInputStream`'s constructor: ~24 bytes
3. A second `TDSReaderMark` when the stream's `close()` invokes `setPositionAfterStreamed()`: ~24 bytes
4. The destination `byte[valueLength]` inside `stream.getBytes()`

…all so that `convertStreamToObject` can call `new String(stream.getBytes(), charset)` and return the string. The stream and both marks are immediately discarded. On a workload with ~40 % string columns this creates substantial transient allocation traffic.

When the call is provably a small synchronous `getString()`, i.e., all five of the following hold:

| Guard | Required because |
|---|---|
| `valueLength > 0 && valueLength <= 4000` | Truly large values stay on the streaming path (avoid one-shot huge allocations); zero-length cells stay on the regular path |
| `streamGetterArgs.streamType == StreamType.NONE` | Stream getters (`getBinaryStream` / `getAsciiStream` / `getCharacterStream`) must return the wrapped stream to the caller |
| `!streamGetterArgs.isAdaptive` | Adaptive paths wrap the stream in `BufferedReader`/etc. and hand it to user code |
| `jdbcType.isTextual()` | Non-textual targets (e.g. `getInt()` on a VARCHAR) require `convertStringToObject` to `trim()`/`parse()` the string |
| `jdbcType != CLOB && jdbcType != NCLOB` | Although textual, `getClob()` / `getNClob()` must return a `SQLServerClob` / `SQLServerNClob` (from the stream path), not a `String` |

…read the bytes directly and construct the `String` inline:

```java
case CHAR:
case VARCHAR:
case NCHAR:
case NVARCHAR: {
    if (valueLength > 0 && valueLength <= 4000
            && StreamType.NONE == streamGetterArgs.streamType
            && !streamGetterArgs.isAdaptive
            && jdbcType.isTextual()
            && JDBCType.CLOB != jdbcType && JDBCType.NCLOB != jdbcType) {
        byte[] bytes = new byte[valueLength];
        tdsReader.readBytes(bytes, 0, valueLength);
        convertedValue = new String(bytes, typeInfo.getCharset());
        break;
    }
    // Fall through to the stream-based path below for large/streaming/non-textual/LOB cases.
}
case TEXT:
case NTEXT:
case IMAGE:
case BINARY:
case VARBINARY:
case TIMESTAMP:
case VECTOR: {
    convertedValue = DDC.convertStreamToObject(
            new SimpleInputStream(tdsReader, valueLength, streamGetterArgs, this),
            typeInfo, jdbcType, streamGetterArgs);
    break;
}
```

When all guards hold (the dominant `rs.getString()` shape), **4 allocations + ~6 method calls per string cell collapse to 1 allocation + 1 method call**.

**Safety.** The guarded execution paths through `DDC.convertStreamToObject` were traced and verified to produce equivalent behavior.

* For textual JDBC types with `streamType == NONE` and `!isAdaptive`, the original ultimately executes `convertStringToObject(new String(stream.getBytes(), charset), charset, jdbcType, NONE)` whose default branch (for textual targets) returns the `String` as-is. The fast path builds the same `String` with the same `charset`.
* `CLOB` / `NCLOB` targets are explicitly excluded so `getClob()` / `getNClob()` (and `getObject(i, Clob.class)` / `getObject(i, NClob.class)`) continue to return `SQLServerClob` / `SQLServerNClob` from the stream path rather than a `String`.
* The Java `switch` fall-through is intentional and is explicitly documented in an inline comment so reviewers do not flag it as a bug.
* All other cell-handling cases (`TEXT`, `NTEXT`, `IMAGE`, `BINARY`, `VARBINARY`, `TIMESTAMP`, `VECTOR`) are untouched.
---

## Why this is safe to land

1. **No public-API changes.** No new methods, no signature changes, no contract changes.
2. **All existing slow paths are preserved unchanged.** Every fast path is gated on a condition that defaults to the original code path whenever the optimization does not apply. Packet-boundary reads, large values, streaming/adaptive access, non-textual conversions, encrypted columns, LOBs, PLP types, scrollable/updatable cursors, and server cursors all continue through the existing implementations.
3. **The reusable `ServerDTVImpl` instance is per-DTV, single-threaded by construction.** Each `DTV` belongs to a single column of a single result set; no cross-column or cross-row aliasing is possible. The instance is `reset()` at the single point of reuse (`acquireServerImpl()`) before being handed out, and `impl == null` between rows preserves the read-side null checks, so a stale value can never leak into the next row.
4. **The `String` fast path is behaviorally equivalent.** Traced through every branch of `convertStreamToObject` for the guarded inputs; The resulting `String` is equivalent to the existing implementation for the guarded inputs.
5. The optimized hot paths eliminate or reuse transient objects where possible and do not introduce additional allocation-heavy intermediate structures.
6. **Defaults unchanged.** No new connection property, no opt-in flag. The optimized paths are used automatically whenever their guard conditions are satisfied.

---

### Methodology

* **Volume:** 1,000,000 rows × 200 iterations per run

|---|---|---|
| Average (ms) | 52,158 | **47,651**  |
| Median (ms) | 51,423 | **46,976**  |
| Min (ms) | 47,062 | **43,984**  |
| Max (ms) | 80,500 | **62,374**  |
| Range (ms) | 33,438 | **18,390**  |
| P90 (ms) | 55,841 | **50,304**  |
| P95 (ms) | 58,239 | **52,756**  |
| StdDev (ms) | 3,344 | **2,704**  |
| CV % | 6.41 % | **5.67 %**  |
| Spikes > 55 s | 53 | **15**  |
| Spikes > 60 s | 15 | **5**  |
| Spikes > 65 s | 3 | **0**  |
| Throughput (rows/sec) | 19,173 | **20,986**  |

### Observations

1. The **~9 % mean reduction is uniform** across both fetch-size configurations (−8.74 % default, −8.53 % at 512), confirming the changes target per-cell decoding cost rather than network batching.
2. **100 % of `> 65 s` outliers are eliminated** (3 → 0), consistent with reduced allocation pressure in young generation GC regions.
3. **Throughput rises +9.5 %** (19,173 → 20,986 rows/sec).

---

## Testing

* **Existing test suite:** all unit and integration tests pass against this PR.
* **Benchmark validation:** the 400-run benchmark above was executed end-to-end against a live SQL Server instance with both the baseline and PR builds; the ~9 % improvement was reproduced across multiple runs.
* Added WideRowReadTests nested class in ResultSetTest that reads a fat table (10 rows x 601 columns) and asserts the exact value of every cell via type-specific getters. Covers 20 SQL types including both decimal decode paths: decimal(18,0) with an 8-byte magnitude and decimal(38,4) with a larger-than-8-byte magnitude.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResultSet.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResultSet.java
@@ -1842,8 +1842,11 @@ public class SQLServerResultSet implements ISQLServerResultSet, java.io.Serializ
         deletedCurrentRow = false;
         if (lastColumnIndex >= 1) {
             initializeNullCompressedColumns();
-            // Discard columns up to, but not including, the last indexed column
-            for (int columnIndex = 1; columnIndex < lastColumnIndex; ++columnIndex)
+
+            // Clear accessed columns to release DTV state for reuse on the next row.
+            // Math.min guards against out-of-bounds if lastColumnIndex exceeds columns.length.
+            int clearUpTo = Math.min(lastColumnIndex, columns.length + 1);
+            for (int columnIndex = 1; columnIndex < clearUpTo; ++columnIndex)
                 getColumn(columnIndex).clear();
 
             // Skip and discard the remainder of the last indexed column
--- a/src/main/java/com/microsoft/sqlserver/jdbc/dtv.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/dtv.java
@@ -3893,10 +3893,13 @@ final class ServerDTVImpl extends DTVImpl {
                     //   - !isAdaptive               : adaptive paths wrap the stream for user code
                     //   - jdbcType.isTextual()      : non-textual targets need convertStringToObject
                     //                                 to trim/parse the value
+                    //   - CLOB/NCLOB excluded        : although textual, getClob()/getNClob() must
+                    //                                 return a SQLServerClob/SQLServerNClob, not a String
                     if (valueLength > 0 && valueLength <= 4000
                             && StreamType.NONE == streamGetterArgs.streamType
                             && !streamGetterArgs.isAdaptive
-                            && jdbcType.isTextual()) {
+                            && jdbcType.isTextual()
+                            && JDBCType.CLOB != jdbcType && JDBCType.NCLOB != jdbcType) {
                         byte[] bytes = new byte[valueLength];
                         tdsReader.readBytes(bytes, 0, valueLength);
                         convertedValue = new String(bytes, typeInfo.getCharset());
--- a/src/main/java/com/microsoft/sqlserver/jdbc/dtv.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/dtv.java
@@ -133,6 +133,15 @@ final class DTV {
     /** The source (app or server) providing the data for this value. */
     private DTVImpl impl;
 
+    /**
+    * Single-slot cache for reusing a {@link ServerDTVImpl} across row reads,
+    * avoiding repeated allocations on the ResultSet hot path. {@link #clear()}
+    * still resets {@link #impl} to {@code null}, preserving the existing
+    * initialization and null-state semantics. {@link AppDTVImpl} is excluded
+    * because its update-path lifecycle differs from {@link ServerDTVImpl}.
+    */
+    private ServerDTVImpl pooledServerImpl;
+
     CryptoMetadata cryptoMeta = null;
     JDBCType jdbcTypeSetByUser = null;
     int valueLength = 0;
@@ -157,20 +166,38 @@ final class DTV {
         impl.setValue(value, javaType);
     }
 
+    /** Returns the pooled {@link ServerDTVImpl} (already reset) or a fresh instance. */
+    private ServerDTVImpl acquireServerImpl() {
+        ServerDTVImpl s = pooledServerImpl;
+        if (null != s) {
+            pooledServerImpl = null;
+            return s;
+        }
+        return new ServerDTVImpl();
+    }
+
     final void clear() {
+        // Pool the outgoing ServerDTVImpl for reuse on the next read, so we
+        // avoid a fresh allocation per column per row. AppDTVImpl is not pooled
+        // (rare update path; would complicate the AppDTV <-> ServerDTV swap).
+        if (impl instanceof ServerDTVImpl) {
+            ServerDTVImpl s = (ServerDTVImpl) impl;
+            s.reset();
+            pooledServerImpl = s;
+        }
         impl = null;
     }
 
     final void skipValue(TypeInfo type, TDSReader tdsReader, boolean isDiscard) throws SQLServerException {
         if (null == impl)
-            impl = new ServerDTVImpl();
+            impl = acquireServerImpl();
 
         impl.skipValue(type, tdsReader, isDiscard);
     }
 
     final void initFromCompressedNull() {
         if (null == impl)
-            impl = new ServerDTVImpl();
+            impl = acquireServerImpl();
 
         impl.initFromCompressedNull();
     }
@@ -250,7 +277,7 @@ final class DTV {
             TypeInfo typeInfo, CryptoMetadata cryptoMetadata, TDSReader tdsReader,
             SQLServerStatement statement) throws SQLServerException {
         if (null == impl) {
-            impl = new ServerDTVImpl();
+            impl = acquireServerImpl();
         } else if (impl.isNull()) {
             return null;
         }
@@ -270,6 +297,13 @@ final class DTV {
      * Called by DTV implementation instances to change to a different DTV implementation.
      */
     void setImpl(DTVImpl impl) {
+        // Preserve an outgoing ServerDTVImpl for reuse when replacing the
+        // current implementation, avoiding a fresh allocation on the next read.
+        if (this.impl instanceof ServerDTVImpl && this.impl != impl) {
+            ServerDTVImpl s = (ServerDTVImpl) this.impl;
+            s.reset();
+            pooledServerImpl = s;
+        }
         this.impl = impl;
     }
 
@@ -3321,6 +3355,19 @@ final class ServerDTVImpl extends DTVImpl {
     private SqlVariant internalVariant;
 
     /**
+     * Resets all wire-side state (length, mark, null indicator, SQL_VARIANT context)
+     * so this instance can be reused for the next row's value. Invoked by
+     * {@link DTV#clear()} on the per-row hot path to avoid allocating a fresh
+     * {@code ServerDTVImpl} for every column of every row.
+     */
+    final void reset() {
+        valueLength = 0;
+        valueMark = null;
+        isNull = false;
+        internalVariant = null;
+    }
+
+    /**
      * Sets the value of the DTV to an app-specified Java type.
      *
      * Generally, the value cannot be stored back into the TDS byte stream (although this could be done for fixed-length
@@ -3833,9 +3880,31 @@ final class ServerDTVImpl extends DTVImpl {
                 // (CHAR/VARCHAR/TEXT/NCHAR/NVARCHAR/NTEXT/BINARY/VARBINARY/IMAGE) -> ANY jdbcType.
                 case CHAR:
                 case VARCHAR:
-                case TEXT:
                 case NCHAR:
                 case NVARCHAR:
+                {
+                    // Fast path: small synchronous rs.getString() reads bytes directly into
+                    // a String, bypassing the SimpleInputStream wrapper and its embedded
+                    // TDSReaderMark / close() machinery (~4 allocations per cell saved).
+                    //
+                    // Guards (all required for byte-identical semantics with the stream path):
+                    //   - valueLength in (0, 4000]  : skips zero-length and overly large values
+                    //   - streamType == NONE        : stream getters need the actual InputStream
+                    //   - !isAdaptive               : adaptive paths wrap the stream for user code
+                    //   - jdbcType.isTextual()      : non-textual targets need convertStringToObject
+                    //                                 to trim/parse the value
+                    if (valueLength > 0 && valueLength <= 4000
+                            && StreamType.NONE == streamGetterArgs.streamType
+                            && !streamGetterArgs.isAdaptive
+                            && jdbcType.isTextual()) {
+                        byte[] bytes = new byte[valueLength];
+                        tdsReader.readBytes(bytes, 0, valueLength);
+                        convertedValue = new String(bytes, typeInfo.getCharset());
+                        break;
+                    }
+                    // Fall through to the stream-based path for large/streaming/non-textual cases.
+                }
+                case TEXT:
                 case NTEXT:
                 case IMAGE:
                 case BINARY:
--- a/src/main/java/com/microsoft/sqlserver/jdbc/dtv.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/dtv.java
@@ -134,11 +134,11 @@ final class DTV {
     private DTVImpl impl;
 
     /**
-     * Single reusable {@link ServerDTVImpl} instance for this column. Because there is one {@link DTV} per column, this
-     * instance is created once per column and reused to fetch every row's value for that column, avoiding a fresh
-     * allocation per row on the ResultSet hot path. {@link #clear()} still resets {@link #impl} to {@code null},
-     * preserving the existing initialization and null-state semantics. {@link AppDTVImpl} is excluded because its
-     * update-path lifecycle differs from {@link ServerDTVImpl}.
+     * A single ServerDTVImpl instance reused for this column. There is one DTV per column, so this instance is created
+     * once (lazily, on the first server read) and reused to fetch every subsequent row's value, avoiding a fresh
+     * allocation per row on the ResultSet hot path. It is reset in acquireServerImpl() before each reuse, so it is
+     * always handed out clean. During an active read, impl points at this same instance. AppDTVImpl is not kept here
+     * because its update-path lifecycle is different from ServerDTVImpl.
      */
     private ServerDTVImpl reusableServerImpl;
 
@@ -166,25 +166,23 @@ final class DTV {
         impl.setValue(value, javaType);
     }
 
-    /** Returns the reusable {@link ServerDTVImpl} (already reset) or a fresh instance if none is retained. */
+    /**
+     * Returns this column's reusable ServerDTVImpl, creating it on first use. The instance is reset here, at the single
+     * point of reuse, so it is always handed out clean regardless of the previous row's state.
+     */
     private ServerDTVImpl acquireServerImpl() {
         ServerDTVImpl s = reusableServerImpl;
         if (null != s) {
-            reusableServerImpl = null;
+            s.reset(); // clear the previous row's state before handing it out
             return s;
         }
-        return new ServerDTVImpl();
+        reusableServerImpl = new ServerDTVImpl();
+        return reusableServerImpl;
     }
 
     final void clear() {
-        // Retain the outgoing ServerDTVImpl for reuse on the next read, so we
-        // avoid a fresh allocation per column per row. AppDTVImpl is not retained
-        // (rare update path; would complicate the AppDTV <-> ServerDTV swap).
-        if (impl instanceof ServerDTVImpl) {
-            ServerDTVImpl s = (ServerDTVImpl) impl;
-            s.reset();
-            reusableServerImpl = s;
-        }
+        // Just clear the active impl reference so isInitialized() and the read-side null checks work as before. The
+        // per-column ServerDTVImpl stays in reusableServerImpl and is reset in acquireServerImpl() before the next row.
         impl = null;
     }
 
@@ -297,13 +295,8 @@ final class DTV {
      * Called by DTV implementation instances to change to a different DTV implementation.
      */
     void setImpl(DTVImpl impl) {
-        // Retain an outgoing ServerDTVImpl for reuse when replacing the
-        // current implementation, avoiding a fresh allocation on the next read.
-        if (this.impl instanceof ServerDTVImpl && this.impl != impl) {
-            ServerDTVImpl s = (ServerDTVImpl) this.impl;
-            s.reset();
-            reusableServerImpl = s;
-        }
+        // No need to save the outgoing ServerDTVImpl here: the per-column instance is already kept in
+        // reusableServerImpl (impl points at it during a server read), so it stays available for the next read.
         this.impl = impl;
     }
 
@@ -3355,10 +3348,9 @@ final class ServerDTVImpl extends DTVImpl {
     private SqlVariant internalVariant;
 
     /**
-     * Resets all wire-side state (length, mark, null indicator, SQL_VARIANT context)
-     * so this instance can be reused for the next row's value. Invoked by
-     * {@link DTV#clear()} on the per-row hot path to avoid allocating a fresh
-     * {@code ServerDTVImpl} for every column of every row.
+     * Resets all wire-side state (length, mark, null indicator, SQL_VARIANT context) so this instance can be reused
+     * for the next row's value. Called from DTV.acquireServerImpl() on the per-row hot path to avoid allocating a
+     * fresh ServerDTVImpl for every column of every row.
      */
     final void reset() {
         valueLength = 0;
--- a/src/main/java/com/microsoft/sqlserver/jdbc/dtv.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/dtv.java
@@ -134,12 +134,12 @@ final class DTV {
     private DTVImpl impl;
 
     /**
-    * Single-slot cache for reusing a {@link ServerDTVImpl} across row reads,
-    * avoiding repeated allocations on the ResultSet hot path. {@link #clear()}
-    * still resets {@link #impl} to {@code null}, preserving the existing
-    * initialization and null-state semantics. {@link AppDTVImpl} is excluded
-    * because its update-path lifecycle differs from {@link ServerDTVImpl}.
-    */
+     * Single-slot cache holding this column's {@link ServerDTVImpl}. Because there is one {@link DTV} per column, this
+     * instance is created once per column and reused to fetch every row's value for that column, avoiding a fresh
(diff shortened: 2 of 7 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResultSet.java
```diff
@@ -1842,8 +1842,11 @@ final void discardCurrentRow() throws SQLServerException {
         deletedCurrentRow = false;
         if (lastColumnIndex >= 1) {
             initializeNullCompressedColumns();
-            // Discard columns up to, but not including, the last indexed column
-            for (int columnIndex = 1; columnIndex < lastColumnIndex; ++columnIndex)
+
+            // Clear accessed columns to release DTV state for reuse on the next row.
```
machavan, line 1846: lets have a mini CRD app , just 10 records as a unit test to exercise these code paths

src/test/java/com/microsoft/sqlserver/jdbc/resultset/ResultSetTest.java
```diff
@@ -2462,4 +2462,206 @@ public void testRefetchStreamValues() throws SQLException, IOException {
...
+                    insert.append(",?");
+                }
+            }
+            create.append(")");
+            insert.append(")");
+
+            try (Connection conn = getConnection(); Statement stmt = conn.createStatement()) {
+                TestUtils.dropTableIfExists(wideTable, stmt);
+                stmt.execute(create.toString());
+            }
```
muskan124947, line 2646: Currently only column count is checked

assertEquals(NUM_COLUMNS, rs.getMetaData().getColumnCount());

If the schema were accidentally reordered but still had 601 columns, this test would still pass until later (and perhaps with confusing failures).

we can have like 
ResultSetMetaData md = rs.getMetaData();
assertEquals("ID", md.getColumnName(1));
assertEquals("C_INT_1", md.getColumnName(2));
assertEquals("C_UID_30", md.getColumnName(NUM_COLUMNS));

src/test/java/com/microsoft/sqlserver/jdbc/resultset/ResultSetTest.java
```diff
@@ -2462,4 +2462,206 @@ public void testRefetchStreamValues() throws SQLException, IOException {
...
+    @Nested
+    class WideRowReadTests {
+
+        private static final int GROUP_COUNT = 30; // 20 columns per group
+        private static final int COLS_PER_GROUP = 20;
+        private static final int NUM_COLUMNS = 1 + GROUP_COUNT * COLS_PER_GROUP;
+        private static final int ROW_COUNT = 10;
+
+        // Column name/type pairs making up one repeating group.
+        private final String[] colNames = {"C_INT", "C_BIGINT", "C_SMALLINT", "C_TINYINT", "C_BIT", "C_CHAR",
```
muskan124947: colNames and colTypes should probably be static final

src/test/java/com/microsoft/sqlserver/jdbc/resultset/ResultSetTest.java
```diff
@@ -2462,4 +2462,206 @@ public void testRefetchStreamValues() throws SQLException, IOException {
...
+                "C_VARCHAR", "C_NCHAR", "C_NVARCHAR", "C_DEC18", "C_DEC38", "C_NUMERIC", "C_FLOAT", "C_REAL", "C_MONEY",
+                "C_SMALLMONEY", "C_DATETIME", "C_DATETIME2", "C_SMALLDATETIME", "C_UID"};
+        private final String[] colTypes = {"int", "bigint", "smallint", "tinyint", "bit", "char(1)", "varchar(64)",
+                "nchar(4)", "nvarchar(32)", "decimal(18,0)", "decimal(38,4)", "numeric(10,2)", "float", "real", "money",
+                "smallmoney", "datetime", "datetime2", "smalldatetime", "uniqueidentifier"};
+
+        // decimal(18,0) 18-digit value fits an 8-byte magnitude; decimal(38,4) 29-digit value needs more.
+        private final BigDecimal dec18 = new BigDecimal("123456789012345678");
```
muskan124947: private static final BigDecimal DEC18 = ...
private static final BigDecimal DEC38 = ...
private static final String UID = ...

src/test/java/com/microsoft/sqlserver/jdbc/resultset/ResultSetTest.java
```diff
@@ -2462,4 +2462,206 @@ public void testRefetchStreamValues() throws SQLException, IOException {
...
+                    assertEquals(NUM_COLUMNS, rs.getMetaData().getColumnCount());
+
+                    int rowsRead = 0;
+                    while (rs.next()) {
+                        int row = rs.getInt(1);
+                        rowsRead++;
+                        int col = 2;
+                        for (int g = 1; g <= GROUP_COUNT; g++) {
+                            int s = seed(row, g);
+                            String m = " at row " + row + " group " + g;
```
muskan124947: can we rename it to message or context

src/main/java/com/microsoft/sqlserver/jdbc/dtv.java
```diff
@@ -157,20 +166,38 @@ final void setValue(Object value, JavaType javaType) {
...
-            impl = new ServerDTVImpl();
+            impl = acquireServerImpl();
 
         impl.skipValue(type, tdsReader, isDiscard);
     }
 
     final void initFromCompressedNull() {
         if (null == impl)
-            impl = new ServerDTVImpl();
+            impl = acquireServerImpl();
```
machavan, line 198: Key Risks
1. [ServerDTVImpl](vscode-file://vscode-app/c:/Users/machavan/AppData/Local/Programs/Microsoft%20VS%20Code/6a49527b96/resources/app/out/vs/code/electron-browser/workbench/workbench.html) pooling (highest risk) — Changes object lifecycle, not just local logic.

[reset()](vscode-file://vscode-app/c:/Users/machavan/AppData/Local/Programs/Microsoft%20VS%20Code/6a49527b96/resources/app/out/vs/code/electron-browser/workbench/workbench.html) zeroes 4 fields correctly today, but any future field added and forgotten = silent stale-state bleeding into the next row. No guard keeps [reset()](vscode-file://vscode-app/c:/Users/machavan/AppData/Local/Programs/Microsoft%20VS%20Code/6a49527b96/resources/app/out/vs/code/electron-browser/workbench/workbench.html) in sync.
Reusing the impl (instead of orphaning it) tightens coupling with adaptive streams/marks that could outlive [next()](vscode-file://vscode-app/c:/Users/machavan/AppData/Local/Programs/Microsoft%20VS%20Code/6a49527b96/resources/app/out/vs/code/electron-browser/workbench/workbench.html). Corruption risk if any stream survives row advance.
Updatable-cursor [setImpl](vscode-file://vscode-app/c:/Users/machavan/AppData/Local/Programs/Microsoft%20VS%20Code/6a49527b96/resources/app/out/vs/code/electron-browser/workbench/workbench.html) swap path is rarely exercised (patch coverage ~77%).
2. getString() bypass of SimpleInputStream — Verified equivalent for guarded inputs, but:

[valueLength <= 4000](vscode-file://vscode-app/c:/Users/machavan/AppData/Local/Programs/Microsoft%20VS%20Code/6a49527b96/resources/app/out/vs/code/electron-browser/workbench/workbench.html) boundary and fast→stream transition untested.
Single string value straddling a TDS packet boundary not asserted.
3. discardCurrentRow bound clamp — Safer, but masks the root cause: if [lastColumnIndex > columns.length](vscode-file://vscode-app/c:/Users/machavan/AppData/Local/Programs/Microsoft%20VS%20Code/6a49527b96/resources/app/out/vs/code/electron-browser/workbench/workbench.html) is truly possible, clamping hides a deeper indexing bug rather than explaining it.

4. Low risk — [readUnsignedByte](vscode-file://vscode-app/c:/Users/machavan/AppData/Local/Programs/Microsoft%20VS%20Code/6a49527b96/resources/app/out/vs/code/electron-browser/workbench/workbench.html)/[readDaysIntoCE](vscode-file://vscode-app/c:/Users/machavan/AppData/Local/Programs/Microsoft%20VS%20Code/6a49527b96/resources/app/out/vs/code/electron-browser/workbench/workbench.html) fast paths (mirror existing pattern), [isLoggable](vscode-file://vscode-app/c:/Users/machavan/AppData/Local/Programs/Microsoft%20VS%20Code/6a49527b96/resources/app/out/vs/code/electron-browser/workbench/workbench.html) hoisting (negligible).

Tests Needed
Guard/test that [ServerDTVImpl.reset()](vscode-file://vscode-app/c:/Users/machavan/AppData/Local/Programs/Microsoft%20VS%20Code/6a49527b96/resources/app/out/vs/code/electron-browser/workbench/workbench.html) stays complete as fields evolve.
Adaptive/streaming reads still return correct data after row advance.
Updatable-resultset impl-swap pooling path.
Encrypted (Always Encrypted) textual columns.
getString() at the 4000-byte boundary + value spanning a packet boundary.
Coverage gap: 82.6% patch coverage; the 8 missing lines are exactly the new fallback/pool branches that differ from the happy path.

src/main/java/com/microsoft/sqlserver/jdbc/dtv.java
```diff
@@ -133,6 +133,15 @@ final class DTV {
...
     private DTVImpl impl;
 
+    /**
+    * Single-slot cache for reusing a {@link ServerDTVImpl} across row reads,
+    * avoiding repeated allocations on the ResultSet hot path. {@link #clear()}
+    * still resets {@link #impl} to {@code null}, preserving the existing
+    * initialization and null-state semantics. {@link AppDTVImpl} is excluded
+    * because its update-path lifecycle differs from {@link ServerDTVImpl}.
+    */
+    private ServerDTVImpl pooledServerImpl;
```
divang: If ServerDTVImpl is intended to hold only a single instance, is there a reason it's named pooledServerImpl? A name like cachedServerDTVImpl might better reflect its behavior and be less confusing.

src/main/java/com/microsoft/sqlserver/jdbc/dtv.java
```diff
@@ -157,20 +166,38 @@ final void setValue(Object value, JavaType javaType) {
...
+        ServerDTVImpl s = pooledServerImpl;
+        if (null != s) {
+            pooledServerImpl = null;
+            return s;
+        }
+        return new ServerDTVImpl();
+    }
+
     final void clear() {
+        // Pool the outgoing ServerDTVImpl for reuse on the next read, so we
```
divang: I think cached would be a more appropriate term than pool here, since it appears to hold a single instance rather than managing a pool of instances. If we decide to use cached instead of pool, we should also update all related comments and documentation wherever the term pool is currently used to keep the naming consistent.

src/test/java/com/microsoft/sqlserver/jdbc/resultset/ResultSetTest.java
```diff
@@ -2462,4 +2484,724 @@ public void testRefetchStreamValues() throws SQLException, IOException {
...
+         * {@code PLPInputStream} captures its owning {@code ServerDTVImpl} and, on {@code close()}, calls
+         * {@code setPositionAfterStreamed()} which mutates that impl. Because the driver closes the active stream while
+         * moving off the column (which nulls the stream's back-reference to the impl) BEFORE {@code clear()} pools the
+         * impl, any later user-issued {@code close()} cannot reach the pooled instance. This test deliberately retains
+         * the stream across {@code next()} and closes it late to exercise exactly that ordering; the earlier version
+         * closed the stream before advancing and therefore never covered this path.
+         */
+        @Test
+        @Tag(Constants.xAzureSQLDW)
+        public void testAdaptiveStreamDoesNotCorruptNextRow() throws Exception {
```
divang, line 3055: Would the StateMachineTest framework be a good fit for this scenario? While the targeted tests added here cover the known adaptive-stream late-close case, a state-machine-based approach could randomly explore combinations of cursor movements and read operations - such as next/previous/absolute, getBytes/getString, partial getBinaryStream reads, retained stream close, and late close under responseBuffering=adaptive. This could help uncover less obvious interleavings while still keeping failures reproducible through seed-based replay.

src/main/java/com/microsoft/sqlserver/jdbc/dtv.java
```diff
@@ -157,20 +166,38 @@ final void setValue(Object value, JavaType javaType) {
         impl.setValue(value, javaType);
     }
 
+    /** Returns the reusable {@link ServerDTVImpl} (already reset) or a fresh instance if none is retained. */
+    private ServerDTVImpl acquireServerImpl() {
+        ServerDTVImpl s = reusableServerImpl;
+        if (null != s) {
+            reusableServerImpl = null;
```
machavan: Though clear is calling reset on this... it would be safer to call a reset from here before returning it ... ?

src/main/java/com/microsoft/sqlserver/jdbc/dtv.java
```diff
@@ -157,20 +166,38 @@ final void setValue(Object value, JavaType javaType) {
...
+            return s;
+        }
+        return new ServerDTVImpl();
+    }
```
machavan, line 181: private ServerDTVImpl acquireServerImpl() {
        ServerDTVImpl s = reusableServerImpl;
        if (null != s) {
            s.reset();
            return s;
        } else {
          reusableServerImpl =   new ServerDTVImpl();
          return reusableServerImpl;
        }        
    }

machavan (approved):

muskan124947 (approved):

PR #2977, opened 2026-06-30
 Add tls-unique channel binding support for Kerberos and NTLM

divang:
**Summary**
This PR adds `tls-unique` channel binding support (RFC 5929) in the JDBC driver so TLS Finished `verify_data` can be incorporated into Java authentication flows used for Extended Protection and related integrations.

**What changed**
The driver now captures `tls-unique` channel binding data from the completed TLS handshake and uses it in authentication paths that support channel binding.

- `TDSChannel` extracts channel binding bytes from `ExtendedSSLSession` after the TLS handshake completes.
- The implementation prefers `getTlsUniqueClientFirstFinishedVerifyData()` and falls back to `getTlsUniqueFirstFinishedVerifyData()` when needed.
- The captured value is normalized as `tls-unique:<verify_data>` per RFC 5929 handling in the client.
- Kerberos integrated authentication now sets GSS channel binding when `tls-unique` data is available.
- NTLM authentication now includes the `MsvAvChannelBindings` AV pair using the MD5 hash of the channel binding structure.
- If the runtime does not expose the required JSSE methods, or if `tls-unique` is unavailable, the driver falls back to the existing no-channel-binding behavior.

**Compatibility**
No driver behavior changes unless the runtime exposes `tls-unique` data and it is enabled in JSSE.

- `sun.security.ssl.enableTlsUniqueChannelBinding` remains disabled by default, so this stays opt-in.
- Existing TLS, Kerberos, and NTLM flows continue unchanged when the feature is unavailable or not enabled.
- On unsupported JDKs or platforms, the driver safely preserves prior behavior.

**Documentation**
Channel Binding issue details doc:
https://github.com/microsoft/mssql-jdbc/wiki/Channel-Binding-Feature-Summary

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
@@ -85,7 +85,6 @@ import com.microsoft.sqlserver.jdbc.dataclassification.SensitivityClassification
 import microsoft.sql.Vector;
 
 final class TDS {
-    // application protocol
     static final String PROTOCOL_TDS80 = "tds/8.0"; // TLS-first connections
 
     // TDS versions
@@ -661,6 +660,9 @@ final class TDSChannel implements Serializable {
 
     // Socket for SSL-encrypted communications with SQL Server
     private transient SSLSocket sslSocket;
+
+    // tls-unique channel binding data from the completed TLS handshake.
+    static byte[] channelBindingInfo = null;
 
     /*
      * Socket providing the communications interface to the driver. For SSL-encrypted connections, this is the SSLSocket
@@ -1938,6 +1940,8 @@ final class TDSChannel implements Serializable {
                 con.addWarning(warningMsg);
             }
 
+            setChannelBindingInfo();
+
             if (logger.isLoggable(Level.FINER))
                 logger.finer(toString() + " SSL enabled");
 
@@ -2001,6 +2005,45 @@ final class TDSChannel implements Serializable {
                 con.terminate(SQLServerException.DRIVER_ERROR_SSL_FAILED, form.format(msgArgs), e);
             }
         }
+    }
+
+    private void setChannelBindingInfo() {
+        channelBindingInfo = createChannelBindingInfo(null == sslSocket ? null : sslSocket.getSession());
+    }
+
+    static byte[] createChannelBindingInfo(javax.net.ssl.SSLSession session) {
+        if (null == session) {
+            return null;
+        }
+
+        byte[] tlsUnique = null;
+        try {
+            Class<?> clazz = Class.forName("javax.net.ssl.ExtendedSSLSession");
+            if (!clazz.isInstance(session)) {
+                return null;
+            }
+
+            java.lang.reflect.Method method = clazz.getMethod("getTlsUniqueChannelBinding");
+            Object value = method.invoke(session);
+            if (value instanceof byte[]) {
+                tlsUnique = (byte[]) value;
+            }
+        } catch (ClassNotFoundException | NoSuchMethodException | IllegalAccessException
+                | InvocationTargetException e) {
+            if (logger.isLoggable(Level.FINER)) {
+                logger.finer("getTlsUniqueChannelBinding is not supported on this platform.");
+            }
+            return null;
+        }
+
+        if (null == tlsUnique || 0 == tlsUnique.length) {
+            return null;
+        }
+
+        byte[] prefix = "tls-unique:".getBytes(java.nio.charset.StandardCharsets.UTF_8);
+        byte[] channelBinding = Arrays.copyOf(prefix, prefix.length + tlsUnique.length);
+        System.arraycopy(tlsUnique, 0, channelBinding, prefix.length, tlsUnique.length);
+        return channelBinding;
     }
 
     /**
@@ -6750,6 +6793,7 @@ final class TDSWriter {
             }
         }
     }
+
 }
 
 
--- a/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
@@ -2023,31 +2023,31 @@ final class TDSChannel implements Serializable {
                 return null;
             }
 
-            // Try getTlsUniqueClientFirstFinishedVerifyData() first (client's Finished
-            // verify_data,
-            // preferred per RFC 5929 tls-unique for the initiating side).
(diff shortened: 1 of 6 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -2003,6 +2007,71 @@ else if (con.getTrustManagerClass() != null) {
...
+                logger.finer("Using " + method.getName() + " for tls-unique channel binding.");
+            }
+
+            Object value = method.invoke(session);
+            if (value instanceof byte[]) {
+                tlsUnique = (byte[]) value;
+                if (logger.isLoggable(Level.FINER)) {
+                    logger.finer("tls-unique verify_data received: " + tlsUnique.length + " bytes.");
+                }
+            }
```
machavan, line 2063: logger.finer to log verify_data not received

src/main/java/com/microsoft/sqlserver/jdbc/NTLMAuthentication.java
```diff
@@ -581,6 +594,12 @@ private byte[] generateClientChallengeBlob(final byte[] clientNonce) {
         token.putShort((short) context.spnUbytes.length);
         token.put(context.spnUbytes, 0, context.spnUbytes.length);
 
+        // Channel binding
+        calculateChannelBindingMD5Hash();
+        token.putShort(NTLM_AVID_MSVAVCHANNELBINDING);
```
machavan: Check for msvAvChannelBindings

machavan (dismissed):

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -662,6 +661,9 @@ final TDSReader getReader(TDSCommand command) {
     // Socket for SSL-encrypted communications with SQL Server
     private transient SSLSocket sslSocket;
```
muskan124947, line 662: High: Channel binding state is global and shared across all connections, which can produce incorrect CBT on concurrent or mixed TLS/non-TLS auth flows.

Why this is risky: one connection can overwrite another’s binding data between handshake and token generation. Also, a prior TLS connection can leave stale data visible to a later non-TLS connection.
Recommended fix: make CBT per-connection/per-channel (instance field, not static) and pass it explicitly into Kerberos/NTLM auth context; clear it when not applicable.

src/test/java/com/microsoft/sqlserver/jdbc/TDSChannelBindingTest.java
```diff
@@ -0,0 +1,72 @@
...
+                return tlsUnique;
+            }
+            if ("getTlsUniqueFirstFinishedVerifyData".equals(invocation.getMethod().getName())) {
+                return tlsUnique;
+            }
+            return Answers.RETURNS_DEFAULTS.answer(invocation);
+        });
+        return (SSLSession) mock;
+    }
+
```
muskan124947, line 124: Can we add a test that simulates two channels/auth contexts in parallel and verifies each auth path consumes only its own CBT value.

src/main/java/com/microsoft/sqlserver/jdbc/NTLMAuthentication.java
```diff
@@ -607,6 +631,27 @@ private byte[] hmacMD5(final byte[] key, final byte[] data) throws InvalidKeyExc
...
+            md5.update(new byte[4]); // Initiator address
+            md5.update(new byte[4]); // Initiator address length
+            md5.update(new byte[4]); // Acceptor address
+            md5.update(new byte[4]); // Acceptor address length
+            md5.update(new byte[] {(byte) channelBindingInfo.length, (byte) 0, (byte) 0, (byte) 0});
+            md5.update(channelBindingInfo);
+
+            msvAvChannelBindings = md5.digest();
+        } catch (NoSuchAlgorithmException e) {
+            throw new RuntimeException(e);
```
Ananya2: Please avoid throwing a bare RuntimeException from the auth path, wrap as a SQLServerException (or reuse the existing error-message pattern) so failures surface consistently with the rest of the driver.

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -2003,6 +2007,73 @@ else if (con.getTrustManagerClass() != null) {
         }
     }
 
+    private void setChannelBindingInfo() {
```
Ananya2, line 2010: Non-TLS connections don't call setChannelBindingInfo(), so verify_data from a previous TLS connection can remain referenced. This both leaves stale state and unnecessarily retains sensitive data. Please clear it for non-TLS connections and after the channel binding token is built.

src/test/java/com/microsoft/sqlserver/jdbc/TDSChannelChannelBindingTest.java
```diff
@@ -0,0 +1,162 @@
...
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+
+import javax.net.ssl.SSLSession;
+
+import org.junit.jupiter.api.Test;
+import org.mockito.Answers;
+import org.mockito.Mockito;
+
+class TDSChannelChannelBindingTest {
```
machavan: TDSChannelChannelBindingTest has a doubled word; TDSChannelBindingTest reads better.

src/main/java/com/microsoft/sqlserver/jdbc/NTLMAuthentication.java
```diff
@@ -607,6 +646,32 @@ private byte[] hmacMD5(final byte[] key, final byte[] data) throws InvalidKeyExc
...
+            md5.update(new byte[4]); // Acceptor address
+            md5.update(new byte[4]); // Acceptor address length
+            md5.update(new byte[] {(byte) channelBindingInfo.length, (byte) 0, (byte) 0, (byte) 0});
+            md5.update(channelBindingInfo);
+
+            msvAvChannelBindings = md5.digest();
+        } catch (NoSuchAlgorithmException e) {
+            throw new SQLServerException("Failed to calculate NTLM channel binding hash.", e);
+        } finally {
+            if (null != channelBindingInfo) {
```
machavan, line 670: calculateChannelBindingMD5Hash() nulls channelBindingInfo in its finally, so it can only be called once per instance. That's fine for the current single NTLM auth exchange, but worth a brief comment so a future caller doesn't get a silent null on a second invocation.

machavan (approved):

muskan124947 (approved):

Ananya2 (approved):

PR #2980, opened 2026-07-07
Fix | Login TLS handshake hangs on Android (Conscrypt): read up to maxBytes during SSL handshake

### LINKED ISSUE #2979: Connection hangs during login TLS handshake on Android (Conscrypt)
## Driver version
Reproduced against 13.5.0 built from `main`. The affected code path is unchanged in current GA releases; any recent `jre8`/`jre11` release reproduces on Android.

## SQL Server version
Microsoft SQL Server 2022 (RTM-CU24) (KB5080999) - 16.0.4245.2 (X64) on Linux (Ubuntu 22.04.5 LTS). The server is not relevant — the defect is in client-side TLS-over-TDS framing and reproduces against any SQL Server.

## Client Operating System
Android 14 (API 34), `google_apis` x86_64 emulator. Android uses **Conscrypt** as its default JSSE/TLS provider, so this affects physical devices as well.

## JAVA/JVM version
Android ART (Android 14 / API 34), TLS provider `AndroidOpenSSL` (Conscrypt). For comparison, the same driver connects successfully on desktop OpenJDK (SunJSSE), even with `-ea`.

## Table schema
N/A — the failure occurs during `Connection` establishment (the login-phase TLS handshake), before any statement executes. The repro simply runs `SELECT @@VERSION` after connecting.

## Problem description
On Android, opening a connection hangs during the login-phase TLS handshake and eventually fails. SQL Server performs a TLS handshake during login even when `encrypt=false`, so both `encrypt=false` and `encrypt=true` fail.

Root cause: `TDSChannel.SSLHandshakeInputStream.readInternal(byte[], int, int)` reads **exactly** `maxBytes` (blocking across TDS packets) and always returns `maxBytes`, instead of reading **up to** `maxBytes` and returning the actual count as the `InputStream.read` contract requires.

- SunJSSE reads the handshake in small, exact TLS-record-sized chunks, so it never requests more bytes than are available and the defect is masked.
- Conscrypt issues a single large read (its buffer size, e.g. 16709 bytes). The driver blocks trying to fill 16709 bytes, but the server has already sent its complete handshake flight (~1.5 KB in a single TDS packet) and is now waiting for the client's next flight. This deadlocks until `loginTimeout`.

Suggested fix in `SSLHandshakeInputStream.readInternal`, after `ensureSSLPayload()`:

    int bytesToRead = Math.min(maxBytes, tdsReader.available());
    tdsReader.readBytes(b, offset, bytesToRead);
    return bytesToRead;

Verified to fix both `encrypt=false` and `encrypt=true` on Android (release and debug builds) with no regression on desktop SunJSSE (encrypt on/off, with and without `-ea`).

## Expected behavior
The connection succeeds on Android just as it does on a desktop JVM.

## Actual behavior
- Release build (assertions off, as in a shipped app): the connect hangs, then throws `SQLServerException: ... could not establish a secure connection ... Read timed out`.
- Debug build (assertions on — AGP/d8 force-enables Java `assert` in `debuggable` builds): fails faster with `AssertionError: numMsgsRcvd:1 should be less than numMsgsSent:1` from `TDSReader.readPacket`. Same root cause, surfaced earlier by an assertion that does not hold because the ClientHello has not yet been sent when the premature read occurs.

## Error message/stack trace
Release (assertions off):

    com.microsoft.sqlserver.jdbc.SQLServerException: "encrypt" property is set to "false" and "trustServerCertificate" property is set to "true" but the driver could not establish a secure connection to SQL Server by using Secure Sockets Layer (SSL) encryption: Error: Read timed out ClientConnectionId:...
        at com.microsoft.sqlserver.jdbc.TDSChannel.enableSSL(IOBuffer.java:2008)
        at com.microsoft.sqlserver.jdbc.SQLServerConnection.connectHelper(SQLServerConnection.java:4467)
        ...
    Caused by: java.net.SocketTimeoutException: Read timed out
        at java.net.SocketInputStream.socketRead0(Native Method)
        at com.microsoft.sqlserver.jdbc.TDSChannel$ProxyInputStream.read(IOBuffer.java:1225)
        at com.android.org.conscrypt.ConscryptEngineSocket$SSLInputStream.readFromSocket(ConscryptEngineSocket.java:983)
        at com.android.org.conscrypt.ConscryptEngineSocket.doHandshake(ConscryptEngineSocket.java:236)
        at com.android.org.conscrypt.ConscryptEngineSocket.startHandshake(ConscryptEngineSocket.java:218)
        at com.microsoft.sqlserver.jdbc.TDSChannel.enableSSL(IOBuffer.java:1897)

Debug (assertions on):

    java.lang.AssertionError: numMsgsRcvd:1 should be less than numMsgsSent:1
        at com.microsoft.sqlserver.jdbc.TDSReader.readPacket(IOBuffer.java:6977)
        at com.microsoft.sqlserver.jdbc.TDSReader.readBytes(IOBuffer.java:7236)
        at com.microsoft.sqlserver.jdbc.TDSChannel$SSLHandshakeInputStream.readInternal(IOBuffer.java:985)
        at com.android.org.conscrypt.ConscryptEngineSocket.doHandshake(ConscryptEngineSocket.java:236)
        at com.microsoft.sqlserver.jdbc.TDSChannel.enableSSL(IOBuffer.java:1890)

Behavior matrix (same driver, same server):

| Environment | encrypt=false | encrypt=true |
| --- | --- | --- |
| Desktop OpenJDK (SunJSSE), assertions off | works | works |
| Desktop OpenJDK (SunJSSE), `-ea` | works | works |
| Android release (Conscrypt), assertions off | Read timed out | Read timed out |
| Android debug (Conscrypt), assertions on | AssertionError | AssertionError |

A minimal, self-contained Android instrumented-test reproducer is attached (`mssql-jdbc-android-tls-repro.zip`). It pulls the driver from Maven Central and connects with configurable instrumentation arguments.

## JDBC trace logs
`FINEST` trace of the handshake (annotated) showing the premature second read:

    (SSLHandshakeOutputStream) Starting new TDS packet...
    TDSWriter Writing 8 bytes            # TDS header
    TDSWriter Writing 237 bytes          # ClientHello buffered
    (SSLHandshakeOutputStream) Ignored a request to flush the stream
    (SSLHandshakeInputStream) Reading 16709 bytes...      # Conscrypt one large read
    ensureSSLPayload: No handshake response bytes available. Flushing SSL handshake output stream.
    endMessage: Finishing TDS message    # ClientHello finally sent
    ensureSSLPayload: Reading first packet of SSL handshake response
    ProxyInputStream Reading 8 bytes     # TDS header
    ProxyInputStream Reading 1534 bytes  # server's full ServerHello flight (one TDS packet)
    TDSReader Reading 1534 bytes from offset 0
    ProxyInputStream Reading 8 bytes     # driver blocks trying to read a SECOND packet...
    (15s later) Reading bytes threw exception: Read timed out   # server had nothing more to send

[mssql-jdbc-android-tls-repro.zip](https://github.com/user-attachments/files/29776494/mssql-jdbc-android-tls-repro.zip)

David-Engel:
## Problem description

On Android (which uses Conscrypt as its default TLS provider), establishing a JDBC connection hangs during the login-phase TLS handshake and eventually fails. SQL Server performs a TLS handshake during login even when `encrypt=false`, so both `encrypt=false` and `encrypt=true` are affected. In a shipped (assertions-off) build the connect blocks until `loginTimeout` and throws `SQLServerException: ... could not establish a secure connection ... Read timed out`. In a debuggable build (where the Android Gradle Plugin / d8 force-enables Java assertions) it fails earlier with `AssertionError: numMsgsRcvd:1 should be less than numMsgsSent:1` from `TDSReader.readPacket` — the same root cause, surfaced earlier by an assertion.

Root cause: `TDSChannel.SSLHandshakeInputStream.readInternal(byte[], int, int)` read exactly `maxBytes` (blocking across TDS packets) and always returned `maxBytes`, instead of reading up to `maxBytes` and returning the actual number of bytes read as the `InputStream.read` contract requires. SunJSSE reads the TLS handshake in small, exact record-sized chunks and never requests more than is available, so the defect is masked. Conscrypt issues a single large read (its buffer size, e.g. 16709 bytes); the driver then blocks trying to fill 16709 bytes while the server has already sent its complete handshake flight (~1.5 KB in one TDS packet) and is waiting for the client's next flight — a deadlock.

Fixes #2979

## Fix Description

In `SSLHandshakeInputStream.readInternal`, after `ensureSSLPayload()` guarantees at least one buffered TDS packet, read `Math.min(maxBytes, tdsReader.available())` bytes and return the actual count instead of blocking for the full `maxBytes`. `TDSReader.available()` reports only non-blocking buffered bytes, so the handshake input stream now honors the read-up-to contract and hands each buffered chunk to the SSL engine, which can then advance the handshake. A `FINEST` log records when a short read occurs. There is no behavior change for SunJSSE, which already read within a single packet.

## New Public APIs

None.

## Testing

- Reproduced and fixed with the Android instrumented-test reproducer attached to #2979 (Android 14 / API 34 emulator, Conscrypt). With the fix, both `encrypt=false` and `encrypt=true` connect successfully on both the release (assertions-off) and debug (assertions-on) variants.
- Desktop regression matrix (OpenJDK / SunJSSE) against the same SQL Server: `encrypt=false` and `encrypt=true`, with and without `-ea`, all continue to connect successfully — no change in the common path.
- Justification for the automated-test gap: the defect is TLS-provider-specific. The JVM-based test suite runs on SunJSSE, which reads the handshake in record-sized chunks and therefore cannot reproduce Conscrypt's single-large-read behavior, so a JVM unit/integration test cannot exercise the failing path. The existing encryption/connection integration tests continue to guard the SunJSSE path against regressions. Happy to add a targeted test if maintainers can suggest a supported way to inject a large-read SSL engine into the suite.
- ADO test runs:
   - https://sqlclientdrivers.visualstudio.com/mssql-jdbc/_build/results?buildId=160194&view=results
   - https://sqlclientdrivers.visualstudio.com/mssql-jdbc/_build/results?buildId=161468&view=results

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
@@ -1037,8 +1037,15 @@ final class TDSChannel implements Serializable {
         }
 
         void endMessage() throws SQLServerException {
-            // We should only be asked to end the message if we have started one
-            assert messageStarted;
+            // Nothing to flush if no handshake output has been buffered since the last flush.
+            // With the "read up to maxBytes" behavior in SSLHandshakeInputStream.readInternal, the
+            // input stream may ask us to ensure payload again while reading a handshake response
+            // that spans multiple TDS packets. At that point no new output has been written, so
+            // there is nothing to end. (Previously this method asserted messageStarted, which
+            // failed for multi-packet handshake responses, e.g. large server certificate chains.)
+            if (!messageStarted) {
+                return;
+            }
 
             if (logger.isLoggable(Level.FINEST))
                 logger.finest(logContext + " Finishing TDS message");
--- a/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
@@ -982,13 +982,24 @@ final class TDSChannel implements Serializable {
             ensureSSLPayload();
 
             try {
-                tdsReader.readBytes(b, offset, maxBytes);
+                // Honor the InputStream contract: read UP TO maxBytes, returning the number of
+                // bytes actually read. Previously this blocked until exactly maxBytes were read,
+                // which deadlocks with SSL providers (e.g. Conscrypt on Android) that issue a
+                // single large read for the whole record: the server sends its handshake flight
+                // in fewer bytes and then waits for the client's response, while the driver waits
+                // for more bytes that never come. SunJSSE happened to avoid this by reading in
+                // exact TLS-record-sized chunks.
+                int bytesToRead = Math.min(maxBytes, tdsReader.available());
+                if (bytesToRead < maxBytes && logger.isLoggable(Level.FINEST))
+                    logger.finest(logContext + " Returning " + bytesToRead
+                            + " buffered bytes reported by tdsReader.available() instead of the " + maxBytes
+                            + " requested to avoid blocking");
+                tdsReader.readBytes(b, offset, bytesToRead);
+                return bytesToRead;
             } catch (SQLServerException e) {
                 logger.finer(logContext + " Reading bytes threw exception:" + e.getMessage());
                 throw new IOException(e.getMessage(), e);
             }
-
-            return maxBytes;
         }
     }
 
--- a/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
@@ -989,6 +989,15 @@ final class TDSChannel implements Serializable {
                 // in fewer bytes and then waits for the client's response, while the driver waits
                 // for more bytes that never come. SunJSSE happened to avoid this by reading in
                 // exact TLS-record-sized chunks.
+                //
+                // ensureSSLPayload() only guarantees a packet was read, not that it carried any
+                // payload. Guard against a zero-length payload packet so we never return 0 for a
+                // non-zero request, which would violate the InputStream.read contract and can make
+                // SSL engines busy-spin. Reading the next packet is bounded: readPacket() blocks
+                // for server data and terminates on premature EOF, so this cannot spin.
+                while (maxBytes > 0 && 0 == tdsReader.available())
+                    ensureSSLPayload();
+
                 int bytesToRead = Math.min(maxBytes, tdsReader.available());
                 if (bytesToRead < maxBytes && logger.isLoggable(Level.FINEST))
                     logger.finest(logContext + " Returning " + bytesToRead


src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -982,13 +982,24 @@ private int readInternal(byte[] b, int offset, int maxBytes) throws IOException
...
+                // in fewer bytes and then waits for the client's response, while the driver waits
+                // for more bytes that never come. SunJSSE happened to avoid this by reading in
+                // exact TLS-record-sized chunks.
+                int bytesToRead = Math.min(maxBytes, tdsReader.available());
+                if (bytesToRead < maxBytes && logger.isLoggable(Level.FINEST))
+                    logger.finest(logContext + " Returning " + bytesToRead
+                            + " buffered bytes reported by tdsReader.available() instead of the " + maxBytes
+                            + " requested to avoid blocking");
+                tdsReader.readBytes(b, offset, bytesToRead);
+                return bytesToRead;
```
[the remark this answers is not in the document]
David-Engel: Good catch — fixed in 8bf18502.

You're right that `readInternal` could return 0 for `maxBytes > 0`: `ensureSSLPayload()` only guarantees a packet was read, not that it carried payload, so a zero-length-payload TDS packet would leave `tdsReader.available() == 0` and make us return 0 — violating the `InputStream.read` contract and risking a busy-spin in the SSL engine (and in our own single-byte `read()`).

I now loop until at least one payload byte is buffered when `maxBytes > 0`:

```java
while (maxBytes > 0 && 0 == tdsReader.available())
    ensureSSLPayload();
int bytesToRead = Math.min(maxBytes, tdsReader.available());
```

This is bounded rather than a spin: `readPacket()` blocks for server data and terminates the connection on premature EOF, so each iteration either makes progress or throws. The normal path (payload already present) never iterates, so there's no behavior change for the common case. Verified with a TLS handshake (`encrypt=true`, assertions enabled) against SQL Server 2022.

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -982,13 +982,33 @@ private int readInternal(byte[] b, int offset, int maxBytes) throws IOException
...
+                // in fewer bytes and then waits for the client's response, while the driver waits
+                // for more bytes that never come. SunJSSE happened to avoid this by reading in
+                // exact TLS-record-sized chunks.
+                //
+                // ensureSSLPayload() only guarantees a packet was read, not that it carried any
+                // payload. Guard against a zero-length payload packet so we never return 0 for a
+                // non-zero request, which would violate the InputStream.read contract and can make
+                // SSL engines busy-spin. Reading the next packet is bounded: readPacket() blocks
+                // for server data and terminates on premature EOF, so this cannot spin.
+                while (maxBytes > 0 && 0 == tdsReader.available())
```
machavan, line 998: Hopefully there are no side effects on the existing mechanism/behavior for SunJSSE path.

David-Engel: Oracle and IBM runtime runs showed no regressions. Should be pretty safe.

ADO test runs:
https://sqlclientdrivers.visualstudio.com/mssql-jdbc/_build/results?buildId=160194&view=results
https://sqlclientdrivers.visualstudio.com/mssql-jdbc/_build/results?buildId=161468&view=results

machavan (approved):

Ananya2 (approved):

PR #2988, opened 2026-07-17

funkyjive:
When troubleshooting a problem with a custom accessTokenCallbackClass, the relevant exception may be missed because of a call to getCause().  If the exception doesn't have a cause exception, it will leave no chained exception to the SQLServerException(), effectively eliminating effective troubleshooting information.  If it does have a cause exception, it should still be listed as the logging prints out the stack trace will all the caused by traces.


funkyjive:
think this was made on the wrong release, will try again.

PR #2989, opened 2026-07-17

funkyjive:
When troubleshooting a problem with a custom accessTokenCallbackClass the relevant exception may be missed because of a call to getCause(). If the exception doesn't have a cause exception, it will leave no chained exception to the SQLServerException(), effectively eliminating effective troubleshooting information. If it does have a cause exception, it should still be listed as the logging prints out the stack trace will all the caused by traces.


divang:
Thanks @funkyjive for the contribution and for taking the time to work on this improvement. This functionality has now been merged via PR #[2990](https://github.com/microsoft/mssql-jdbc/pull/2990), and your contribution has been acknowledged in the merged change.

Since PR #2990 is now the merged implementation, I'm closing this PR. Thanks again for your contribution and support of the JDBC project!

PR #2990, opened 2026-07-20

divang:
**Summary**:
This change improves error visibility for custom accessTokenCallbackClass failures by preserving the original exception instead of only its cause.

**What changed:**
In [SQLServerConnection.java, the thrown SQLServerException now uses e instead of e.getCause().
If callback exception has no cause, the real exception is no longer lost.
If callback exception has a cause, full chained details are still available through the original exception.

**Scope:**
One file changed.
No functional auth flow change, only better troubleshooting/debugging context.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
@@ -6945,8 +6945,7 @@ public class SQLServerConnection implements ISQLServerConnection, java.io.Serial
                     } catch (Exception e) {
                         MessageFormat form = new MessageFormat(
                                 SQLServerException.getErrString("R_InvalidAccessTokenCallbackClass"));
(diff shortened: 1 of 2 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java
```diff
@@ -6945,8 +6945,7 @@ void onFedAuthInfo(SqlFedAuthInfo fedAuthInfo, TDSTokenHandler tdsTokenHandler)
                     } catch (Exception e) {
                         MessageFormat form = new MessageFormat(
                                 SQLServerException.getErrString("R_InvalidAccessTokenCallbackClass"));
-                        throw new SQLServerException(form.format(new Object[] { accessTokenCallbackClass }),
```
machavan, line 6948: Lets add a test for this

machavan (approved):

Ananya2 (approved):

muskan124947 (approved):

PR #2991, opened 2026-07-20
Perf: Allocation-free MONEY and DECIMAL processing with scratch-buffer String reads in the ResultSet read path

Ananya2:
## Summary

This PR adds a third batch of targeted ResultSet perf optimizations focused on **`MONEY`/`SMALLMONEY` decoding**, **`BigDecimal` byte encoding**, **synchronous `String` reads**, and **NBCROW null-bitmap initialization**. Each change removes throwaway per-cell allocations or redundant per-column work while preserving the exact decoded value and the existing lazy-decoding architecture.

On wide-row bulk reads, where every column of every row is decoded, these per-cell temporaries (intermediate `BigInteger` objects, per-cell `byte[]` buffers, and repeated per-column method calls) multiply into billions of short-lived objects, creating young-generation GC pressure that shows up as both lower throughput and periodic latency spikes.

> This is part of a series of performance improvement PRs. The type-agnostic changes (DTV state pooling and the length-prefix reader fast path) landed in #2974, and a separate PR covers the `DECIMAL`/`NUMERIC` decoder fast path. This PR is the home for the remaining datatype-specific decode/encode optimizations and the scratch-buffer String read.

**Measured result** on a 601-column x 1M-row x 200-iteration scan with 400 runs (default + fetchSize=512):

|---|---|---|---|
| Average fetch (ms) | 51,776 | **44,157** | **14.72 % faster** |
| Median fetch (ms) | 50,593 | **43,566** | **13.89 % faster** |
| Min fetch (ms) | 47,514 | **40,553** | **14.65 % faster** |
| Max fetch (ms) | 71,026 | **58,180** | **18.09 % faster** |
| Range (ms) | 23,512 | **17,627** | **25.03 % lower** |
| P90 (ms) | 56,235 | **47,433** | **15.65 % lower** |
| P95 (ms) | 60,745 | **50,143** | **17.45 % lower** |
| Standard deviation (ms) | 3,775 | **2,916** | **22.75 % lower** |
| CV % | 7.29 % | **6.60 %** | lower variance |
| Spikes > 55 s | 49 | **7** | **86 % fewer** |
| Spikes > 60 s | 25 | **0** | **100 % eliminated** |
| Spikes > 65 s | 9 | **0** | **100 % eliminated** |
| Throughput (rows/sec) | 19,314 | **22,646** | **17.3 % higher** |

The tail-latency improvement is a notable secondary benefit: fewer allocations slow young-generation growth, reduce minor GC frequency, and shrink the GC-induced latency tail. This PR **eliminates 100 % of the `> 60 s` and `> 65 s` catastrophic outliers**.

## Key improvements

* Decode `MONEY`/`SMALLMONEY` without an intermediate `BigInteger` (both the plaintext and Always Encrypted paths).
* Encode `BigDecimal` to wire bytes without an intermediate `BigInteger.toByteArray()` allocation when the magnitude fits in a `long`.
* Decode short synchronous `getString()` values into the per-reader scratch buffer, avoiding a per-cell `byte[]`.
* Skip the redundant per-column NBCROW bitmap-initialization call once the row is already initialized.
* Preserve existing behavior for large-magnitude values, streaming/adaptive access, and the Always Encrypted decode path.

---

### 1. Allocation-free `MONEY`/`SMALLMONEY` decode (`IOBuffer.java`, `dtv.java`)

`readMoney()` decodes every `MONEY`/`SMALLMONEY` column. The original wrapped the 8-byte (money) or 4-byte (smallmoney) unscaled value in a `BigInteger` before constructing the `BigDecimal`. The money/smallmoney unscaled magnitude always fits in a `long`, so build the `BigDecimal` directly with `BigDecimal.valueOf(long, scale)` and skip the intermediate `BigInteger` allocation.

```java
// IOBuffer.java - readMoney()
long unscaledValue;
switch (valueLength) {
    case 8: // money
        ...
        unscaledValue = ((long) intBitsHi << 32) | (intBitsLo & 0xFFFFFFFFL);
        break;
    case 4: // smallmoney
        ...
        unscaledValue = readInt();
        break;
    ...
}
// money/smallmoney unscaled magnitude always fits in a long.
return DDC.convertBigDecimalToObject(BigDecimal.valueOf(unscaledValue, 4), jdbcType, streamType);
```

The same transformation is applied to the Always Encrypted decode path in `ServerDTVImpl` (`SMALLMONEY` and `MONEY` cases), replacing `new BigDecimal(BigInteger.valueOf(...), 4)` with `BigDecimal.valueOf(..., 4)`.

**Safety.** Value-identical. `BigDecimal.valueOf(long, 4)` produces the same compact, `long`-backed `BigDecimal` the original path produced for these magnitudes.

---

### 2. Allocation-free `BigDecimal` byte encoding (`DDC.java`)

When encoding a `BigDecimal` to its TDS wire representation, the original always called `BigInteger.toByteArray()` to obtain the magnitude bytes. When the non-negative magnitude fits in a `long`, emit its little-endian bytes directly, matching the minimal two's-complement length `toByteArray()` would produce, without allocating the intermediate `byte[]`. Larger magnitudes fall through to the unchanged `BigInteger` path.

```java
int bitLength = bi.bitLength();
if (bitLength < Long.SIZE) {
    long mag = bi.longValue();                 // >= 0 and exact because bitLength < 64
    int numMagBytes = bitLength / 8 + 1;       // == bi.toByteArray().length for non-negative bi
    valueBytes = new byte[numMagBytes + 3];
    int k = 0;
    valueBytes[k++] = (byte) bigDecimalVal.scale();
    valueBytes[k++] = (byte) (numMagBytes + 1); // data length + sign
    valueBytes[k++] = (byte) (isNegative ? 0 : 1);
    for (int i = 0; i < numMagBytes; i++)
        valueBytes[k++] = (byte) (mag >> (8 * i)); // little-endian magnitude
    return valueBytes;
}
```

**Safety.** Byte-identical output for the guarded inputs. `numMagBytes` is computed to match exactly what `BigInteger.toByteArray().length` yields for a non-negative magnitude, and larger magnitudes retain the original path.

---

### 3. Scratch-buffer decode for short synchronous `String` reads (`IOBuffer.java`, `dtv.java`)

Building on #2974's inline `getString()` fast path, this replaces the per-cell `byte[]` allocation with a new `readStringFromBytes()` helper that decodes short values from the per-reader scratch buffer. The scratch buffer is only read by the `String` constructor, so it does not escape.

```java
// IOBuffer.java
final String readStringFromBytes(int valueLength, Charset charset) throws SQLServerException {
    if (valueLength <= valueBytes.length) {
        readBytes(valueBytes, 0, valueLength);
        return new String(valueBytes, 0, valueLength, charset);
    }
    byte[] bytes = new byte[valueLength];
    readBytes(bytes, 0, valueLength);
    return new String(bytes, 0, valueLength, charset);
}
```

```java
// dtv.java - ServerDTVImpl.getValue(), CHAR/VARCHAR/NCHAR/NVARCHAR fast path
convertedValue = tdsReader.readStringFromBytes(valueLength, typeInfo.getCharset());
```

**Safety.** The produced `String` is identical to `new String(bytes, charset)`; values larger than the scratch buffer still allocate a right-sized `byte[]`. The existing guards from #2974 (`valueLength <= 4000`, `streamType == NONE`, `!isAdaptive`, `jdbcType.isTextual()`) are unchanged.

---

### 4. Skip redundant per-column NBCROW initialization (`SQLServerResultSet.java`)

`loadColumn()` called `initializeNullCompressedColumns()` for every column access. For an NBCROW, the null bitmap is read on the first accessed column and a flag is set; subsequent columns re-entered the method only to return immediately. Guard the call with the existing `areNullCompressedColumnsInitialized` flag to skip the method call entirely once the row is initialized.

```java
if (!areNullCompressedColumnsInitialized)
    initializeNullCompressedColumns();
```

**Safety.** Identical behavior. The guard replicates the method's own early-return condition. The flag is reset per row.

---

## Why this is safe to land

1. **No public-API changes.** No new public methods, no signature changes, no contract changes.
2. **All fast paths are guarded and fall back to the original code** whenever the optimization does not apply (large magnitudes, values larger than the scratch buffer, non-textual/streaming/adaptive string access).
3. **Decoded and encoded values are identical.** `BigDecimal.valueOf(long, scale)` and the direct byte-emit path produce the same results as the original `BigInteger`-based code for the guarded inputs.
4. **No shared mutable state escapes.** The scratch buffer is only read by the `String` constructor within `readStringFromBytes()` and never handed to callers.
5. **Defaults unchanged.** No new connection property, no opt-in flag. The optimized paths are used automatically whenever their guard conditions are satisfied.

---

### Methodology

* **Volume:** 1,000,000 rows x 200 iterations per run

|---|---|---|
| Average (ms) | 51,776 | **44,157** |
| Median (ms) | 50,593 | **43,566** |
| Min (ms) | 47,514 | **40,553** |
| Max (ms) | 71,026 | **58,180** |
| Range (ms) | 23,512 | **17,627** |
| P90 (ms) | 56,235 | **47,433** |
| P95 (ms) | 60,745 | **50,143** |
| StdDev (ms) | 3,775 | **2,916** |
| CV % | 7.29 % | **6.60 %** |
| Spikes > 55 s | 49 | **7** |
| Spikes > 60 s | 25 | **0** |
| Spikes > 65 s | 9 | **0** |
| Throughput (rows/sec) | 19,314 | **22,646** |

### Improvement vs Baseline (negative % = faster)

| Config | Avg | Median | Max | P95 | StdDev |
|---|---|---|---|---|---|
| Combined | -14.72 % | -13.89 % | -18.09 % | -17.45 % | -22.75 % |
| Default fetchSize | -14.67 % | -13.97 % | -18.09 % | -15.17 % | -13.37 % |
| fetchSize = 512 | -14.76 % | -13.50 % | -16.93 % | -18.47 % | -30.32 % |

### Observations

1. **~14.7 % mean reduction** (approximately 7.6 seconds faster per fetch), consistent across both fetch-size configurations (-14.67 % default, -14.76 % at 512), confirming the changes target per-cell decoding cost rather than network batching.
2. **100 % of `> 60 s` and `> 65 s` outliers are eliminated** (25 to 0 and 9 to 0), consistent with reduced young-generation GC pressure.
3. **Throughput rises +17.3 %** (19,314 to 22,646 rows/sec).
4. **Reproducible:** three consecutive July 16 datasets landed at -14.89 %, -14.90 %, and -14.72 % (1,056 total runs at ~15 %).

---

## Testing

* **Existing test suite:** all unit and integration tests pass against this PR. The `MONEY`/`SMALLMONEY`, `DECIMAL`/`NUMERIC`, and string decode paths are already exercised across precision/scale combinations, NBCROW null handling, packet-boundary spanning reads, and Always Encrypted columns.
* **Benchmark validation:** the 400-run benchmark above was executed end-to-end against a live SQL Server instance with both the baseline and this PR builds; the ~14.7 % improvement was reproduced across multiple runs.
* The `WideRowReadTests` nested class added in #2974 (reads a fat table of 10 rows x 601 columns and asserts the exact value of every cell via type-specific getters, covering 20 SQL types including both decimal decode paths) continues to cover the changes here.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/DDC.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/DDC.java
@@ -330,6 +330,22 @@ final class DDC {
             if (isNegative)
                 bi = bi.negate();
 
+            // Fast path: magnitude fits in a long, so emit its bytes directly instead of
+            // allocating an intermediate byte[] via BigInteger.toByteArray().
+            int bitLength = bi.bitLength();
+            if (bitLength < Long.SIZE) {
+                long mag = bi.longValue(); // non-negative and exact since bitLength < 64
+                int numMagBytes = bitLength / 8 + 1; // matches bi.toByteArray().length for non-negative bi
+                valueBytes = new byte[numMagBytes + 3];
+                int k = 0;
+                valueBytes[k++] = (byte) bigDecimalVal.scale();
+                valueBytes[k++] = (byte) (numMagBytes + 1); // data length + sign
+                valueBytes[k++] = (byte) (isNegative ? 0 : 1); // 1 = +ve, 0 = -ve
+                for (int i = 0; i < numMagBytes; i++)
+                    valueBytes[k++] = (byte) (mag >> (8 * i)); // little-endian magnitude
+                return valueBytes;
+            }
+
             byte[] unscaledBytes = bi.toByteArray();
 
             valueBytes = new byte[unscaledBytes.length + 3];
--- a/src/main/java/com/microsoft/sqlserver/jdbc/DDC.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/DDC.java
@@ -311,6 +311,20 @@ final class DDC {
         }
     }
 
+    /**
+     * Writes the 3-byte TDS decimal/numeric header (scale, data length including the sign byte, and sign) at the start
+     * of {@code out}. Shared by both encode paths in {@link #convertBigDecimalToBytes} so the header layout cannot
+     * drift between them.
+     *
+     * @return the offset just past the header (always 3)
+     */
+    private static int writeDecimalHeader(byte[] out, int scale, int magnitudeLength, boolean isNegative) {
+        out[0] = (byte) scale;
+        out[1] = (byte) (magnitudeLength + 1); // data length + sign
+        out[2] = (byte) (isNegative ? 0 : 1); // 1 = +ve, 0 = -ve
+        return 3;
+    }
+
     static final byte[] convertBigDecimalToBytes(BigDecimal bigDecimalVal, int scale) {
         byte[] valueBytes;
 
@@ -335,12 +349,11 @@ final class DDC {
             int bitLength = bi.bitLength();
             if (bitLength < Long.SIZE) {
                 long mag = bi.longValue(); // non-negative and exact since bitLength < 64
+                // Zero magnitude is handled here: bitLength() == 0 -> numMagBytes == 1, emitting a single zero byte,
+                // which matches BigInteger.ZERO.toByteArray() == {0}. Keep the + 1 to preserve this.
                 int numMagBytes = bitLength / 8 + 1; // matches bi.toByteArray().length for non-negative bi
                 valueBytes = new byte[numMagBytes + 3];
-                int k = 0;
-                valueBytes[k++] = (byte) bigDecimalVal.scale();
-                valueBytes[k++] = (byte) (numMagBytes + 1); // data length + sign
-                valueBytes[k++] = (byte) (isNegative ? 0 : 1); // 1 = +ve, 0 = -ve
+                int k = writeDecimalHeader(valueBytes, bigDecimalVal.scale(), numMagBytes, isNegative);
                 for (int i = 0; i < numMagBytes; i++)
                     valueBytes[k++] = (byte) (mag >> (8 * i)); // little-endian magnitude
                 return valueBytes;
@@ -349,10 +362,7 @@ final class DDC {
             byte[] unscaledBytes = bi.toByteArray();
 
             valueBytes = new byte[unscaledBytes.length + 3];
-            int j = 0;
-            valueBytes[j++] = (byte) bigDecimalVal.scale();
-            valueBytes[j++] = (byte) (unscaledBytes.length + 1); // data length + sign
-            valueBytes[j++] = (byte) (isNegative ? 0 : 1); // 1 = +ve, 0 = -ve
+            int j = writeDecimalHeader(valueBytes, bigDecimalVal.scale(), unscaledBytes.length, isNegative);
             for (int i = unscaledBytes.length - 1; i >= 0; i--)
                 valueBytes[j++] = unscaledBytes[i];
         }
--- a/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
@@ -7403,6 +7403,21 @@ final class TDSReader implements Serializable {
         return valueBytes;
     }
 
+    /**
+     * Reads {@code valueLength} bytes and decodes them into a String using the given charset. Short values reuse the
+     * per-reader scratch buffer to avoid a per-cell byte[]; the buffer is only read by the String constructor, so it
+     * does not escape.
+     */
+    final String readStringFromBytes(int valueLength, Charset charset) throws SQLServerException {
+        if (valueLength <= valueBytes.length) {
+            readBytes(valueBytes, 0, valueLength);
+            return new String(valueBytes, 0, valueLength, charset);
+        }
+        byte[] bytes = new byte[valueLength];
+        readBytes(bytes, 0, valueLength);
+        return new String(bytes, 0, valueLength, charset);
+    }
+
     final Object readDecimal(int valueLength, TypeInfo typeInfo, JDBCType jdbcType,
             StreamType streamType) throws SQLServerException {
         if (valueLength > valueBytes.length) {
@@ -7418,7 +7433,7 @@ final class TDSReader implements Serializable {
     }
 
     final Object readMoney(int valueLength, JDBCType jdbcType, StreamType streamType) throws SQLServerException {
-        BigInteger bi;
+        long unscaledValue;
         switch (valueLength) {
             case 8: // money
             {
@@ -7432,7 +7447,7 @@ final class TDSReader implements Serializable {
                     return value;
                 }
 
-                bi = BigInteger.valueOf(((long) intBitsHi << 32) | (intBitsLo & 0xFFFFFFFFL));
+                unscaledValue = ((long) intBitsHi << 32) | (intBitsLo & 0xFFFFFFFFL);
                 break;
             }
 
@@ -7443,7 +7458,7 @@ final class TDSReader implements Serializable {
                     return value;
                 }
 
-                bi = BigInteger.valueOf(readInt());
+                unscaledValue = readInt();
                 break;
 
             default:
@@ -7451,7 +7466,9 @@ final class TDSReader implements Serializable {
                 return null;
         }
 
-        return DDC.convertBigDecimalToObject(new BigDecimal(bi, 4), jdbcType, streamType);
+        // money/smallmoney unscaled magnitude always fits in a long, so build the BigDecimal
+        // directly and skip the intermediate BigInteger.
+        return DDC.convertBigDecimalToObject(BigDecimal.valueOf(unscaledValue, 4), jdbcType, streamType);
     }
 
     final Object readReal(int valueLength, JDBCType jdbcType, StreamType streamType) throws SQLServerException {
--- a/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
@@ -7404,9 +7404,12 @@ final class TDSReader implements Serializable {
     }
 
     /**
-     * Reads {@code valueLength} bytes and decodes them into a String using the given charset. Short values reuse the
-     * per-reader scratch buffer to avoid a per-cell byte[]; the buffer is only read by the String constructor, so it
-     * does not escape.
+     * Reads {@code valueLength} bytes and decodes them into a String using the given charset. Values that fit in the
+     * per-reader scratch buffer ({@code valueBytes}, 256 bytes) are decoded in place to avoid a per-cell byte[]; the
+     * buffer is only read by the String constructor, so it does not escape. Longer values (the caller admits up to
+     * 4000 bytes) allocate a right-sized byte[] instead. Either way the value bypasses the stream-wrapper path, so the
+     * larger allocation-free win (no SimpleInputStream / marks) still applies; only the byte[] reuse is limited to
+     * values &le; 256 bytes.
      */
     final String readStringFromBytes(int valueLength, Charset charset) throws SQLServerException {
         if (valueLength <= valueBytes.length) {
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResultSet.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResultSet.java
@@ -837,7 +837,10 @@ public class SQLServerResultSet implements ISQLServerResultSet, java.io.Serializ
     private Column loadColumn(int index) throws SQLServerException {
         assert 1 <= index && index <= columns.length;
 
-        initializeNullCompressedColumns();
+        // Skip the call once the row's null-compressed columns are initialized. For an NBCROW the
+        // bitmap is read on the first accessed column and the flag is set; the flag is reset per row.
+        if (!areNullCompressedColumnsInitialized)
+            initializeNullCompressedColumns();
 
         // Skip any columns between the last indexed column and the target column,
         // retaining their values so they can be retrieved later.
--- a/src/main/java/com/microsoft/sqlserver/jdbc/dtv.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/dtv.java
@@ -3627,17 +3627,17 @@ final class ServerDTVImpl extends DTVImpl {
                 }
 
             case SMALLMONEY:
-                return DDC.convertMoneyToObject(new BigDecimal(BigInteger.valueOf(Util.readInt(decryptedValue, 4)), 4),
+                return DDC.convertMoneyToObject(BigDecimal.valueOf(Util.readInt(decryptedValue, 4), 4),
                         JDBCType.VARBINARY == jdbcType ? baseSSType.getJDBCType() : jdbcType, // use jdbc type from
                                                                                               // baseTypeInfo if using
                                                                                               // getObject()
                         streamGetterArgs.streamType, 4);
 
             case MONEY:
-                BigInteger bi = BigInteger.valueOf(((long) Util.readInt(decryptedValue, 0) << 32)
-                        | (Util.readInt(decryptedValue, 4) & 0xFFFFFFFFL));
+                long moneyUnscaled = ((long) Util.readInt(decryptedValue, 0) << 32)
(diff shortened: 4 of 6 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/DDC.java
```diff
@@ -330,6 +330,22 @@ static final byte[] convertBigDecimalToBytes(BigDecimal bigDecimalVal, int scale
             if (isNegative)
                 bi = bi.negate();
 
+            // Fast path: magnitude fits in a long, so emit its bytes directly instead of
+            // allocating an intermediate byte[] via BigInteger.toByteArray().
+            int bitLength = bi.bitLength();
+            if (bitLength < Long.SIZE) {
+                long mag = bi.longValue(); // non-negative and exact since bitLength < 64
+                int numMagBytes = bitLength / 8 + 1; // matches bi.toByteArray().length for non-negative bi
```
divang, line 354: This correctly handles bi == 0 (bitLength() -> 0, so numMagBytes = 1, emitting one zero byte, matching BigInteger.ZERO.toByteArray(), which is {0}). Since zero is the least obvious case and the most fragile under future refactors of the + 1, could we add an inline comment noting that zero magnitude is intentionally handled here, and ideally a unit-test assertion for it?

src/main/java/com/microsoft/sqlserver/jdbc/DDC.java
```diff
@@ -330,6 +330,22 @@ static final byte[] convertBigDecimalToBytes(BigDecimal bigDecimalVal, int scale
...
+                valueBytes = new byte[numMagBytes + 3];
+                int k = 0;
+                valueBytes[k++] = (byte) bigDecimalVal.scale();
+                valueBytes[k++] = (byte) (numMagBytes + 1); // data length + sign
+                valueBytes[k++] = (byte) (isNegative ? 0 : 1); // 1 = +ve, 0 = -ve
```
divang: The sign byte (byte) (isNegative ? 0 : 1) and the length header are now written in two places - here in the fast path and again at lines 353 - 355 in the BigInteger.toByteArray() fall through. If these ever drift, only large-magnitude values would break, which is hard to catch. Consider extracting the shared scale/length/sign header write so the two branches can not diverge.

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -7282,6 +7282,21 @@ final byte[] readWrappedBytes(int valueLength) throws SQLServerException {
...
+     */
+    final String readStringFromBytes(int valueLength, Charset charset) throws SQLServerException {
+        if (valueLength <= valueBytes.length) {
+            readBytes(valueBytes, 0, valueLength);
+            return new String(valueBytes, 0, valueLength, charset);
+        }
+        byte[] bytes = new byte[valueLength];
+        readBytes(bytes, 0, valueLength);
+        return new String(bytes, 0, valueLength, charset);
+    }
```
divang, line 7422: The scratch-buffer reuse is gated on valueLength <= valueBytes.length here, while the caller in dtv.java gates on valueLength <= 4000. valueBytes is new byte[256] (see line 6851), so any value between 257 and 4000 bytes silently takes the new byte[valueLength] fallback rather than the scratch buffer. Worth confirming this is intended - with a 256-byte scratch buffer the allocation-free path only applies to fairly short strings, so the optimization may fire less often than the PR description implies. Consider documenting the relationship between the 256-byte buffer and the 4000-byte caller guard.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResultSet.java
```diff
@@ -837,7 +837,10 @@ private void initializeNullCompressedColumns() throws SQLServerException {
     private Column loadColumn(int index) throws SQLServerException {
         assert 1 <= index && index <= columns.length;
 
-        initializeNullCompressedColumns();
+        // Skip the call once the row's null-compressed columns are initialized. For an NBCROW the
+        // bitmap is read on the first accessed column and the flag is set; the flag is reset per row.
+        if (!areNullCompressedColumnsInitialized)
```
machavan, line 842: areNullCompressedColumnsInitialized is already being checked in the method initializeNullCompressedColumns  at the start ?

src/test/java/com/microsoft/sqlserver/jdbc/UtilTest.java
```diff
@@ -101,4 +103,48 @@ private void writeAndReadLong(long valueToTest) {
...
+                new BigDecimal(new BigInteger("123456789012345678901234567890"), 4)};
+
+        for (BigDecimal v : values) {
+            assertArrayEquals(referenceBigDecimalBytes(v), DDC.convertBigDecimalToBytes(v, v.scale()),
+                    "encoding mismatch for " + v);
+        }
+    }
+
+    /** Reference TDS decimal encoding using the straightforward BigInteger.toByteArray() path. */
+    private static byte[] referenceBigDecimalBytes(BigDecimal bigDecimalVal) {
```
machavan, line 131: Lets have a functional unit test, if we could adapt the crd app unit test to accomodate this , that will be good too

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResultSet.java
Ananya2: Yes, the check already exists inside `initializeNullCompressedColumns()`, so this doesn't change the behavior. The intent was simply to avoid repeatedly invoking `initializeNullCompressedColumns()` on the `loadColumn()` hot path. For an NBCROW, the bitmap is read only once per row, and subsequent calls immediately return after checking the flag. This guard skips those redundant method calls once the row is initialized.

src/test/java/com/microsoft/sqlserver/jdbc/UtilTest.java
Ananya2: A functional round-trip test is already included in this PR series: ResultSetTest.WideRowReadTests.testWideRowMultiTypeRead. It's a 601-column wide table that inserts and reads back every value with exact-value assertions and a small fetch size to force multiple round-trips. It covers the paths optimized here, including decimal(18,0), decimal(38,4), money/smallmoney, and the string read path. Combined with the byte-level referenceBigDecimalBytes tests in UtilTest, this provides both end-to-end functional coverage and encoding validation.

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -7432,7 +7450,7 @@ final Object readMoney(int valueLength, JDBCType jdbcType, StreamType streamType
                     return value;
                 }
 
-                bi = BigInteger.valueOf(((long) intBitsHi << 32) | (intBitsLo & 0xFFFFFFFFL));
+                unscaledValue = ((long) intBitsHi << 32) | (intBitsLo & 0xFFFFFFFFL);
```
divang, line 7453: The switch from new BigDecimal(BigInteger, 4) to BigDecimal.valueOf(unscaledValue, 4) has no test covering the decode path. Could we add a round-trip test for both money (8-byte) and smallmoney (4-byte) that asserts negative values decode correctly - smallmoney depends on readInt() being sign-extended into a long, and money on ((long) intBitsHi << 32) | (intBitsLo & 0xFFFFFFFFL)? Long.MIN_VALUE, -0.0001, and Integer.MIN_VALUE (smallmoney) are the risky boundary values, since an accidental unsigned read would silently turn negative amounts into large positives.

src/main/java/com/microsoft/sqlserver/jdbc/dtv.java
```diff
@@ -3627,17 +3627,17 @@ Object denormalizedValue(byte[] decryptedValue, JDBCType jdbcType, TypeInfo base
...
                         streamGetterArgs.streamType, 4);
 
             case MONEY:
-                BigInteger bi = BigInteger.valueOf(((long) Util.readInt(decryptedValue, 0) << 32)
-                        | (Util.readInt(decryptedValue, 4) & 0xFFFFFFFFL));
+                long moneyUnscaled = ((long) Util.readInt(decryptedValue, 0) << 32)
+                        | (Util.readInt(decryptedValue, 4) & 0xFFFFFFFFL);
 
-                return DDC.convertMoneyToObject(new BigDecimal(bi, 4),
+                return DDC.convertMoneyToObject(BigDecimal.valueOf(moneyUnscaled, 4),
```
divang, line 3640: The same BigInteger -> BigDecimal.valueOf change on the Always Encrypted decode path isn't tested, and it's a separate code path from readMoney(), so those tests won't protect it. Could we add coverage here specifically for negative SMALLMONEY, where Util.readInt(decryptedValue, 4) is sign-extended? Without it, an encrypted money column could regress independently with no failing test.

src/main/java/com/microsoft/sqlserver/jdbc/IOBuffer.java
```diff
@@ -7403,6 +7403,24 @@ final byte[] readWrappedBytes(int valueLength) throws SQLServerException {
...
+    /**
+     * Reads {@code valueLength} bytes and decodes them into a String using the given charset. Values that fit in the
+     * per-reader scratch buffer ({@code valueBytes}, 256 bytes) are decoded in place to avoid a per-cell byte[]; the
+     * buffer is only read by the String constructor, so it does not escape. Longer values (the caller admits up to
+     * 4000 bytes) allocate a right-sized byte[] instead. Either way the value bypasses the stream-wrapper path, so the
+     * larger allocation-free win (no SimpleInputStream / marks) still applies; only the byte[] reuse is limited to
+     * values &le; 256 bytes.
+     */
+    final String readStringFromBytes(int valueLength, Charset charset) throws SQLServerException {
+        if (valueLength <= valueBytes.length) {
```
divang, line 7415: This method has two branches - the in-place scratch buffer for valueLength <= 256 and the right-sized allocation above that - but nothing asserts they produce identical Strings. Could we add tests just under 256 bytes, exactly 256, and above 256 (up to 4000), including a multibyte charset with a character straddling the 256-byte boundary? That boundary is where a subtle decode difference between the two branches would surface.

src/main/java/com/microsoft/sqlserver/jdbc/dtv.java
```diff
@@ -3875,26 +3875,22 @@ Object getValue(DTV dtv, JDBCType jdbcType, int scale, InputStreamGetterArgs str
...
+                    // All guards are required for byte-identical semantics with the stream path:
+                    //   valueLength in (0, 4000] - skip zero-length and overly large values
+                    //   streamType == NONE       - stream getters need the actual InputStream
+                    //   !isAdaptive              - adaptive paths wrap the stream for user code
                     //   - jdbcType.isTextual()      : non-textual targets need convertStringToObject
                     //                                 to trim/parse the value
-                    //   - CLOB/NCLOB excluded        : although textual, getClob()/getNClob() must
+                    //   - CLOB/NCLOB excluded       : although textual, getClob()/getNClob() must
                     //                                 return a SQLServerClob/SQLServerNClob, not a String
                     if (valueLength > 0 && valueLength <= 4000
```
divang, line 3888: The guard set (valueLength in (0,4000], streamType == NONE, !isAdaptive, isTextual(), excluding CLOB/NCLOB) decides between the fast path and the stream fallback, but no test confirms the fallback still triggers for the excluded cases. Could we assert byte-identical results for valueLength == 0, valueLength > 4000, adaptive access, and CLOB/NCLOB targets? Otherwise a future change to a guard could silently route these through the fast path and return wrong or wrongly-typed values.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResultSet.java
divang, line 842: Guarding initializeNullCompressedColumns() behind !areNullCompressedColumnsInitialized isn't covered by any test. Could we add an NBCROW test that reads multiple columns in a row, including out-of-order access, and asserts null vs non-null values are still correct after the first accessed column initializes the bitmap? If the flag were ever set before the bitmap is fully read or not reset per row, later columns would report the wrong null state - data corruption that only shows on NBCROW rows.

divang (approved):

muskan124947 (approved):

PR #2993, opened 2026-07-23
Enhance Always Encrypted VSM/HGS enclave attestation validation

Ananya2:
**Summary**

This change enhances the Always Encrypted v2 enclave attestation flow for the VSM/HGS provider (`enclaveAttestationProtocol=HGS`) by adding validation of the enclave public key binding, bringing it in line with the AAS provider.

The signed VBS enclave report carries a commitment to the enclave public key in its Enclave Data field. This change validates that the commitment matches the enclave public key used for the session, ensuring the verified attestation report corresponds to the key used during key exchange. The AAS path already performs an equivalent validation.

**Notes**

No public API or connection-string changes. Behavior is unchanged for genuine HGS/VBS deployments (verified end-to-end against a live HGS environment). AAS deployments are unaffected.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResource.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResource.java
@@ -535,6 +535,7 @@ public final class SQLServerResource extends ListResourceBundle {
         {"R_InvalidHealthCert", "Enclave attestation failed, the health report certificate provided in the enclave was not signed by the HGS - see https://go.microsoft.com/fwlink/?linkid=2160553 for more details."},
         {"R_InvalidSignedStatement", "Enclave attestation failed, the statement bytes were not signed by the health certificate - see https://go.microsoft.com/fwlink/?linkid=2160553 for more details."},
         {"R_InvalidDHKeySignature", "Enclave attestation failed, the DH public key signature can't be verified with the enclave public key - see https://go.microsoft.com/fwlink/?linkid=2157649 for more details."},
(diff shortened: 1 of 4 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/SQLServerResource.java
```diff
@@ -535,6 +535,7 @@ protected Object[][] getContents() {
         {"R_InvalidHealthCert", "Enclave attestation failed, the health report certificate provided in the enclave was not signed by the HGS - see https://go.microsoft.com/fwlink/?linkid=2160553 for more details."},
         {"R_InvalidSignedStatement", "Enclave attestation failed, the statement bytes were not signed by the health certificate - see https://go.microsoft.com/fwlink/?linkid=2160553 for more details."},
         {"R_InvalidDHKeySignature", "Enclave attestation failed, the DH public key signature can't be verified with the enclave public key - see https://go.microsoft.com/fwlink/?linkid=2157649 for more details."},
```
machavan, line 537: Could we reuse the existing error  R_InvalidDHKeySignature ? or is it used for some other purpose ?

Ananya2: @machavan - R_InvalidDHKeySignature is used for a different validation (validateDHPublicKey()) and is shared by both the VSM/HGS and AAS providers. Although both checks involve enclavePK, they validate different trust relationships. validateDHPublicKey() verifies that the DH public key was signed by enclavePK, assuming enclavePK is already trusted whereas the new check verifies that enclavePK itself is the key committed to by the signed attestation report (EnclaveData).

machavan (approved):

muskan124947 (approved):

PR #3002, opened 2026-08-03
Fix getGeneratedKeys() returns a non-empty ResultSet for tables without generated keys

### LINKED ISSUE #3000: getGeneratedKeys() incorrectly returns a non-empty ResultSet for tables without generated keys.
## Driver version
13.4.0.jre11

## SQL Server version

- Microsoft SQL Server 2022 (RTM-CU25) (KB5081477) - 16.0.4255.1 (X64)   Apr 23 2026 22:38:54   Copyright (C) 2022 Microsoft Corporation  Express Edition (64-bit) on Windows 10 Pro 10.0 <X64> (Build 26200: ) (Hypervisor) 
- Microsoft SQL Server 2022 (RTM-CU25) (KB5081477) - 16.0.4255.1 (X64) 
	Apr 23 2026 22:38:54 
	Enterprise Edition (64-bit) on Windows Server 2022 Datacenter 10.0 <X64> (Build 20348: ) (Hypervisor)

## Client Operating System
Windows / Linux

## JAVA/JVM version

```
openjdk version "21.0.8" 2025-07-15 LTS
OpenJDK Runtime Environment Temurin-21.0.8+9 (build 21.0.8+9-LTS)
OpenJDK 64-Bit Server VM Temurin-21.0.8+9 (build 21.0.8+9-LTS, mixed mode, sharing)
```

## Table schema

```sql
CREATE TABLE TEST_TABLE (
	ID INT NOT NULL
);
```

## Problem description

Incorrect return when using `getGeneratedKeys()` on a `Statement`. It returns a non-empty `ResultSet` for a table without any generated keys.

## Expected behavior

According to the JDBC Javadoc, `If this Statement object did not generate any keys, an empty ResultSet object is returned.` Therefore, I expected `getGeneratedKeys()` to return a `ResultSet` for which calling `next()` would return false.

(Link : https://docs.oracle.com/en/java/javase/21/docs/api/java.sql/java/sql/Statement.html#getGeneratedKeys())

## Actual behavior

Calling `next()` on the `ResultSet` should return false

## Error message/stack trace

N/A

Little snippet :
```java
import java.sql.Connection;
import java.sql.PreparedStatement;
import java.sql.ResultSet;
import java.sql.SQLException;
import java.sql.Statement;

public class TestGeneratedKey {
	public static void main(String[] args) throws SQLException {
		String url = "jdbc:sqlserver://localhost:1433;" +
				"databaseName=test;" +
				"trustServerCertificate=true";

		String user = "test_usr";
		String password = "¤¤¤¤¤¤¤¤¤¤";

		try (Connection c = DriverManager.getConnection(url, user, password)) {
			String insertSQL = "INSERT INTO TEST_TABLE(ID) VALUES(?)";

			try (PreparedStatement preparedStatement = c.prepareStatement(insertSQL, Statement.RETURN_GENERATED_KEYS)) {
				preparedStatement.setInt(1, 123456);

				int affectedRows = preparedStatement.executeUpdate();

				if (affectedRows == 0) {
					throw new SQLException("Insert entity failed, no rows affected.");
				}

				try (ResultSet generatedKeys = preparedStatement.getGeneratedKeys()) {
					if (generatedKeys.next()) { // that should return false !
						System.out.println(generatedKeys.getObject(1));
						System.out.println("EXPECTED NOT GO HERE !!!");
					}
				}
			}
		}
	}
}
```

## JDBC trace logs
N/A

Ananya2:
## Problem

When executing an `INSERT` with `Statement.RETURN_GENERATED_KEYS` against a table that has **no identity/generated key column**, `getGeneratedKeys()` returned a **non-empty** `ResultSet` — a single row whose only value was `null`:

```java
CREATE TABLE TEST_TABLE (ID INT NOT NULL); // no identity column

try (PreparedStatement ps = c.prepareStatement(
        "INSERT INTO TEST_TABLE(ID) VALUES(?)", Statement.RETURN_GENERATED_KEYS)) {
    ps.setInt(1, 123456);
    ps.executeUpdate();

    try (ResultSet keys = ps.getGeneratedKeys()) {
        keys.next();          // returned true  — should be false
        keys.getObject(1);    // returned null
    }
}
```

Per the [JDBC specification](https://docs.oracle.com/en/java/javase/21/docs/api/java.sql/java/sql/Statement.html#getGeneratedKeys()), if the `Statement` did not generate any keys, an **empty** `ResultSet` must be returned, so that a call to `ResultSet.next()` returns `false`.

## Root cause

The driver does not receive generated keys inline with the `INSERT`. Instead, for an `INSERT` executed with `RETURN_GENERATED_KEYS`, it appends an internal query to the user's statement and exposes that query's result through `getGeneratedKeys()`:

```java
// SQLServerStatement.java
final static String IDENTITY_QUERY = " select SCOPE_IDENTITY() AS GENERATED_KEYS";
```

`SCOPE_IDENTITY()` returns the last identity value generated in the current scope. When the target table has **no identity column**, `SCOPE_IDENTITY()` evaluates to `NULL` — but the `SELECT` still produces **one row** containing that `NULL`. That row was surfaced verbatim through `getGeneratedKeys()`, so `next()` returned `true` and `getObject(1)` returned `null`, violating the spec.

## Fix

Add a `WHERE` clause to the internal identity query so it yields a row **only** when a key was actually generated:

```java
final static String IDENTITY_QUERY =
        " select SCOPE_IDENTITY() AS GENERATED_KEYS where SCOPE_IDENTITY() is not null";
```

- Table **with** identity column → `SCOPE_IDENTITY()` is non-null → the predicate passes → one row is returned (**unchanged behavior**).
- Table **without** identity column → `SCOPE_IDENTITY()` is `NULL` → the predicate filters it out → **zero rows** → empty `ResultSet` → `next()` returns `false` (**spec-compliant**).

## Tests

- Added `testGetGeneratedKeysNoIdentityColumn` in `StatementTest` (`TCStatementParam`), covering both the `PreparedStatement` and `Statement` paths, asserting that `getGeneratedKeys().next()` returns `false` for a table without generated keys.
- Added regression coverage in StatementTest (TCStatementParam) and MultiStatementResultTraversalTest (GeneratedKeysMatrix) validating that getGeneratedKeys() returns an empty ResultSet for tables without a generated key, across:
    1. Statement and PreparedStatement paths, via execute, executeUpdate, and executeLargeUpdate.
    2. Multi-row and batched inserts.
    3. Non-identity server-generated columns (DEFAULT NEWID(), computed columns).
    4. The SCOPE_IDENTITY vs @@IDENTITY case — a trigger inserting into a separate identity table must not leak its key.
    5. A positive check that identity tables still return the correct generated key (unchanged behavior).

Fixes [#3000](https://github.com/microsoft/mssql-jdbc/issues/3000).

### DIFF
--- a/src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/StatementTest.java
+++ b/src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/StatementTest.java
@@ -1826,6 +1826,243 @@ public class StatementTest extends AbstractTest {
             }
         }
 
+        /**
+         * Tests that an INSERT into a table that has an IDENTITY column still returns a non-empty
+         * generated-keys ResultSet with the correct key value, covering the positive counterpart to
+         * testGetGeneratedKeysNoIdentityColumn.
+         *
+         * Regression guard for GitHub issue #3000.
+         *
+         * @throws Exception
+         */
+        @Test
+        @Tag(Constants.xAzureSQLDW)
+        public void testGetGeneratedKeysWithIdentityColumn() throws Exception {
+            try (Connection con = getConnection(); Statement stmt = con.createStatement()) {
+                stmt.executeUpdate(
+                        "CREATE TABLE " + tableName + " (ID int NOT NULL IDENTITY(1,1) PRIMARY KEY, col2 varchar(32))");
+
+                // PreparedStatement path - must surface the generated identity value.
+                try (PreparedStatement pstmt = con.prepareStatement("INSERT INTO " + tableName + " (col2) VALUES (?)",
+                        Statement.RETURN_GENERATED_KEYS)) {
+                    pstmt.setString(1, "a");
+                    assertEquals(1, pstmt.executeUpdate(), TestResource.getResource("R_valueNotMatch"));
+
+                    try (ResultSet generatedKeys = pstmt.getGeneratedKeys()) {
+                        assertTrue(generatedKeys.next(),
+                                "getGeneratedKeys() must return a row for a table with an identity column");
+                        assertEquals(1, generatedKeys.getInt(1), "first identity value must be 1");
+                        assertFalse(generatedKeys.next(), "exactly one generated key must be returned");
+                    }
+                }
+
+                // Statement path - must surface the generated identity value.
+                assertEquals(1, stmt.executeUpdate("INSERT INTO " + tableName + " (col2) VALUES ('b')",
+                        Statement.RETURN_GENERATED_KEYS), TestResource.getResource("R_valueNotMatch"));
+
+                try (ResultSet generatedKeys = stmt.getGeneratedKeys()) {
+                    assertTrue(generatedKeys.next(),
+                            "getGeneratedKeys() must return a row for a table with an identity column");
+                    assertEquals(2, generatedKeys.getInt(1), "second identity value must be 2");
+                    assertFalse(generatedKeys.next(), "exactly one generated key must be returned");
+                }
+            }
+        }
+
+        /**
+         * Tests that a multi-row INSERT into a table without an identity column returns an empty
+         * generated-keys ResultSet and does not surface a single NULL row.
+         *
+         * Regression guard for GitHub issue #3000.
+         *
+         * @throws Exception
+         */
+        @Test
+        @Tag(Constants.xAzureSQLDW)
+        public void testGetGeneratedKeysNoIdentityColumnMultiRowInsert() throws Exception {
+            try (Connection con = getConnection(); Statement stmt = con.createStatement()) {
+                stmt.executeUpdate("CREATE TABLE " + tableName + " (ID int NOT NULL)");
+
+                try (PreparedStatement pstmt = con.prepareStatement(
+                        "INSERT INTO " + tableName + " (ID) VALUES (?), (?)", Statement.RETURN_GENERATED_KEYS)) {
+                    pstmt.setInt(1, 1);
+                    pstmt.setInt(2, 2);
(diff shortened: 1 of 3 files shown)


src/test/java/com/microsoft/sqlserver/jdbc/unit/statement/StatementTest.java
```diff
@@ -1783,6 +1783,49 @@ public void testRowError() throws Exception {
             }
         }
 
+        /**
```
divang, line 1786: worthwhile test scenarios consider adding:

Positive regression in the same test area - an INSERT into a table with an IDENTITY column asserting next()==true and the correct key value, right next to the negative case, so both branches of the new predicate are covered in one place.

Batch inserts (addBatch/executeBatch with RETURN_GENERATED_KEYS) into a no-identity table - batching uses a different code path and is a common real-world usage.

executeLargeUpdate(sql, RETURN_GENERATED_KEYS) into a no-identity table - the large-update path is separate from executeUpdate.

Multi-row INSERT (INSERT ... VALUES (?),(?)) into a no-identity table - verify it stays empty and doesn't surface a single NULL row.

Table with a non-identity default/computed column (e.g. DEFAULT NEWID() or a computed column) - confirm these are correctly treated as "no generated key" (they don't populate SCOPE_IDENTITY()), matching driver semantics.

INSERT that fires an AFTER trigger which itself inserts into an identity table - verify SCOPE_IDENTITY() (scope-limited, unlike @@IDENTITY) still yields an empty ResultSet for the outer no-identity table. This is the classic SCOPE_IDENTITY vs @@IDENTITY gotcha and a valuable guard.

divang (approved):

muskan124947 (approved):

PR #3004, opened 2026-08-03
Validate reflective class names for connection properties

divang:
**Description:**
This change adds early validation for reflective class-name values supplied through connection properties such as socketFactoryClass and accessTokenCallbackClass. Malformed values are now rejected before reflective loading, which prevents confusing downstream failures and improves security posture.

**What changed**
Added validation in Util.newInstance to ensure the supplied value is a valid Java binary class name.
Added a dedicated driver error message for invalid reflective class names.
Added regression tests covering malformed socketFactoryClass and accessTokenCallbackClass values.

**Why**
Previously, malformed class-name values could reach reflective loading and fail later with less informative errors. This patch fails fast and makes the behavior explicit.

**Validation**
Verified with Maven compile
Verified with targeted regression tests: 2 tests run


src/main/java/com/microsoft/sqlserver/jdbc/Util.java
```diff
@@ -1038,6 +1040,45 @@ static <T> T newInstance(Class<?> returnType, String className, String construct
...
+            return false;
+        }
+
+        String[] tokens = className.split("\\.", -1);
+        for (String token : tokens) {
+            if (token.isEmpty()) {
+                return false;
+            }
+
+            int cp = token.codePointAt(0);
```
machavan: Lets find a way to validate the entire classname is a valid java identifer at once ?

src/main/java/com/microsoft/sqlserver/jdbc/Util.java
```diff
@@ -1038,6 +1040,45 @@ static <T> T newInstance(Class<?> returnType, String className, String construct
...
+            return;
+        }
+
+        String propertyName = (null != msgArgs && msgArgs.length > 0 && null != msgArgs[0]) ? msgArgs[0].toString()
+                : "unknown";
+        MessageFormat form = new MessageFormat(SQLServerException.getErrString("R_invalidClassNameForProperty"));
+        throw new IllegalArgumentException(form.format(new Object[] { propertyName, className }));
+    }
+
+    private static boolean isValidJavaBinaryClassName(String className) {
```
Ananya2, line 1053: We can simplify this to a single whole-string validation using a precompiled regex:

```
private static final Pattern JAVA_BINARY_CLASS_NAME = Pattern.compile(
    "\\p{javaJavaIdentifierStart}\\p{javaJavaIdentifierPart}*(\\.\\p{javaJavaIdentifierStart}\\p{javaJavaIdentifierPart}*)*");
```

Then validate with `JAVA_BINARY_CLASS_NAME.matcher(name).matches()`. This validates the entire qualified class name at once, naturally rejects empty segments, and preserves the same Unicode semantics as `Character.isJavaIdentifier*`.

src/test/java/com/microsoft/sqlserver/jdbc/SQLServerConnectionTest.java
```diff
@@ -1913,6 +1913,30 @@ public void testInvalidAccessTokenCallbackClassPreservesCause() throws Exception
         assertNotNull(exception.getCause());
     }
 
+    @Test
```
Ananya2: Please add regression tests for `trustManagerClass` as well, since it follows the same `Util.newInstance` path as `socketFactoryClass` and `accessTokenCallbackClass` (`IOBuffer.java:1764`).

It would also be good to cover:
- Invalid class name edge cases: `null`, empty/blank values, `com..example.Foo`, leading/trailing `.`, leading digits, embedded spaces, `/`, `:`, and `http://...`.
- A valid class name that fails the expected type check (e.g. `java.lang.String` for `SocketFactory`) to verify that the syntax validation doesn't weaken the existing `isAssignableFrom` check.

muskan124947:
`testUtilNewInstanceRejectsInvalidSocketFactoryClassName` and `testUtilNewInstanceRejectsInvalidAccessTokenCallbackClassName` are added to `SQLServerConnectionTest`, which extends `AbstractTest` and requires SQL Server test config — yet they only exercise `Util.newInstance`. They also duplicate the `UtilTest` parameterized coverage. Move them to `UtilTest` (no DB required) or delete them.

Ananya2 (approved):

machavan (approved):

PR #3005, opened 2026-08-04
Update vector tests for server-side implicit FLOAT16/FLOAT32 conversion

muskan124947:
## Summary

SQL Server now supports **implicit and explicit conversion between FLOAT16 and FLOAT32 vector subtypes**. This PR updates existing tests that expected `SQLException` for cross-type operations to instead expect success, and adds comprehensive new test coverage for all conversion scenarios.

## Motivation

The server-side change enables direct conversion between `vector(N, float16)` and `vector(N, float32)` without requiring intermediate `varchar` conversion. This improves usability and performance for vector operations across different precision levels.

Previously, cross-type operations like `INSERT INTO f32_table SELECT * FROM f16_table` or `CAST(f16_col AS VECTOR(3, float32))` would fail with:
> "Conversion of vector from data type float16 to float32 is not allowed."

With the server update, these operations now succeed with implicit conversion.

### Updated existing tests (expect success instead of failure)

- **VectorFloat16Test.java**: Cross-type INSERT...SELECT and setObject tests now verify data arrives correctly instead of catching `SQLException`.
- **VectorFloat16BulkCopyTest.java**: Bulk copy cross-type tests (`testBulkCopyFloat16ToFloat32Table`, `testBulkCopyFloat32ToFloat16Table`, `testMixedBulkCopyColumnMappingSwapped`) now verify successful transfer with implicit conversion.
- **VectorTest.java**: Updated expected error message for exceeding max vector dimensions.

### New test coverage

| Test | Scenario |
|------|----------|
| `testExplicitCastFloat16ToFloat32` | `CAST(v AS VECTOR(3, float32))` with data assertion |
| `testExplicitCastFloat32ToFloat16` | `CAST(v AS VECTOR(3, float16))` with data assertion |
| `testConvertFloat16ToFloat32` | `CONVERT(VECTOR(3, float32), v)` with data assertion |
| `testConvertFloat32ToFloat16` | `CONVERT(VECTOR(3, float16), v)` with data assertion |
| `testAlterColumnFloat16ToFloat32` | `ALTER TABLE ... ALTER COLUMN v VECTOR(3, float32)` with data assertion |
| `testAlterColumnFloat32ToFloat16` | `ALTER TABLE ... ALTER COLUMN v VECTOR(3, float16)` with data assertion |
| `testCrossTypeUpdateFloat32ColWithFloat16Value` | UPDATE a FLOAT32 column with a FLOAT16 Vector object, assert values |
| `testCrossTypeUpdateFloat16ColWithFloat32Value` | UPDATE a FLOAT16 column with a FLOAT32 Vector object, assert values |
| `testStoredProcedureCrossTypeParameterPassing` | SP declares `@vec VECTOR(3)`, caller passes FLOAT16, assert values |
| `testUdfCrossTypeParameterPassing` | UDF declares `@vec VECTOR(3, float16)`, called with FLOAT32 source, assert values |
| `testTvfCrossTypeReturn` | Inline TVF returns `CAST(v AS VECTOR(3, float32))` from FLOAT16 source, assert values |
| `testCrossTypeRoundTripPrecision` | F32→F16→F32 round-trip preserves values within FLOAT16 precision |
| `testFloat16PrecisionLossWithNonRepresentableValues` | F32→F16 with 0.1/0.3/0.7 verifies lossy rounding contract |
| `testCrossTypeDimensionMismatchFails` | VECTOR(3, float16)→VECTOR(4, float32) rejects dimension mismatch |
| `testNullVectorCrossTypeInsertSelect` | NULL vector via cross-type INSERT...SELECT remains NULL |
| `testNullVectorCrossTypeSetObject` | setObject(null) into FLOAT16 column stores NULL correctly |
| `testGetObjectCrossReadFloat16AsFloat32RoundTrip` | Read FLOAT16 → setObject into FLOAT32 col → verify values end-to-end |
| `testBulkCopyCrossTypeCSVFloat16ToFloat32` | CSV declared as FLOAT16 imported into FLOAT32, per-row value assertion |
| `testBulkCopyCrossTypeCSVFloat32ToFloat16` | CSV declared as FLOAT32 imported into FLOAT16, per-row value assertion |

### New test resource

- `BulkCopyCSVTestInputCrossTypeVector.csv` — 3-row CSV with VECTOR(3) data for cross-type bulk copy tests.

## Server capability covered

| Capability | Tests |
|-----------|-------|
| Implicit conversion in assignments (INSERT...SELECT) | `testInsertFloat32FromFloat16Table`, `testInsertFloat16FromFloat32Table` |
| Implicit conversion in parameter passing (setObject) | `testSetObjectFloat32VectorIntoFloat16Column`, `testSetObjectFloat16VectorIntoFloat32Column` |
| Explicit CAST | `testExplicitCastFloat16ToFloat32`, `testExplicitCastFloat32ToFloat16` |
| Explicit CONVERT | `testConvertFloat16ToFloat32`, `testConvertFloat32ToFloat16` |
| DDL (ALTER COLUMN) | `testAlterColumnFloat16ToFloat32`, `testAlterColumnFloat32ToFloat16` |
| SP cross-type parameter passing | `testStoredProcedureCrossTypeParameterPassing` |
| UDF cross-type parameter passing | `testUdfCrossTypeParameterPassing` |
| TVF cross-type return | `testTvfCrossTypeReturn` |
| UPDATE with cross-type value | `testCrossTypeUpdateFloat32ColWithFloat16Value`, `testCrossTypeUpdateFloat16ColWithFloat32Value` |
| Round-trip precision fidelity | `testCrossTypeRoundTripPrecision` |
| Precision-loss (lossy F32→F16) | `testFloat16PrecisionLossWithNonRepresentableValues` |
| Dimension mismatch still rejected | `testCrossTypeDimensionMismatchFails` |
| NULL vector cross-type handling | `testNullVectorCrossTypeInsertSelect`, `testNullVectorCrossTypeSetObject` |
| JDBC-layer cross-read round-trip | `testGetObjectCrossReadFloat16AsFloat32RoundTrip` |
| Bulk copy cross-type (table-to-table) | `testBulkCopyFloat16ToFloat32Table`, `testBulkCopyFloat32ToFloat16Table`, `testMixedBulkCopyColumnMappingSwapped` |
| Bulk copy cross-type (CSV) | `testBulkCopyCrossTypeCSVFloat16ToFloat32`, `testBulkCopyCrossTypeCSVFloat32ToFloat16` |

## Testing

All tests are tagged with `@Tag(Constants.vectorFloat16Test)` and require a server with vector version 2 support. Tests are excluded from default CI runs via `excludedGroups` in `pom.xml`.


src/test/java/com/microsoft/sqlserver/jdbc/datatypes/vector/bulkcopy/VectorFloat16BulkCopyTest.java
```diff
@@ -719,17 +719,16 @@ public void testMixedBulkCopyDifferentDimensions() throws SQLException {
         }
 
         // ================================================================
-        // Bulk Copy Column Mapping Mismatch
```
divang, line 722: Please validate below test scenarios: 

Value correctness is not asserted in most new tests. testExplicitCastFloat16ToFloat32, testConvertFloat16ToFloat32, the CAST/CONVERT/ALTER/UPDATE tests, and both SP/UDF/TVF tests only assert getVectorDimensionType() and getDimensionCount() - never the actual element values. A conversion that silently zeroed or reordered elements would still pass. Only testCrossTypeRoundTripPrecision checks values. Recommend asserting (Float[]) result.getData() against expected values (with FLOAT16 tolerance where relevant).

No FLOAT16 precision-loss assertion. testConvertFloat32ToFloat16 / testExplicitCastFloat32ToFloat16 / testAlterColumnFloat32ToFloat16 use exactly-representable values, so they never prove the lossy F32 -> F16 path actually rounds. A test with a non-representable value (e.g. 0.1) asserting the expected rounded FLOAT16 result would cover the precision contract.

CSV bulk-copy tests don't validate values or per-row correctness. testBulkCopyCrossTypeCSVFloat16ToFloat32 / ...Float32ToFloat16 only count 3 rows and check type. Since ORDER BY v is used, you can deterministically assert each row's contents.

No dimension-mismatch cross-type case. All tests use VECTOR(3) on both sides. A cross-type operation combined with a dimension mismatch (e.g. source VECTOR(3) -> dest VECTOR(4)) should still fail - worth a negative test to confirm only the subtype conversion became permissive, not dimension checks.

No NULL vector cross-type case. Inserting/selecting a NULL vector across F16/F32 (INSERT...SELECT and setObject with a null Vector) isn't covered and is a common edge case.

No getObject cross-read of a stored value into the other subtype at the JDBC layer. All reads use the column's native type. Consider reading a FLOAT16 column then asserting round-trip through a FLOAT32 Vector object end-to-end.

machavan (approved):

Ananya2 (approved):

PR #3006, opened 2026-08-04
Feature | Send uniqueidentifier columns natively in bulk copy

### LINKED ISSUE #2986: SQLServerBulkCopy always declares uniqueidentifier columns as CHAR(n), forcing a CONVERT_IMPLICIT on every row
## Driver version

12.8.1.jre11 (root cause re-verified identical on 13.4.0 and `main` as of 2026-07 — unaffected by driver version)

## SQL Server version

```
Microsoft SQL Server 2022 (RTM-CU25) (KB5081477) - 16.0.4255.1 (X64)
	Apr 23 2026 22:38:54
	Developer Edition (64-bit) on Linux (Ubuntu 22.04.5 LTS) <X64>
```

Also confirmed the same behavior in production against SQL Server 2019 (on-prem) and Azure SQL.

## Client Operating System

Repro run on macOS (Apple Silicon) against a Linux (Ubuntu 22.04) SQL Server container. Also occurs in our production deployment, where the client JVM itself runs on Linux (Flink task managers on `apache/flink:1.17.2-jammy`).

## JAVA/JVM version

Repro run with:
```
openjdk version "17.0.19" 2026-04-21 LTS
OpenJDK Runtime Environment Corretto-17.0.19.10.1 (build 17.0.19+10-LTS)
OpenJDK 64-Bit Server VM Corretto-17.0.19.10.1 (build 17.0.19+10-LTS, mixed mode, sharing)
```
Production usage is on Java 11 (driver classifier `mssql-jdbc-12.8.1.jre11`) — behavior is identical, this isn't JVM-version-dependent.

## Table schema

```sql
CREATE TABLE dbo.GuidBulkCopyRepro (
    id uniqueidentifier NOT NULL
);
```

Same behavior reproduces on any table with a `uniqueidentifier` column when populated via `SQLServerBulkCopy`.

## Problem description

`SQLServerBulkCopy` never sends the native `uniqueidentifier` TDS wire type for a bulk-inserted GUID column, even when the source explicitly declares `microsoft.sql.Types.GUID` via `ISQLServerBulkData`. The `INSERT BULK` column list always shows the source column as `CHAR(36)`, so SQL Server has to run `CONVERT_IMPLICIT(uniqueidentifier, ...)` on every single row. At high row volumes (our case: tens of millions of rows/day through this path) this is a measurable, avoidable per-row CPU cost, and `PlanAffectingConvert` is also a documented cardinality-estimation red flag that can push the optimizer toward bad plans.

This looks closely related to #1999 (same root cause, different symptom — that one hard-fails on Azure Synapse instead of silently converting).

### Root cause

Traced in `SQLServerBulkCopy.java` (line refs from 12.8.1, re-verified identical on 13.4.0 and `main`):

1. **`getDestTypeFromSrcType(...)`** — the `case microsoft.sql.Types.GUID:` branch unconditionally returns `SSType.CHAR.toString() + "(" + bulkPrecision + ")"`. There is no branch that returns `"uniqueidentifier"`. This string builds the `INSERT BULK` column list sent to the server, so the source column is always declared `CHAR(36)`, regardless of the destination column's real type.

   ```java
       // For char the value has to be between 0 to 8000.
       return SSType.CHAR.toString() + "(" + bulkPrecision + ")";
   ```

2. **`writeTypeInfo(...)`** — the native `TDSType.GUID` token is only emitted when `isBaseType && (SSType.GUID == destSSType)`:

   ```java
   case java.sql.Types.CHAR:
       if (isBaseType && (SSType.GUID == destSSType)) {
           tdsWriter.writeByte(TDSType.GUID.byteValue());
           tdsWriter.writeByte((byte) 0x10);
       } else {
           // ... falls through to NCHAR/BIGCHAR
       }
   ```

3. **`isBaseType` is only ever `true` for the Always Encrypted `BaseTypeInfo` sub-token** (the plaintext-type descriptor written alongside `CryptoMetaData`, used purely for encrypt/decrypt bookkeeping — the actual on-wire value for an AE column is sent as VARBINARY ciphertext separately). Every ordinary (non-encrypted) bulk column goes through the other call site, which hardcodes `isBaseType = false`. So for a plain `uniqueidentifier` destination column, the native GUID token is architecturally unreachable — there's no connection property, driver version, or `ISQLServerBulkData`/`ISQLServerBulkRecord` declaration that changes this.

Also confirmed empirically: declaring the source column as `java.sql.Types.BINARY(16)` instead of `GUID` does **not** help — SQL Server still runs an implicit convert (`binary(16)` → `uniqueidentifier`), just a cheaper one (16-byte reinterpret instead of a 36-char string parse). The fix has to be on the wire-type/TDS-token side, not the source JDBC type declaration.

Feels closely related to #1582 → #2324 (reverted) → #2370, which added native `TDSType.GUID (0x24)` support for the **prepared-statement / RPC path** (`dtv.java`, `SQLServerConnection.java`) in 12.7.0. That work proves the driver can correctly encode a native GUID token outside the AE path — it just never got extended to `SQLServerBulkCopy`.

## Expected behavior

`INSERT BULK` declares `[id] UNIQUEIDENTIFIER` for a `uniqueidentifier` destination column, with no server-side conversion.

## Actual behavior

`INSERT BULK` declares `[id] CHAR(36)`, and the server runs `CONVERT_IMPLICIT(uniqueidentifier, ...)` on every row.

Verified with a minimal repro (below): create a table with one `uniqueidentifier` column, start a `plan_affecting_convert` Extended Events session, bulk-insert a single row through `SQLServerBulkCopy` declaring the source column as `microsoft.sql.Types.GUID`, then read the XE ring buffer.

Actual captured output from the repro:
```
Captured plan_affecting_convert columns: [id]
REPRODUCED: server ran CONVERT_IMPLICIT(uniqueidentifier, ...) on column 'id'
```

And from a production `INSERT BULK` capture (abridged, column names changed to match this repro's schema):
```
INSERT BULK dbo.GuidBulkCopyRepro ([id] CHAR(36) COLLATE SQL_Latin1_General_CP1_CI_AS)  with (ROWS_PER_BATCH = 1048576)
plan_affecting_convert: CONVERT_IMPLICIT(uniqueidentifier,[!BulkInsert].[id],0)
```

## Error message/stack trace

No exception — this is a silent behavioral/performance issue, not a crash. The evidence is server-side (an Extended Events `plan_affecting_convert` event and the `INSERT BULK` column-type declaration), not a client-side stack trace. See the repro output above and the runnable test below.

Minimal standalone repro (no test framework dependency — plain JDBC + `ISQLServerBulkData`). Set `MSSQL_JDBC_TEST_CONNECTION_PROPERTIES` (same env var name this repo's own `AbstractTest` uses) or pass the connection string as `args[0]`, then run.

<details>
<summary>GuidBulkCopyImplicitConvertRepro.java</summary>

```java
import com.microsoft.sqlserver.jdbc.ISQLServerBulkData;
import com.microsoft.sqlserver.jdbc.SQLServerBulkCopy;
import com.microsoft.sqlserver.jdbc.SQLServerBulkCopyOptions;

import java.sql.Connection;
import java.sql.ResultSet;
import java.sql.Statement;
import java.util.Set;
import java.util.UUID;
import java.util.stream.Collectors;

import javax.xml.parsers.DocumentBuilderFactory;
import org.w3c.dom.Document;
import org.w3c.dom.Element;
import org.w3c.dom.NodeList;
import java.nio.charset.StandardCharsets;

public class GuidBulkCopyImplicitConvertRepro {

    private static final String TABLE = "dbo.GuidBulkCopyRepro";
    private static final String XE_SESSION = "GuidBulkCopyReproXE";

    public static void main(String[] args) throws Exception {
        String connString = args.length > 0 ? args[0] : System.getenv("MSSQL_JDBC_TEST_CONNECTION_PROPERTIES");
        if (connString == null) {
            System.err.println("Pass a JDBC connection string as arg[0] or set MSSQL_JDBC_TEST_CONNECTION_PROPERTIES");
            System.exit(1);
        }

        try (Connection connection = DriverManager.getConnection(connString)) {
            setUp(connection);
            try {
                runRepro(connection);
            } finally {
                tearDown(connection);
            }
        }
    }

    private static void setUp(Connection connection) throws Exception {
        try (Statement st = connection.createStatement()) {
            st.execute("IF OBJECT_ID('" + TABLE + "') IS NOT NULL DROP TABLE " + TABLE);
            st.execute("CREATE TABLE " + TABLE + " (id uniqueidentifier NOT NULL)");

            st.execute("IF EXISTS (SELECT 1 FROM sys.server_event_sessions WHERE name = '"
                    + XE_SESSION + "') DROP EVENT SESSION " + XE_SESSION + " ON SERVER");
            st.execute(
                    "CREATE EVENT SESSION " + XE_SESSION + " ON SERVER " +
                    "ADD EVENT sqlserver.plan_affecting_convert( " +
                    "    WHERE sqlserver.database_name = N'" + connection.getCatalog() + "') " +
                    "ADD TARGET package0.ring_buffer " +
                    "WITH (MAX_DISPATCH_LATENCY = 1 SECONDS)");
            st.execute("ALTER EVENT SESSION " + XE_SESSION + " ON SERVER STATE = START");

            st.execute("DBCC FREEPROCCACHE");
        }
    }

    private static void tearDown(Connection connection) throws Exception {
        try (Statement st = connection.createStatement()) {
            st.execute("IF EXISTS (SELECT 1 FROM sys.server_event_sessions WHERE name = '"
                    + XE_SESSION + "') DROP EVENT SESSION " + XE_SESSION + " ON SERVER");
            st.execute("IF OBJECT_ID('" + TABLE + "') IS NOT NULL DROP TABLE " + TABLE);
        }
    }

    private static void runRepro(Connection connection) throws Exception {
        try (SQLServerBulkCopy bulkCopy = new SQLServerBulkCopy(connection)) {
            bulkCopy.setBulkCopyOptions(new SQLServerBulkCopyOptions());
            bulkCopy.setDestinationTableName(TABLE);
            bulkCopy.writeToServer(new SingleGuidColumnRecord(UUID.randomUUID()));
        }

        Thread.sleep(1500);

        Set<String> convertedColumns = readConvertedColumnsFromRingBuffer(connection);

        System.out.println("Captured plan_affecting_convert columns: " + convertedColumns);
        if (convertedColumns.contains("id")) {
            System.out.println("REPRODUCED: server ran CONVERT_IMPLICIT(uniqueidentifier, ...) on column 'id'");
        } else {
            System.out.println("NOT REPRODUCED: no convert captured on column 'id' (bug fixed, or XE session missed the event)");
        }
    }

    private static final class SingleGuidColumnRecord implements ISQLServerBulkData {
        private final UUID value;
        private boolean consumed = false;

        SingleGuidColumnRecord(UUID value) {
            this.value = value;
        }

        @Override
        public Set<Integer> getColumnOrdinals() {
            return Set.of(1);
        }

        @Override
        public String getColumnName(int column) {
            return "id";
        }

        @Override
        public int getColumnType(int column) {
            return microsoft.sql.Types.GUID;
        }

        @Override
        public int getPrecision(int column) {
            return 36;
        }

        @Override
        public int getScale(int column) {
            return 0;
        }

        @Override
        public Object[] getRowData() {
            return new Object[] { value.toString() };
        }

        @Override
        public boolean next() {
            if (consumed) {
                return false;
            }
            consumed = true;
            return true;
        }
    }

    private static Set<String> readConvertedColumnsFromRingBuffer(Connection connection) throws Exception {
        String xml;
        try (Statement st = connection.createStatement();
             ResultSet rs = st.executeQuery(
                     "SELECT CAST(t.target_data AS NVARCHAR(MAX)) AS target_data " +
                     "FROM sys.dm_xe_session_targets t " +
                     "JOIN sys.dm_xe_sessions s ON s.address = t.event_session_address " +
                     "WHERE s.name = '" + XE_SESSION + "'")) {
            rs.next();
            xml = rs.getString("target_data");
        }

        Document doc = DocumentBuilderFactory.newInstance().newDocumentBuilder()
                .parse(new ByteArrayInputStream(xml.getBytes(StandardCharsets.UTF_8)));

        // Each <event name="plan_affecting_convert"> carries an "expression" data field like:
        //   CONVERT_IMPLICIT(uniqueidentifier,[!BulkInsert].[id],0)
        NodeList dataNodes = doc.getElementsByTagName("data");
        return java.util.stream.IntStream.range(0, dataNodes.getLength())
                .mapToObj(i -> (Element) dataNodes.item(i))
                .filter(el -> "expression".equals(el.getAttribute("name")))
                .map(el -> el.getElementsByTagName("value").item(0).getTextContent())
                .filter(expr -> expr.contains("CONVERT_IMPLICIT(uniqueidentifier"))
                .flatMap(expr -> java.util.regex.Pattern.compile("\\.\\[(\\w+)\\]\\s*,0\\)")
                        .matcher(expr).results().map(m -> m.group(1)))
                .collect(Collectors.toSet());
    }
}
```

</details>

### Suggested fix

In `SQLServerBulkCopy.java`:
- `getDestTypeFromSrcType`: when `destSSType == SSType.GUID`, return `"uniqueidentifier"` instead of `CHAR(n)`.
- `writeTypeInfo`: drop the `isBaseType &&` gate for the GUID case — emit `TDSType.GUID` whenever `destSSType == SSType.GUID`, regardless of AE.
- Value-writing path (`writeColumnToTdsWriter`): for that branch, write the 16-byte GUID layout (the driver already has this — `Util.asGuidByteArray(UUID.fromString(value))`, currently only invoked from the AE-related value-conversion path) instead of the CHAR string bytes, and handle NULL as a single `0x00` length byte (matching the GUID-null convention used elsewhere in this file) rather than the CHAR `0xFF 0xFF` NULL marker.

Happy to help test against a candidate fix — this is currently forcing us to consider a class-shadowing workaround on our end, which we'd much rather not maintain long-term.

## JDBC trace logs

Not applicable here in the usual sense — the client-side JDBC trace log doesn't surface this, since the implicit conversion happens server-side after the `INSERT BULK` statement is sent; it isn't a client-observable error or warning. The definitive evidence is the server-side Extended Events capture shown above (`plan_affecting_convert` / `CONVERT_IMPLICIT`), reproduced by the attached repro. Happy to also provide the driver's own TDS-level trace output (enabled per the driver tracing docs) showing the outgoing `CHAR(36)` column-type declaration if that's useful in addition.

huzaus:
## Brief Description

Bulk copy declared GUID columns as `CHAR(36)` on the wire, so the server converted every single row on the way into a `uniqueidentifier` column. This sends the native type instead.

What changed in `SQLServerBulkCopy`:

- `getDestTypeFromSrcType` now returns `uniqueidentifier` rather than `CHAR(n)`
- `writeTypeInfo` emits the `TDSType.GUID` token. It could only be reached via the `isBaseType` branch before, which is written for Always Encrypted base type metadata only, so normal columns always fell through to `BIGCHAR`
- values go out as the 16 mixed endian bytes from `Util.asGuidByteArray`, same as `writeRPCUUID`; null is a `0x00` length byte
- a value that is not a GUID is reported as a `SQLServerException` using `R_errorConvertingValue`, chaining the `IllegalArgumentException` from `UUID.fromString`, so nothing unchecked escapes `writeToServer` in the middle of a batch

Only applies when the source declares `microsoft.sql.Types.GUID` and the destination is `uniqueidentifier`. `CHAR`/`VARCHAR` sources and the Always Encrypted path are untouched.

There is no connection property. Following the review, the registry format in braces is normalized rather than rejected, so this stays a wire format change.

Fixes #2986

## New Public APIs

None.

## Behavior

A source column of type `java.sql.Types.CHAR` with the same precision reproduces the previous character wire format exactly, which makes the old and the new behavior comparable inside one build. Measured on SQL Server 2022, per rendering of the same GUID:

| Rendering | Character format | Native format |
| --- | --- | --- |
| `6f9619ff-...-00c04fc964ff` | stored | stored |
| `6F9619FF-...-00C04FC964FF` | stored | stored |
| `{6f9619ff-...-00c04fc964ff}` declared 38 | stored | stored |
| **`{6f9619ff-...-00c04fc964ff}` declared 36** | **rejected** | **stored** |
| ` 6f9619ff-...-00c04fc964ff ` | rejected | rejected |
| `6f9619ff8b86d011b42d00c04fc964ff` | rejected | rejected |
| `(...)`, `urn:uuid:...`, `not-a-guid`, empty | rejected | rejected |

The one difference is the row in bold above: the character format truncated the value to the declared precision before the server converted it, so the registry format did not fit a column declared with 36 characters. Parsing on the client no longer depends on the declared precision. Braces are accepted at every precision, whitespace is rejected as before, and rejections that used to come from the server now come from the driver.

## Verification

New `BulkCopyGuidTest`, 29 cases, run against SQL Server 2022 (16.0.4265.3):

- `testBulkCopyGuidNativeFormatKeepsCharacterFormatBehavior` — every rendering above through both wire formats, asserting equal outcomes
- `testBulkCopyGuidNativeFormatIgnoresDeclaredPrecision` — pins the single deliberate difference
- `testBulkCopyGuidUnparsableValueFailsOnClient` — 7 unparsable renderings, asserting the exact driver message; `testBulkCopyGuidByteArrayValueFailsOnClient` covers a `byte[]` value. These run everywhere, including Azure SQL Database, and are the deterministic check that the native format is in use, since the character format would have failed on the server instead
- `testBulkCopyGuidConversionOnServer` — Extended Events on `plan_affecting_convert`, parameterized over both source types: nothing is reported against `[!BulkInsert]` for a GUID source, and a conversion is reported for a CHAR source. The CHAR case is the positive control, so the GUID case cannot pass by capturing nothing at all. A conversion issued by the test acts as a dispatch barrier, and the session is closed through `AutoCloseable`, so a failed `STATE = START` cannot leave a server scoped session behind. The whole test is skipped unless `HAS_PERMS_BY_NAME` reports both permissions on a server that is not Azure SQL Database
- `testBulkCopyGuidIntoCharacterDestination` — a GUID source into `char`, `varchar`, `nchar` and `nvarchar` destinations still goes out as a character string
- `testBulkCopyGuidRoundTripsValues` — `UUID`, lowercase string and null through `ISQLServerBulkData`

New `BulkCopyGuidAETest`, run against a local Java key store column master key:

- `testBulkCopyGuidIntoEncryptedColumn` — an encrypted `uniqueidentifier` column still receives ciphertext, verified by reading the column over a connection without column encryption
- `testBulkCopyGuidSourceIntoEncryptedColumnIsRejected` — pins that a GUID source into an encrypted column is rejected, which is also the behavior on `main`, since `validateDataTypeConversions` checks against the base type of the encrypted column

Against pristine `main`, `testBulkCopyGuidConversionOnServer` fails for the GUID source and passes for the CHAR source, reporting:

```
INSERT BULK [guidConversionDest_...] ([id] CHAR(36) )
CONVERT_IMPLICIT(uniqueidentifier,[!BulkInsert].[id],0)
```

After the change the GUID source reports no conversion.

Full bulk copy suite: 113 tests, the only failure being the pre-existing temporal flake in `BulkCopyColumnMappingTest`, which passes on a rerun with unchanged code. A GUID source into a `varchar(max)` destination fails identically on `main` and is left alone.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 
 ### Added
 
+- **Send uniqueidentifier Columns Natively in Bulk Copy** [#2986](https://github.com/microsoft/mssql-jdbc/issues/2986)
+  - **What was added**: `SQLServerBulkCopy` now declares a source column of type `microsoft.sql.Types.GUID` as `uniqueidentifier` on the wire and sends its value in the native 16 byte representation, instead of declaring it as `CHAR(n)` and sending a character string.
+  - **Who benefits**: Applications bulk copying into `uniqueidentifier` columns, for which the server previously had to run an implicit conversion on every row.
+  - **Impact**: Removes the per-row `CONVERT_IMPLICIT(uniqueidentifier, ...)` and the cardinality estimate warning it produced. Values must be valid GUIDs, either `java.util.UUID` or a string accepted by `UUID.fromString`. The new connection property `sendGuidAsStringForBulkCopy`, when set to `true`, restores the previous character string behavior.
+
 - **Add tls-unique Channel Binding Support for Kerberos and NTLM** [#2977](https://github.com/microsoft/mssql-jdbc/pull/2977)
   - **What was added**: tls-unique channel binding support for Kerberos and NTLM authentication flows.
   - **Who benefits**: Applications using integrated Kerberos or NTLM authentication that require channel binding for enhanced protection against relay attacks.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 - **Send uniqueidentifier Columns Natively in Bulk Copy** [#2986](https://github.com/microsoft/mssql-jdbc/issues/2986)
   - **What was added**: `SQLServerBulkCopy` now declares a source column of type `microsoft.sql.Types.GUID` as `uniqueidentifier` on the wire and sends its value in the native 16 byte representation, instead of declaring it as `CHAR(n)` and sending a character string.
   - **Who benefits**: Applications bulk copying into `uniqueidentifier` columns, for which the server previously had to run an implicit conversion on every row.
-  - **Impact**: Removes the per-row `CONVERT_IMPLICIT(uniqueidentifier, ...)` and the cardinality estimate warning it produced. Values must be valid GUIDs, either `java.util.UUID` or a string accepted by `UUID.fromString`.
+  - **Impact**: Removes the per-row `CONVERT_IMPLICIT(uniqueidentifier, ...)` and the cardinality estimate warning it produced. Values are `java.util.UUID` or a string in one of the renderings the server accepted before, which includes the registry format in braces. A value that is not a GUID is now reported by the driver as a conversion error instead of by the server. Since the value is no longer truncated to the declared precision before being converted, a rendering longer than that precision, such as the registry format in a column declared with 36 characters, is accepted where the server rejected it before.
 
 - **Add tls-unique Channel Binding Support for Kerberos and NTLM** [#2977](https://github.com/microsoft/mssql-jdbc/pull/2977)
   - **What was added**: tls-unique channel binding support for Kerberos and NTLM authentication flows.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 - **Send uniqueidentifier Columns Natively in Bulk Copy** [#2986](https://github.com/microsoft/mssql-jdbc/issues/2986)
   - **What was added**: `SQLServerBulkCopy` now declares a source column of type `microsoft.sql.Types.GUID` as `uniqueidentifier` on the wire and sends its value in the native 16 byte representation, instead of declaring it as `CHAR(n)` and sending a character string.
   - **Who benefits**: Applications bulk copying into `uniqueidentifier` columns, for which the server previously had to run an implicit conversion on every row.
-  - **Impact**: Removes the per-row `CONVERT_IMPLICIT(uniqueidentifier, ...)` and the cardinality estimate warning it produced. Values must be valid GUIDs, either `java.util.UUID` or a string accepted by `UUID.fromString`. The new connection property `sendGuidAsStringForBulkCopy`, when set to `true`, restores the previous character string behavior.
+  - **Impact**: Removes the per-row `CONVERT_IMPLICIT(uniqueidentifier, ...)` and the cardinality estimate warning it produced. Values must be valid GUIDs, either `java.util.UUID` or a string accepted by `UUID.fromString`.
 
 - **Add tls-unique Channel Binding Support for Kerberos and NTLM** [#2977](https://github.com/microsoft/mssql-jdbc/pull/2977)
   - **What was added**: tls-unique channel binding support for Kerberos and NTLM authentication flows.
--- a/src/main/java/com/microsoft/sqlserver/jdbc/ISQLServerDataSource.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/ISQLServerDataSource.java
@@ -1297,21 +1297,6 @@ public interface ISQLServerDataSource extends javax.sql.CommonDataSource {
     void setSendTemporalDataTypesAsStringForBulkCopy(boolean sendTemporalDataTypesAsStringForBulkCopy);
 
     /**
-     * Returns the current flag for value sendGuidAsStringForBulkCopy
-     *
-     * @return 'sendGuidAsStringForBulkCopy' property value.
-     */
-    boolean getSendGuidAsStringForBulkCopy();
-
-    /**
-     * Specifies the flag to send uniqueidentifier values as String for Bulk Copy.
-     *
-     * @param sendGuidAsStringForBulkCopy
-     *        boolean value for 'sendGuidAsStringForBulkCopy'.
-     */
-    void setSendGuidAsStringForBulkCopy(boolean sendGuidAsStringForBulkCopy);
-
-    /**
      * 
      * Returns the value for the connection property 'AADSecurePrincipalId'.
      * 
--- a/src/main/java/com/microsoft/sqlserver/jdbc/ISQLServerDataSource.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/ISQLServerDataSource.java
@@ -1297,6 +1297,21 @@ public interface ISQLServerDataSource extends javax.sql.CommonDataSource {
     void setSendTemporalDataTypesAsStringForBulkCopy(boolean sendTemporalDataTypesAsStringForBulkCopy);
 
     /**
+     * Returns the current flag for value sendGuidAsStringForBulkCopy
+     *
+     * @return 'sendGuidAsStringForBulkCopy' property value.
+     */
+    boolean getSendGuidAsStringForBulkCopy();
+
+    /**
+     * Specifies the flag to send uniqueidentifier values as String for Bulk Copy.
+     *
+     * @param sendGuidAsStringForBulkCopy
+     *        boolean value for 'sendGuidAsStringForBulkCopy'.
+     */
+    void setSendGuidAsStringForBulkCopy(boolean sendGuidAsStringForBulkCopy);
+
+    /**
      * 
      * Returns the value for the connection property 'AADSecurePrincipalId'.
      * 
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
@@ -1080,7 +1080,7 @@ public class SQLServerBulkCopy implements java.lang.AutoCloseable, java.io.Seria
 
             case microsoft.sql.Types.GUID:
             case java.sql.Types.CHAR: // 0xAF
-                if (isBaseType && (SSType.GUID == destSSType)) {
+                if ((SSType.GUID == destSSType) && (isBaseType || sendGuidAsNative(srcJdbcType))) {
                     tdsWriter.writeByte(TDSType.GUID.byteValue());
                     tdsWriter.writeByte((byte) 0x10);
                 } else {
@@ -1463,6 +1463,9 @@ public class SQLServerBulkCopy implements java.lang.AutoCloseable, java.io.Seria
                 return SSType.NUMERIC.toString() + "(" + bulkPrecision + ", " + bulkScale + ")";
 
             case microsoft.sql.Types.GUID:
+                if ((SSType.GUID == destSSType) && sendGuidAsNative(bulkJdbcType)) {
+                    return SSType.GUID.toString();
+                }
                 // For char the value has to be between 0 to 8000.
                 return SSType.CHAR.toString() + "(" + bulkPrecision + ")";
             case java.sql.Types.CHAR:
@@ -2256,6 +2259,30 @@ public class SQLServerBulkCopy implements java.lang.AutoCloseable, java.io.Seria
         }
     }
 
+    /**
+     * Returns true if a source column declared as microsoft.sql.Types.GUID is sent as a native uniqueidentifier, which
+     * the connection property sendGuidAsStringForBulkCopy reverts to sending it as a character string.
+     */
+    private boolean sendGuidAsNative(int srcJdbcType) {
+        return (microsoft.sql.Types.GUID == srcJdbcType) && !connection.getSendGuidAsStringForBulkCopy();
+    }
+
+    /**
+     * Writes the value of a uniqueidentifier destination column in the native 16 byte representation, which spares the
+     * server a conversion from a character string for every row.
+     */
+    private void writeGuidToTdsWriter(TDSWriter tdsWriter, Object colValue) throws SQLServerException {
+        if (null == colValue) {
+            tdsWriter.writeByte((byte) 0);
+            return;
+        }
+
+        UUID guidValue = (colValue instanceof UUID) ? (UUID) colValue : UUID.fromString(colValue.toString().trim());
+
+        tdsWriter.writeByte((byte) 0x10);
+        tdsWriter.writeBytes(Util.asGuidByteArray(guidValue));
+    }
+
     private void writeNullToTdsWriter(TDSWriter tdsWriter, int srcJdbcType,
             boolean isStreaming) throws SQLServerException {
 
@@ -2525,6 +2552,11 @@ public class SQLServerBulkCopy implements java.lang.AutoCloseable, java.io.Seria
                 case java.sql.Types.CHAR: // Fixed-length, non-Unicode string data.
                 case java.sql.Types.VARCHAR: // Variable-length, non-Unicode string data.
                 case microsoft.sql.Types.JSON:
+                    if ((SSType.GUID == destSSType) && sendGuidAsNative(bulkJdbcType)) {
+                        writeGuidToTdsWriter(tdsWriter, colValue);
+                        break;
+                    }
+
                     if (isStreaming) // PLP
                     {
                         // PLP_BODY rule in TDS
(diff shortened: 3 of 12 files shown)


huzaus:

muskan124947:
@huzaus 
I don’t think we need to introduce a new connection string option for this change. The existing behavior/configuration already covers the intended scenario, and adding another option would unnecessarily expand the public configuration surface and increase maintenance and documentation overhead. Could we implement this using the existing behavior instead?

huzaus:
@muskan124947 Property removed,  uniqueidentifier is now always sent natively. Ready for review.

muskan124947:
@huzaus 
Adding below comments, please have a look:
1. **`writeGuidToTdsWriter` — wrap the parse failure in `SQLServerException`**
`UUID.fromString(...)` throws an unchecked `IllegalArgumentException`, which escapes `writeToServer()` mid-batch. The driver should only surface `SQLServerException`. Please wrap it using `R_errorConvertingValue`, following the pattern in `SQLServerBulkCSVFileRecord.getRowData()` for `BIGINT`. Same applies when `colValue` is a `byte[]` — `toString()` yields `"[B@..."` and hits the same path.

2. **Behavior change needs to be non-breaking**
Values the server previously accepted, e.g. `{0E984725-...}`, now fail client-side, and with the property removed there is no escape hatch. To answer your open question: please **normalize rather than reject** — strip braces before `UUID.fromString` — so this is a pure wire-format optimization with zero user-visible behavior change. That keeps the connection property unnecessary.

3. **`testBulkCopyGuidDoesNotConvertOnServer` needs a positive control**
The test asserts an absence, so it passes vacuously if the XE harness misbehaves. The control was deleted along with the property. Please add one: a `java.sql.Types.CHAR` source into the same `uniqueidentifier` destination, asserting the conversion **is** captured.

4. **Test cleanup issues**
- If `CREATE EVENT SESSION` succeeds but `ALTER ... STATE = START` fails, `canCaptureImplicitConversions` returns `false`, the assumption aborts, and `dropEventSession` never runs — orphaned server-scoped session.
- In the `finally`, if `dropEventSession` throws, `dropTableIfExists` is skipped. Please nest them.
- `catch (SQLException e) { return false; }` hides real failures.
- Server-scoped XE isn't available on Azure SQL DB, so this guard silently skips in most CI legs. Consider adding a deterministic assertion on `getDestTypeFromSrcType` that always runs.

5. **Missing test cases**
Invalid GUID error path, `{braces}` input, CHAR source → `uniqueidentifier` dest (no-regression), and an AE-encrypted `uniqueidentifier` column — the `writeTypeInfo` condition being modified is the AE one, so it should have coverage.

Also, please update the PR description

huzaus:
@muskan124947 All five done in `ec5724dd`.

1. Fixed. Wrapped in `SQLServerException` with `R_errorConvertingValue`, chaining the `IllegalArgumentException`. Same for a `byte[]` value.

3. Fixed. Control added back. The test now runs twice: GUID source, expecting no conversion, and CHAR source, expecting one. The CHAR run confirms the session really captures events, so an empty result in the GUID run means something. Checked on `main`: GUID run fails, CHAR run passes.

4. Fixed. Session is `AutoCloseable` and drops itself if `STATE = START` fails, table drop is nested in its own `finally`, and the swallowed catch is replaced by a `HAS_PERMS_BY_NAME` plus engine edition guard. For the check that always runs, including on Azure SQL Database, I used the error path instead of the private `getDestTypeFromSrcType`: an unparsable value fails on the client, where the old format failed on the server.

5. Fixed. Invalid GUID, braces, CHAR into `uniqueidentifier`, and an AE encrypted column are covered, plus GUID into `char`/`varchar`/`nchar`/`nvarchar`.

2. Braces are stripped now, as asked. One correction though: braces were not accepted before either, in the usual case. I compared the two formats by declaring the source as `java.sql.Types.CHAR` with the same precision, which reproduces the old wire format. It truncates the value to the declared precision before the server converts, so `{0E984725-...}` needs the column declared with 38, and was rejected at the usual 36. Full table is in the PR description. The two formats now agree everywhere except braces at precision 36, which is accepted now because client side parsing ignores the declared precision. Pinned by a test, noted in the CHANGELOG.

Same comparison showed whitespace was always rejected, so the `trim()` I had would have quietly accepted more than before. Removed.

On AE: a GUID source into an encrypted column is rejected on `main` too, because `validateDataTypeConversions` checks against the base type. So the case that can be tested is a CHAR source, and `testBulkCopyGuidIntoEncryptedColumn` confirms over a non encrypting connection that the column really holds ciphertext. The rejection is pinned separately.

Tests moved to `BulkCopyGuidTest`, so `BulkCopyAllTypesTest` is back to its `main` state. One thing I noticed and left alone, since it behaves the same on `main`: a GUID source into a `varchar(max)` destination fails with "Invalid column type from bcp client for colid 1". PR description updated.

muskan124947:
Thanks @huzaus , will take a look

huzaus:
@muskan124947 Hi! Just a friendly nudge on this PR 🙂 Is there anything else needed from my side to help move it forward?

muskan124947:
Thanks @huzaus for your contribution! We’ve reviewed the PR, and everything looks good. There’s nothing pending from your side.

The PR has been tagged for the 13.7.0 milestone.

PR #3008, opened 2026-08-07
Test: Port legacy FX non-AE test scenarios to JUnit under legacyFx tag

Ananya2:
## Description

This PR ports functional test scenarios from the legacy JDBC FX test suite that were previously **not covered** by the JUnit test suite. It focuses on the **non-Always-Encrypted** gaps identified during a scenario-by-scenario audit of the FX suite against the current JUnit suite.

## What's added

**48 `legacyFx`-tagged tests across 11 existing classes**, covering 16 FX gap areas:

| Area | Tests | Added to |
|------|-------|----------|
| Blob/Clob `position()`, `truncate()`, partial `setBytes()`, serialization contract | 6 | `unit/lobs/LobsTest` |
| JDBC `{fn ...}` escape scalar/numeric functions | 19 | `unit/statement/LimitEscapeTest` |
| DatabaseMetaData: `getTypeInfo` type IDs, `supportsConvert` matrix, `getPseudoColumns`, `getMaxLogicalLobSize`, `generatedKeyAlwaysReturned`, UDT-family, `supportsRefCursors`, schema-collation (VSTS#532970) | 8 | `databasemetadata/DatabaseMetaDataTest` |
| Connection-string parsing: duplicate keywords, trailing token separators | 2 | `UtilTest` |
| Negative stream lengths, stream resettability, mega-select, stream-read-failure (Repro39941) | 4 | `unit/lobs/LobsStreamingTest` |
| Rollback of uncommitted work on pooled-connection close (VSTS#247260) | 1 | `connection/PoolingTest` |
| Clob 4000–8000 char roundtrip with `sendStringParametersAsUnicode=false` (VSTS#197731) | 1 | `preparedStatement/RegressionTest` |
| REF_CURSOR `registerOutParameter` exception | 1 | `callablestatement/CallableStatementTest` |
| Bulk copy: CHECK-constraint enforcement, transaction rollback/commit, GB18030 encoding | 3 | `SQLServerBulkCopyTest` |
| SSL: hostname-in-certificate mismatch failure | 1 | `ssl/TLSProtocolTest` |
| Read-only routing: double-hop, SQL Azure routing | 2 | `connection/ReadonlyRoutingTest` |

## Notes

- Each ported test reproduces the **functional behavior** of its FX origin (assertions on values/exceptions), not just the API call.

### DIFF
--- a/src/test/java/com/microsoft/sqlserver/jdbc/connection/ReadonlyRoutingTest.java
+++ b/src/test/java/com/microsoft/sqlserver/jdbc/connection/ReadonlyRoutingTest.java
@@ -143,4 +143,39 @@ public class ReadonlyRoutingTest extends AbstractTest {
             }
         }
     }
+
+    /**
+     * Verifies that a read-only-intent connection is established and usable; if the availability group
+     * re-routes, the driver still yields a single working connection.
+     */
+    @Test
+    public void testDoubleHopRouting() throws SQLException {
+        String url = connectionString + ";applicationIntent=ReadOnly";
+        try (Connection conn = PrepUtil.getConnection(url)) {
+            assertNotNull(conn);
+            try (Statement stmt = conn.createStatement(); ResultSet rs = stmt.executeQuery("SELECT 1")) {
+                assertTrue(rs.next());
+            }
+        } catch (SQLServerException e) {
+            // If AG not configured, routing may not be available.
+            assertNotNull(e.getMessage());
+        }
+    }
+
+    /**
+     * Verifies that a read-only-intent connection with encryption against SQL Azure connects and is
+     * usable, or fails cleanly if routing is not configured.
+     */
+    @Test
+    public void testSQLAzureRouting() throws SQLException {
+        String url = connectionString + ";applicationIntent=ReadOnly;encrypt=true;trustServerCertificate=true";
+        try (Connection conn = PrepUtil.getConnection(url)) {
+            assertNotNull(conn);
+            try (Statement stmt = conn.createStatement(); ResultSet rs = stmt.executeQuery("SELECT 1")) {
+                assertTrue(rs.next());
+            }
+        } catch (SQLServerException e) {
+            assertNotNull(e.getMessage());
+        }
+    }
 }
--- a/src/test/java/com/microsoft/sqlserver/jdbc/connection/ReadonlyRoutingTest.java
+++ b/src/test/java/com/microsoft/sqlserver/jdbc/connection/ReadonlyRoutingTest.java
@@ -157,14 +157,13 @@ public class ReadonlyRoutingTest extends AbstractTest {
                 assertTrue(rs.next());
             }
         } catch (SQLServerException e) {
-            // If AG not configured, routing may not be available.
-            assertNotNull(e.getMessage());
+            fail("Read-only routing test requires an AlwaysOn AG routing configuration: " + e.getMessage());
         }
     }
 
     /**
      * Verifies that a read-only-intent connection with encryption against SQL Azure connects and is
-     * usable, or fails cleanly if routing is not configured.
+     * usable.
      */
     @Test
     public void testSQLAzureRouting() throws SQLException {
@@ -175,7 +174,7 @@ public class ReadonlyRoutingTest extends AbstractTest {
                 assertTrue(rs.next());
             }
         } catch (SQLServerException e) {
-            assertNotNull(e.getMessage());
+            fail("Read-only routing test requires an AlwaysOn AG routing configuration: " + e.getMessage());
         }
     }
 }
--- a/src/test/java/com/microsoft/sqlserver/jdbc/connection/ReadonlyRoutingTest.java
+++ b/src/test/java/com/microsoft/sqlserver/jdbc/connection/ReadonlyRoutingTest.java
@@ -157,7 +157,8 @@ public class ReadonlyRoutingTest extends AbstractTest {
                 assertTrue(rs.next());
             }
         } catch (SQLServerException e) {
-            fail("Read-only routing test requires an AlwaysOn AG routing configuration: " + e.getMessage());
+            fail("Connection was expected to be usable with applicationIntent=ReadOnly specified: "
+                    + e.getMessage());
         }
     }
 
@@ -174,7 +175,8 @@ public class ReadonlyRoutingTest extends AbstractTest {
                 assertTrue(rs.next());
             }
         } catch (SQLServerException e) {
-            fail("Read-only routing test requires an AlwaysOn AG routing configuration: " + e.getMessage());
+            fail("Connection was expected to be usable with applicationIntent=ReadOnly specified: "
+                    + e.getMessage());
         }
     }
 }
--- a/src/test/java/com/microsoft/sqlserver/jdbc/databasemetadata/DatabaseMetaDataTest.java
+++ b/src/test/java/com/microsoft/sqlserver/jdbc/databasemetadata/DatabaseMetaDataTest.java
@@ -2960,8 +2960,9 @@ public class DatabaseMetaDataTest extends AbstractTest {
                 try {
                     stmt.executeUpdate("DROP DATABASE " + escapedDbName);
                 } catch (SQLException dropEx) {
-                    // Cleanup failure is best-effort; the afterAll/other cleanup will not run for this
-                    // temp DB, but suppressing here preserves any in-flight assertion failure.
+                    // Cleanup failure is best-effort; log it so the leaked temp database is visible
+                    // while still preserving any in-flight assertion failure.
+                    System.out.println("Failed to drop temporary database " + escapedDbName + ": " + dropEx.getMessage());
                 }
             }
         }
--- a/src/test/java/com/microsoft/sqlserver/jdbc/databasemetadata/DatabaseMetaDataTest.java
+++ b/src/test/java/com/microsoft/sqlserver/jdbc/databasemetadata/DatabaseMetaDataTest.java
@@ -2831,9 +2831,13 @@ public class DatabaseMetaDataTest extends AbstractTest {
         try (Connection conn = getConnection()) {
             DatabaseMetaData dbmd = conn.getMetaData();
             assertNotNull(dbmd);
+            // SQL Server supports conversion between all of these types, so both the parameterless and
+            // the from/to overloads must report true for every combination.
+            assertTrue(dbmd.supportsConvert(), "supportsConvert() should return true");
             for (int from : types) {
                 for (int to : types) {
-                    dbmd.supportsConvert(from, to);
+                    assertTrue(dbmd.supportsConvert(from, to),
+                            "supportsConvert(" + from + ", " + to + ") should return true");
                 }
             }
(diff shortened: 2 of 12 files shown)


divang (dismissed):

src/test/java/com/microsoft/sqlserver/jdbc/databasemetadata/DatabaseMetaDataTest.java
```diff
@@ -2783,4 +2784,163 @@ public void testGetFunctionColumnsWithValidation() throws SQLException {
...
+
+    /**
+     * Verifies getTypeInfo reports the correct DATA_TYPE ids for the temporal types (datetimeoffset =
+     * -155, time = 92, date = 91, datetime2 = 93).
+     */
+    @Test
+    @Tag(Constants.legacyFx)
+    @Tag(Constants.legacyFxMetadata)
+    public void testTemporalTypeInfo() throws SQLException {
+        Map<String, Integer> expected = new HashMap<>();
```
machavan, line 2800: In [DataTypesTest.java:2058-2061](vscode-file://vscode-app/c:/Users/machavan/AppData/Local/Programs/Microsoft%20VS%20Code/6a49527b96/resources/app/out/vs/code/electron-browser/workbench/workbench.html), the test only checks that a value was returned, so it would not catch incorrect or corrupted data; please compare each result with its expected value.

In [DatabaseMetaDataTest.java:2827-2839](vscode-file://vscode-app/c:/Users/machavan/AppData/Local/Programs/Microsoft%20VS%20Code/6a49527b96/resources/app/out/vs/code/electron-browser/workbench/workbench.html), the supportsConvert results are ignored, so the test should assert the expected conversion support values.

In [DatabaseMetaDataTest.java:2842-2854](vscode-file://vscode-app/c:/Users/machavan/AppData/Local/Programs/Microsoft%20VS%20Code/6a49527b96/resources/app/out/vs/code/electron-browser/workbench/workbench.html), getPseudoColumns is only checked for a non-null result set; please validate its metadata and expected contents.

In [DatabaseMetaDataTest.java:2872-2879](vscode-file://vscode-app/c:/Users/machavan/AppData/Local/Programs/Microsoft%20VS%20Code/6a49527b96/resources/app/out/vs/code/electron-browser/workbench/workbench.html), the return value of generatedKeyAlwaysReturned() is ignored, so the test does not verify the actual behavior.

In [TLSProtocolTest.java:193-208](vscode-file://vscode-app/c:/Users/machavan/AppData/Local/Programs/Microsoft%20VS%20Code/6a49527b96/resources/app/out/vs/code/electron-browser/workbench/workbench.html), the test only checks that an exception occurs and should also verify that it is caused by the certificate hostname mismatch.

In [DatabaseMetaDataTest.java:2929-2943](vscode-file://vscode-app/c:/Users/machavan/AppData/Local/Programs/Microsoft%20VS%20Code/6a49527b96/resources/app/out/vs/code/electron-browser/workbench/workbench.html), a failure in DROP DATABASE could hide the original test failure; cleanup should preserve the initial exception.

In [PoolingTest.java:314-350](vscode-file://vscode-app/c:/Users/machavan/AppData/Local/Programs/Microsoft%20VS%20Code/6a49527b96/resources/app/out/vs/code/electron-browser/workbench/workbench.html), the temporary table is not cleaned up if pooled connection creation fails, so table cleanup should be placed in an outer finally block.

src/test/java/com/microsoft/sqlserver/jdbc/connection/ReadonlyRoutingTest.java
```diff
@@ -143,4 +143,38 @@ public void testDefaultApplicationIntent() throws SQLException {
...
+    @Test
+    public void testDoubleHopRouting() throws SQLException {
+        String url = connectionString + ";applicationIntent=ReadOnly";
+        try (Connection conn = PrepUtil.getConnection(url)) {
+            assertNotNull(conn);
+            try (Statement stmt = conn.createStatement(); ResultSet rs = stmt.executeQuery("SELECT 1")) {
+                assertTrue(rs.next());
+            }
+        } catch (SQLServerException e) {
+            fail("Read-only routing test requires an AlwaysOn AG routing configuration: " + e.getMessage());
```
machavan: Is this failure message misleading ? can it be simple saying the connection was expected to be usable with applicationIntent specified ?

src/test/java/com/microsoft/sqlserver/jdbc/connection/ReadonlyRoutingTest.java
```diff
@@ -143,4 +143,38 @@ public void testDefaultApplicationIntent() throws SQLException {
...
+    @Test
+    public void testSQLAzureRouting() throws SQLException {
+        String url = connectionString + ";applicationIntent=ReadOnly;encrypt=true;trustServerCertificate=true";
+        try (Connection conn = PrepUtil.getConnection(url)) {
+            assertNotNull(conn);
+            try (Statement stmt = conn.createStatement(); ResultSet rs = stmt.executeQuery("SELECT 1")) {
+                assertTrue(rs.next());
+            }
+        } catch (SQLServerException e) {
+            fail("Read-only routing test requires an AlwaysOn AG routing configuration: " + e.getMessage());
```
machavan: Is this failure message misleading ? can it be simple saying the connection was expected to be usable with applicationIntent specified ?

divang (commented):
Please have a look on below suggestion -

ReadonlyRoutingTest.testDoubleHopRouting / testSQLAzureRouting - beyond the misleading message machavan flagged, these two tests have no environment gate. If the test connection string doesn't point at an AlwaysOn AG with read-only routing, the catch (SQLServerException) -> fail(...) branch turns "environment not configured" into a hard failure. The legacy FX origin was an external-setup scenario; these should be tagged @ Tag(Constants.reqExternalSetup) (and likely xAzureSQLDW/xAzureSQLDB as appropriate), or use an assumption to skip when routing isn't configured - otherwise they'll fail on ordinary CI.

ReadonlyRoutingTest new tests are untagged - every other ported test carries @ Tag(Constants.legacyFx), but testDoubleHopRouting and testSQLAzureRouting have no legacyFx tag, so they won't be included/excluded consistently with the rest of the ported suite.

TLSProtocolTest.testHostNameInCertificateMismatchRejected - the assertion falls back to message.toLowerCase().contains("certificate"), which is broad enough that an unrelated certificate error (e.g. untrusted chain) would also pass. Since the intent is specifically hostname mismatch, prefer asserting on the R_certNameFailed text ("Failed to validate the server name") only, or additionally gate with reqExternalSetup/encryption availability.

LobsStreamingTest and LobsTest end without a trailing newline (\ No newline at end of file). Minor, but worth fixing to avoid noisy future diffs.
SQLServerBulkCopyTest.testBulkCopyCheckConstraintEnforcement - asserts ex.getMessage().toUpperCase().contains("CHECK"). That's a bit fragile across locales/SQL versions; consider also asserting the constraint name (chkCity) or the SQLState to make the check more specific.

DatabaseMetaDataTest.testSupportsConvertMatrix - asserts every from/to pair returns true. SQL Server does not support every conversion in that 16×16 matrix (e.g. some binary - temporal combinations), so this may be over-asserting and could be flaky against driver behavior; verify against the actual supportsConvert implementation rather than assuming all-true.

muskan124947:
Please check below :
## Blocking


**1. `LobsTest.testSerializeBlobThrows` asserts the opposite of the class contract**

`SQLServerBlob extends SQLServerLob implements java.io.Serializable`. I verified it serializes fine (338 bytes) via a direct probe. The test only "passes" incidentally: with the default `delayLoadingLobs=true`, `getBlob()` returns the `SQLServerBlob(BaseInputStream)` form and the **non-transient** `activeStreams` list holds a non-serializable stream. Set `delayLoadingLobs=false`, or call `free()`/materialize first, and it fails. If the intent is "an unmaterialized stream-backed LOB can't be serialized", say that explicitly and pin `delayLoadingLobs=true`; otherwise drop it.


**2. `LimitEscapeTest.testEscapeFnFloorCeiling` will almost certainly fail**

SQL Server's `FLOOR`/`CEILING` preserve the argument's scale — `{fn FLOOR(3.7)}` → `3.0` and `{fn CEILING(3.2)}` → `4.0`, so `getString()` returns `"3.0"`/`"4.0"`, not `"3"`/`"4"`. Compare numerically like the sibling `SQRT`/`ROUND` tests do.


## 🟡 Tests that assert nothing meaningful


**3. `ReadonlyRoutingTest.testDoubleHopRouting` / `testSQLAzureRouting`** are line-for-line duplicates of the existing `testReadOnlyIntentConnection` / `testReadOnlyWithEncrypt`. Neither touches double-hop or Azure routing, and there's no routing assertion (e.g. `SERVERPROPERTY('ServerName')` differing from the requested server, or `DATABASEPROPERTYEX(...,'Updateability')`). Recommend removing or making them actually verify redirection.


**4. `DatabaseMetaDataTest.testSupportsConvertMatrix`** — `supportsConvert(int,int)` is `checkClosed(); return true;`. 256 assertions against a hardcoded constant. Same for `testGeneratedKeyAlwaysReturned`, `testSupportsRefCursors`, `testGetMaxLogicalLobSize`.


**5. `TLSProtocolTest`** — the `|| message.toLowerCase().contains("certificate")` fallback makes it pass on nearly any TLS error. Match `R_certNameFailed` only.


## 🟢 Nits



- **Localized-message coupling**: `"not valid"` (LobsStreamingTest), `"CHECK"` (bulk copy), `"REF_CURSOR is not supported"`. These break under a non-English resource bundle. Prefer `SQLServerException.getErrString("R_invalidLength")` / SQL error 547.

- **`DataTypesTest.testAllTypesServerPrepareRoundTrip`**: javadoc claims prepexec, but with default `enablePrepareOnFirstPreparedStatementCall=false` a single execution uses `sp_executesql` — prepexec is never hit. Also no `try/finally` around the per-type table, so a mid-loop failure leaks temp tables, and 27 types in one method means the first failure masks the rest (`@ParameterizedTest`).

- **`RegressionTest.testVSTS197731`**: the second `rs.next()` return is unchecked, and `ORDER BY (SELECT NULL)` gives no deterministic order — "row 2" is arbitrary. Add an identity column or assert both rows.

- **`LobsStreamingTest`**: new tests reassign the shared static `tableName` that `@BeforeEach` already sets. Use locals.

- **`DatabaseMetaDataTest.testSchemaCollation`**: empty `catch (SQLException dropEx)` silently leaks the test database — at least log it.

- **`SQLServerBulkCopyTest.testBulkCopyTransactionRollbackAndCommit`**: `setAutoCommit(true)` only runs on the happy path; move to `finally`.

- Mixed JUnit 4 (`org.junit.Assert.assertTrue("msg", cond)`) and Jupiter (`assertTrue(cond, "msg")`) asserts in `DataTypesTest`/`CallableStatementTest`/`UtilTest` — compiles, but the reversed argument order is a footgun. Standardize on Jupiter.

- `LobsTest.java` and `LobsStreamingTest.java` now end without a trailing newline.

machavan (dismissed):

machavan (approved):

divang (approved):

PR #3012, opened 2026-08-11
Enhance input validation for table names in BulkCopy and PreparedStatement batch insert

muskan124947:
## Description

Bulk copy takes the destination table name as a string and concatenates it into the SQL it builds internally — the `sp_executesql` metadata probe and the `INSERT BULK` command. Anything the caller passes therefore lands in query text as-is, so a name that is not really a name is not treated as one.

This adds `Util.sanitizeIdentifier()`, which parses the name into its parts and re-emits each one bracket-quoted. The result can only be read as an object name: text that is not a valid name identifies one non-existent object instead of contributing SQL. Applications that pass a name from configuration or user input no longer depend on having validated it themselves.

## Changes

**`Util.java`** — added `sanitizeIdentifier(String)` and its helpers.

The name is split on dots, ignoring dots inside a part the caller delimited with `[]` or `""`, and each part is re-quoted with `escapeSQLId()`. A part that is already a single well-formed delimited identifier keeps its identifier and is not quoted twice; anything else is quoted whole.

Both the splitter and the unwrapper decide what counts as a delimited part with the same `delimitedLength()` primitive, so they cannot disagree. That matters for a part like `[a] PRINT 1 --[b]`, which starts and ends with brackets but is not one identifier: it is quoted whole rather than unwrapped into `a] PRINT 1 --[b`.

**`SQLServerBulkCopy.java`** — applied at both places the name reaches query text:
- `getDestinationMetadata()` — the `sp_executesql` FMTONLY probe, and the `OBJECT_ID('...')` lookups it feeds
- `createInsertBulkCommand()` — the `INSERT BULK` command

**`SQLServerPreparedStatement.java`** — applied in `executeBatch()` and `executeLargeBatch()`, which build the same metadata query from the table name parsed out of the INSERT when `useBulkCopyForBatchInsert=true`.

## Compatibility

Quoting rather than rejecting, so that names that work today keep working:

| Name | Result |
|---|---|
| `table`, `schema.table`, `db.schema.table` | quoted per part |
| `srv.db.dbo.table` | accepted — four parts, matching SqlClient's `MultipartIdentifier` limit |
| `db..table` | preserved as `[db]..[table]` — an empty qualifier defers to the default schema |
| `[dbo].[My Table]`, `"dbo"."My Table"` | already delimited, not quoted twice |
| `my]table`, `O'Brien`, `#temp` | quoted, not rejected |
| `a.b.c.d.e` | rejected — more parts than `[server].[database].[schema].[object]` |
| empty, or no object part (`dbo.`) | rejected |

Only the object part is required, and only names with more parts than SQL Server has cannot identify an object, so those are the only rejections. The part count is not a security control — `escapeSQLId` makes every part inert regardless of how many there are — so it is not used as one.

## Tests

**`UtilTest.java`** — 8 parameterized tests, 63 cases:
- 27 quoting cases: one to four part names, `db..table`, caller-delimited (`[]` and `""`), escaped delimiters, whitespace, `[]` and `[ ]`, temp and global temp tables, unterminated and mismatched delimiters, a delimiter that does not start its part, Unicode
- 10 payload cases, each asserting the exact quoted result rather than just that something changed
- 12 rejection cases: empty, whitespace, no object part, five-part in four spellings, null
- properties that hold for any input: the result is nothing but bracket-quoted parts joined by dots, and every remaining single quote is doubled so the result cannot close the `N'...'` literal it is embedded in
- a 128-character name is not truncated, and a name containing a single quote survives both quoting steps

**`BulkCopyInputValidationTest.java`** (new) — 6 tests, 22 cases:
- 7 payloads run against both the bulk copy and the `useBulkCopyForBatchInsert` batch path. Each payload creates a marker object first, so the marker existing is direct proof it executed; each asserts the failure is specifically error 208, since any other error would mean the copy failed for an unrelated reason
- 5 valid name forms end to end, including `[db]..[table]` and a database-qualified name
- a table whose name contains `]` and `.` round-trips
- batch insert still reaches its destination
- a five-part name is rejected


src/main/java/com/microsoft/sqlserver/jdbc/Util.java
```diff
@@ -1049,6 +1049,58 @@ static String escapeSingleQuotes(String name) {
         return name.replace("'", "''");
     }
 
+    /**
+     * Parses a multi-part SQL Server identifier using ThreePartName and bracket-quotes each part
+     * to produce a safe identifier string. Prevents SQL injection through identifier concatenation.
+     */
+    static String sanitizeIdentifier(String name) throws SQLServerException {
```
divang, line 1075: should Util.escapeSingleQuotes not part of this new method? So we do not need to call twice those Uitls methods
Util.escapeSingleQuotes(Util.sanitizeIdentifier( ....

muskan124947: they should stay separate because sanitizeIdentifier is used in two different contexts:

- Inside a string literal (sp_executesql N'...FROM [table] ...') — needs both sanitizeIdentifier + escapeSingleQuotes because a bracket-quoted name like [O'Brien] still has a single quote that would break the N'...' string.
- As a direct identifier (INSERT BULK [table]) — only needs sanitizeIdentifier, NOT escapeSingleQuotes. Adding single-quote escaping here would corrupt the identifier (turning [O'Brien] into [O''Brien]).
```

// String literal context — needs both
"sp_executesql N'SET FMTONLY ON SELECT * FROM " + Util.escapeSingleQuotes(Util.sanitizeIdentifier(name)) + " '"

// Direct identifier context — only sanitize
"INSERT BULK " + Util.sanitizeIdentifier(name)
```

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyInputValidationTest.java
```diff
@@ -0,0 +1,334 @@
+/*
```
divang: Please have a look on below scenarios and add test case if missing:

Unicode/non-ASCII identifier names - No tests for identifiers with special Unicode characters (e.g., [Tïñés], Chinese characters, emoji). These should be preserved and re-bracketed correctly.

Identifiers with mixed delimiters - No test for malformed inputs like "dbo].[table" (mixed double-quote and bracket delimiters). 

Very long identifier names - No test for SQL Server identifier length limits (~128 chars). The method should still work, but it's worth explicitly verifying no truncation or buffer issues occur.

Whitespace-only parts after parsing - No test for inputs like [   ].[table] or dbo.   .table. ThreePartName may parse these, but the result of stripBrackets("   ") should be validated.

Consecutive dots without delimiters - No test for dbo..table (empty middle part). The parser may treat this differently than dbo.table.

Bracket/quote escaping edge cases - While [table]]name] is tested, there's no test for:

Double-quote escaping: "my""table" should become [my"table] (single quote, then re-escaped as ]])

Mixed escaped characters in a single part: [a]]b]]c]

Empty parts: [] or [""]

Integration test for useBulkCopyForBatchInsert with complex payloads - The current integration tests use simple payloads. Missing: injection via comment syntax (--, /* */), or GO batch separator attempts.

Column name injection in BulkCopy - The PR focuses on table names, but BulkCopy also uses column names. No tests verify that column names in BulkCopyColumnMapping.getSourceColumnName() or similar are also protected (if applicable).

Concurrent execution test - No test for thread-safety or race conditions if multiple threads call sanitizeIdentifier() simultaneously.

Round-trip test - No test verifying that a sanitized identifier, when used in a query and re-parsed later, still refers to the same table.

muskan124947: Added
below scenarios are not required
Column name injection   Out of scope — this PR is table name sanitization only
Concurrent execution    sanitizeIdentifier is stateless (no shared mutable state), inherently thread-safe
Round-trip test Needs DB connection; the method is a pure string transformation

src/main/java/com/microsoft/sqlserver/jdbc/Util.java
```diff
@@ -1049,6 +1049,58 @@ static String escapeSingleQuotes(String name) {
...
+        }
+
+        ThreePartName parsed = ThreePartName.parse(name);
+
+        // Reject 4+ part names (e.g. server.db.schema.table) — ThreePartName only splits up to 3 parts,
+        // so an undelimited dot in the last part means there were more parts than supported.
+        String procPart = parsed.getProcedurePart();
+        if (procPart != null && !procPart.startsWith("[") && !procPart.startsWith("\"")
+                && procPart.contains(".")
+                && (parsed.getDatabasePart() != null || parsed.getOwnerPart() != null)) {
```
Ananya2: This can be bypassed when an intermediate part is delimited. For example:

- srv.db.schema.[table] : rejected
- srv.db.[schema].table : accepted
- a.b.[c].d : accepted
- [a].[b].[c].[d] : accepted

For srv.db.[schema].table, ThreePartName.parse() leaves procPart as [schema].table. Since procPart starts with [, the 4-part check is skipped even though there is still another dotted component.

As a result, these inputs will be silently mangled rather than rejected:

- srv.db.[schema].table -> [srv].[db].[[schema].table]
- a.b.[c].d              -> [a].[b].[[c].d]
- [a].[b].[c].[d]        -> [a].[b].[c]].[d]

This is a correctness issue because the validation/sanitization method should not silently transform a 4-part name into a different identifier. Even if the resulting string does not create an injection vulnerability, accepting and corrupting the input is undesirable at a security validation boundary.

Could we avoid inferring the part count from procPart.startsWith(...) and procPart.contains("."), and instead explicitly determine the number of identifier parts using the same parsing rules and reject anything with more than three parts? This would also correctly handle bracketed/quoted components without relying on the formatting of the leftover substring.

src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyInputValidationTest.java
```diff
@@ -0,0 +1,334 @@
...
+                TestUtils.dropTableIfExists(escapedName, stmt);
+            }
+        }
+    }
+
+    /**
+     * Tests that the PreparedStatement executeBatch() path with useBulkCopyForBatchInsert=true
+     * sanitizes the table name parsed from the INSERT SQL, preventing injection via sp_executesql.
+     */
+    @Test
```
Ananya2: Could we add regression coverage for the following cases?

1. srv.db.[schema].table
2. a.b.[c].d
3. [a].[b].[c].[d]

All of these are 4-part names and should be rejected. The existing testSanitizeIdentifierFourPartNameRejected covers the undelimited case, but these delimited variants can bypass the current check.

Issue #2631, opened 2025-03-10
stored procedure call doesnt work without brackets {}


DmitriyGod:
## Driver version
latest

## SQL Server version
mcr.microsoft.com/mssql/server:2022-latest

## Client Operating System
any

## JAVA/JVM version
any

## Table schema
create procedure test_bigdecimal
@big_decimal_type      decimal(15, 5)       ,
@big_decimal_type_o    decimal(15, 5) output
as begin
   set @big_decimal_type_o = @big_decimal_type;
end;

## Problem description
Hi! MSSQL driver don't allow to use prepareCall without brackets {}

## Expected behavior
Like this works in pg, oracle, hsqldb, mariadb, mysql, snowflake, db2.
Code below shouldn't cause exception:
var call = connection.prepareCall("""
    call test_bigdecimal(100.24112, ?)""");

call.registerOutParameter(1, Types.DECIMAL);
call.execute();

## Actual behavior
prepareCall works only with brackets
var call = connection.prepareCall("""
    {call test_bigdecimal(100.24112, ?)}""");
call.registerOutParameter(1, Types.DECIMAL);
call.execute();

## Error message/stack trace
Incorrect syntax near '@P0'

## Any other details that can be helpful
-

## JDBC trace logs
-

muskan124947:
Thank you for reaching out.

As per the JDBC specification, the {} braces are required in the call escape syntax for stored procedure execution. Our driver follows this compliance to correctly parse and execute stored procedure calls.
You can find more details in the JDBC documentation [here](https://download.oracle.com/otn-pub/jcp/jdbc-4_3-mrel3-eval-spec/jdbc4.3-fr-spec.pdf)

DmitriyGod:
Yeah, page 116 shows an example of code with escaping
But this example is in paragraph 13.4.4 which is in chapter 13.4 which says 1 and 2:
1. The SQL string used in a Statement object **_may_** include JDBC escape syntax.

2. Special escape processing might be needed for the following:
- commonly used features that do not have standard syntax defined by SQL, or where the native syntax supported by the underlying data source varies widely among vendors.
**Not the case because the SQL-standard supports calling without escaping and major drivers already support it: pg, oracle, hsqldb, mariadb, mysql, snowflake, db2**
- features that are not supported by the underlying data source but are implemented by the driver. 
**obviously not the case**

Making calls work without escaping won't break spec compliance.
[...]

PR #3016, opened 2026-08-14
Users/hattorihenzo/support procedure call without curly braces

HattoriHenzo:
## Description

The issue: https://github.com/microsoft/mssql-jdbc/issues/2631

This pull request enhances the parsing and validation of SQL call syntax in the JDBC driver, ensuring correct handling of braces in stored procedure calls. It updates the regular expressions and logic to support both braced and unbraced syntax, adds validation for matching braces, and introduces comprehensive tests to verify correct and incorrect usage.

**Allowed syntax**:  `{call procedure(x,y)} `or `call procedure(x,y)`

## What's added

* Updated the `SQL_IDENTIFIER_PART` regex in `SQLServerStatement.java` to more accurately match SQL identifiers, including those outside parentheses and braces.
* Modified the `JDBC_CALL_SYNTAX` regex to support both optional opening and closing braces in procedure call syntax.
* Added a check in the `translate` method to ensure SQL statements have matching braces before applying regex parsing, throwing a `SQLServerException` if mismatched.
* Introduced the `hasMatchingBraces` helper method to encapsulate brace validation logic.

## Testing

* Added tests in `CallableStatementTest.java` to verify that procedure calls with and without braces work correctly, and that malformed calls with mismatched braces throw exceptions as expected.
* Cleaned up test imports in `CallableStatementTest.java` for better readability.


muskan124947:
@HattoriHenzo 
## `hasMatchingBraces` breaks all JDBC escape syntax


`translate()` is called from `SQLServerConnection.parseAndCacheSQL()`, which runs for **every** `Statement.execute()` and **every** `PreparedStatement` (`SQLServerStatement.java:1015-1017`, `SQLServerPreparedStatement.java:342-348`) — not just callable statements. So this new guard gates *all* SQL:


Java




```
boolean hasOpenBrace  = trimmedSql.startsWith("{");
boolean hasCloseBrace = trimmedSql.endsWith("}");
return hasOpenBrace == hasCloseBrace;

```





Any statement that merely *ends* with `}` without *starting* with `{` now throws `R_invalidSQL`. That includes the driver's own escape sequences, handled ~20 lines below in the same method:


SQL




```
select * from t1 {limit 3543}                          -- LimitEscapeTest.java:318
select * from t1 {limit ( (  (2)))}                     -- LimitEscapeTest.java:291
SELECT * FROM t WHERE d < {ts '2024-01-01 00:00:00'}    -- ODBC/JDBC escape
SELECT {fn UCASE(name)}                                 -- ODBC scalar fn escape

```





`LimitEscapeTest` should fail outright with this change. Conversely `{call p(?)} SELECT 1` (previously valid — note the trailing `.*+` in the old regex) now throws.


The method also isn't brace *matching* at all — it compares the first and last characters. `{call a(?)}{` passes; `{call a(?)}` inside a string literal isn't considered.

HattoriHenzo:
@muskan124947 Thanks for your feedback, I will take a look on it.

HattoriHenzo:
@muskan124947 I have to narrow the control of the SQL query syntax just only for `CallableStatement`? What do you think about implementing the control in this method `SQLServerConnection#prepareCall`, like at the line `SQLServerConnection.java:5870` with the call of the method `SQLServerConnection#checkClosed`?

muskan124947:
@HattoriHenzo 
Scoping this to the callable path is the right direction — it removes the `{limit ...}` / `{ts ...}` / `{fn ...}` regression, since `translate()` currently runs for *every* `Statement` and `PreparedStatement` via `parseAndCacheSQL()`, not just calls. Two notes on placement:


**1. `prepareCall(String, int, int)` isn't the only entry point.** There are two construction sites — `SQLServerConnection.java:5872` and `:8149` (the holdability overload). A check at line 5870 misses the second. The single choke point both funnel through is the `SQLServerCallableStatement` constructor (`SQLServerCallableStatement.java:100`); `SQLServerConnection43` and `SQLServerConnectionPoolProxy` delegate down to these, so they're covered for free.


Keeping the check out of `translate()` also avoids a callable-only rule leaking into `parsedSQLCache`, which is keyed on SQL text and shared across all statement types.


**2. The heuristic still has false positives, even scoped.** `startsWith("{") == endsWith("}")` rejects valid callable SQL that works today:


Java




```
call p({ts '2024-01-01 00:00:00'})   // ends '}' → rejected (and it's the unbraced form you want!)
{call p(?)} -- trailing comment      // ends 't' → rejected
{call p(?)};                         // ends ';' → rejected

```

PR #3017, opened 2026-08-14
Remove sqljdbc_xa.dll: drop legacy XA extended-procedure DLL

Ananya2:
## Summary
Removes all references to the legacy server-side `sqljdbc_xa.dll` from the driver's documentation and diagnostics. XA distributed transactions now rely solely on the `xp_sqljdbc_xa_*` extended stored procedures that are built into SQL Server 2017 CU16 and later (enabled via `EXEC sp_sqljdbc_xa_install`). The DLL is no longer shipped.

## Motivation
`sqljdbc_xa.dll` was only required for **SQL Server 2016 and earlier** — the sp_xa procedures have been built into the SQL Server engine since **SQL Server 2017 CU16**. The server-side sp_xa sprocs originally derived from the code in this JDBC DLL, so a recently reported MSRC vulnerability in the server procedures is likely to exist in the DLL as well. With SQL Server 2016 leaving the 5+5 support lifecycle this year, the DLL can be safely dropped rather than serviced.

## Changes
- **README.md** — "Download the DLLs" no longer lists `sqljdbc_xa`; documents that XA procedures are built into SQL Server 2017 CU16+ (enable via `sp_sqljdbc_xa_install`) and that the standalone DLL is no longer shipped.
- **DLLException.java** — Class comment now references only the `mssql-jdbc_auth` DLL.
- **XAStateTest.java** — Diagnostic hints updated to point to `EXEC sp_sqljdbc_xa_install` on SQL Server 2017 CU16+.

**Not changed (intentionally)**
- `SQLServerXAResource.java` and the `xp_sqljdbc_xa_*` calls — the XA feature remains fully functional against SQL Server 2017 CU16+.
- `R_failedToInitializeXA` resource string and its localized translations.

## Testing
No runtime code changed — only docs, one comment, and test diagnostic strings. XA tests continue to run against SQL Server 2017 CU16+ using the built-in stored procedures.

## Related
Companion PR in the internal build repo removes the DLL source, build, and packaging.

### DIFF
--- a/src/test/java/com/microsoft/sqlserver/jdbc/statemachinetest/xa/XAStateTest.java
+++ b/src/test/java/com/microsoft/sqlserver/jdbc/statemachinetest/xa/XAStateTest.java
@@ -195,7 +195,7 @@ public class XAStateTest extends AbstractTest {
             if ((msg != null && msg.toLowerCase().contains("xp_sqljdbc_xa")) ||
                     (fullMsg != null && fullMsg.toLowerCase().contains("xp_sqljdbc_xa"))) {
                 System.out.println("XA not supported: XA stored procedures not found");
-                System.out.println("  Solution: Enable XA on SQL Server 2017+ by running EXEC sp_sqljdbc_xa_install");
+                System.out.println("  Solution: Enable XA on SQL Server 2017 CU16+ by running EXEC sp_sqljdbc_xa_install");
                 return true;
             }
 
@@ -207,7 +207,7 @@ public class XAStateTest extends AbstractTest {
                             fullMsg.toLowerCase().contains("xa control connection") ||
                             fullMsg.toLowerCase().contains("could not load the dll")))) {
                 System.out.println("XA not supported: " + msg);
-                System.out.println("  Solution: Enable XA on SQL Server 2017+ by running EXEC sp_sqljdbc_xa_install");
+                System.out.println("  Solution: Enable XA on SQL Server 2017 CU16+ by running EXEC sp_sqljdbc_xa_install");
                 return true;
             }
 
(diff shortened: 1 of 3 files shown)


src/test/java/com/microsoft/sqlserver/jdbc/statemachinetest/xa/XAStateTest.java
```diff
@@ -195,7 +195,7 @@ private static boolean isXANotInstalledError(Throwable e) {
             if ((msg != null && msg.toLowerCase().contains("xp_sqljdbc_xa")) ||
                     (fullMsg != null && fullMsg.toLowerCase().contains("xp_sqljdbc_xa"))) {
                 System.out.println("XA not supported: XA stored procedures not found");
-                System.out.println("  Solution: Run xa_install.sql script on SQL Server");
+                System.out.println("  Solution: Enable XA on SQL Server 2017 CU16+ by running EXEC sp_sqljdbc_xa_install");
```
divang, line 198: Please have a look on below obersvations:

XAStateTest.java - second isXANotInstalledError branch still triggers on "could not load the dll" message, but now gives the same advice as the non-DLL branch. The condition fullMsg.toLowerCase().contains("could not load the dll") was originally meaningful when the DLL was a real installation step. Now that the DLL is gone, this message path likely dead code for current SQL Server versions. A comment or a dedicated test verifying this branch is no longer reachable (or removing the condition) would improve clarity.

XAStateTest.java - No negative test for SQL Server 2016 or older behavior. The PR drops support guidance for SQL Server 2016 and earlier. There's no test (or @ Disabled test with a note) asserting that the driver correctly fails/errors for these versions, or that the minimum supported version is now effectively SQL Server 2017 CU16. Adding an assumeTrue guard or a documented test gap would make the intent explicit.

XAStateTest.java - No test asserting sp_sqljdbc_xa_install is the canonical install step. The existing isXANotInstalledError helper detects setup failures at runtime, but there's no positive test that validates the driver successfully connects and executes XA transactions when sp_sqljdbc_xa_install has been run on SQL Server 2017 CU16+. Consider adding or tagging an integration test for this flow.

muskan124947 (approved):

divang (approved):

machavan (approved):

PR #3018, opened 2026-08-14
Add getCurrentApplicationName to PerformanceLogCallback

### LINKED ISSUE #3009: [FEATURE REQUEST] Add application name while publishing performance log
Most application use a connection pool to manage connections. Connections are bundled by the pool name. Driver started emitting performance logs in the recent version, but there is no way to identify which pool's connection logs are getting emitted. 

To solve the problem, pool name should be part of the event published by the performance logger. ApplicationName attribute can be used to pass pool name to the mssql connections.

muskan124947:
Applications using connection pools had no way to identify which pool a published performance event originated from. This adds a getCurrentApplicationName() default method to PerformanceLogCallback, returning the applicationName connection property of the connection that produced the event, so a pool can be identified by setting applicationName on its data source.

The value is supplied through a new PerformanceLog.currentApplicationName ThreadLocal, set immediately before callback.publish(...) and cleared in the finally block, mirroring the existing currentUserSql lifecycle. To provide the value, PerformanceLog.Scope now holds the SQLServerConnection, just as it already holds the SQLServerStatement.

The application name is resolved lazily at publish time rather than at scope creation time. The CONNECTION scope is opened at the top of connect(), before activeConnectionProperties has been parsed, so an eagerly captured value would be null for connection-level activities and would miss the DEFAULT_APP_NAME fallback when the property is not set.

Existing callback implementations are unaffected: publish() signatures are unchanged and the method is a default. The java.util.logging output format is also unchanged, and the application name is only resolved when a callback is registered.

Fixes #3009

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/PerformanceLog.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/PerformanceLog.java
@@ -23,6 +23,7 @@ class PerformanceLog {
     // ThreadLocal to hold current SQL text and statement type for the duration of a publish callback
     static final ThreadLocal<String> currentUserSql = new ThreadLocal<>();
     static final ThreadLocal<StatementType> currentStatementType = new ThreadLocal<>();
+    static final ThreadLocal<String> currentApplicationName = new ThreadLocal<>();
 
     /**
      * Register a callback for performance log events.
@@ -52,6 +53,7 @@ class PerformanceLog {
 
     public static class Scope implements AutoCloseable {
         private Logger logger;
+        private SQLServerConnection con;
         private int connectionId;
         private int statementId;
         private PerformanceActivity activity;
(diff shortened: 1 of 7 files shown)


src/main/java/com/microsoft/sqlserver/jdbc/PerformanceLog.java
```diff
@@ -64,19 +66,20 @@ public static class Scope implements AutoCloseable {
...
-        public Scope(Logger logger, int connectionId, int statementId,
+        public Scope(Logger logger, SQLServerConnection con, int statementId,
                      SQLServerStatement stmt, String userSql, PerformanceActivity activity) {
             this.enabled = logger.isLoggable(Level.FINE) || (callback != null);
             this.useNanos = cachedUseNanos;
 
             if (enabled) {
                 this.logger = logger;
-                this.connectionId = connectionId;
+                this.con = con;
```
joyaljohn, line 81: Do we really need to store a complete connection here? I am just trying to understand why can't we store only the app name. I don't see any other attribute from connection is used here

muskan124947: The CONNECTION scope wraps `connectInternal`, so at construction there is no app name to copy. PRELOGIN/LOGIN/statement scopes are created *inside* `connectInternal`, which is why they'd work fine either way.

On keeping a `String` field plus a `setApplicationName()` setter called after line 2879. That works, but it adds plumbing into `connectInternal`, introduces mutable scope state, and creates a window where the value is wrong if anything publishes early. Holding the reference gets correctness for free.

muskan124947: Today we already need two things from the connection — connection ID and application name — so passing the `SQLServerConnection` reference covers both in one parameter and keeps `Scope` extendible if we need more connection-level context later, without changing signatures again. It's also the existing pattern: `Scope` already holds the `SQLServerStatement` for the same lazy-resolution reason.

machavan (approved):

Ananya2 (approved):

PR #3031, opened 2026-08-31
Improve JDBC URL parsing validation for port and instance name segments

Ananya2:
### Summary

Strengthens input validation in `Util.parseUrl` for the server portion of the JDBC URL. The port segment is now validated as a strictly numeric value in the legal TCP range, and the instance-name segment now rejects malformed values containing `=` (mirroring the existing serverName check). This makes URL parsing stricter and more predictable, and rejects badly formed URLs early with a clear error instead ofsilently accepting them.

### Motivation

The URL parser previously accepted arbitrary characters in the port segment until the next `;`, and did not validate the instance-name segment for `=`. Enforcing the expected grammar for these segments improves robustness and gives clearer feedback for
malformed connection strings.

### Changes

- **`Util.java`**
  - New helper `validatePortNumber(...)`: enforces a non-empty, digits-only port in the range `0–65535`.
  - Invoked in both the mid-string `inPort` state and the end-of-URL `inPort` exit state.
  - `inInstanceName` (mid-string and exit) now rejects instance names containing `=`.
- **`SQLServerResource.java`**
  - Added `R_errorInstanceName` ("The instanceName connection property value {0} is badly formed."). Reuses existing `R_invalidPortNumber`.
- **`UtilTest.java`**
  - `testParseUrlRejectsInjectedPortNumber` — non-numeric / out-of-range port is rejected.
  - `testParseUrlValidPortStillParses` — valid numeric port still parses (no regression).
  - `testParseUrlRejectsInjectedInstanceName` — instance name with `=` is rejected.
- **`docs/security/jdbc-url-property-injection.md`**
  - Documents the validation behavior and recommends typed APIs (`SQLServerDataSource` setters / `Properties`) instead of concatenating externally-influenced values into URL strings.

### Behavioral impact / compatibility

- Well-formed URLs are unaffected.
- Previously-tolerated malformed inputs (non-numeric ports, instance names with `=`) now fail fast with a descriptive error. This is stricter but aligned with documented URL syntax.

### Testing

- `mvn -Pjre11 -Dtest=UtilTest test` — new and existing `parseUrl` tests pass.

### DIFF
--- a/src/main/java/com/microsoft/sqlserver/jdbc/Util.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/Util.java
@@ -39,6 +39,9 @@ final class Util {
 
     final static String ACTIVITY_ID_TRACE_PROPERTY = "com.microsoft.sqlserver.jdbc.traceactivity";
 
+    /** Maximum valid TCP port number. */
+    static final int MAX_PORT_NUMBER = 65535;
+
     private Util() {
         throw new UnsupportedOperationException(SQLServerException.getErrString("R_notSupported"));
     }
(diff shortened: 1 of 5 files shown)


muskan124947:
1. acceptsURL silently swallows this — worse diagnostics, not better.

SQLServerDriver.acceptsURL (line 1520-1525) catches SQLServerException from parseUrl and returns false. So for jdbc:sqlserver://localhost:1433x;..., DriverManager.getConnection now throws "No suitable driver found" instead of the previous "The port number 1433x is not valid." (which SQLServerConnection line 3400 already produced via NumberFormatException). That's the opposite of the PR's stated goal. Either validate in connect() rather than parseUrl, or make acceptsURL only fail on prefix mismatch.

2. Character.isDigit does not mean "digits-only". It accepts Unicode digits, and Integer.parseInt accepts them too — I verified "١٤٣٣" passes both and parses to 1433. So the "strictly numeric" claim isn't enforced, and the raw non-ASCII string is what gets stored in portNumber. Use an explicit ch < '0' || ch > '9' check.

3. Undocumented behavior change. The exit-state inInstanceName now does property.toLowerCase(Locale.US). It's a consistency fix with the mid-string branch (line 382), but it's not in the PR description and has no test. Same for the newly added exit-state serverName = check — also untested. Either call both out explicitly or split them out.

4. PR description doesn't match the diff. It claims a docs/security/jdbc-url-property-injection.md file; no such file is in the changeset.

-------------------

validatePortNumber has dead code: after a digits-only pass, port < 0 is unreachable and NumberFormatException only fires on >10-digit overflow. The whole thing collapses to a length + ASCII-digit-range check.
65535 is now hardcoded in a third place (SQLServerConnection 3393 and 4383 already have it). Pull it into a constant or SQLServerDriverIntProperty.PORT_NUMBER bounds.
Empty port (jdbc:sqlserver://localhost:;db=x) is now rejected at parse time. Previously it flowed through and failed at connect. Add a test pinning the intended behavior.

src/main/java/com/microsoft/sqlserver/jdbc/Util.java
```diff
@@ -275,6 +278,37 @@ static void writeLong(long value, byte[] valueBytes, int offset) {
         valueBytes[offset] = (byte) ((value >> 56) & 0xFF);
     }
 
+    /**
+     * Validates that the port segment parsed from a JDBC URL is a strictly numeric (ASCII digits only)
+     * value within the legal TCP port range (0-65535).
```
divang, line 283: does 0 port number allowed?

Ananya2: yes

PR #3037, opened 2026-09-04
Fix silent drop of trailing SQL after VALUES in bulk copy batch insert

machavan:
## Problem

With `useBulkCopyForBatchInsert=true`, the driver parses a batched `INSERT` into its table, columns, and values and sends rows through Bulk Copy instead of executing the SQL text.

The parser stops after the first `VALUES (...)` tuple. Previously, `checkAdditionalQuery()` stripped comments, whitespace, and semicolons but did not reject remaining SQL. Trailing query hints such as `OPTION (RECOMPILE)` could therefore be ignored, and additional tuples in a multi-row `VALUES` list could be silently lost.

## Changes

- Reject meaningful SQL remaining after the first values tuple and use the existing normal prepared-batch fallback, preserving the complete statement.
- Validate trailing SQL only during parsing. Cached parsing results mean later executions have not consumed `localUserSQL`, so checking that full string again would incorrectly reject supported inserts.
- Remember rejected trailing SQL in `bulkCopyHasUnsupportedTrailingSQL`. Both `executeBatch()` and `executeLargeBatch()` bypass Bulk Copy on every subsequent execution of that statement.
- Continue accepting trailing comments, whitespace, and semicolons, including whitespace-separated semicolons such as `; ;`.

### Why the rejection must persist

The values list is cached before trailing-SQL validation runs. Throwing without remembering rejection only falls back for the current batch: the next batch skips parsing and validation and can silently drop the trailer again.

The rejection flag is set before throwing and lasts for the lifetime of the prepared statement, whose SQL text does not change. It survives `clearBatch()`, switching batch APIs, and failures in normal execution. It is separate from the configured Bulk Copy option and is not set for database or runtime failures. Supported inserts continue to reuse cached parsing and use Bulk Copy.

Focused code comments explain this state lifetime, why rejection must be recorded before throwing, and why regression tests check the execution path rather than relying on row counts or repeated fallback log messages alone.

## Expected behavior

| SQL / execution | Result |
|---|---|
| Trailing `OPTION (...)` | Normal batch execution preserves the hint on every batch |
| Additional `VALUES` tuples | Normal batch execution preserves all tuples and update counts |
| Reused statement with rejected trailing SQL | Remains on the normal path, including when switching batch APIs |
| Supported plain insert, including repeated batches | Continues using Bulk Copy |
| Only trailing comments, whitespace, or semicolons | Continues using Bulk Copy |

No public API or connection-string changes. **This PR does not modify `CHANGELOG.md`.**

## Regression coverage

`BatchExecutionWithBulkCopyTest` covers:

- Rejected and permitted trailers.
- Three successive batches with a query hint or an additional values tuple, using either batch API or alternating between them in either order.
- Exact update counts and inserted data, plus verification that rejected SQL never creates a Bulk Copy operation. Row counts alone would not detect a discarded query hint.
- Supported inserts retaining Bulk Copy across repeated and alternating batch calls.

Database-independent `BulkCopyTrailingSQLTest` verifies that rejected SQL stays on the normal batch path after simulated execution failures, `clearBatch()`, and switching batch APIs.

## Validation

The regression tests were written before the persistent-state fix: all 16 unsupported-trailer cases failed on statement reuse, while the four supported-insert cases passed.

After the behavioral fix and Eclipse formatting:

- `mvn -Pjre21 -Dtest=BulkCopyTrailingSQLTest,BatchExecutionWithBulkCopyTest test`: **57/57 cases passed** (49 integration-class cases and 8 isolated cases).
- `mvn -Pjre8 -Dmaven.test.skip=true clean compile`: passed on the installed JDK 21.

The final follow-up removes the changelog entry and adds explanatory comments only; it does not change executable code. Changed comments were formatted with the repository Eclipse configuration.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -97,6 +97,11 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 
 ### Fixed
 
+- **Fix Silent Drop of Trailing SQL After VALUES in Bulk Copy Batch Insert** [#3037](https://github.com/microsoft/mssql-jdbc/pull/3037)
+  - **What was fixed**: Trailing content after the `VALUES` list, such as an `OPTION (...)` query hint or an extra value tuple, now triggers the existing fallback to the regular batch execution path instead of being discarded.
+  - **Who benefits**: Applications using `useBulkCopyForBatchInsert=true` with batched `INSERT` statements that carry trailing clauses.
+  - **Impact**: The full statement is sent to the server verbatim; statements with only trailing comments, whitespace, or semicolons continue to use Bulk Copy.
+
 - **Fix Closed-Statement NullPointerException in buildParamTypeDefinitions** [#2995](https://github.com/microsoft/mssql-jdbc/pull/2995)
   - **What was fixed**: Checked the statement's closed state before accessing a null parameter array in `buildParamTypeDefinitions()`.
   - **Who benefits**: Applications that invoke parameter-related methods after a prepared or callable statement is closed.
--- a/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
+++ b/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerPreparedStatement.java
@@ -3043,6 +3043,16 @@ public class SQLServerPreparedStatement extends SQLServerStatement implements IS
 
     private void checkAdditionalQuery() {
         while (checkAndRemoveCommentsAndSpace(true)) {}
+
+        // Once the VALUES list has been consumed, only comments, whitespace or semicolons may remain.
+        // Any other trailing content (for example an OPTION (...) query hint) cannot be honored by the
+        // Bulk Copy API. Throw so the caller falls back to the regular batch execution path, which sends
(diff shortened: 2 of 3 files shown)


sylvesterkaczmarek (commented):
Hi, I think the trailing-SQL fallback only survives the first batch on a reused PreparedStatement. `bcOperationValueList` is populated before `checkAdditionalQuery()` throws, and the catch falls back without clearing it; on the next `executeBatch()` the null check skips `checkAdditionalQuery()` and Bulk Copy can drop the same trailing `OPTION` clause again. Could the fallback persist an unsupported-for-bulk-copy state (or clear the cached parse) and add a repeated-batch regression with trailing SQL?

sylvesterkaczmarek (approved):
Re-reviewed current `f1c277a0`. The repeated-batch issue I raised is fixed: unsupported trailing SQL now sets persistent statement state before fallback, both `executeBatch()` and `executeLargeBatch()` bypass Bulk Copy thereafter, and the new regression covers repeated and alternating batch APIs even after execution failure. I do not see a remaining blocker in this area.

divang (approved):

PR #3041, opened 2026-09-09
Feature | Send uniqueidentifier columns natively in bulk copy

### LINKED ISSUE #2986: SQLServerBulkCopy always declares uniqueidentifier columns as CHAR(n), forcing a CONVERT_IMPLICIT on every row

muskan124947:
> [!NOTE]
> Internal mirror of community PR [#3006](https://github.com/microsoft/mssql-jdbc/pull/3006) by @huzaus for internal validation and merge. Contributor commits and authorship are preserved.
>
> Source: `huzaus/mssql-jdbc:feature/native-uniqueidentifier-bulkcopy` at `ec5724ddcb1902ca1f22f7c89e1fd08eedac8bc7`

## Brief Description

Bulk copy declared GUID columns as `CHAR(36)` on the wire, so the server converted every single row on the way into a `uniqueidentifier` column. This sends the native type instead.

What changed in `SQLServerBulkCopy`:

- `getDestTypeFromSrcType` now returns `uniqueidentifier` rather than `CHAR(n)`
- `writeTypeInfo` emits the `TDSType.GUID` token. It could only be reached via the `isBaseType` branch before, which is written for Always Encrypted base type metadata only, so normal columns always fell through to `BIGCHAR`
- values go out as the 16 mixed endian bytes from `Util.asGuidByteArray`, same as `writeRPCUUID`; null is a `0x00` length byte
- a value that is not a GUID is reported as a `SQLServerException` using `R_errorConvertingValue`, chaining the `IllegalArgumentException` from `UUID.fromString`, so nothing unchecked escapes `writeToServer` in the middle of a batch

Only applies when the source declares `microsoft.sql.Types.GUID` and the destination is `uniqueidentifier`. `CHAR`/`VARCHAR` sources and the Always Encrypted path are untouched.

There is no connection property. Following the review, the registry format in braces is normalized rather than rejected, so this stays a wire format change.

Fixes #2986

## New Public APIs

None.

## Behavior

A source column of type `java.sql.Types.CHAR` with the same precision reproduces the previous character wire format exactly, which makes the old and the new behavior comparable inside one build. Measured on SQL Server 2022, per rendering of the same GUID:

| Rendering | Character format | Native format |
| --- | --- | --- |
| `6f9619ff-...-00c04fc964ff` | stored | stored |
| `6F9619FF-...-00C04FC964FF` | stored | stored |
| `{6f9619ff-...-00c04fc964ff}` declared 38 | stored | stored |
| **`{6f9619ff-...-00c04fc964ff}` declared 36** | **rejected** | **stored** |
| ` 6f9619ff-...-00c04fc964ff ` | rejected | rejected |
| `6f9619ff8b86d011b42d00c04fc964ff` | rejected | rejected |
| `(...)`, `urn:uuid:...`, `not-a-guid`, empty | rejected | rejected |

The one difference is the row in bold above: the character format truncated the value to the declared precision before the server converted it, so the registry format did not fit a column declared with 36 characters. Parsing on the client no longer depends on the declared precision. Braces are accepted at every precision, whitespace is rejected as before, and rejections that used to come from the server now come from the driver.

## Verification

New `BulkCopyGuidTest`, 29 cases, run against SQL Server 2022 (16.0.4265.3):

- `testBulkCopyGuidNativeFormatKeepsCharacterFormatBehavior` — every rendering above through both wire formats, asserting equal outcomes
- `testBulkCopyGuidNativeFormatIgnoresDeclaredPrecision` — pins the single deliberate difference
- `testBulkCopyGuidUnparsableValueFailsOnClient` — 7 unparsable renderings, asserting the exact driver message; `testBulkCopyGuidByteArrayValueFailsOnClient` covers a `byte[]` value. These run everywhere, including Azure SQL Database, and are the deterministic check that the native format is in use, since the character format would have failed on the server instead
- `testBulkCopyGuidConversionOnServer` — Extended Events on `plan_affecting_convert`, parameterized over both source types: nothing is reported against `[!BulkInsert]` for a GUID source, and a conversion is reported for a CHAR source. The CHAR case is the positive control, so the GUID case cannot pass by capturing nothing at all. A conversion issued by the test acts as a dispatch barrier, and the session is closed through `AutoCloseable`, so a failed `STATE = START` cannot leave a server scoped session behind. The whole test is skipped unless `HAS_PERMS_BY_NAME` reports both permissions on a server that is not Azure SQL Database
- `testBulkCopyGuidIntoCharacterDestination` — a GUID source into `char`, `varchar`, `nchar` and `nvarchar` destinations still goes out as a character string
- `testBulkCopyGuidRoundTripsValues` — `UUID`, lowercase string and null through `ISQLServerBulkData`

New `BulkCopyGuidAETest`, run against a local Java key store column master key:

- `testBulkCopyGuidIntoEncryptedColumn` — an encrypted `uniqueidentifier` column still receives ciphertext, verified by reading the column over a connection without column encryption
- `testBulkCopyGuidSourceIntoEncryptedColumnIsRejected` — pins that a GUID source into an encrypted column is rejected, which is also the behavior on `main`, since `validateDataTypeConversions` checks against the base type of the encrypted column

Against pristine `main`, `testBulkCopyGuidConversionOnServer` fails for the GUID source and passes for the CHAR source, reporting:

```
INSERT BULK [guidConversionDest_...] ([id] CHAR(36) )
CONVERT_IMPLICIT(uniqueidentifier,[!BulkInsert].[id],0)
```

After the change the GUID source reports no conversion.

Full bulk copy suite: 113 tests, the only failure being the pre-existing temporal flake in `BulkCopyColumnMappingTest`, which passes on a rerun with unchanged code. A GUID source into a `varchar(max)` destination fails identically on `main` and is left alone.


src/test/java/com/microsoft/sqlserver/jdbc/bulkCopy/BulkCopyGuidTest.java
```diff
@@ -0,0 +1,453 @@
...
+            bulkCopy.writeToServer(new GuidBulkRecord(srcJdbcType, declaredPrecision, values));
+        }
+    }
+
+    private static SQLServerBulkCSVFileRecord guidFileRecord(String guid) throws Exception {
+        byte[] bytes = ("guidcol\n" + guid + "\n").getBytes(StandardCharsets.UTF_8);
+        try (InputStream inputStream = new ByteArrayInputStream(bytes)) {
+            return new SQLServerBulkCSVFileRecord(inputStream, Constants.UTF8, Constants.COMMA, true);
+        }
+    }
```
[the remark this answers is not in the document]
muskan124947: Updated `guidFileRecord()` so the CSV header is `id`, matching both the explicit column metadata and the destination column.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
```diff
@@ -1080,7 +1080,7 @@ private void writeTypeInfo(TDSWriter tdsWriter, int srcJdbcType, int srcScale, i
 
             case microsoft.sql.Types.GUID:
             case java.sql.Types.CHAR: // 0xAF
-                if (isBaseType && (SSType.GUID == destSSType)) {
+                if ((SSType.GUID == destSSType) && (isBaseType || microsoft.sql.Types.GUID == srcJdbcType)) {
```
machavan, line 1083: How does this work for CSV bulk load ? and a test for it ?

src/test/java/com/microsoft/sqlserver/jdbc/AlwaysEncrypted/BulkCopyGuidAETest.java
```diff
@@ -0,0 +1,180 @@
...
+ */
+@RunWith(JUnitPlatform.class)
+@Tag(Constants.xSQLv11)
+@Tag(Constants.xSQLv12)
+@Tag(Constants.xAzureSQLDW)
+@Tag(Constants.reqExternalSetup)
+@Tag(Constants.alwaysEncrypted)
+public class BulkCopyGuidAETest extends AESetup {
+
+    private static final int GUID_TEXT_LENGTH = 36;
```
machavan, line 49: -The new parser accepts malformed GUIDs that SQL Server previously rejected, silently converting them into padded UUIDs.
-Valid GUIDs followed by trailing text or spaces, previously accepted by SQL Server, now fail.
-Add regression tests for both cases and preserve existing behavior.

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
```diff
@@ -2258,6 +2267,62 @@ private void validateColumnMappings() throws SQLServerException {
...
+        if (null == colValue) {
+            tdsWriter.writeByte((byte) 0);
+            return;
+        }
+
+        UUID guidValue;
+        try {
+            if (colValue instanceof UUID) {
+                // UUID objects previously used their 36-character rendering on the CHAR wire path.
+                if (precision < 36) {
```
divang: NIT: good to have repeat int hardcoded values as constant: 
private static final int GUID_TEXT_LENGTH = 36;
private static final int BRACED_GUID_TEXT_LENGTH = 38;
private static final int BRACED_GUID_CLOSING_BRACE_INDEX = 37;

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
```diff
@@ -2527,6 +2596,11 @@ else if (null != sourceCryptoMeta) {
                 case java.sql.Types.CHAR: // Fixed-length, non-Unicode string data.
                 case java.sql.Types.VARCHAR: // Variable-length, non-Unicode string data.
                 case microsoft.sql.Types.JSON:
+                    if ((SSType.GUID == destSSType) && (microsoft.sql.Types.GUID == bulkJdbcType)) {
```
muskan124947: Possible regression: the column header and the row data can disagree about the type here.

Scenario: the source is a `SQLServerResultSet` on a connection with Always Encrypted enabled, the source column is an encrypted `char`/`varchar`/`nvarchar`, and the destination is a plaintext `uniqueidentifier`.

- Column header (`writeColumnMetaDataToken`): `bulkJdbcType` comes from the source metadata, and `getColumnType` returns the decrypted type (`CHAR`/`VARCHAR`). The new `writeTypeInfo` check (L1086) needs a `GUID` source, so the header declares `BIGCHAR`.
- Row data (here): because `sourceCryptoMeta != null`, `bulkJdbcType` was replaced with the destination's type (`GUID`) a few lines above, so this branch writes the native `0x10` + 16 bytes.

The server expects a 2-byte-length char value and gets a native GUID instead, which corrupts the TDS stream. Before this change both sides used the CHAR format, so this copy worked.

Suggested fix: use the original source type on both sides, e.g.
```java
if (SSType.GUID == destSSType
        && microsoft.sql.Types.GUID == srcColumnMetadata.get(srcColOrdinal).jdbcType) {
```
A test that copies an encrypted `varchar(36)` source into a plaintext `uniqueidentifier` destination would cover this. `BulkCopyGuidAETest` only covers an encrypted destination.

_Generated by GitHub Copilot (run by Muskan)._

muskan124947: Fixed in ab84634a. The native GUID row path now also requires the original source column type (`srcColumnMetadata.get(srcColOrdinal).jdbcType`) to be `GUID`, which is the same check `writeTypeInfo` uses for the header, so the two can't disagree anymore.

Added `testBulkCopyEncryptedSourceIntoPlaintextGuidColumn` to `BulkCopyGuidAETest`. It copies from encrypted `varchar(36)`, `char(36)` and `uniqueidentifier` columns into a plaintext `uniqueidentifier` column, with a value row and a null row. Without the fix it fails with a TDS error; with the fix it passes.

_Generated by GitHub Copilot (run by Muskan)._

src/main/java/com/microsoft/sqlserver/jdbc/SQLServerBulkCopy.java
```diff
@@ -2258,6 +2270,63 @@ private void validateColumnMappings() throws SQLServerException {
...
+                // UUID objects previously used their 36-character rendering on the CHAR wire path.
+                if (precision < GUID_TEXT_LENGTH) {
+                    throw new IllegalArgumentException();
+                }
+                guidValue = (UUID) colValue;
+            } else {
+                guidValue = parseGuid(colValue.toString(), precision);
+            }
+        } catch (IllegalArgumentException ex) {
+            MessageFormat form = new MessageFormat(SQLServerException.getErrString("R_errorConvertingValue"));
```
muskan124947: **Blocking - preserve the transaction outcome on GUID conversion failure.** Moving this failure to the client changes more than the reported exception: earlier work in an existing transaction can now be committed when the same bulk-copy error previously rolled it back.

I reproduced this with the exact baseline `347d7755cdd46a1c01e3b23b02c707986264424d` and head `ab84634ac882e1375b22c5c4d10e9b96202ef1f5` on SQL Server 16.0.1200.5 / Windows JDK 21: use an externally supplied connection, call `setAutoCommit(false)`, insert a sentinel row into a separate table, then bulk-copy a GUID source (precision 36, default batch size/internal-transaction settings) containing `not-a-guid`. Catch the exception and call `commit()` on that connection. Baseline leaves **0 sentinel rows**; head commits **1 sentinel row**. The result is the same with the malformed value first, after one valid row, or after 4,096 valid rows. The failed bulk batch has zero rows and subsequent queries succeed in both versions, so this is not a connection-reuse or partial-bulk-insert claim.

Previously SQL Server performed the conversion and reported error 8169; this throw instead takes `TDSCommand.execute`'s incomplete-request cancellation path, leaving the caller's earlier writes intact. A caller relying on the previous conversion-error rollback can therefore persist work after an upgrade that it did not persist before. The existing tests compare only stored/rejected outcomes and close the connection before checking this subsequent operation.

Please preserve the legacy behavior for existing callers, or obtain an explicit compatibility/migration decision covering this transaction change before making the native path unconditional. A legacy-default/opt-in path is one possible approach; adding an unconditional rollback of caller-owned transactions is **not** a safe fix. Add a regression covering prior work and the next commit/rollback, not just the conversion exception.

muskan124947 (commented):
Reviewed the native GUID bulk-copy change against merge-base `347d7755cdd46a1c01e3b23b02c707986264424d`, including the parsing/precision and encrypted-source follow-up fixes. One new compatibility finding is attached: a reproduced change in the committed outcome of a caller-owned transaction after a malformed GUID, beyond the previously discussed parsing and exception changes.

Validation included the standalone parser suite, 311 non-Extended-Events integration cases, and identical baseline/head transaction probes on Windows/JDK 21 with SQL Server 16.0.1200.5. Always Encrypted execution, other runtime/platform combinations, and a quantitative performance comparison were not rerun; source review is not proof of compatibility or performance success.

PR #3046, opened 2026-09-13
Fix spelling of occurred in JDBC error messages

tanvir-ux:
## Summary
- Correct `occured` → `occurred` in `SQLServerFMTQuery` fine log and `SQLServerResource` `R_AasJWTError` message.

## Test plan
- [x] Message/log-string only; no behavior change

Happy to adjust if anything looks off.


divang:
Thanks for catching and fixing these spelling errors! We’ve addressed the same corrections through an internal PR (#3051), so we’ll close this PR to avoid duplicating the change.

We’ll make sure your contribution is credited and that you remain recognized as the original contributor. We appreciate you taking the time to report this!

Issue #3047, opened 2026-09-15
[PERFORMANCE] Reduce number of allocations in executeBatch


dolfinus:
I'm using Spark 3.5.9 with MSSQL JDBC Driver 13.4.0 to write data into a table (24 columns of mostly varchar and datetime). It takes a lot of time - almost an hour for 58M rows (6 threads, batchsize 20k). Raw data is just 5.2GiB in ORC format, without compression.

Under the hood, Spark uses `addBatch` + `executeBatch`, like this:
https://github.com/apache/spark/blob/v3.5.9/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala#L728-L760

Using [Pyroscope](https://grafana.com/docs/pyroscope/latest/configure-client/language-sdks/java/) to get flamegraph reveals a lot of allocations:
| Mehod | Self | Total |
| --- | --- | --- |
| com/microsoft/sqlserver/jdbc/SQLServerConnection$CityHash128Key.<init> | 117.04 GiB | 137.64 GiB |
| com/microsoft/sqlserver/jdbc/DTV.sendByRPC | 48.18 GiB | 187.16 GiB |
| java/util/Arrays.copyOfRange | 46.68 GiB | 46.68 GiB |
| com/microsoft/sqlserver/jdbc/Parameter.cloneForBatch | 43.14 GiB | 43.14 GiB |
| java/util/Calendar.<init> | 37.03 GiB | 37.03 GiB |
| java/util/TimeZone.clone | 31.03 GiB | 31.03 GiB |
| java/lang/AbstractStringBuilder.<init> | 28.85 GiB | 28.85 GiB |
| com/microsoft/sqlserver/jdbc/DTV.setValue | 28.62 GiB | 42.95 GiB |
| com/sun/crypto/provider/CipherCore.prepareInputBuffer | 26.75 GiB | 26.75 GiB |
| com/microsoft/sqlserver/jdbc/DTV$SendByRPCOp.sendTemporal | 19.78 GiB | 109.69 GiB |
| mssql/googlecode/cityhash/CityHash.weakHashLen32WithSeeds | 19.45 GiB | 19.45 GiB |
| com/microsoft/sqlserver/jdbc/Parameter.setValue | 18.99 GiB | 61.94 GiB |
| java/lang/String.valueOf | 18.23 GiB | 37.26 GiB |
[...]

PR #3048, opened 2026-09-15
Avoid generating new CityHash128Key if connection pooling disabled

dolfinus:
See #3047. By default, `disableStatementPooling=true`, so PreparedStatementHandle is never added into the cache, and just allocated and calculated CityHash128Key is discarded.

<details>
<summary>Before</summary>

669 GiB of allocations

Снимок экрана_20260915_135101
Снимок экрана_20260915_135128
</details>

<details>
<summary>After (-34% of allocations)</summary>

438 GiB of allocations

Снимок экрана_20260915_154227
Снимок экрана_20260915_154300
</details>


dolfinus:
Setting `disableStatementPooling=false` and non-zero `statementPoolingCacheSize` also reduces number of allocations (as cached statementHandle is accessed without generating cache key). But this may have side effects.
Снимок экрана_20260915_160719

